PR #13, opened 2017-07-18
Various fixes for MVP

jonathan-gramain:
- set the destination owner ID and display name in metadata before
   putting to the destination

 - transmit the owner ID in the 'x-scal-canonical-id' header when
   putting data

 - parse the comma-separated source and destination roles from bucket
   replication configuration

 - use an abstraction to handle authentication for 'account' and
   'role' auth types. For account type, match the configured local
   user with the role ARN read in the bucket replication
   configuration.

 - rename auth types to 'user' or 'role' for local user (or account)
   authentication or role-based.

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -15,6 +15,67 @@ const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
 const BackbeatClient = require('../../../lib/clients/BackbeatClient');
 const QueueEntry = require('../utils/QueueEntry');
 
+class _AccountAuthManager {
+    constructor(authConfig, log) {
+        assert.strictEqual(authConfig.type, 'account');
+
+        this.log = log;
+        const accountInfo = authdata.accounts.find(
+            account => account.name === authConfig.account);
+        if (accountInfo === undefined) {
+            throw Error(`No such account registered: ${authConfig.account}`);
+        }
+        if (accountInfo.arn === undefined) {
+            throw Error(`Configured account ${authConfig.account} has no ` +
+                        '"arn" property defined');
+        }
+        if (accountInfo.canonicalID === undefined) {
+            throw Error(`Configured account ${authConfig.account} has no ` +
+                        '"canonicalID" property defined');
+        }
+        if (accountInfo.displayName === undefined) {
+            throw Error(`Configured account ${authConfig.account} has no ` +
+                        '"displayName" property defined');
+        }
+        this.accountArn = accountInfo.arn;
+        this.canonicalID = accountInfo.canonicalID;
+        this.displayName = accountInfo.displayName;
+        this.credentialProvider = new AWS.CredentialProviderChain([
+            new AWS.Credentials(accountInfo.keys.access,
+                                accountInfo.keys.secret),
+        ]);
+    }
+
+    getCredentialProvider() {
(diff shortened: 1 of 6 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -15,6 +15,67 @@ const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
...
+
+class _RoleAuthManager {
+    constructor() {
+        throw Error('Role-based authentication not yet implemented');
+    }
+}
+
+
+function _extractUserFromRole(role) {
+    return role.split(':').slice(0, 5).join(':');
```
rahulreddy: This gives back the same string, I am not sure if this is intentional. Also, role arn contains accountId not user.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -88,7 +140,34 @@ class QueueProcessor {
...
+                }
+                return cb(null, roles[0], roles[1]);
+            });
+    }
+
+    _processRoles(sourceEntry, sourceRole, destEntry, targetRole, cb) {
+        this.log.debug('processing role for destination',
+                       { entry: destEntry.getLogInfo(),
+                         role: targetRole });
+        const targetUserId = _extractUserFromRole(targetRole);
```
rahulreddy: You would be extracting target accountId here not user.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -15,6 +15,67 @@ const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
...
 
+class _UserAuthManager {
+    constructor(authConfig, log) {
+        assert.strictEqual(authConfig.type, 'user');
+
+        this.log = log;
+        const userInfo = authdata.users.find(
+            user => user.name === authConfig.user);
+        if (userInfo === undefined) {
+            throw Error(`No such user registered: ${authConfig.user}`);
```
jonathan-gramain: This error is about a misconfiguration: the friendly "user" name configured in config.json to reference a source or target account configured in authdata.json is not found in the latter, which should not happen if the product is configured correctly, so I think it's appropriate to throw here.

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: Normally, unless I used a bad role arn as an example, this should return everything up to the last colon before the role name. Will check this.

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: OK, will rename.

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: Maybe when we add the support for multiple users for the open source version, and a more clever lookup to match each role, we would want to be more gentle with misconfiguration and just ignore those entries not matching a registered user, to allow the user to miss some user info entries and still be able to replicate the other ones, for now I think it's ok since there can be only one configured user.

rahulreddy: Ok, since this impacts only open-source users and it would happen only when there is misconfiguration, let's keep it and cleanup later.

PR #14, opened 2017-07-18
FT: S3C-350 Use roles to lookup canonicalId

rahulreddy:
Depends on

~~-  #13~~
~~https://github.com/scality/Vault/pull/757~~

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -8,6 +8,7 @@ const Logger = require('werelogs').Logger;
 
 const errors = require('arsenal').errors;
 const jsutil = require('arsenal').jsutil;
+const Vaultclient = require('vaultclient');
 
 const authdata = require('../../../conf/authdata.json');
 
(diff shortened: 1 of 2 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -66,8 +67,31 @@ class _UserAuthManager {
...
 class _RoleAuthManager {
-    constructor() {
-        throw Error('Role-based authentication not yet implemented');
+    constructor(authConfig, log) {
+        this._log = log;
+        const { host, port } = authConfig.vault;
+        this._vaultclient = new Vaultclient(host, port);
+    }
+
+    getCredentialProvider() {
```
jonathan-gramain: This function returns `undefined`, but it may be temporary until we implement the credentials properly. (actually I don't know if `getCredentialProvider()` is going to stay with the credentials cache etc.).

rahulreddy: Yeah I am not sure, it was something that I worked with last night. I will make changes as needed.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -264,6 +254,37 @@ class QueueProcessor {
         });
     }
 
+    _setupClients(sourceRole, targetRole) {
```
jonathan-gramain: I would rather decouple the auth managers from the `QueueProcessor` object, because we may want to keep the possibility to parallelize synchronizations later, which is not possible if cred managers are stored in the main queue processor object. Could be done by embedding all necessary info in a `ReplicationContext` object for example.

rahulreddy: I agree, I did this just to have a working mvp. Once, we have a working copy we can cleanup the setup.

jonathan-gramain: ok

PR #15, opened 2017-07-19
enhance replication error management

jonathan-gramain:
Split error management in two categories: retryable and permanent
failures. The choice is currently based on the 'retryable' field of
the error sent back by the AWS SDK, but it could be refined later.

 - retryable errors get retried indefinitely with an exponential
   back-off algorithm (using 'backo' open-source module, it's simple
   but non-trivial to do properly so makes sense to use a third-party
   module IMO). We may later think about giving up after some (long
   enough) time without success and set the replication status to
   FAILED, not sure we want to though, if the errors are really
   retryable they should eventually succeed.

 - permanent errors lead to the status FAILED being written on the
   source object version.

 - failures to write the replication status also lead to retries.

Tests should be written, but they will be in a later PR as the code
will need instrumentation that will likely conflict with other pending
PRs.

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -4,6 +4,7 @@ const async = require('async');
 const assert = require('assert');
 const AWS = require('aws-sdk');
 const http = require('http');
+const BackOff = require('backo');
 
 const Logger = require('werelogs').Logger;
 
@@ -18,7 +19,7 @@ const BackbeatClient = require('../../../lib/clients/BackbeatClient');
 const QueueEntry = require('../utils/QueueEntry');
 const CredentialsManager = require('../../../credentials/CredentialsManager');
 
-const LIMIT = 10;
+const MPU_CONC_LIMIT = 10;
 
 class _AccountAuthManager {
     constructor(authConfig, log) {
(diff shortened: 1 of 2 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -354,97 +425,150 @@ class QueueProcessor {
...
+                      'status',
+                      { entry: updatedSourceEntry.getLogInfo(),
+                        replicationStatus:
+                        updatedSourceEntry.getReplicationStatus() });
+            // Rely on AWS SDK notion of retryable error to decide if
+            // we should retry or give up updating the status.
+            if (err.retryable) {
+                return this._retryUpdateReplicationStatus(
+                    updatedSourceEntry, backoffCtx, log, done);
+            }
```
philipyoo: I see `_retryUpdateReplicationStatus(updatedSourceEntry, params, done)` as the function definition. Does `backoffCtx, log` need to be wrapped as an object?

jonathan-gramain:
@philipyoo The suggestion you make about a max retry limit is done in https://github.com/scality/backbeat/pull/31 :)

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: right, good catch, will fix in this PR so that it's mergeable, though the next retry PR does get rid of this code.

PR #20, opened 2017-07-20
bf: fix value of initial logOffset zk nodes

jonathan-gramain:
The logOffset zookeeper nodes from queue populator state are now
created with the expected initial value '1', instead of null, which
was interpreted as 0 when converted to an integer and was causing
issues with the Raft log API.


jonathan-gramain:
This was not the root cause of the issue, and the changes here are unnecessary since default 'null' value when read initializes the `logOffset` counter to 1 as expected.

PR #22, opened 2017-07-24
raft session provisioning

jonathan-gramain:
Support raft ID auto provisioning with zookeeper watchers.


jonathan-gramain:
replaced by https://github.com/scality/backbeat/pull/23, please review the latter instead.

PR #23, opened 2017-07-24
raft session provisioning

jonathan-gramain:
Support raft ID auto provisioning with zookeeper watchers.

Replaces https://github.com/scality/backbeat/pull/22

### DIFF
--- a/lib/provisioning/ProvisionDispatcher.js
+++ b/lib/provisioning/ProvisionDispatcher.js
@@ -1,0 +1,348 @@
+'use strict'; // eslint-disable-line
+
+const async = require('async');
+const crypto = require('crypto');
+const path = require('path');
+const zookeeper = require('node-zookeeper-client');
+
+const Logger = require('werelogs').Logger;
+
+function _genRandomHexChars(len) {
+    return crypto.randomBytes(Math.ceil(len / 2))
+        .toString('hex') // convert to hexadecimal format
+        .slice(0, len);  // return required number of characters
+}
+
+const LEADERS = '/leaders';
+const LEADER = '/leader';
+const OWNERS = '/owners';
+const PROVISIONS = '/provisions';
+
+class ProvisionDispatcher {
+
+    /**
+     * Create a new provision dispatcher object
+     *
+     * @constructor
+     * @param {object} zkConfig - zookeeper config object
+     * @param {string} zkConfig.endpoint - zookeeper endpoint string
+     * where provisions are managed
+     * (e.g. "localhost:2181/shared-tasks")
+     * @param {object} logConfig - logging configuration
+     * @param {string} logConfig.logLevel - logging level
+     * @param {string} logConfig.dumpLevel - dump level
+     */
+    constructor(zkConfig, logConfig) {
+        this._zkEndpoint = zkConfig.endpoint;
+        this._client = zookeeper.createClient(zkConfig.endpoint);
+        this._client.connect();
+        this._connectWaitList = [];
+        this._client.once('connected', () => {
+            this._log.debug('connected to the ZK server');
+            this._connectWaitList.forEach(cb => cb());
+            this._connectWaitList = null;
+        });
+        this._myName = null;
+        this._myLeaderName = null;
+        this._isLeader = false;
+        this._owners = null;
+        this._provisions = null;
+
+        this._log = new Logger('Backbeat:ProvisionDispatcher',
+                              { level: logConfig.logLevel,
+                                dump: logConfig.dumpLevel });
+    }
+
+    /**
+     * Add new provisions to the list of provisions to be dispatched
+     *
+     * @param {array} provisionList - list of provision items as strings
+     * @param {function} cb - callback when done
+     * @return {undefined}
+     */
+    addProvisions(provisionList, cb) {
+        this._log.debug('add new provisions', { provisionList });
+        async.series([
+            next => this._waitConnected(next),
+            next => this._populateZkHierarchy(next),
+            next => this._addProvisionZkNodes(provisionList, next),
+        ], cb);
+    }
+
+    subscribe(cb) {
+        this._waitConnected(() => {
+            async.parallel([
+                done => this._registerLeader(done),
+                done => this._registerOwner(done),
+            ], err => {
+                if (err) {
+                    return cb(err);
+                }
+                this._onLeaderChange();
+                this._onOwnerChange();
+                this._onMyselfChange(cb);
+                return this._onProvisionChange();
+            });
+        });
+    }
+
+    unsubscribe(cb) {
+        async.parallel([
+            next => this._unregisterOwner(next),
+            next => this._unregisterLeader(next),
+        ], cb);
+    }
+
+    _waitConnected(cb) {
+        if (this._connectWaitList === null) {
+            return process.nextTick(cb);
+        }
+        return this._connectWaitList.push(cb);
+    }
+
+    _populateZkHierarchy(cb) {
+        async.each([
+            LEADERS,
+            OWNERS,
+            PROVISIONS,
+        ], (zkPath, done) => this._client.create(zkPath, err => {
+            if (err && err.getCode() !== zookeeper.Exception.NODE_EXISTS) {
+                this._log.error('error populating zk node',
+                               { zkPath: `${this._zkEndpoint}${zkPath}`,
+                                 error: err });
+                return done(err);
+            }
+            this._log.debug('populated zk node', { zkPath });
+            return done();
+        }), cb);
+    }
+
+    _addProvisionZkNodes(provisionList, cb) {
+        async.eachLimit(
+            provisionList.map(
+                item => `${PROVISIONS}/${item}`),
+            20,
+            (zkPath, done) => this._client.create(zkPath, err => {
+                if (err &&
+                    err.getCode() !== zookeeper.Exception.NODE_EXISTS) {
+                    this._log.error('error adding provision node',
+                                   { zkPath: `${this._zkEndpoint}${zkPath}`,
+                                     error: err });
+                    return done(err);
+                }
+                this._log.debug('added new provision zk node',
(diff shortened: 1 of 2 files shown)


jonathan-gramain:
Updated PR to include only the zookeeper part of the feature, because there are significant new changes on the queue populator related to provisioning which makes the present changes obsolete. Another PR will come soon for those new changes.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,355 @@
...
+     * @param {object} logConfig - logging configuration
+     * @param {string} logConfig.logLevel - logging level
+     * @param {string} logConfig.dumpLevel - dump level
+     */
+    constructor(zkConfig, logConfig) {
+        this.zkEndpoint = zkConfig.endpoint;
+        this.client = zookeeper.createClient(zkConfig.endpoint);
+        this.client.connect();
+        this.connectWaitList = [];
+        this.client.once('connected', () => {
```
jonathan-gramain: In the handler of `connected` event, we set `connectWaitList` to `null`, so that next calls to `_waitConnected` return immediately. AFAIK, if disconnects occur next, it's all up to the client zookeeper lib to attempt to reconnect automatically, transparently to the user. With the use of `once` handler, we'll just ignore next instances of `connected` event.

If we lose the zookeeper session though, we would have to restart the queue populator since we don't attempt to reestablish a new session when such event occurs. I think it's ok for now as the queue populator is local to a zookeeper node but it may have to be done in the future.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,355 @@
...
+        this.myName = null;
+        this.myLeaderName = null;
+        this.isLeader = false;
+        this.owners = null;
+        this.provisions = null;
+
+        this.logger = new Logger('Backbeat:ProvisionDispatcher',
+                                 { level: logConfig.logLevel,
+                                   dump: logConfig.dumpLevel });
+        this.log = this.logger.newRequestLogger();
```
jonathan-gramain: Oh, ok, good to know, it indeed should be fixed.

How it should be done then? I'm not sure using the `logger` object works in this case, is there another call to use instead or is the `logger` object supposed to be used for logging calls?

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,355 @@
...
+     * @param {array} provisionList - list of provision items as strings
+     * @param {function} cb - callback when done
+     * @return {undefined}
+     */
+    addProvisions(provisionList, cb) {
+        this.log.debug('add new provisions', { provisionList });
+        async.series([
+            done => this._waitConnected(done),
+            done => this._populateZkHierarchy(done),
+            done => async.eachLimit(
```
jonathan-gramain: Makes sense, will do

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,355 @@
...
+                               { zkPath: this.zkEndpoint + zkPath });
+                return cb(null, zkPath);
+            });
+    }
+
+    _unregisterOwner(cb) {
+        const zkPath = `${ProvisionDispatcher.constants.OWNERS}/` +
+                  `${this.myName}`;
+        this.client.remove(zkPath, -1, err => {
+            if (err) {
```
jonathan-gramain: I think it's ok to forward the error, and let the client handle it appropriately. Normally he should only call `unsubscribe()` if he already has called `subscribe()` earlier, in which case there shall be no error, or if he calls `unsubscribe()` in a "safety net" fashion without knowing if he subscribed earlier, he should then deliberately ignore the `NO_NODE` errors.

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: It seems to work by using the logger object, so will go this way.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,356 @@
...
+        this._waitConnected(() => {
+            this._registerLeader(err => {
+                if (err) {
+                    return cb(err);
+                }
+                return this._onLeaderChange();
+            });
+            this._registerOwner(err => {
+                if (err) {
+                    return cb(err);
```
rahulreddy: This has a potential of callback getting called twice.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,356 @@
...
+                }
+                this._onOwnerChange();
+                this._onMyselfChange(cb);
+                return undefined;
+            });
+            return this._onProvisionChange();
```
rahulreddy: Shouldn't this method use a callback since it's async?

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,356 @@
...
+        async.parallel([done => this._unregisterOwner(done),
+                        done => this._unregisterLeader(done)],
+                       cb);
+    }
+
+    _waitConnected(cb) {
+        if (this.connectWaitList === null) {
+            return process.nextTick(cb);
+        }
+        return this.connectWaitList.push(cb);
```
rahulreddy: The callback is never called in this case?

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,356 @@
...
+                                 event });
+                this._onMyselfChange(cb);
+            },
+            (err, data) => {
+                if (err &&
+                    err.getCode() !== zookeeper.Exception.NO_NODE) {
+                    this.log.error('error in getData',
+                                   { zkPath: this.zkEndpoint + myPath,
+                                     error: err });
+                    return undefined;
```
rahulreddy: This code path should call the callback with an error.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    _onProvisionChange() {
+        const zkPath = ProvisionDispatcher.constants.PROVISIONS;
+        this.client.getChildren(zkPath, event => {
+            this.log.debug('got provision event', {
+                zkPath: this.zkEndpoint + zkPath,
+                event,
+            });
+            this._onProvisionChange();
```
rahulreddy: Is there a deep exit condition since this is recursive?

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: This is ok, because it's more of a listener callback than a completion callback, getting called whenever changes happen on the provisioning.

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: same thing, this call is just to register event based on if the provisions list changes.

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: In this case, it gets called when the actual connection occurs, see in the constructor, there's a loop over `connectWaitList` to notify waiters that we're actually connected.

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: This is only triggered when there's an actual change on the children (the second argument of `getChildren` is an event watcher, not the actual callback).

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: We can, it's not a strong functional requirement but it would allow to notify the caller of errors so that he can log them as well as here (or do some other action). I will do the change.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -58,36 +58,34 @@ class ProvisionDispatcher {
     addProvisions(provisionList, cb) {
         this._log.debug('add new provisions', { provisionList });
         async.series([
-            done => this._waitConnected(done),
-            done => this._populateZkHierarchy(done),
+            this._waitConnected,
```
jonathan-gramain: I think you need to `.bind(this)` in order to provide a closure rather than the bare class method

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -58,36 +58,34 @@ class ProvisionDispatcher {
...
         this._waitConnected(() => {
-            this._registerLeader(err => {
-                if (err) {
-                    return cb(err);
-                }
-                return this._onLeaderChange();
-            });
-            this._registerOwner(err => {
+            async.parallel([
+                this._registerLeader,
```
jonathan-gramain: same here

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -58,36 +58,34 @@ class ProvisionDispatcher {
...
-            return this._onProvisionChange();
         });
     }
 
     unsubscribe(cb) {
-        async.parallel([done => this._unregisterOwner(done),
-                        done => this._unregisterLeader(done)],
-                       cb);
+        async.parallel([
+            this._unregisterOwner,
```
jonathan-gramain: same

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -346,11 +345,4 @@ class ProvisionDispatcher {
     }
 }
 
-ProvisionDispatcher.constants = {
```
jonathan-gramain: Did you move the definition of those constants elsewhere? As far as I can see they're still used.

rahulreddy: Ah I did partial changes. Let me update that.

lib/provisioning/ProvisionDispatcher.js
rahulreddy: It's shorthand, I don't think we have to bind this here.

PR #25, opened 2017-07-25
Integrate ProvisionDispatcher class in queue populator

jonathan-gramain:
Follow-up of https://github.com/scality/backbeat/pull/23.

Use the ProvisionDispatcher class to configure the queue populator so that it manages one or more raft session IDs. Make it dynamic so that a running queue populator can switch to another set of raft session
IDs any time (if the dispatcher reassigns a new one, after a failure for example), or can become idle as well if no raft ID is assigned from zookeeper.

Add support for graceful handling of SIGTERM, to unsubscribe from the dispatcher so that the dispatcher can quickly notify other queue populators to take over the unmanaged raft IDs.

### DIFF
--- a/tests/behavior/extensions/replication/queuePopulator.js
+++ b/tests/behavior/extensions/replication/queuePopulator.js
@@ -21,9 +21,11 @@ const s3config = {
 describe('queuePopulator', () => {
     let queuePopulator;
     let s3;
-    let latestLastProcessedSeq;
 
-    before(done => {
+    before(function setup(done) {
+        this.timeout(30000); // may take some time to keep up with the
+                             // log entries
+
         s3 = new S3(s3config);
         async.waterfall([
             next => {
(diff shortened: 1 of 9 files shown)


tests/behavior/extensions/replication/queuePopulator.js
```diff
@@ -56,20 +59,17 @@ describe('queuePopulator', () => {
...
+                    testConfig.zookeeper,
+                    testConfig.replication.source,
+                    testConfig.replication,
+                    testConfig.log);
                 queuePopulator.open(next);
             },
             next => {
                 queuePopulator.processAllLogEntries({ maxRead: 10 }, next);
             },
             (counters, next) => {
```
bennettbuchanan: Is there a reason to keep this step of the waterfall?

rahulreddy: I guess not, good catch.

PR #26, opened 2017-07-27
queue processor functional test

jonathan-gramain:
test: end-to-end queue processor functional test
    
Using mocks for S3 and Vault, check that the normal successful path works correctly to replicate an object.


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -180,11 +179,12 @@ class QueueProcessor {
         return this.S3source.getBucketReplication(
             { Bucket: entry.getBucket() }, (err, data) => {
                 if (err) {
-                    this.log.error('error response getting replication ' +
+                    this.log.error('error getting replication ' +
                                    'configuration from S3',
                                    { entry: entry.getLogInfo(),
                                      origin: this.sourceConfig.s3,
-                                     error: err,
+                                     error: { message: err.message },
```
rahulreddy: ELK expects this property to be a string, so using `error: err.message` makes sense. Please update other references as well.

PR #28, opened 2017-07-28
FT: Support mpu for big files

bennettbuchanan:
Depends on https://github.com/scality/S3/pull/840.

I will be adding functional tests to validate this.


extensions/replication/utils/QueueEntry.js
```diff
@@ -135,6 +135,18 @@ class QueueEntry {
         };
     }
 
+    getLocation() {
+        return this.objMd.location;
+    }
+
+    getPartNumber(location) {
+        return Number(location.dataStoreETag.split(':')[0], 10);
```
jonathan-gramain: I think you have to use `Number.parseInt()` to provide the base, the constructor seems to only take one argument.

tests/functional/replication/queueProcessor.js
```diff
@@ -81,6 +81,7 @@ class S3Mock {
                         size: constants.body.length,
                         start: 0,
                         dataStoreName: 'file',
+                        dataStoreETag: `1:${this.contentMd5}`,
```
jonathan-gramain: We should keep the test without the `dataStoreETag` set, because it appears non-MPU objects do not have this property. But you may already have planned to do this for the tests you're currently writing :)

You could e.g. duplicate the `kafkaEntry` object (deep copy) and append this property for the MPU test, or have a function generate the `kafkaEntry` from a set of input params.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -221,48 +223,66 @@ class QueueProcessor {
...
-        const incomingMsg = req.createReadStream();
-        req.on('error', err => {
-            this.log.error('error response getting data from S3',
-                           { entry: entry.getLogInfo(),
-                             origin: this.sourceConfig.s3,
-                             error: err.message,
-                             httpStatus: err.statusCode });
-            incomingMsg.emit('error', err);
+        const locations = entry.getLocation();
+        return async.timesLimit(locations.length, LIMIT, (i, next) => {
```
jonathan-gramain: I think you could use `async.eachLimit(locations, LIMIT, (location, next) => {...})` instead.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -18,6 +18,8 @@ const BackbeatClient = require('../../../lib/clients/BackbeatClient');
 const QueueEntry = require('../utils/QueueEntry');
 const CredentialsManager = require('../../../credentials/CredentialsManager');
 
+const LIMIT = 10;
```
jonathan-gramain: It could be interesting to have this configurable. I heard that some of our customers will probably need MPU for performance, and the parallelism factor will be an important variable to control in order to optimize performance for big objects.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -221,48 +223,66 @@ class QueueProcessor {
...
-        });
-        this.backbeatDest.putData({
-            Bucket: entry.getBucket(),
-            Key: entry.getObjectKey(),
-            CanonicalID: entry.getOwnerCanonicalId(),
-            ContentLength: entry.getContentLength(),
-            ContentMD5: entry.getContentMD5(),
-            Body: sourceStream,
-        }, (err, data) => {
+        return async.timesLimit(streams.length, LIMIT, (i, next) => {
```
jonathan-gramain: Same than above about `eachLimit`

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: For performance reasons, and to avoid an unnecessary "barrier" effect, it could be interesting to replace the "get all streams first and put next" where gets likely involve some metadata lookups on the source for each part and take some non-negligible time, by a "get and start putting ASAP" for each individual stream. This would reduce wait times and increase throughput by some amount.

Basically, the async processing callback would englobe the two operations `get` and `put` serialized in a single processing unit for each part (you may use `async.waterfall` but not sure it's worth it for just two calls).

extensions/replication/utils/QueueEntry.js
bennettbuchanan: Thanks, that's what I was meaning.

tests/functional/replication/queueProcessor.js
rahulreddy: I think I tested this case yesterday. `dataStoreETag` will appear for all objects, mpu or not.

extensions/replication/queueProcessor/QueueProcessor.js
rahulreddy: I think this translates to rate limiting for us. Our rate limiting api will/can probably tune this, we can make it configurable then.

tests/functional/replication/queueProcessor.js
bennettbuchanan: Yes, with the change is S3, it is included for regular object's MD.

extensions/replication/queueProcessor/QueueProcessor.js
bennettbuchanan: I've used `timesLimit` here since I want an array of results at the end, which is not included with `eachLimit`. Also I want the `i` to get each part number.

rahulreddy: You are right, that would be a nice optimization. I think `async.queue` is a good fit here. You can keep adding streams to the queue, set the concurrency of how many you want to process in parallel and the "processor" is invoked as soon as something is added to the queue.

tests/functional/replication/queueProcessor.js
jonathan-gramain: ok, no issue then

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: @bennettbuchanan apparently you don't need `i` to get the part number since you're calling `entry.getPartNumber()` ;)

To get the array of results at the end, it seems like `mapLimit` would be the right answer here. But if we go towards doing the perf optimization I mentioned, keeping the whole list of streams will not be needed anymore.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -221,48 +223,66 @@ class QueueProcessor {
...
+                               { entry: entry.getLogInfo(),
+                                 origin: this.sourceConfig.s3,
+                                 error: err, httpStatus: err.statusCode });
+                incomingMsg.emit('error', err);
+            });
+            return next(null, {
+                partETag: entry.getPartETag(locations[i]),
+                sourceStream: incomingMsg,
+            });
+        }, (err, results) => {
```
jonathan-gramain: This callback looks unnecessary

tests/functional/replication/queueProcessor.js
jonathan-gramain: just make sure we're compatible with old data as well (if they don't have this field in metadata stored prior to the introduction of `dataStoreETag`)

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: @rahulreddy since, as you mention, a queue would allow to keep adding streams to it, but since we already have the whole list of locations by fetching it from the entry, I would think than `eachLimit` is simpler for this use case. Maybe I'm missing other features of the queue which could be useful.

tests/functional/replication/queueProcessor.js
bennettbuchanan: Updated to return error. InvalidObjectState if old object.

jonathan-gramain: :+1:

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -219,50 +221,58 @@ class QueueProcessor {
...
+                                 error: err, httpStatus: err.statusCode });
+                incomingMsg.emit('error', err);
+            });
+            this.log.debug('putting data', { entry: destEntry.getLogInfo() });
+            incomingMsg.on('error', err => {
+                this.log.error('error from source S3 server',
+                               { entry: destEntry.getLogInfo(), error: err });
+                return done(err);
+            });
+            return this.backbeatDest.putData({
```
rahulreddy: This needs to be hooked into the req's response event. Otherwise we may start putting even before we receive a response from the source S3.

jonathan-gramain: I don't think that works, because as far as I can remember the `response` event is only emitted when all the data has been received from the source, which means we'll get stuck waiting for it if we don't start writing to the destination.

Ideally we would hook when only the headers have been received from S3 but I don't know if this is feasible.

jonathan-gramain: Actually, I remember I was listening to the `success` event, which explains that it had to wait for the rest of the request. There may be the possibility to wait on the HTTP connection `response` event but I did not figure this out.

jonathan-gramain: I think I found what's needed, S3, sends a `httpHeaders` event when headers have been sent by the remote.

See http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html for details.

rahulreddy: My bad it was not clear. I meant in the sense of an event not the actual response event.

tests/functional/replication/queueProcessor.js
```diff
@@ -21,6 +21,7 @@ const constants = {
         secretKey: 'verySecretKey1',
         canonicalId: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
         dataKey: '6d808697fbaf9f16fb32b94be189b80b3b9b2890',
+        dataKeyTmp: 'aaaaaaaabbbbbbbasdcasdlksdjfalksdjfli',
```
jonathan-gramain: May just rename both to `dataKeyPart1` and `dataKeyPart2`

tests/functional/replication/queueProcessor.js
```diff
@@ -30,19 +31,43 @@ const constants = {
         accessKey: 'accessKey2',
         secretKey: 'verySecretKey2',
         canonicalId: 'bac4bea1bac4bea1bac4bea1bac4bea1bac4bea1bac4bea1bac4bea1bac4bea1',
-        dataKey: '7d808697fbaf9f16fb32b94be189b80b3b9b2890',
+        dataKey: '6d808697fbaf9f16fb32b94be189b80b3b9b2891',
```
jonathan-gramain: same (and they can/should actually be different from the source keys, for test purposes it can be interesting to make them different).

tests/functional/replication/queueProcessor.js
```diff
@@ -34,15 +34,38 @@ const constants = {
     },
     key: 'key_to_replicate',
     versionId: '98498980852335999999RG001  100',
-    body: 'some contents to be replicated',
+    partOneBody: 'some contents to be replicated',
+    partTwoBody: 'some more contents to be replicated',
     roleSessionName: 'backbeat-replication',
 };
 
+const partsContent = [constants.partOneBody, constants.partTwoBody];
```
jonathan-gramain: Those two should be part of the `S3Mock` object, built in the constructor (the idea is being able to spawn multiple `S3Mock` objects at the same time with different params).

tests/functional/replication/queueProcessor.js
```diff
@@ -362,28 +378,50 @@ describe('queue processor error management with mocking', () => {
...
-
-        s3mock = new S3Mock();
-        httpServer = http.createServer(
-            (req, res) => s3mock.onRequest(req, res));
-        httpServer.listen(7777);
     });
 
     afterEach(() => {
         s3mock.resetTest();
+        httpServer.close();
```
jonathan-gramain: Not sure why you moved this in `afterEach`, I think it should stay in `after`.

tests/functional/replication/queueProcessor.js
```diff
@@ -362,28 +378,50 @@ describe('queue processor error management with mocking', () => {
...
-    it('should complete a replication end-to-end', done => {
-        queueProcessor.processKafkaEntry(s3mock.kafkaEntry, err => {
-            assert.ifError(err);
-            assert(s3mock.hasPutTargetData);
-            assert(s3mock.hasPutTargetMd);
-            assert(s3mock.hasPutSourceMd);
-            done();
+    describe('object with multiple parts', () => {
+        before(() => {
+            // don't call start() on the queue processor, so that we don't
```
jonathan-gramain: This comment is only meaningful when creating the kafka queue, should be removed here.

tests/functional/replication/queueProcessor.js
```diff
@@ -34,15 +34,38 @@ const constants = {
...
+    {
+        key: constants.target.dataKey,
+        start: 0,
+        size: constants.partOneBody.length,
+        dataStoreName: 'file',
+        dataStoreETag: `1:${crypto.createHash('md5')
+            .update(constants.partOneBody).digest('hex')}`,
+    },
+    {
+        key: constants.target.dataKey,
```
jonathan-gramain: IMO the keys should be different for both locations, for test purpose.

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -38,6 +38,10 @@
                         "locationName": "Content-Length",
                         "type": "long"
                     },
+                    "PartNumber": {
+                        "location": "header",
```
jonathan-gramain: as discussed offline, please provide the location header name starting with `X-Scal-` such as `X-Scal-Part-Number`.

tests/functional/replication/queueProcessor.js
bennettbuchanan: Updated.

tests/functional/replication/queueProcessor.js
bennettbuchanan: Updated.

tests/functional/replication/queueProcessor.js
bennettbuchanan: Updated.

tests/functional/replication/queueProcessor.js
bennettbuchanan: Updated.

tests/functional/replication/queueProcessor.js
bennettbuchanan: Removed.

tests/functional/replication/queueProcessor.js
bennettbuchanan: Since I'm creating a server for each test, needed to free up the 7777 address.

jonathan-gramain: ok, that works, I think you could also reassign a new request handler for each test to avoid respawning the server, anyway I'm reworking the instrumentation as part of the error management work so that I think it will change a bit how things are setup.

PR #30, opened 2017-08-02
Ft/scripts for automation

philipyoo:
Running 2 separate instances of S3 and Vault on separate host and ports, this script:
- creates source bucket, target bucket
- enables versioning for source and target bucket
- creates roles on source and target with assume role trust policy
- creates resource policies allowing replication actions for BB on source and target
- policies attached to the roles created in the prior step
- enables replication configuration on the source bucket

This script sanity checks:
- source and target are reachable and exist
- source and target buckets have versioning enabled
- source bucket has replication config enabled
- roles mentioned in the replication config are available in IAM

Right now, the error logs are really long and I believe it is splitting multiple messages.
I will look into thinning this down as a next step 👍 
Any recommendations on changes, improvements, additions are welcome!

### DIFF
--- a/bin/setupReplication.js
+++ b/bin/setupReplication.js
@@ -1,0 +1,507 @@
+const async = require('async');
+const program = require('commander');
+const { S3, IAM, SharedIniFileCredentials } = require('aws-sdk');
+
+const { Logger } = require('werelogs');
+
+const config = require('../conf/Config');
+const { proxyIAMPath } = require('../extensions/replication/constants');
+
+const trustPolicy = {
+    Version: '2012-10-17',
+    Statement: [
+        {
+            Effect: 'Allow',
+            Principal: {
+                Service: 'backbeat',
+            },
+            Action: 'sts:AssumeRole',
+        },
+    ],
+};
+
+function _buildResourcePolicy(source, target) {
+    return {
+        Version: '2012-10-17',
+        Statement: [
+            {
+                Effect: 'Allow',
+                Action: [
+                    's3:GetObjectVersion',
+                    's3:GetObjectVersionAcl',
+                ],
+                Resource: [
+                    `arn:aws:s3:::${source}/*`,
+                ],
+            },
+            {
+                Effect: 'Allow',
+                Action: [
+                    's3:ListBucket',
+                    's3:GetReplicationConfiguration',
+                ],
+                Resource: [
+                    `arn:aws:s3:::${source}`,
+                ],
+            },
+            {
+                Effect: 'Allow',
+                Action: [
+                    's3:ReplicateObject',
+                    's3:ReplicateDelete',
+                ],
+                Resource: `arn:aws:s3:::${target}/*`,
+            },
+        ],
+    };
+}
+
+function _setupS3Client(transport, endpoint, profile) {
+    const credentials = new SharedIniFileCredentials({ profile });
+    return new S3({
+        endpoint: `${transport}://${endpoint}`,
+        sslEnabled: transport === 'https',
+        credentials,
+        s3ForcePathStyle: true,
+    });
+}
+
+
+function _setupIAMClient(transport, endpoint, profile) {
+    const credentials = new SharedIniFileCredentials({ profile });
+    return new IAM({
+        endpoint: `${transport}://${endpoint}`,
+        sslEnabled: transport === 'https',
+        credentials,
+        maxRetries: 0,
+        region: 'us-east-1',
+        signatureCache: false,
+        httpOptions: {
+            timeout: 1000,
+            proxy: proxyIAMPath,
+        },
+    });
+}
+
+class _SetupReplication {
+    /**
+     * This class sets up two buckets for replication.
+     * @constructor
+     * @param {String} sourceBucket - Source Bucket Name
+     * @param {String} targetBucket - Target Bucket Name
+     * @param {String} sourceProfile - Source Credentials Profile
+     * @param {String} targetProfile - Target Credentials Profile
+     * @param {Object} log - Werelogs Request Logger object
+     * @param {Object} config - bucket configurations
+     */
+    constructor(sourceBucket, targetBucket, sourceProfile, targetProfile, log,
+        config) {
+        const { source, destination } = config.extensions.replication;
+        this._log = log;
+        this._sourceBucket = sourceBucket;
+        this._targetBucket = targetBucket;
+        // TODO: support bootstrap list failover
+        const destEndpoint = destination.bootstrapList[0].servers[0];
+        this._s3Clients = {
+            source: _setupS3Client(source.transport,
+                `${source.s3.host}:${source.s3.port}`,
+                sourceProfile),
+            target: _setupS3Client(destination.transport, destEndpoint,
+                targetProfile),
+        };
+        this._iamClients = {
+            source: _setupIAMClient(source.transport,
+                `${source.auth.vault.host}:${source.auth.vault.iamPort}`,
+                sourceProfile),
+            target: _setupIAMClient(source.transport, destEndpoint,
+                targetProfile),
+        };
+    }
+
+    checkSanity(cb) {
+        return async.waterfall([
+            next => this._isValidBucket('source', next),
+            next => this._isValidBucket('target', next),
+            next => this._isVersioningEnabled('source', next),
+            next => this._isVersioningEnabled('target', next),
+            next => this._isReplicationEnabled('source', next),
+            (arns, next) => this._arnParser(arns, next),
+            (arnObj, next) => this._isValidRole('source', arnObj, next),
+            (arnObj, next) => this._isValidRole('target', arnObj, next),
+        ], cb);
+    }
+
+    _arnParser(arns, cb) {
+        const [src, des] = arns.split(',');
+
+        return cb(null, {
+            source: src,
+            target: des,
+        });
+    }
+
+    _isValidBucket(where, cb) {
+        // Does the bucket exist and is it reachable?
+        const bucket = where === 'source' ? this._sourceBucket :
+            this._targetBucket;
+        this._s3Clients[where].headBucket({ Bucket: bucket }, err => {
+            if (err) {
+                this._log.error('bucket sanity check error', {
+                    error: err.message,
+                    method: '_SetupReplication._isValidBucket',
+                    errorStack: err.stack,
+                });
+                return cb(err);
+            }
+            return cb();
+        });
+    }
+
+    _isVersioningEnabled(where, cb) {
+        // Does the bucket have versioning enabled?
+        const bucket = where === 'source' ? this._sourceBucket :
+            this._targetBucket;
+        this._s3Clients[where].getBucketVersioning({ Bucket: bucket },
+            (err, res) => {
+                if (err) {
+                    this._log.error('versioning sanity check error: ' +
+                        'Cannot retrieve versioning configuration', {
+                            error: err.message,
+                            method: '_SetupReplication._isVersioningEnabled',
+                            errorStack: err.stack,
+                        }
+                    );
+                    return cb(err);
+                }
+                if (res.Status === 'Disabled') {
+                    const error = new Error('Expected bucket versioning to ' +
+                        'be Enabled. Status is still Disabled.');
+                    this._log.error('versioning sanity check error: ' +
+                        'Status Disabled', {
+                            error: error.message,
+                            method: '_SetupReplication._isVersioningEnabled',
+                        }
+                    );
+                    return cb(error);
+                }
+                return cb();
+            }
+        );
+    }
+
+    _isValidRole(where, arnObj, cb) {
+        // Is the role mentioned in the replication config available in IAM
+
+        // Goal is to get Role given known ARN.
+        // If err, there is no matching role
+        const arn = arnObj[where];
+        const roleName = arn.split('/').pop();
+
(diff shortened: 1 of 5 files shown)


bin/setupReplication.js
```diff
@@ -0,0 +1,449 @@
...
+     * @param {String} targetBucket - Destination Bucket Name
+     * @param {Object} log - Werelogs Request Logger object
+     * @param {Object} config - bucket configurations
+     */
+    constructor(sourceBucket, targetBucket, log, config) {
+        const { source, destination } = config.extensions.replication;
+        this._log = log;
+        this._sourceBucket = sourceBucket;
+        this._targetBucket = targetBucket;
+        this._s3 = {
```
bennettbuchanan: Since this is for multiple s3 clients perhaps a name like `this._s3Clients` is suitable. That would also apply to `this._iam`.

bin/setupReplication.js
```diff
@@ -0,0 +1,449 @@
...
+                'backbeatsource'),
+            target: _setupIAMClient(destination.auth.vault.host,
+                 destination.auth.vault.iamPort,
+                'backbeattarget'),
+        };
+    }
+
+    _checkSanity(cb) {
+        return async.waterfall([
+            next => this._isValidBucket('source', (err, res) => {
```
bennettbuchanan: Minor: these can be reduced to `next => this._isValidBucket('source', next),`

bin/setupReplication.js
```diff
@@ -0,0 +1,449 @@
...
+            return cb(err, tgtArn);
+        });
+    }
+
+    _isReplicationEnabled(src, cb) {
+        // Is the Replication config enabled?
+        this._s3[src].getBucketReplication({ Bucket: this._sourceBucket },
+            (err, res) => {
+                const r = res.ReplicationConfiguration;
+                if (err || r.Rules[0].Status === 'Disabled') {
```
bennettbuchanan: I'm not sure I can think of a case where the status would have changed to 'Disabled', but then again I suppose this is meant as a sanity check.

bin/setupReplication.js
```diff
@@ -0,0 +1,449 @@
...
+                targetRole.RoleName, 'target', next),
+            next => this._enableReplication(roleArns, next),
+        ], (err, res) => {
+            cb(err, res);
+        });
+    }
+
+    run(cb) {
+        async.waterfall([
+            next => this._parallelTasks((err, setupInfo) => {
```
bennettbuchanan: Duplicate: Could just pass `next` here as the second arg.

bin/setupReplication.js
```diff
@@ -0,0 +1,449 @@
...
+
+    _parallelTasks(cb) {
+        async.parallel({
+            sourceBucket: next => this._createBucket('source', next),
+            targetBucket: next => this._createBucket('target', next),
+            sourceRole: next => this._createRole('source', next),
+            targetRole: next => this._createRole('target', next),
+            sourcePolicy: next => this._createPolicy('source', next),
+            targetPolicy: next => this._createPolicy('target', next),
+        }, (err, res) => {
```
bennettbuchanan: Could just pass `cb` here.

conf/config.json
```diff
@@ -36,15 +37,16 @@
             "destination": {
                 "s3": {
                     "host": "127.0.0.2",
-                    "port": 9000,
+                    "port": 8100,
```
bennettbuchanan: I guess the change since it conflicts with the source's bucketd port?

bin/setupReplication.js
philipyoo: Ya I agree, especially since the params for replication are nearly hard-coded in. I wasn't sure if there would be scenarios where this might ever be 'Disabled'. If it's cool, I'll remove it and only check on `err`?

bennettbuchanan: It definitely doesn't hurt, but it would only change to 'Disabled' if you put another replication config specifying that.

conf/config.json
philipyoo: Yup

bin/setupReplication.js
```diff
@@ -0,0 +1,449 @@
...
+
+commander
+  .version('0.1.0')
+  .arguments('<source> <destination>')
+  .action((source, destination) => {
+      const log = new Logger('BackbeatSetup').newRequestLogger();
+      const s = new _SetupReplication(source, destination, log, config);
+      s.run((err, res) => {
+          if (err) {
+              log.info(err);
```
bennettbuchanan: Shouldn't this be `log.error`? (Also, we would need only one here with the message and error object.)

bin/setupReplication.js
```diff
@@ -0,0 +1,449 @@
...
+              return log.info('replication script failed');
+          }
+          log.info(res);
+          return log.info('replication setup successful');
```
bennettbuchanan: Probably one `log.info` with the message `'replication script failed'` would suffice, and then include the `res` in the object following the message, as you've done in the `log.error` calls.

bin/setupReplication.js
philipyoo: Hmm, k sounds good. I'll keep it in then for now

bin/setupReplication.js
philipyoo: oops, I think i put that there for debugging haha. I think errors might be logging multiple times, so let me double check on the error logs and see. Thanks!

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+     * @param {Object} log - Werelogs Request Logger object
+     * @param {Object} config - bucket configurations
+     */
+    constructor(sourceBucket, targetBucket, log, config) {
+        const { source, destination } = config.extensions.replication;
+        this._log = log;
+        this._sourceBucket = sourceBucket;
+        this._targetBucket = targetBucket;
+        this._s3Clients = {
+            source: _setupS3Client(source.s3.host, source.s3.port,
```
jonathan-gramain: very minor: maybe use a separator for the profile name like `backbeat-source`

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+    _isValidBucket(where, cb) {
+        // Does the bucket exist and is it reachable?
+        const bucket = where === 'source' ? this._sourceBucket :
+            this._targetBucket;
+        this._s3Clients[where].headBucket({ Bucket: bucket }, err => {
+            if (err) {
+                this._log.error('bucket sanity check error', {
+                    error: err,
+                    method: '_SetupReplication._isValidBucket',
+                    errorMessage: err.message,
```
jonathan-gramain: I think by convention we log the error message in the `error` attribute itself. @rahulreddy is that correct?

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    _isVersioningEnabled(where, cb) {
+        // Does the bucket have versioning enabled?
+        const bucket = where === 'source' ? this._sourceBucket :
+            this._targetBucket;
+        this._s3Clients[where].getBucketVersioning({ Bucket: bucket },
+            (err, res) => {
+                if (err || res.Status === 'Disabled') {
+                    this._log.error('versioning sanity check error', {
```
jonathan-gramain: I think it's worth having two separate logs for both cases to help troubleshooting: error to retrieve versioning configuration or versioning disabled.

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+    _isValidRole(where, arn, cb) {
+        // Is the role mentioned in the replication config available in IAM
+
+        // Goal is to get Role given known ARN.
+        // If err, there is no matching role
+        let roleName;
+        let tgtArn;
+
+        if (where === 'source') {
+            [roleName, tgtArn] = arn.split(',');
```
jonathan-gramain: Rather than doing this parsing here, I would rather put it outside `_isValidRole`, in the main waterfall block after `_getBucketReplication`, and calling twice `_isValidRole` then with source and target parts of the comma-separated roles. I know this could spoil the nice layout of function calls in the waterfall :face_with_head_bandage: but it's better IMO to have `_isValidRole` always taking the same kind of `arn` argument which is a single role arn, rather than a different one if source or target.

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+        } else if (where === 'target') {
+            roleName = arn;
+        }
+        const namesOnly = roleName.split('/').pop();
+
+        this._iamClients[where].getRole({ RoleName: namesOnly }, (err, res) => {
+            if (err || roleName !== res.Role.Arn) {
+                this._log.error('role validation sanity check error', {
```
jonathan-gramain: similar comment than earlier, it's better to have separate logs for error case or mismatch case.

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    _isReplicationEnabled(src, cb) {
+        // Is the Replication config enabled?
+        this._s3Clients[src].getBucketReplication(
+            { Bucket: this._sourceBucket },
+            (err, res) => {
+                const r = res.ReplicationConfiguration;
+                if (err || r.Rules[0].Status === 'Disabled') {
+                    this._log.error('replication enabled sanity check error', {
```
jonathan-gramain: ditto

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+            }
+            this._log.debug('Bucket replication enabled', {
+                response: res,
+                method: '_enableReplication',
+            });
+            return cb(null, res);
+        });
+    }
+
+    _parallelTasks(cb) {
```
jonathan-gramain: This function may be better named with a more descriptive name like `_createEntities` which tells more about what it does (external info) than how it does it (internal info).

Alternatively it could also be clearer to put the function body in the main `run()` method, since all steps appearing in a single place are easier to follow (and since all of them are one-liner calls to sub methods it should not grow the `run` method beyond a reasonable size)

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+        return async.parallel({
+            sourceBucket: next => this._createBucket('source', next),
```
jonathan-gramain: very minor style comment: you could use an array of operations instead of an object here, and you could use `done` for callback names since calls are unordered.

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+            targetBucket: next => this._createBucket('target', next),
+            sourceRole: next => this._createRole('source', next),
+            targetRole: next => this._createRole('target', next),
+            sourcePolicy: next => this._createPolicy('source', next),
+            targetPolicy: next => this._createPolicy('target', next),
+        }, cb);
+    }
+
+    _seriesTasks(data, cb) {
```
jonathan-gramain: ditto about naming based on what the function does

bin/setupReplication.js
```diff
@@ -0,0 +1,431 @@
...
+                targetRole.RoleName, 'target', next),
+            next => this._enableReplication(roleArns, next),
+        ], cb);
+    }
+
+    run(cb) {
+        return async.waterfall([
+            next => this._parallelTasks(next),
+            (setupInfo, next) => this._seriesTasks(setupInfo, next),
+            (args, next) => this._checkSanity(next),
```
jonathan-gramain: I think the `_checkSanity` function is very useful, though I would think it fits better in an alternative script that checks an existing configuration than here, where the configuration has just been applied, and is supposedly correct because that's the script's job and it already checks for errors (if not correct, it would mean the script is buggy, not that the administrator has failed to configure replication correctly).

We could if we want to run the sanity check script just after this one, but I don't think it's really necessary. @rahulreddy may have another opinion on this.

rahulreddy: I agree, it makes sense to separate the sanity check script into another command.
This is how I am envisioning this.
```
node bin/replication setup <source bucket> <destination bucket>

node bin/replication check-sanity <source bucket> <destination bucket>
```

bin/setupReplication.js
philipyoo: For error logging, I followed https://github.com/scality/S3/pull/845#discussion_r131225448
Not too sure which way to do this, going to wait on this one until we get a more definitive answer

bin/setupReplication.js
philipyoo: I wasn't sure how to pass targetArn after using `_isValidRole` for sourceArn. Instead, I used an object with keys "source" and "target". I will push up with changes soon so you can see

bin/setupReplication.js
philipyoo: Nice! So I tried moving everything into `run()`. Let me know what you think

bin/setupReplication.js
philipyoo: I think using an object here would be better because when calling `async.series`, I'm using the key names for accessing corresponding data. I think it's more descriptive this way vs accessing these from an array

lib/config/configItems.joi.js
```diff
@@ -5,6 +5,7 @@ const joi = require('joi');
 const hostPortJoi = joi.object({
     host: joi.string().required(),
     port: joi.number().greater(0).required(),
+    iamPort: joi.number().greater(0).optional(),
```
jonathan-gramain: A better place to put this would be by appending `.keys({ iamPort: ...})` to the `hostPortJoi` where it's used, this one shall remain generic.

bin/setupReplication.js
rahulreddy: This is the format I recommend since ElasticSearch indexes err as string
```
error: err.message,
errStack: err.stack
```

PR #36, opened 2017-08-04
fix: g++ version for circleci

philipyoo:


DavidPineauScality:
Looking at the build results, if you look at the NPM install section, it was actually failing since the requested command does not exist: `G++-4.9`: http://ci.ironmann.io/gh/scality/backbeat/476

It should actually be `g++-4.9`

alexandre-merle:
to avoid that in the future, you may want to put ```rm -rf node_modules``` before the npm install :)

PR #37, opened 2017-08-04
FIX: Set log consumer for BucketFileLogReader

bennettbuchanan:
Depends on https://github.com/scality/backbeat/pull/37.


rahulreddy:
@ironman-machine r+

PR #38, opened 2017-08-04
fix: update config to reflect production

rahulreddy:


extensions/replication/queuePopulator/QueuePopulator.js
```diff
@@ -146,7 +146,7 @@ class QueuePopulator {
 
     _setupProducer(done) {
         const producer = new BackbeatProducer({
-            zookeeper: this.zkConfig,
+            zookeeper: this.zkConfig.connectionString,
```
jonathan-gramain: Here you still want to pass `this.zkConfig`

rahulreddy: Good catch!

PR #39, opened 2017-08-07
ft: changes to support bootstrap list for proxy

rahulreddy:


rahulreddy:
Opening this PR early so that everyone can see the incoming changes to `config.json` and comment early on to propose any changes.

conf/config.json
```diff
@@ -40,31 +32,16 @@
...
-                        "hosts": [
-                            {
-                                "host": "127.0.0.2",
-                                "port": 9500
-                            }
-                        ]
-                    }
-                },
+                "transport": "https",
+                "bootstrapList": ["127.0.0.1:8443"],
```
jonathan-gramain: This looks ok for enterprise, but we may want to keep the possibility to configure S3 and vault separately, especially for open source zenko which users may not want to install nginx as a requirement?

rahulreddy: I thought about that and realized that for opensource we don't have vault, as long as we pass the destination endpoint as an array this will just work.

PR #41, opened 2017-08-08
fix: g++ syntax for circleci

philipyoo:
Oops, please see following comment for reference:
https://github.com/scality/backbeat/pull/36#issuecomment-320962020

Also following @alexandre-merle advice, removing node_modules and reinstalling

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,7 +9,11 @@ machine:
     services:
         - docker
     environment:
-        CXX: G++-4.9
+        CXX: g++-4.9
+
+dependencies:
+    pre:
+        - rm -rf node_modules
 
 test:
     override:


circle.yml
```diff
@@ -9,7 +9,12 @@ machine:
     services:
         - docker
     environment:
-        CXX: G++-4.9
+        CXX: g++-4.9
+
+dependencies:
+    override:
```
DavidPineauScality: you could use "pre" instead, right ?

philipyoo: Cool, moved only removal of node_modules to "pre". Also realized dependencies build had a few env var exports, so ya, shouldn't use "override"

philipyoo:
I'm not too familiar with CircleCI and how the build environments are setup, but I noticed that tests might be failing because Ubuntu versions differ. Looking at other builds, I see `CIRCLE_BUILD_IMAGE=ubuntu-12.04`

Looking at this failing build: `CIRCLE_BUILD_IMAGE=ubuntu-14.04`

Could this be a reason why `g++-4.9` is a package that is not found? Environment is somehow different?

DavidPineauScality:
@philipyoo I took a look at the public CI's failure.
As I understand, the override we do with environment.CXX=g++-4.9 is only changing the command, but it never installs the package of the compiler itself.
As you noticed, the difference of Ubuntu version might be at play here, but the public circleCI allows choosing which of their supported OSes to use for testing.

As such  I see two possibilities:
1. Explicitly install the right G++ package
2. Select the same OS as the one used in our private 

Now, whether this will work or not, I cannot promise :/

philipyoo:
@DavidPineauScality I tried out your second recommendation, using same OS, and the build passed. Waiting to test out with another PR :)

rahulreddy:
Looks like the public CI is doing fine now. We don't need these changes anymore?

PR #44, opened 2017-08-09
target hosts failover

jonathan-gramain:
use bootstrapList with RoundRobin to select hosts

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -10,6 +10,7 @@ const Logger = require('werelogs').Logger;
 
 const errors = require('arsenal').errors;
 const jsutil = require('arsenal').jsutil;
+const RoundRobin = require('arsenal').network.RoundRobin;
 const VaultClient = require('vaultclient').Client;
 const { proxyPath } = require('../constants');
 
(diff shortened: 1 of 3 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -448,51 +509,56 @@ class QueueProcessor {
...
+            this._createAuthManager('source', this.sourceConfig.auth,
+                                    sourceRole, log);
 
         // Disable retries, use our own retry policy (mandatory for
         // putData route in order to fetch data again from source).
 
+        const sourceS3 = this.sourceConfig.s3;
         this.S3source = new AWS.S3({
-            endpoint: `${this.sourceConfig.transport}://` +
+            endpoint: `${sourceS3.transport}://` +
```
philipyoo: Tests are using it based on `transport` defined in `sourceS3`, so maybe move `transport` on `config.json` within `source`?

rahulreddy: You are right, we made these config changes in an earlier PR, the changes may have been reverted during conflict resolution.
@jonathan-gramain Can you rever the changes for transport config

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -448,51 +509,56 @@ class QueueProcessor {
...
                 `${sourceS3.host}:${sourceS3.port}`,
-            credentials: this.s3sourceAuthManager.getCredentials(),
-            sslEnabled: this.sourceConfig.transport === 'https',
+            credentials:
+            this.s3sourceAuthManager.getCredentials(),
+            sslEnabled: true,
```
rahulreddy: We need to handle this based on config.

PR #45, opened 2017-08-09
ft: support new bootstraplist format

rahulreddy:


rahulreddy:
PR #44 will use introduce the changes to use all hosts in the bootstrap list.

lib/config/configItems.joi.js
```diff
@@ -7,7 +7,12 @@ const hostPortJoi = joi.object({
     port: joi.number().greater(0).required(),
 });
 
-const bootstrapListJoi = joi.array().min(1).items(joi.string());
+const bootstrapListJoi = joi.array().min(1).items(
+    joi.object({
+        site: joi.string().required(),
+        servers: joi.array().min(1).items(joi.string()),
```
bennettbuchanan: Maybe it is overkill, but we could add the regex validating the string is of the form "host:port", similar to zookeeperJoi validation.

rahulreddy: Yeah, I am not sure how to do that in joi yet. We can address this later.

PR #54, opened 2017-08-10
bf: handle 404 error from source gracefully

jonathan-gramain:
don't update source metadata, just log and skip the entry

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -378,6 +378,9 @@ class QueueProcessor {
             PartNumber: partNumber,
         });
         req.on('error', err => {
+            if (err.ObjNotFound || err.code === 'ObjNotFound') {
+                return doneOnce(errors.ObjNotFound);
+            }
             // eslint-disable-next-line no-param-reassign
             err.origin = 'source';
             log.error('an error occurred on getObject from S3',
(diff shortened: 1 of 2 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -539,6 +545,12 @@ class QueueProcessor {
                             error: err.description });
                 return done();
             }
+            if (err.ObjNotFound) {
+                log.info('replication skipped: ' +
```
philipyoo: Minor, but maybe use `log.error`?

jonathan-gramain: It's actually not an error, just a singular case of not replicating anything. (the user can create and remove an object version on the source before the queue populator has a chance to look at it, it's a normal situation).

jonathan-gramain: I meant, before the queue processor processes it

jonathan-gramain:
@rahulreddy the PENDING status is written on each version, so if the version has been deleted, it used to have the PENDING status but not anymore?

On access denied, we just skip because we don't have the access rights to update the replication status (unless access denied is only for reading but not for writing, which would be weird).

jonathan-gramain:
Or you're thinking about the master version replication status?

rahulreddy:
> the PENDING status is written on each version, so if the version has been deleted, it used to have the PENDING status but not anymore?
On access denied, we just skip because we don't have the access rights to update the replication status (unless access denied is only for reading but not for writing, which would be weird).

I understand better now, if we cannot write to the versioned copy we skip.

PR #61, opened 2017-08-15
bf: no reset of '_ready' flag on producer error

jonathan-gramain:
Since errors may occur at any time, especially at startup when kafka
is not yet listening, we should not reset the _ready flag which is
initially set to true when zookeeper is up and running on port 2181
and could return kafka's configuration with success, even though port
9092 is not yet ready.

We observed a backtrace saying 'producer is not ready yet' in a loop
on the queue populator, which required restarting the container, we
suspect that this is the cause but haven't confirmed yet. In any case
this fixes a real bug that may happen in other cases of errors.


rahulreddy:
@ironman-machine r+

PR #62, opened 2017-08-16
Bf/edit setup script

philipyoo:


bin/setupReplication.js
```diff
@@ -100,12 +117,24 @@ class _SetupReplication {
         this._sourceBucket = sourceBucket;
         this._targetBucket = targetBucket;
         // TODO: support bootstrap list failover
-        const destEndpoint = destination.bootstrapList[0].servers[0];
+        const destEndpoint = '217.182.11.96:8000';
```
philipyoo: When building Federation client-template, I did not set `destination.bootstrapList` so I hard-coded this value in for testing. I think this is set in `env/client-template/group_vars/all` using `env_replication_endpoints`?

bin/setupReplication.js
```diff
@@ -100,12 +117,24 @@ class _SetupReplication {
...
+
+	console.log('sourceBucket:', this._sourceBucket);
+	console.log('targetBucket:', this._targetBucket);
+	console.log(source.s3);
+	console.log('source.auth.vault:', source.auth.vault);
+	console.log('destEndpoint', destEndpoint);
+	console.log('verifySource:', verifySourceProfile);
+	console.log('verifyTarget:', verifyTargetProfile);
+	source.auth.vault.host = 'localhost';
+	source.s3.host = 'localhost';
```
philipyoo: These were set as the server name, i.e. `philz-07`. I manually changed to `localhost` for testing

bin/setupReplication.js
```diff
@@ -62,23 +66,36 @@ function _setupS3Client(transport, endpoint, profile) {
         endpoint: `${transport}://${endpoint}`,
         sslEnabled: transport === 'https',
         credentials,
+        region: 'us-east-1',
```
philipyoo: Changed to `us-east-1`. Saw `us-east-2` was not defined in `/conf/locationConfig.json` in S3 server

philipyoo:
Only thing online I could find about using `httpOptions.proxy` was here: https://aws.amazon.com/releasenotes/3118686131819314

Looks like a full url, including transport is used: `proxy: 'http://localhost:8080'`

rahulreddy:
Let's address this as a two-step fix.
1. Hardcode the destination IAM port to 8600 and don't use any proxy path. This was we can get the script working.
2. Use aws4 module to build our own signature so the request goes to the proxy server at the proxy path but the signer would use / as path instead.

bin/replication.js
```diff
@@ -260,18 +273,28 @@ class _SetupReplication {
         const bucket = where === 'source' ? this._sourceBucket :
             this._targetBucket;
         this._s3Clients[where].createBucket({ Bucket: bucket }, (err, res) => {
-            if (err) {
+            if (err && err.code !== 'BucketAlreadyOwnedByYou') {
                 this._log.error('error creating a bucket', {
                     error: err.message,
+                    errCode: err.code,
```
bennettbuchanan: We could probably go ahead and log the entire error instead of just the error code.

bin/replication.js
```diff
@@ -403,7 +429,7 @@ class _SetupReplication {
 
     setupReplication(cb) {
         return async.waterfall([
-            next => async.parallel({
```
bennettbuchanan: Minor: seems like `parallel` would be fine here.

bin/replication.js
```diff
@@ -260,18 +273,28 @@ class _SetupReplication {
...
                 });
                 return cb(err);
             }
-            this._log.debug('Created bucket', {
-                bucket: where,
-                response: res,
-                method: '_createBucket',
-            });
+
+            if (err && err.code === 'BucketAlreadyOwnedByYou') {
```
bennettbuchanan: Good idea 👍

bin/replication.js
```diff
@@ -2,10 +2,14 @@ const async = require('async');
 const program = require('commander');
 const { S3, IAM, SharedIniFileCredentials } = require('aws-sdk');
 
-const { Logger } = require('werelogs');
-
+const werelogs = require('werelogs');
+const { RoundRobin } = require('arsenal').network;
+const Logger = werelogs.Logger;
```
bennettbuchanan: Minor: You might move this below the requires. Alternatively, couldn't we keep `const { Logger } = require('werelogs');`?

bin/replication.js
rahulreddy: This change was intentional, the idea was that we want to stop immediately if something fails and it translates to a smaller headache while debugging.

bin/replication.js
rahulreddy: It was a lot of text, for a command line client this should be sufficient.

bin/replication.js
```diff
@@ -99,25 +109,28 @@ class _SetupReplication {
...
-
+        source.auth.vault.host = 'localhost';
+        source.s3.host = 'localhost';
+        const destHost = this.destHosts.pickHost().host;
         this._s3Clients = {
             source: _setupS3Client(source.transport,
                 `${source.s3.host}:${source.s3.port}`,
                 verifySourceProfile),
-            target: _setupS3Client(destination.transport, destEndpoint,
+            target: _setupS3Client(destination.transport, destHost,
```
rahulreddy: So the idea is that for destination it is sending request directly to the s3-nginx which runs on port 80. I will do a followup PR later where we could calculate the signature using a proxy path.

bin/replication.js
```diff
@@ -59,26 +63,32 @@ function _buildResourcePolicy(source, target) {
 function _setupS3Client(transport, endpoint, profile) {
     const credentials = new SharedIniFileCredentials({ profile });
     return new S3({
-        endpoint: `${transport}://${endpoint}`,
+        endpoint: `${endpoint}`,
```
rahulreddy: transport will be handled in the httpAgent setup.

bin/replication.js
rahulreddy: It's because I wanted to configure Werelogs (the new way) log levels.
```
werelogs.configure({
    level: config.log.logLevel,
     dump: config.log.dumpLevel,
 });
```

bin/replication.js
```diff
@@ -99,25 +109,28 @@ class _SetupReplication {
         this._log = log;
         this._sourceBucket = sourceBucket;
         this._targetBucket = targetBucket;
-        // TODO: support bootstrap list failover
-        const destEndpoint = destination.bootstrapList[0].servers[0];
+        this.destEndpoint =
```
rahulreddy: Ah yes, good catch!

rahulreddy:
@ironman-machine r+

PR #63, opened 2017-08-16
ft: increase fetchMaxBytes for replication consumer

rahulreddy:


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -22,6 +22,7 @@ const QueueEntry = require('../utils/QueueEntry');
 const CredentialsManager = require('../../../credentials/CredentialsManager');
 
 const MPU_CONC_LIMIT = 10;
+const CONSUMER_FETCH_MAX_BYTES = 5000020; // a little over 5MB
```
jonathan-gramain: Could you explain why we need a little over 5MB? I think it deserves a comment especially since 5MB looks like a magic value here.

rahulreddy: I will add a comment. Basically it comes down to the fact the the biggest object json from S3 would be 1.6 MB, so we padded it and made up the Kafka's max message limit size that 5MB. For the consumer to read from Kafka, say for example an entry of 5MB, the fetchMaxBytes has to be >= 5MB.

PR #64, opened 2017-08-16
FT: Support MPUs when part copied from MPU

bennettbuchanan:


rahulreddy:
@electrachong Good idea. I think an Integration test would also help.

bennettbuchanan:
Thanks for the feedback. I will add an Integration test for this case since it does depend on location objMD that is originally built in S3. 

Also, I believe we can assume that the locations MD of an MPU will be sequentially ordered by part number. I also think it is possible that S3 could be the one to build a location obj MD that suits our needs. At the moment, though, I am not sure how these location elements are being used, or if they can be modified. Such a change could help with optimization down the line.

bennettbuchanan:
@ironman-machine try

rahulreddy:
I don't think Backbeat is enabled on ultron. Let me do that change.

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

bennettbuchanan:
E2E passed with master, and Integration tests using uploadCopyPart.
* http://ci.ironmann.io/gh/scality/Integration/14443
* http://ci.ironmann.io/gh/scality/Integration/14439

I've now pushed unit tests but have not altered the lib code.

bennettbuchanan:
@ironman-machine 
try

rahulreddy:
@ironman-machine r+

PR #65, opened 2017-08-17
FT: Use key and dataStoreName returned from putData

bennettbuchanan:
Depends on https://github.com/scality/S3/pull/874.


bennettbuchanan:
E2E passed with dependent branch: http://ci.ironmann.io/gh/scality/Integration/14486

bennettbuchanan:
@ironman-machine try

rahulreddy:
@ironman-machine try

PR #67, opened 2017-08-18
cleanup: move cli script to separate file

philipyoo:
Moving command line script to another file in anticipation for building a UI portion.
Added a way for a user to define location of an AWS credentials file. User must use `AWS_SHARED_CREDENTIALS_FILE`, or we can change this to something shorter as well

### DIFF
--- a/bin/replication.js
+++ b/bin/replication.js
@@ -1,15 +1,12 @@
 const async = require('async');
-const program = require('commander');
+const os = require('os');
 const { S3, IAM, SharedIniFileCredentials } = require('aws-sdk');
 
-const werelogs = require('werelogs');
 const { RoundRobin } = require('arsenal').network;
-const Logger = werelogs.Logger;
-const config = require('../conf/Config');
-werelogs.configure({
-    level: config.log.logLevel,
-    dump: config.log.dumpLevel,
-});
+
+// User can define where to find credentials file. Relative Path is used.
+const AWS_CRED_LOC = process.env.AWS_SHARED_CREDENTIALS_FILE ||
+    `${os.homedir()}/.aws/credentials`;
 
 const trustPolicy = {
     Version: '2012-10-17',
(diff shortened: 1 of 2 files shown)


philipyoo:
@bennettbuchanan awesome thank you! Added edits :)

rahulreddy:
This is good. As Bennett mentioned, please make sure `AWS_SHARED_CREDENTIALS_FILE` is optional but it gets first priority if it is defined. We will fallback to the `~/.aws/credentials` if the env var is not defined.

bin/replication.js
```diff
@@ -61,7 +54,10 @@ function _buildResourcePolicy(source, target) {
 }
 
 function _setupS3Client(transport, endpoint, profile) {
-    const credentials = new SharedIniFileCredentials({ profile });
+    const filename = process.env.AWS_SHARED_CREDENTIALS_FILE ||
+        `${os.homedir()}/.aws/credentials`;
+    const credentials = new SharedIniFileCredentials({ profile, filename });
```
jonathan-gramain: I think you can make `credentials` a global variable, that would avoid reconstructing it for different clients. Globals are usually fine to expose global configuration values.

bin/replication.js
```diff
@@ -115,8 +114,10 @@ class _SetupReplication {
             'default' : sourceProfile;
         const verifyTargetProfile = targetProfile === undefined ?
             'default' : targetProfile;
-        source.auth.vault.host = 'localhost';
-        source.s3.host = 'localhost';
+
+        // TODO: Federation fix made #1125?
+        // source.auth.vault.host = 'localhost';
+        // source.s3.host = 'localhost';
         const destHost = this.destHosts.pickHost().host;
```
jonathan-gramain: If I'm not mistaken, you should pick both `host` and `port` from the result of `pickHost`, then construct the endpoint string as `host:port`.

philipyoo: For this, Rahul wanted to use s3-nginx port 80 https://github.com/scality/backbeat/pull/62#discussion_r133603964

bin/replication.js
philipyoo: Thanks! Ya that sounds much better :) Will make this change and rebase soon

bin/replication.js
rahulreddy: Yes, for now we are hardcoding the destination S3/IAM ports.

jonathan-gramain: OK, I understand, I think it deserves a comment explaining the reason though, and/or a FIXME for later.

PR #68, opened 2017-08-18
FT: Add backbeat local quickstart guide

bennettbuchanan:
This guide is intended to help users get started with using backbeat's replication extension. It uses Vault (a private repository), so it is primarily targeted for internal usage at Scality. However, getting kafka and zookeeper running and configured would allow users to quickly develop by reusing backbeat's consumer and producer classes. 

From here, we might consider adding guides for running zookeeper and kafka with docker (with spotfy/kafka, and evenutally Scality's container). Also, a guide for other OS is in order. 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,3 +28,236 @@ updates in a FIFO order.
 ## DESIGN
 
 Please refer to the ****[Design document](/DESIGN.md)****
+
+## QUICKSTART
+
+This guide assumes the following:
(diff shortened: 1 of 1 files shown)


rahulreddy:
Awesome 🎉

rahulreddy:
This can go in the README directly, instead of having to link from there.

PR #69, opened 2017-08-21
FIX: Use empty array for zero byte objects

bennettbuchanan:


bennettbuchanan:
@ironman-machine try

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

PR #70, opened 2017-08-22
bf: don't crash if dataStoreETag missing

jonathan-gramain:
Restore graceful handling when an entry's location info does not have the
dataStoreETag property.

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -423,10 +423,18 @@ class QueueProcessor {
     // Object metadata may contain multiple elements for a single part if
     // the part was originally copied from another MPU. Here we reduce the
     // locations array to a single element for each part.
-    _getLocations(sourceEntry) {
+    _getLocations(sourceEntry, log) {
         const locations = sourceEntry.getLocation();
         const reducedLocations = [];
         let partTotal = 0;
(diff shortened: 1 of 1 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -424,10 +424,18 @@ class QueueProcessor {
     // Object metadata may contain multiple elements for a single part if
     // the part was originally copied from another MPU. Here we reduce the
     // locations array to a single element for each part.
-    _getLocations(sourceEntry) {
+    _getLocations(sourceEntry, log) {
         const locations = sourceEntry.getLocation();
         const reducedLocations = [];
         let partTotal = 0;
+        if (locations.some(
```
bennettbuchanan: Could we add this check to the for loop below? That way we avoid looping through a second time.

jonathan-gramain: It would add complexity because we will need to check both the current and next part then, because we're calling `getPartNumber()` for each of them, which requires the `dataStoreETag` to be set. I think it's simpler to just do a loop first for this reason, I don't think it will cost any significant additional CPU, because when there are many parts, it's a big object so it will take a lot of time anyways.

rahulreddy:
@ironman-machine r+

PR #74, opened 2017-08-22
FIX: Encode and decode object keys

bennettbuchanan:


extensions/replication/utils/QueueEntry.js
```diff
@@ -57,11 +57,10 @@ class QueueEntry {
     static createFromKafkaEntry(kafkaEntry) {
         try {
             const record = JSON.parse(kafkaEntry.value);
+            const key = QueueEntry._extractVersionedBaseKey(record.key);
```
jonathan-gramain: I think this line and the following one should be reversed, otherwise the extract call will return the key containing the version id since the version delimiter has been encoded by `encodeURIComponent`.

Could you add a functional test, making sure you can write such utf8 key in kafka and read it back correctly? There are tests using kafka already for BackbeatProducer and BackbeatConsumer.

jonathan-gramain:
Nevermind about the test, I realized you did it in integration.

rahulreddy:
@ironman-machine r+

PR #79, opened 2017-08-26
fix: disable timeout on AWS SDK to avoid socket hangups

rahulreddy:


rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

PR #81, opened 2017-08-29
bf: Fix required fields for script command options

philipyoo:


philipyoo:
Not a bf, but more a doc fix. Making changes soon

PR #84, opened 2017-08-31
ft: auto-create zookeeper base path for zenko

jonathan-gramain:
Auto-create queue populator base path on zookeeper if it does not
exist, when starting up the queue populator deployed previously with
docker swarm for zenko.

Since docker and docker swarm do not provide an easy way to run
specific commands at deployment time, especially conditioned to
containers being up and running first, in order to ensure zookeeper
base paths are created initially, we wrap the zookeeper client handle
creation with some custom code to do this.

This is enabled only in zenko mode with docker swarm deployment,
default options (including Federation-based deployment) do not change
the existing behavior which is to not create the base path if it does
not exist (and then further zookeeper commands shall fail with NO_NODE
errors).

### DIFF
--- a/lib/clients/zookeeper.js
+++ b/lib/clients/zookeeper.js
@@ -1,0 +1,57 @@
+const zookeeper = require('node-zookeeper-client');
+
+/**
+ * wrapper around createClient() from node-zookeeper-client module,
+ * with the following enhancements:
+ *
+ *  - emits a 'ready' event when the zookeeper client is ready
+ *  - takes an additional option "autoCreateNamespace"
+ *
+ * @param {string} connectionString - connection string to zookeeper
+ *   (e.g. 'localhost:2181/backbeat')
+ * @param {object} [options] - an object to set the client
+ *   options. Currently available options are:
+ * @param {number} [options.sessionTimeout] - session timeout in
+ *   milliseconds, defaults to 30 seconds.
+ * @param {number} [options.spinDelay] - the delay (in milliseconds)
+ *   between each connection attempts.
+ * @param {number} [options.retries] - the number of retry attempts
+ *   for connection loss exception.
+ * @param {boolean} [options.autoCreateNamespace] - when true, ensure
+ *   namespace (chroot) base path is created if not exists
+ *
+ * @return {node-zookeeper-client.Client} a zookeeper client handle
+ */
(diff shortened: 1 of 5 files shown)


lib/clients/zookeeper.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {number} [options.spinDelay] - the delay (in milliseconds)
+ *   between each connection attempts.
+ * @param {number} [options.retries] - the number of retry attempts
+ *   for connection loss exception.
+ * @param {boolean} [options.autoCreateNamespace] - when true, ensure
+ *   namespace (chroot) base path is created if not exists
+ *
+ * @return {node-zookeeper-client.Client} a zookeeper client handle
+ */
+function createClient(connectionString, options) {
```
rahulreddy: Two things. I recommend handling error conditions first. I recommend emitting unknown errors as an error event. Otherwise the invoker of the function will be in a perpetual wait state waiting for the `ready` event.
```js
function createClient(connectionString, options) {
    const zkClient = zookeeper.createClient(connectionString, options);
    zkClient.once('connected', () => {
        // for some reason zkClient.exists() does not return
        // NO_NODE when base path does not exist, hence use
        // getData() instead
        zkClient.getData('/', err => {
            if (err && err.name !== 'NO_NODE') {
                return zkClient.emit('error', err);
            }
            // NO_NODE error and autoCreateNamespace is enabled
            if (err && options && options.autoCreateNamespace) {
                const nsIndex = connectionString.indexOf('/');
                const hostPort = connectionString.slice(0, nsIndex);
                const namespace = connectionString.slice(nsIndex);
                const rootZkClient = zookeeper.createClient(hostPort, options);
                rootZkClient.connect();
                return rootZkClient.mkdirp(namespace, err => {
                    if (err && err.name !== 'NODE_EXISTS') {
                        return zkClient.emit('error');
                    }
                    return zkClient.emit('ready');
                });
            }
            return zkClient.emit('ready');
        });
    });
    return zkClient;
}
```

jonathan-gramain: Done, thanks for the snippet of code :)

jonathan-gramain:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #85, opened 2017-09-06
ft: skip entry if target site not configured

jonathan-gramain:
Do not process entries and log a warning if no target site is
configured (no bootstrap list), instead of aborting with an
exception.

Also set the replication status to FAILED.


rahulreddy:
@ironman-machine r+

PR #86, opened 2017-09-06
use default port 80 in round-robin

jonathan-gramain:
Use port 80 when not specified in bootstrap list items for backbeat target.

Depends on Arsenal PR https://github.com/scality/Arsenal/pull/332


jonathan-gramain:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #88, opened 2017-09-08
add 'npm start' rule to run populator and processor

jonathan-gramain:
This to not have to use supervisord to run both queue populator and queue processor in the same container.


jonathan-gramain:
The current idea is instead to run the queue populator in a separate container from the other services (replication processor, clueso etc.) so it will make the 'npm start' command unnecessary.

PR #89, opened 2017-09-09
ft: build files for zenko deployment

jonathan-gramain:
Add necessary files to build and deploy a zenko backbeat container.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -22,9 +22,10 @@ if [[ "$LOG_LEVEL" ]]; then
     fi
 fi
 
-if [[ $JQ_FILTERS_CONFIG != "." ]]; then
-    jq "$JQ_FILTERS_CONFIG" conf/config.json > conf/config.json.tmp
-    mv conf/config.json.tmp conf/config.json
+if [[ ! "$BACKBEAT_CONFIG_FILE" ]]; then
+    BACKBEAT_CONFIG_FILE=conf/config.json
 fi
+jq "$JQ_FILTERS_CONFIG" "$BACKBEAT_CONFIG_FILE" > /tmp/config.json.tmp
+mv /tmp/config.json.tmp "$BACKBEAT_CONFIG_FILE"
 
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -0,0 +1,30 @@
...
+if [[ "$LOG_LEVEL" ]]; then
+    if [[ "$LOG_LEVEL" == "info" || "$LOG_LEVEL" == "debug" || "$LOG_LEVEL" == "trace" ]]; then
+        JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .log.logLevel=\"$LOG_LEVEL\""
+        echo "Log level has been modified to $LOG_LEVEL"
+    else
+        echo "The log level you provided is incorrect (info/debug/trace)"
+    fi
+fi
+
+if [[ $JQ_FILTERS_CONFIG != "." ]]; then
```
philipyoo: Was thinking we don't really need the if conditional check here since the variable is set at the start with values

docker-entrypoint.sh
```diff
@@ -0,0 +1,30 @@
...
+    jq "$JQ_FILTERS_CONFIG" conf/config.json > conf/config.json.tmp
```
philipyoo: Should we extend this to use `BACKBEAT_CONFIG_FILE` if set?

docker-entrypoint.sh
jonathan-gramain: right, will remove it

docker-entrypoint.sh
jonathan-gramain: I think we should indeed, since the docker stack deployment may have this env var set.

jonathan-gramain:
Note: when https://github.com/scality/backbeat/pull/98 gets merged it will require a change in jq rules 
 in docker-entrypoint.sh (dmd).

jonathan-gramain:
Nothing of value anymore here, replaced by forward/orbit implementation, closing then.

PR #90, opened 2017-09-11
ft: increase concurrency of Queue Processor

rahulreddy:
This commit increases the concurrency of the queue processor to process multiple entries (currently set at 10) at once. It's possible that the requests arrive at the destination out of order but that should be ok since Versioning algorithm supports out of order updates. One more side effect of this change is that Consumer offsets are committed per batch, which means if we fail processing an entry in a batch we may re-process the batch. This shifts the expectation from EXACTLY ONCE to AT LEAST ONCE delivery which should be ok in our case since metadata updates are immutable. However data orphans may be an issue which can be handled by a different strategy.

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -799,8 +799,7 @@ class QueueProcessor {
             zookeeper: this.zkConfig,
             topic: this.repConfig.topic,
             groupId: this.repConfig.queueProcessor.groupId,
-            concurrency: 1, // replication has to process entries in
-                            // order, so one at a time
+            concurrency: 10, // versioning can support out of order updates
             queueProcessor: this.processKafkaEntry.bind(this),
             fetchMaxBytes: CONSUMER_FETCH_MAX_BYTES,
         });


rahulreddy:
@ironman-machine try

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -799,8 +799,7 @@ class QueueProcessor {
             zookeeper: this.zkConfig,
             topic: this.repConfig.topic,
             groupId: this.repConfig.queueProcessor.groupId,
-            concurrency: 1, // replication has to process entries in
-                            // order, so one at a time
+            concurrency: 10, // versioning can support out of order updates
```
jonathan-gramain: You will also need to spawn separate contexts for every parallel processing "thread". Look for `// per-request state variables` comment in the constructor, normally the member variables following are to be duplicated per replicated entry. The cleanest way would likely be to create a new class containing those variables and the methods around processing a single entry, then the QueueProcessor class would become the orchestrator only.

Also please double-check that we're not getting into trouble with version-specific operations (I can think of put-object-tagging, or put-object-acl, which if done out-of-order may lead to inconsistent replication).

Otherwise once we sort out details I'm all for this improvement, as it should allow much fewer partitions and processes running for the same throughput performance (so less memory consumed, CPU overhead etc.) and simpler config since we will care less about adjusting the number of partitions to the workload.

rahulreddy: Ah yes, I now remember the conversation we had earlier. Thanks for pointing it out. We should isolate the state for each entry while sharing the same client instances (may be).

jonathan-gramain:
Replaced by https://github.com/scality/backbeat/pull/93

PR #91, opened 2017-09-12
Chore: upgrade eslint

philipyoo:
I tried to separate my commits by each type of eslint change, so it might be easiest to view by each commit.

Let me know what you think, if we should add or remove certain rules. Current rules can be viewed:
https://github.com/scality/Guidelines/blob/chore/upgradeEslint/index.js

List of Eslint rules: https://eslint.org/docs/rules/

Note: Indent fixes were mainly done by the linter --fix command

Note: Will squash commits after making any changes based on peoples' feedback

### DIFF
--- a/extensions/replication/queuePopulator/BucketFileLogReader.js
+++ b/extensions/replication/queuePopulator/BucketFileLogReader.js
@@ -7,7 +7,7 @@ class BucketFileLogReader extends LogReader {
     constructor(params) {
         const { zkClient, kafkaProducer, dmdConfig, logger } = params;
         super({ zkClient, kafkaProducer, logConsumer: null,
-                logId: `bucketFile_${dmdConfig.logName}`, logger });
+            logId: `bucketFile_${dmdConfig.logName}`, logger });
 
         this._dmdConfig = dmdConfig;
         this._log = logger;
--- a/extensions/replication/queuePopulator/LogReader.js
+++ b/extensions/replication/queuePopulator/LogReader.js
@@ -44,20 +44,20 @@ class LogReader {
      */
     setup(done) {
         this.log.debug('setting up log source',
-                       { method: 'LogReader.setup',
-                         logSource: this.getLogInfo() });
+            { method: 'LogReader.setup',
+                logSource: this.getLogInfo() });
         this._readLogOffset((err, offset) => {
             if (err) {
                 this.log.error('log source setup failed',
-                               { method: 'LogReader.setup',
-                                 logSource: this.getLogInfo() });
+                    { method: 'LogReader.setup',
+                        logSource: this.getLogInfo() });
                 return done(err);
             }
             this.logOffset = offset;
             this.log.info('log reader is ready to populate replication ' +
                           'queue',
-                          { logSource: this.getLogInfo(),
-                            logOffset: this.logOffset });
+                { logSource: this.getLogInfo(),
+                    logOffset: this.logOffset });
             return done();
         });
     }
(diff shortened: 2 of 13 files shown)


extensions/replication/queuePopulator/BucketFileLogReader.js
```diff
@@ -7,7 +7,7 @@ class BucketFileLogReader extends LogReader {
     constructor(params) {
         const { zkClient, kafkaProducer, dmdConfig, logger } = params;
         super({ zkClient, kafkaProducer, logConsumer: null,
-                logId: `bucketFile_${dmdConfig.logName}`, logger });
+            logId: `bucketFile_${dmdConfig.logName}`, logger });
```
jonathan-gramain: I think we should disable the linter rule leading to this indent change, although I'm not sure what it's complaining about I don't see a value here and it is going to cause conflict resolution troubles (and possibly more annoyance depending on which editor people use).

philipyoo: Slight problems :/
I made a post here:
https://github.com/scality/Guidelines/pull/59#issuecomment-328968923

New lint version wants specific changes, but since we are ignoring most of them, one of the problems we face is indent issues. Will talk with Rahul about this tomorrow

package.json
```diff
@@ -29,17 +29,18 @@
...
     "bucketclient": "scality/bucketclient",
     "commander": "^2.11.0",
-    "eslint": "^2.4.0",
-    "eslint-config-airbnb": "^6.0.0",
-    "eslint-config-scality": "scality/Guidelines",
-    "eslint-plugin-react": "^4.2.3",
     "joi": "^10.6",
     "kafka-node": "^1.6.0",
     "node-schedule": "^1.2.0",
+    "node-zookeeper-client": "^0.2.2",
```
bennettbuchanan: Why adding `node-zookeeper-client`?

philipyoo: We were requiring the module in QueuePopulator and ProvisionDispatcher class files and the linter complained we were not including the module in our `package.json`

extensions/replication/queuePopulator/LogReader.js
```diff
@@ -118,23 +118,24 @@ class LogReader {
 
     _writeLogOffset(done) {
         const pathToLogOffset = this.pathToLogOffset;
+        const offsetString = this.logOffset.toString();
         this.zkClient.setData(
             pathToLogOffset,
-            new Buffer(this.logOffset.toString()), -1,
+            Buffer.alloc(offsetString.length, offsetString), -1,
```
rahulreddy: This may be incorrect length. You would have to use `Buffer.byteLength`

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -155,9 +155,10 @@ class ProvisionDispatcher {
         const _provisionOwner = (owner, done) => {
             const zkPath = `${OWNERS}/${owner}`;
             this._log.debug('set new provisions to owner',
-                           { zkPath: `${this._zkEndpoint}${zkPath}`,
-                             provisionList: provisionsByOwner[owner] });
-            const data = new Buffer(JSON.stringify(provisionsByOwner[owner]));
+                { zkPath: `${this._zkEndpoint}${zkPath}`,
+                    provisionList: provisionsByOwner[owner] });
+            const dataString = JSON.stringify(provisionsByOwner[owner]);
+            const data = Buffer.alloc(dataString.length, dataString);
```
rahulreddy: Same here. You should use `Buffer.byteLength`

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine clean try

philipyoo:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #93, opened 2017-09-14
 ft: code changes to handle processor concurrency

jonathan-gramain:
Put queue processor task state in a QueueProcessorTask class instance,
separate from the main QueueProcessor class which is now able to
process multiple entries at once.
    
Also rework _getLocations helper: merge it into QueueEntry class as
getReducedLocations() to ease usage and testing. Unit tests are also
merged into QueueEntry.spec.js.

Note for reviewers: may be easier to review commit by commit.


jonathan-gramain:
@ironman-machine try

jonathan-gramain:
Note that this code does not account for out-of-order updates on version-specific requests (e.g. put-object-acl). For this reason I think it needs to be refined further, or at least discussed to know the limitations.

Setting review only for now.

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

extensions/replication/utils/QueueEntry.js
```diff
@@ -159,6 +163,30 @@ class QueueEntry {
         return location.size;
     }
 
+    // Object metadata may contain multiple elements for a single part if
+    // the part was originally copied from another MPU. Here we reduce the
+    // locations array to a single element for each part.
+    getReducedLocations() {
```
rahulreddy: I would suggest making this change in another PR as this is out of scope. I think there is more context to this as in it may impact the behavior of the GET Object ?partNumber on the target bucket.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -163,469 +60,42 @@ class QueueProcessor {
...
+            queueProcessor: this.processKafkaEntry.bind(this),
+            fetchMaxBytes: CONSUMER_FETCH_MAX_BYTES,
         });
-    }
-
-    _setupDestClients(targetRole, log) {
-        this.s3destAuthManager =
-            this._createAuthManager('target', this.destConfig.auth,
-                                    targetRole, log);
+        consumer.on('error', () => {});
```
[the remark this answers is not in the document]
jonathan-gramain: They are already logged by BackbeatConsumer, and we cannot do much about it (for now at least). By the way this is not new, just moved code.

extensions/replication/utils/QueueEntry.js
jonathan-gramain: This is just the function code moved to a member of QueueEntry and function renamed from `_getLocations` to `_getReducedLocations`. It has unit tests as well (pre-existing ones, moved to QueueEntry.spec.js), so I think it's safe to keep in this PR (actually it would take some time to revert the changes as it's part of the main commit, not sure it's worth it).

jonathan-gramain:
@ironman-machine try

extensions/replication/utils/QueueEntry.js
rahulreddy: Ah I see. Let's keep it then.

rahulreddy:
@ironman-machine r+

PR #94, opened 2017-09-19
bf: fix use of keyed partitioning

jonathan-gramain:
The kafka-node API was misused, resulting in keyed partitioning being
not used: the key has to be part of the individual payload items, not
the 'messages' list. Also fix the provided key to include the bucket
name (may lead to slightly better dispersion across partitions).

### DIFF
--- a/lib/BackbeatProducer.js
+++ b/lib/BackbeatProducer.js
@@ -128,13 +128,17 @@ class BackbeatProducer extends EventEmitter {
                 cb(errors.InternalError);
             });
         }
-        const messages = this._partition === undefined ?
-            entries.map(item => new KeyedMessage(item.key, item.message)) :
-            entries.map(item => item.message);
-        const payload = [{
-            topic: this._topic,
-            messages,
-        }];
+        const payload = this._partition === undefined ?
+                  entries.map(item => ({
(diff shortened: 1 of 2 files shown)


lib/BackbeatProducer.js
```diff
@@ -128,13 +128,17 @@ class BackbeatProducer extends EventEmitter {
...
-        const messages = this._partition === undefined ?
-            entries.map(item => new KeyedMessage(item.key, item.message)) :
-            entries.map(item => item.message);
-        const payload = [{
-            topic: this._topic,
-            messages,
-        }];
+        const payload = this._partition === undefined ?
+                  entries.map(item => ({
+                      topic: this._topic,
```
rahulreddy: We don't necessarily need to set topic or key for the message as they are optional. Topic is set at producer level. Was is impacting the KeyedMessage?

jonathan-gramain: The key is needed here for the message to be sent to the proper partition with keyed partitioning. As for the topic, it was there before so I didn't change this. Actually I don't see where else it's set so i think it's necessary to provide it here.

rahulreddy:
@ironman-machine r+

PR #98, opened 2017-09-25
extensible queue populator

jonathan-gramain:
Create an extension system for the queue populator where the extension:
    
 - subclasses QueuePopulatorExtension to receive new metadata log
   entries
    
 - calls QueuePopulatorExtension.publish() to publish new kafka
   entries
    
Extensions must be located in the extensions/ directory, and are
enabled based on their presence in config.json "extensions" section.
    
Moved configuration section of the queue populator on top of
config.json


lib/queuePopulator/QueuePopulatorExtension.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    setBatch(batch) {
+        this._batch = batch;
+    }
+
+    /**
+     * Prepare a kafka entry for publishing
+     *
+     * This function is meant to be used in subclasses' filter()
```
LaurenSpiegel: I think you mean _onEntry()?

jonathan-gramain: It's a left-over, actually I mean that it's meant to be called after `entry` events are received.

jonathan-gramain:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=ft/S3C-849-genericQueuePopulator
try

jonathan-gramain:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=ft/S3C-849-genericQueuePopulator
try

jonathan-gramain:
@ironman-machine
SCALITY_FEDERATION_BRANCH=ft/S3C-849-genericQueuePopulator
try

jonathan-gramain:
@ironman-machine
SCALITY_FEDERATION_BRANCH=ft/S3C-849-genericQueuePopulator
try

jonathan-gramain:
Needs more refinement on error management and tests before considering a merge, but early comments are still welcome :)

jonathan-gramain:
@ironman-machine 
SCALITY_BACKBEAT_BRANCH=master
try

jonathan-gramain:
OK should be good enough for now. May need refinement on error management later when we get multiple queue populator extensions running concurrently, but for now it should not be necessary.

rahulreddy:
Thanks. @bennettbuchanan @philipyoo Let's start reviewing.

lib/queuePopulator/QueuePopulatorExtension.js
```diff
@@ -0,0 +1,100 @@
...
+     * This method must be implemented by subclasses of QueuePopulatorExtension
+     * @param {Object} entry - metadata log entry
+     * @param {string} entry.bucket - bucket name
+     * @param {string} entry.type - entry type ('put'|'del')
+     * @param {string} entry.key - object key in log
+     * @param {string} entry.value - object value in log
+     * @return {undefined}
+     */
+    filter() {
+        assert(false,
```
bennettbuchanan: Maybe add a brief comment to note that this assertion is meant to ensure the extension has its own `filter` method.

DESIGN.md
```diff
@@ -187,3 +187,62 @@ the producers/consumers adding records for non-replicated and replicated
...
+backbeat asynchronous processor. E.g. Asynchronous Replication is one
+of the extensions available for backbeat.
+
+Extensions are located in the extensions/ directory, with a
+sub-directory named after the extension name (lowercase).
+
+### Extending the Queue Populator
+
+The queue populator is a core backbeat process which reads entries
+from metadata log periodically and provides them to all running
```
bennettbuchanan: ~from metadata log~ from a metadata log

lib/queuePopulator/QueuePopulatorExtension.js
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * This method must be implemented by subclasses of QueuePopulatorExtension
+     * @param {Object} entry - metadata log entry
+     * @param {string} entry.bucket - bucket name
+     * @param {string} entry.type - entry type ('put'|'del')
+     * @param {string} entry.key - object key in log
+     * @param {string} entry.value - object value in log
+     * @return {undefined}
+     */
+    filter() {
```
bennettbuchanan: For consistency with the JSDoc, you might consider including the `entry` param.

conf/config.json
```diff
@@ -6,6 +6,20 @@
     "kafka": {
         "hosts": "127.0.0.1:9092"
     },
+    "queuePopulator": {
+        "cronRule": "*/5 * * * * *",
+        "batchMaxRead": 10000,
+        "zookeeperPath": "/queue-populator",
```
bennettbuchanan: Should we change this path for a deployment as well, in Federation? Looking here: https://github.com/scality/Federation/pull/1158/files#diff-b106b10a40ec0b4affd4efa2884f0b9fR11.

DESIGN.md
```diff
@@ -187,3 +187,62 @@ the producers/consumers adding records for non-replicated and replicated
...
+To achieve this, in the extension module directory, create an
+'index.js' file exporting the attributes defining the extension
+metadata and entrypoints, e.g. to create a "mycoolext" extension, put
+in extensions/mycoolext/index.js:
+
+```
+module.exports = {
+    name: 'my cool extension',
+    version: '1.0.0',
+    QueuePopulatorExtension: require('./MyCoolExtensionPopulator.js'),
```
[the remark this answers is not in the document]
jonathan-gramain: right, good catch

bin/queuePopulator.js
```diff
@@ -3,13 +3,13 @@ const schedule = require('node-schedule');
...
 const zkConfig = config.zookeeper;
-const repConfig = config.extensions.replication;
-const sourceConfig = config.extensions.replication.source;
-const QueuePopulator = require('./QueuePopulator');
+const extConfigs = config.extensions;
+const qpConfig = config.queuePopulator;
+const QueuePopulator = require('../lib/queuePopulator/QueuePopulator');
 
-const log = new werelogs.Logger('Backbeat:Replication:task');
+const log = new werelogs.Logger('Backbeat:QueuePopulator');
```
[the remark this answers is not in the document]
jonathan-gramain: I though it would add more confusion than benefits, being more an internal split in the code structure than a really different module, so I removed the `:task` on purpose to have the same log prefix than in the `QueuePopulator` class. We could also consider passing this log object to the `QueuePopulator` constructor.

lib/queuePopulator/LogReader.js
```diff
@@ -292,38 +305,67 @@ class LogReader {
...
-                logRes.info.start + logStats.nbLogRecordsRead;
-            this.log.debug(
-                'entries published successfully',
-                { method: 'LogReader._processPublishEntries',
-                    entryCount: entriesToPublish.length,
-                    logOffset: batchState.nextLogOffset });
-            return done();
-        });
+            return async.series([
+                done => this._setupProducer(topic, done),
```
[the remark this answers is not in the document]
jonathan-gramain: The inner `done` is guaranteed to be used since it's the closest scope, I personally don't mind though if you guys think it's confusing I can rework to avoid duplicate callback names.

lib/queuePopulator/QueuePopulator.js
```diff
@@ -172,6 +152,35 @@ class QueuePopulator {
...
+                                   { error: err });
+                }
+                done(err);
+            });
+        }, done);
+    }
+
+    _unloadExtensions(done) {
+        async.each(this._extensions, (ext, done) => ext.shutdown(done),
+                   done);
```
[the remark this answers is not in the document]
jonathan-gramain: I originally did this to prepare integration of some setup/teardown from clueso. After some thought we decided not to put this code in the queue populator extensions but as other standalone tasks, and keep the extension focused to record filtering jobs.

I though it would not hurt to keep these methods as part of the extensions API, "just in case", but it may be better to remove them since the queue populator extensions should remain simple synchronous filters I doubt we will ever need async setup/teardown.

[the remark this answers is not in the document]
jonathan-gramain: To answer your question more directly, yes `this._extensions` is an array of references to extensions as class instances, which are valid and can be accessed also for asynchronous processing.

lib/queuePopulator/QueuePopulatorExtension.js
jonathan-gramain: Wanted to do this originally, but the linter would complain that the parameter is not used. I can disable the warning though.

lib/queuePopulator/QueuePopulatorExtension.js
jonathan-gramain: Since it's already explicit in the assert error message, and it's also in the JSDoc above, I think it would be double redundant :)

DESIGN.md
jonathan-gramain: will say `from the metadata log` since normally there's only one log to read from (or none).

conf/config.json
jonathan-gramain: I think it's better to leave it as the old name `replication-populator` because deployments started and changing it would mean using a new state and restarting log replication from the offset 0 (as well as having the old state lying around in zookeeper). I think it's ok to keep the old name for Federation deployments.

lib/queuePopulator/QueuePopulator.js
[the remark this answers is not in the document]
philipyoo: cool! Thank you for the explanation :)

jonathan-gramain:
@philipyoo indeed behavior tests are not run in the current setup. They would be a good fit in Integration tests/HOMER (because they require kafka and S3 server to be available), but they need some changes for this first (e.g. use the proper access/secret key).
 
Created https://scality.atlassian.net/browse/S3C-987 for tracking.

rahulreddy:
@ironman-machine clean r+

rahulreddy:
@ironman-machine r- r+

rahulreddy:
@ironman-machine r- r+

rahulreddy:
@ironman-machine r- r+

rahulreddy:
@ironman-machine r- after=https://github.com/scality/Federation/pull/1158

rahulreddy:
@ironman-machine r- clean r+

PR #99, opened 2017-09-27
extension config management

jonathan-gramain:
ft: manage extensions configuration
    
Each extension is now responsible to manage its config stored in the
global config.json, through a custom configValidator function.

Built on top of https://github.com/scality/backbeat/pull/98

### DIFF
--- a/conf/Config.js
+++ b/conf/Config.js
@@ -4,8 +4,8 @@ const fs = require('fs');
 const path = require('path');
 const joi = require('joi');
 
+const extensions = require('../extensions');
 const backbeatConfigJoi = require('./config.joi.js');
-const { adminCredsJoi } = require('../lib/config/configItems.joi.js');
 
 class Config {
     constructor() {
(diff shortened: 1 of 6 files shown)


conf/Config.js
```diff
@@ -31,33 +31,27 @@ class Config {
         const parsedConfig = joi.attempt(config, backbeatConfigJoi,
                                          'invalid backbeat config');
 
+        if (parsedConfig.extensions) {
+            Object.keys(parsedConfig.extensions).forEach(extName => {
+                const index = require(`../extensions/${extName}/index.js`);
```
rahulreddy: I recommend looping through `extensions` folder and require all extensions at the top.

jonathan-gramain: If just a style preference, I would rather leave it here as I like the fact that I can just browse a config validated by joi first, guaranteed to have the fields I need to browse.

If there are other good reasons to move `require` to the top (caching or other) I can do the necessary manual checks on the extensions array, then require extensions in a global array and then browse that array here to check each extension config.

jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine r- r+

jonathan-gramain:
@ironman-machine r- r+

jonathan-gramain:
@ironman-machine r- r+

PR #101, opened 2017-10-03
ft: add eve yaml file and worker specification

MikaYuoadas:


rayene:
@rahulreddy I've just rebased and fixed the conflicts on this PR. Could you please review it and eventually merge it ?

package.json
```diff
@@ -1,45 +0,0 @@
-{
```
jonathan-gramain: If this is now a link, why is the root package.json linking to the one in eve directory, shouldn't it be the opposite way?

eve/workers/unit_and_feature_tests/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+#
+ARG BUILDBOT_VERSION=0.9.1
+
+RUN pip install buildbot-worker==$BUILDBOT_VERSION
+ADD supervisor/buildbot_worker.conf /etc/supervisor/conf.d/
+
+#
+# Configure build env
+#
+ENV CXX=g++-4.9
```
jonathan-gramain: Why is it needed to specify the version of gcc? If we can avoid this it would ease maintenance since eventually we will want to use a more recent gcc.

eve/workers/unit_and_feature_tests/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+#
+# Install nodejs packages needed by backbeat
+#
+COPY package.json /usr/lib/package.json
+WORKDIR /usr/lib/
```
jonathan-gramain: `/usr/lib` seems to be a rather uncommon root directory for our installs, what about `/opt` or even `~`?

eve/workers/unit_and_feature_tests/supervisor/buildbot_worker.conf
```diff
@@ -0,0 +1,4 @@
+[program:buildbot_worker]
+command=/bin/sh -c 'buildbot-worker create-worker . "%(ENV_BUILDMASTER)s:%(ENV_BUILDMASTER_PORT)s" "%(ENV_WORKERNAME)s" "%(ENV_WORKERPASS)s"  && buildbot-worker start --nodaemon'
+autostart=true
+autorestart=false
```
jonathan-gramain: missing newline

eve/workers/unit_and_feature_tests/Dockerfile
jonathan-gramain: Or rather, `/opt/backbeat` or `~/backbeat`

eve/workers/unit_and_feature_tests/Dockerfile
ssalaues: We can change it to a newer version or eliminate it all together but the goal was for consistency during testing.

eve/workers/unit_and_feature_tests/Dockerfile
ssalaues: We can change it to whichever is preferred

eve/workers/unit_and_feature_tests/supervisor/buildbot_worker.conf
ssalaues: Where exactly should we have the newline?

eve/workers/unit_and_feature_tests/Dockerfile
jonathan-gramain: We're also using `/usr/src/app` for zenko deployments.

eve/workers/unit_and_feature_tests/supervisor/buildbot_worker.conf
jonathan-gramain: A newline at the very end of the file (no big deal just noticing and github diff shows it with a special symbol).

eve/workers/unit_and_feature_tests/Dockerfile
jonathan-gramain: If it works without, I would personally rather remove it, better use the system compiler version I think unless it's shown to cause issues.

eve/workers/unit_and_feature_tests/Dockerfile
jonathan-gramain: @rahulreddy @RachedBenMustapha what would you advise? I was just surprised to see `/usr/lib` but I'm not sure where it should go instead.

package.json
rahulreddy: Apparently the `npm install` step in `Dockerfile` could not use symlinks. So @rayene did it the other way around.

rahulreddy: We have to check, how npm install works though.

MikaYuoadas: Well it's not a problem with npm, but rather with docker.
Docker doesn't suppert symlink. I think it is a safety to avoid being able to point to file outside of the docker context (the directory where the docker build is run).

eve/workers/unit_and_feature_tests/Dockerfile
MikaYuoadas: g

Well, we install the dependencies in /usr/lib so they are available globally on the worker.
Because at that point (worker creation) we don't yet have the backbeat repo cloned so we can't install them locally in the repo files.
This way we don't have to bother about installing dependecies or changing directory before runnig our npm commands in the ci (eve/main.yml).

MikaYuoadas:
I've updated the branch a bit according to the comments on this PR.

MikaYuoadas:
Do these PR needs more changes before merging?

PR #102, opened 2017-10-04
ft: Add backbeat api route healthcheck

philipyoo:
Add `/_/healthcheck` route

In order to run the server, it is required to run zookeeper and kafka

NOTE: This PR is still using the SOHU-Co/kafka-node module

Sample output:
`
[{"0":{"nodeId":0,"host":"na-node1","port":9092},"1":{"nodeId":1,"host":"na-node2","port":9092},"2":{"nodeId":2,"host":"na-node3","port":9092},"3":{"nodeId":3,"host":"na-node4","port":9092},"4":{"nodeId":4,"host":"na-node5","port":9092}},{"metadata":{"0":{"topic":"backbeat-replication","partition":0,"leader":4,"replicas":[0,1,4],"isr":[1,4,0]},"1":{"topic":"backbeat-replication","partition":1,"leader":1,"replicas":[0,1,2],"isr":[1,2,0]},"2":{"topic":"backbeat-replication","partition":2,"leader":1,"replicas":[1,2,3],"isr":[1,2,3]},"3":{"topic":"backbeat-replication","partition":3,"leader":2,"replicas":[2,3,4],"isr":[3,2,4]},"4":{"topic":"backbeat-replication","partition":4,"leader":3,"replicas":[0,3,4],"isr":[4,3,0]}}},{"internalConnections":{"isrHealth":"ok","zookeeper":{"status":"ok","details":{"name":"SYNC_CONNECTED","code":3}},"kafkaProducer":{"status":"ok"}}}]
`

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,17 @@
+'use strict'; // eslint-disable-line strict
+
+const werelogs = require('werelogs');
+
+const runServer = require('./lib/api/BackbeatServer');
+
+const Config = process.env.TEST_SWITCH !== '1' ? require('./conf/Config') :
+    require('./tests/config.json');
+
+const Logger = werelogs.Logger;
+
+werelogs.configure({
+    level: Config.log.logLevel,
+    dump: Config.log.dumpLevel,
+});
+
+runServer(Config, Logger);
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -1,0 +1,209 @@
+'use strict'; // eslint-disable-line strict
+
+const async = require('async');
+const zookeeper = require('node-zookeeper-client');
+
+const { errors } = require('arsenal');
+
+const BackbeatProducer = require('../BackbeatProducer');
+const routes = require('./routes');
+
+// In-Sync Replicas
+const ISRS = 3;
+
+/**
+ * Class representing Backbeat API endpoints and internals
+ *
+ * @class
+ */
+class BackbeatAPI {
+    /**
+     * @constructor
+     * @param {object} config - configurations for setup
+     * @param {werelogs.Logger} logger - Logger object
+     */
+    constructor(config, logger) {
+        this._zkConfig = config.zookeeper;
+        this._repConfig = config.extensions.replication;
+        this._crrTopic = this._repConfig.topic;
+        this._crrStatusTopic = this._repConfig.replicationStatusTopic;
+        this._metricsTopic = config.metrics.topic;
+        this._queuePopulator = config.queuePopulator;
+        this._kafkaHost = config.kafka.hosts;
+        this._logger = logger;
+
+        this._crrProducer = null;
+        this._crrStatusProducer = null;
+        this._metricProducer = null;
+        this._zkClient = null;
+    }
+
+    /**
+     * Check if incoming request is valid
+     * @param {string} route - request route
+     * @return {boolean} true/false
+     */
+    isValidRoute(route) {
+        return routes.some(r => r.path.includes(route));
+    }
+
+    /**
+     * Check if Zookeeper and Producer are connected
+     * @return {boolean} true/false
+     */
+    isConnected() {
+        return this._zkClient.getState().name === 'SYNC_CONNECTED'
+            && this._checkProducersReady();
+    }
+
+    _checkProducersReady() {
+        return this._crrProducer.isReady() && this._metricProducer.isReady()
+            && this._crrStatusProducer.isReady();
+    }
+
+    _getConnectionDetails() {
+        return {
+            zookeeper: {
+                status: this._zkClient.getState().name === 'SYNC_CONNECTED' ?
+                    'ok' : 'error',
+                details: this._zkClient.getState(),
+            },
+            kafkaProducer: {
+                status: this._checkProducersReady() ? 'ok' : 'error',
+            },
+        };
+    }
+
+    /**
+     * Checks health of in-sync replicas
+     * @param {object} md - topic metadata object
+     * @return {boolean} true if ISR health is ok
+     */
+    _checkISRHealth(md) {
+        // eslint-disable-next-line consistent-return
+        const keys = Object.keys(md);
+        for (let i = 0; i < keys.length; i++) {
+            if (md[keys[i]].isr && md[keys[i]].isr.length !== ISRS) {
+                return 'error';
+            }
+        }
+        return 'ok';
+    }
+
+    /**
+     * Get Kafka healthcheck
+     * @param {function} cb - callback(error, data)
+     * @return {undefined}
+     */
+    healthcheck(cb) {
+        const client = this._crrProducer.getKafkaClient();
+
+        // TODO: refactor by calling specific topic
+        client.loadMetadataForTopics([], (err, res) => {
+            if (err) {
+                this._logger.error('error getting healthcheck');
+                return cb(errors.InternalError
+                    .customizeDescription('error getting healthcheck metadata' +
+                    ' for topics'));
+            }
+            const response = res.map(i => (Object.assign({}, i)));
+            const connections = {};
+            const topicMD = {};
+            response.forEach((obj, idx) => {
+                if (obj.metadata && obj.metadata[this._repConfig.topic]) {
+                    let copy;
+                    try {
+                        copy = JSON.parse(JSON.stringify(obj.metadata[
+                            this._repConfig.topic]));
+                        topicMD.metadata = copy;
+                        response.splice(idx, 1);
+                    } catch (e) {
+                        this._logger.error('error getting topic metadata');
+                    }
+                }
+            });
+            response.push(topicMD);
+
+            if (topicMD.metadata) {
+                connections.isrHealth = this._checkISRHealth(topicMD.metadata);
+            }
+
+            Object.assign(connections, this._getConnectionDetails());
+            response.push({
+                internalConnections: connections,
+            });
+
+            return cb(null, response);
+        });
+    }
+
+    setupInternals(cb) {
+        async.parallel([
+            done => this._setZookeeper(done),
+            done => this._setProducer(this._metricsTopic, (err, producer) => {
+                if (err) {
+                    return done(err);
+                }
+                this._metricProducer = producer;
+                return done();
+            }),
+            done => this._setProducer(this._crrTopic, (err, producer) => {
+                if (err) {
+                    return done(err);
+                }
+                this._crrProducer = producer;
+                return done();
+            }),
+            done => this._setProducer(this._crrStatusTopic, (err, producer) => {
+                if (err) {
+                    return done(err);
+                }
+                this._crrStatusProducer = producer;
+                return done();
+            }),
+        ], err => {
+            if (err) {
+                this._logger.error('error setting up internal clients');
+                return cb(err);
+            }
+            this._logger.info('BackbeatAPI setup ready');
+            return cb();
+        });
+    }
+
+    _setProducer(topic, cb) {
+        const producer = new BackbeatProducer({
+            zookeeper: { connectionString: this._zkConfig.connectionString },
+            topic,
+        });
+
+        producer.once('error', cb);
+        producer.once('ready', () => {
+            producer.removeAllListeners('error');
+            producer.on('error', error => {
+                this._logger.error('error from backbeat producer', { error });
+            });
+            return cb(null, producer);
+        });
+    }
+
+    _setZookeeper(cb) {
+        const populatorZkPath = this._queuePopulator.zookeeperPath;
+        const zookeeperUrl =
+            `${this._zkConfig.connectionString}${populatorZkPath}`;
+
+        const zkClient = zookeeper.createClient(zookeeperUrl, {
+            autoCreateNamespace: this._zkConfig.autoCreateNamespace,
+        });
+        zkClient.connect();
+
+        zkClient.once('error', cb);
+        zkClient.once('connected', () => {
+            zkClient.removeAllListeners('error');
+            this._zkClient = zkClient;
+            return cb();
+        });
+    }
+}
+
+module.exports = BackbeatAPI;
--- a/lib/api/BackbeatRequest.js
+++ b/lib/api/BackbeatRequest.js
@@ -1,0 +1,124 @@
+'use strict'; // eslint-disable-line strict
+
+/**
+ * Class representing a Backbeat API Request
+ *
+ * @class
+ */
+class BackbeatRequest {
+    /**
+     * @constructor
+     * @param {http.IncomingMessage} req - request object
+     * @param {http.ServerResponse} res - response object
+     * @param {Logger.newRequestLogger} log - logger object
+     */
+    constructor(req, res, log) {
+        this._request = req;
+        this._response = res;
+        this._log = log;
+        this._route = req.url;
+        this._statusCode = 0;
+        this._error = null;
+    }
+
+    /**
+     * Get logger object
+     * @return {object} Logger object
+     */
+    getLog() {
+        return this._log;
+    }
+
+    /**
+     * Set logger object
+     * @param {object} log - new Logger object
+     * @return {BackbeatRequest} itself
+     */
+    setLog(log) {
+        this._log = log;
+        return this;
+    }
+
+    /**
+     * Get http request object
+     * @return {object} Http request object
+     */
+    getRequest() {
+        return this._request;
+    }
+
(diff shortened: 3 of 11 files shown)


rahulreddy:
FYI, we are just evaluating `node-rdkafka` which is why Producer and Consumer are being instantiated instead of using BackbeatProducer  & BackbeatConsumer.

lib/api/BackbeatAPI.js
```diff
@@ -0,0 +1,131 @@
...
+        });
+        this._consumer.connect();
+
+        this._producer = new Kafka.Producer({
+            'metadata.broker.list': this._kafkaHost,
+            'dr_cb': true,  // delivery report
+        });
+        this._producer.connect();
+
+        this.xproducer = new Kafka.Producer({
```
jonathan-gramain: why duplicating the producer?

lib/api/BackbeatAPI.js
```diff
@@ -0,0 +1,131 @@
...
+     * Get Kafka healthcheck
+     * @param {BackbeatRequest} backbeatRequest - Backbeat request object
+     * @param {function} cb - callback for response
+     * @return {undefined}
+     */
+    healthcheck(backbeatRequest, cb) {
+        this._producer.getMetadata({}, (err, meta) => {
+            if (err) {
+                // setup backbeatRequest error properly
+                return cb(backbeatRequest, err);
```
jonathan-gramain: Conventionally the `err` should be returned as first argument, and the rest of the args follows.

lib/api/BackbeatAPI.js
```diff
@@ -0,0 +1,131 @@
...
+
+        this._producer.produce(testTopic, partition, value, key);
+        this._consumer.subscribe([testTopic]);
+
+        const start = Date.now();
+        let end;
+        this._consumer.consume(1, () => {
+            end = Date.now();
+
+            cb(backbeatRequest, { difference: end - start });
```
jonathan-gramain: I guess this is work in progress, still a few considerations:

 - `elapsedMs` would be a better name here
 - the function measures the latency of a single operation rather than throughput, which is not be that useful of a metric since it does not reflect real workload
 - the function still does a useful healthcheck, so the function may be renamed to `consumerHealthcheck` for example
 - for real testing we would want to know how many operations per second we can achieve on kafka, but this requires a proper test environment and a specific test function that is not a good fit for a generic API.

lib/api/BackbeatRequest.js
```diff
@@ -0,0 +1,137 @@
...
+        return this;
+    }
+
+    /**
+     * Status check to see if valid request
+     * @return {boolean} valid status check
+     */
+    getStatus() {
+        return ((this.getStatusCode >= 200 && this.getStatusCode < 300)
+            && Object.keys(this.getError()).length === 0);
```
jonathan-gramain: You should check what is error type first before considering browsing the keys, because it may be `null`. Why not just add a `!this._error` check instead?

lib/api/BackbeatServer.js
```diff
@@ -0,0 +1,233 @@
...
+     * Check if incoming request is valid
+     * @param {object} req - incoming request object
+     * @param {BackbeatRequest} backbeatRequest - Backbeat request object
+     * @return {boolean} true if no errors
+     */
+    _isValidRequest(req, backbeatRequest) {
+        const allowIp = ipCheck.ipMatchCidrList(
+            this._config.healthChecks.allowFrom, req.socket.remoteAddress);
+        if (!allowIp) {
+            backbeatRequest.setError({
```
jonathan-gramain: You should set this to an arsenal error object instead (from `require('arsenal').errors`).

lib/api/BackbeatServer.js
```diff
@@ -0,0 +1,233 @@
...
+                error: 'Access Denied',
+                code: 401,
+            });
+            return this._errorResponse(backbeatRequest);
+        }
+
+        if (!this.backbeatAPI.isValidRoute(backbeatRequest.getRoute())
+            || !backbeatRequest.getRoute().startsWith('/_/')) {
+            backbeatRequest.setError({
```
jonathan-gramain: For those errors as well it's preferable to use a base arsenal error, and if needed use `customizeDescription()` call to add a custom error message.

lib/api/BackbeatServer.js
```diff
@@ -0,0 +1,233 @@
...
+        if (!this.backbeatAPI.isConnected()) {
+            backbeatRequest.setError({
+                error: 'Internal Error',
+                code: 500,
+            });
+            return this._errorResponse(backbeatRequest);
+        }
+
+        return true;
+        // Returns 500 when Kafka is unreachable
```
jonathan-gramain: Those comments can be in the JSDoc of the function.

lib/api/BackbeatServer.js
```diff
@@ -0,0 +1,233 @@
...
+        const res = backbeatRequest.getResponse();
+        log.trace('writing HTTP response', {
+            method: 'BackbeatServer._response',
+        });
+        const code = backbeatRequest.getStatusCode();
+        const payload = Buffer.from(JSON.stringify(data), 'utf8');
+
+        res.writeHead(code, {
+            'content-type': 'application/json',
+            'content-length': payload.length,
```
jonathan-gramain: The proper way to get the message byte length is `payload.byteLength` property instead. `payload.length` returns the allocated length which depending on allocation strategy may not be the message byte length.

https://nodejs.org/api/buffer.html#buffer_class_method_buffer_bytelength_string_encoding

lib/api/BackbeatServer.js
```diff
@@ -0,0 +1,233 @@
...
+        });
+        res.write(payload);
+        BackbeatServer.logRequestEnd(log, req, res);
+        return res.end();
```
jonathan-gramain: minor: you can avoid the `res.write` by providing the contents to `res.end` directly.

lib/api/BackbeatServer.js
```diff
@@ -0,0 +1,233 @@
...
+        this.server = http.createServer((req, res) => {
+            this._requestListener(req, res);
+        });
+
+        this.server.on('listening', () => {
+            const addr = this.server.address() || {
+                address: '0.0.0.0',
+                port: this._port,
+            };
+            this._logger.trace('server started', {
```
jonathan-gramain: I think this log deserves an `info` level.

lib/api/BackbeatServer.js
```diff
@@ -0,0 +1,233 @@
...
+/**
+ * start - start the backbeat API server
+ * @param {object} config - configurations required
+ * @return {undefined}
+ */
+function run(config) {
+    const logger = new Logger('BackbeatServer');
+    const loggerAPI = new Logger('BackbeatAPI');
+    const workers = os.cpus().length;
+    const cluster = new Clustering(workers, logger);
```
jonathan-gramain: Using clustering for this server sounds like a premature optimization to me, I'm not sure we will ever have CPU intensive operations in this server that would require multiple CPU assignment. We may actually not want to allocate too much resources for an admin/monitoring interface so that other processes can benefit more, but if we plan to plug actual heavy work into this server later, why not. @rahulreddy?

lib/api/BackbeatAPI.js
jonathan-gramain: Note that as a healthcheck, having the latency of one operation is useful to detect abnormally high latency, in case the system has a problem (but as said, not useful to really know what is the system performance).

lib/api/BackbeatServer.js
philipyoo: Using `trace` here because Arsenal already provides a log message on startup saying a worker has started. Trace seemed good here for debugging purposes if for some reason one of the workers failed to start server

lib/api/BackbeatServer.js
jonathan-gramain: I think you can ignore this comment: `byteLength` and `length` properties shall return the same length. `Buffer.byteLength()` is a more generic way to get the byte length of any buffer/string/etc., while the `.length` property is buffer-specific and reflects the same length. Not sure about the `.byteLength` property as I mentioned but I think it's also buffer-specific.

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

lib/api/BackbeatAPI.js
```diff
@@ -0,0 +1,208 @@
...
+        const zookeeperUrl =
+            `${this._zkConfig.connectionString}${populatorZkPath}`;
+
+        const zkClient = zookeeper.createClient(zookeeperUrl, {
+            autoCreateNamespace: this._zkConfig.autoCreateNamespace,
+        });
+        zkClient.connect();
+
+        zkClient.once('error', cb);
+        zkClient.once('state', event => {
```
jonathan-gramain: You can also use the `'connected'` event, so you don't have to check the events (and you may receive other kinds of events that you want to ignore rather than trigger an error if not what you're waiting for).

tests/functional/api/BackbeatServer.js
```diff
@@ -0,0 +1,102 @@
...
+function getUrl(options, path) {
+    return `http://${options.host}:${options.port}${path}`;
+}
+
+describe('Backbeat Server', () => {
+    it('should get a response', done => {
+        const url = getUrl(defaultOptions, '/');
+
+        http.get(url, res => {
+            assert(res.statusCode);
```
jonathan-gramain: You could check that status code is what you expect (200?)

wait_for_local_port.bash
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
jonathan-gramain: I would rather put this script in the tests/utils/ directory to avoid cluttering the root.

tests/functional/api/BackbeatServer.js
philipyoo: I think this was a silly test because I was just checking to see if any response was received from the server. But, this is sort of checked in following tests. If error in connection refused, then all tests will just fail. So I'll just remove this test

index.js
```diff
@@ -0,0 +1,17 @@
+'use strict'; // eslint-disable-line strict
+
+const werelogs = require('werelogs');
+
+const runServer = require('./lib/api/BackbeatServer');
```
bennettbuchanan: I think it would be worthwhile to put the run function in another file because a file name like `./lib/api/BackbeatServer` seems like it is exporting the class `BackbeatServer`.

lib/api/BackbeatAPI.js
```diff
@@ -0,0 +1,205 @@
...
+        this._zkClient = null;
+    }
+
+    /**
+     * Check if incoming request is valid
+     * @param {string} route - request route
+     * @return {boolean} true/false
+     */
+    isValidRoute(route) {
+        return routes.map(route => route.path).includes(route);
```
bennettbuchanan: Can use `some()` method here instead of mapping and then checking the resulting mapped array.

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

index.js
philipyoo: I agree, i'd like to implement this in a refactor after everything is merged though. There is some refactoring, renaming, and reorganizing to be done :S

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

conf/Config.js
```diff
@@ -48,15 +48,18 @@ class Config {
...
-        Object.assign(this, parsedConfig);
-
         // whitelist IP, CIDR for health checks
-        this.healthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
-        const allowedHealthChecks = parsedConfig.server.healthChecks.allowFrom;
-        if (allowedHealthChecks) {
-            this.healthChecks.allowFrom.concat(allowedHealthChecks);
+        const defaultHealthChecks = ['127.0.0.1/8', '::1'];
+        const healthChecks = parsedConfig.server.healthChecks;
+        if (healthChecks.allowFrom.length > 0) {
```
jonathan-gramain: You should not need this special case check, since `concat` on an empty array should work as well (though not if `allowFrom` is not an array, of course)

lib/api/BackbeatRequest.js
```diff
@@ -0,0 +1,118 @@
...
+    setRoute(route) {
+        this._route = route;
+        return this;
+    }
+
+    /**
+     * Status check to see if valid request
+     * @return {boolean} valid status check
+     */
+    getStatus() {
```
jonathan-gramain: For later refactoring: a better name could be `isStatusOk`

tests/config.json
```diff
@@ -13,24 +17,25 @@
             "logName": "s3-recordlog"
         }
     },
-    "kafka": {
```
jonathan-gramain: It's odd to see these lines removed in this PR, aren't they needed for some of the tests?

jonathan-gramain: Actually talking about `s3` section since kafka has been moved up.

philipyoo: hmm, not exactly sure what I did here honestly. Sorry about this, will add back in

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
SCALITY_INTEGRATION_BRANCH=fix/add-cos-parameters-in-sproxyd-location-constraints
try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-630-backbeat-healthcheck
SCALITY_INTEGRATION_BRANCH=rel/7.4
try

PR #103, opened 2017-10-05
FT: Update README for local usage

bennettbuchanan:
Some README changes needed when starting from new repositories, and using replication locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,12 +125,13 @@ chmod 400 ./tests/utils/keyfile && \
 VAULT_DB_BACKEND=MEMORY node vaultd.js
 ```
 
-In a new shell, start the Scality S3 server:
+In a new shell, start the Scality S3 server with the record log enabled:
 
 ```
 git clone https://github.com/scality/s3 ~/replication/s3 && \
 cd ~/replication/s3 && \
 npm i && \
+sed -i '' 's/"enabled": false.*/"enabled": true,/' config.json \
 S3BACKEND=file S3VAULT=scality npm start
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -123,12 +123,13 @@ chmod 400 ./tests/utils/keyfile && \
...
 ```
 
-In a new shell, start the Scality S3 server:
+In a new shell, start the Scality S3 server with the record log enabled:
 
 ```
 git clone https://github.com/scality/s3 ~/replication/s3 && \
 cd ~/replication/s3 && \
 npm i && \
+sed -i '' 's/"enabled": false.*/"enabled": true,/' config.json \
```
jonathan-gramain: This is too hacky IMO, I would advise to split this in three steps, with explaining what to change in the config manually in between and an explanation of what this change does to inform the user.

`jq` would be a nice alternative though it messes up the sections and is not part of the bare unix command line tools so has to be installed first, but why not. Also `sed` has a way to define boundaries as regexp where to do the substitutions which would be safer but still hacky.

bennettbuchanan: Agree. I will update this to just explain the change manually.

README.md
```diff
@@ -186,10 +187,20 @@ Default region name [None]:
...
+~/replication/backbeat/bin/localReplication.js \
+```
+
 Set up replication on your buckets:
 
 ```
-node ~/replication/backbeat/bin/setupReplication.js setup \
+node ~/replication/backbeat/bin/localReplication.js setup \
 --source-bucket source-bucket \
 --source-profile backbeatuser \
```
bennettbuchanan: 'backbeatuser' should be changed to add profiles 'bart' for source and 'lisa' for destination

dora-korpar (commented):
~~Could you also update references to the kafka version? It looks like `kafka_2.11-0.11.0.0` is no longer available. It's `kafka_2.11-0.11.0.2` now.
References are on lines 63, 65, 66, 72, and 78~~

Done!

bennettbuchanan (commented):
We need to also update the `npm run queue_processor` command to take a site argument, and also run the `npm run replication_status_processor` in parallel.

bennettbuchanan:
The README would require removal of these changes now in any case. Closing this.

PR #104, opened 2017-10-10
setup replication script to lib

jonathan-gramain:
Refactor replication.js to put all logic code into a separate library file, keeping only the script UI and profile lookup in replication.js.

This should allow code reuse for echo mode.

Easier to review commits separately: first one is basically just code move, second one does some minor refactoring to transfer profile lookup responsibility back to replication.js.


rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

rahulreddy:
@ironman-machine r+

PR #105, opened 2017-10-10
FT: Add task to replicate data to AWS

bennettbuchanan:
Depends on https://github.com/scality/S3/pull/935.

Handles replication of single-part objects to AWS.

Forthcoming replication operation support:
* MPU objects
* Metadata-only operations (zero-byte objects, putting object tagging, etc.)
* Object deletion

### DIFF
--- a/extensions/replication/queueProcessor/MultipleBackendTask.js
+++ b/extensions/replication/queueProcessor/MultipleBackendTask.js
@@ -1,0 +1,195 @@
+const async = require('async');
+
+const errors = require('arsenal').errors;
+const jsutil = require('arsenal').jsutil;
+
+const QueueProcessorTask = require('./QueueProcessorTask');
+const attachReqUids = require('../utils/attachReqUids');
+
+class MultipleBackendTask extends QueueProcessorTask {
+
+    _setupRolesOnce(entry, log, cb) {
+        log.debug('getting bucket replication', { entry: entry.getLogInfo() });
+        const entryRolesString = entry.getReplicationRoles();
+        let errMessage;
+        let entryRoles;
+        if (entryRolesString !== undefined) {
+            entryRoles = entryRolesString.split(',');
+        }
+        if (entryRoles === undefined || entryRoles.length !== 1) {
+            errMessage = 'expecting a single role in bucket replication ' +
+                'configuration when replicating to an external location';
+            log.error(errMessage, {
+                method: 'MultipleBackendTask._setupRolesOnce',
+                entry: entry.getLogInfo(),
+                roles: entryRolesString,
+            });
+            return cb(errors.BadRole.customizeDescription(errMessage));
+        }
+        this.sourceRole = entryRoles[0];
+
+        this._setupSourceClients(this.sourceRole, log);
+
+        const req = this.S3source.getBucketReplication({
+            Bucket: entry.getBucket(),
+        });
+        attachReqUids(req, log);
+        return req.send((err, data) => {
+            if (err) {
+                log.error('error getting replication configuration from S3', {
+                    method: 'MultipleBackendTask._setupRolesOnce',
+                    entry: entry.getLogInfo(),
+                    origin: 'source',
+                    peer: this.sourceConfig.s3,
+                    error: err.message,
+                    httpStatus: err.statusCode,
+                });
+                // eslint-disable-next-line no-param-reassign
+                err.origin = 'source';
+                return cb(err);
+            }
+            const replicationEnabled = data.ReplicationConfiguration.Rules
+                .some(rule => rule.Status === 'Enabled' &&
+                    entry.getObjectKey().startsWith(rule.Prefix));
+            if (!replicationEnabled) {
+                errMessage = 'replication disabled for object';
+                log.debug(errMessage, {
+                    method: 'MultipleBackendTask._setupRolesOnce',
+                    entry: entry.getLogInfo(),
+                });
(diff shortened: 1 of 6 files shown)


extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -0,0 +1,175 @@
...
+                return cb(errors.BadRole);
+            }
+            return cb(null, roles[0]);
+        });
+    }
+
+    _getAndPutPartOnce(sourceEntry, destEntry, part, log, done) {
+        log.debug('getting object part', { entry: sourceEntry.getLogInfo() });
+        const doneOnce = jsutil.once(done);
+        // TODO handle zero-byte objects which have no parts
```
jonathan-gramain: is this still relevant?

extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -0,0 +1,175 @@
...
+            entryRoles = entryRolesString.split(',');
+        }
+        if (entryRoles === undefined || entryRoles.length !== 1) {
+            log.error('expecting a single role in entry replication ' +
+            'configuration when replicating to an external location', {
+                method: 'MultipleBackendTask._setupRolesOnce',
+                entry: entry.getLogInfo(),
+                roles: entryRolesString,
+            });
+            return cb(errors.BadRole);
```
rahulreddy: Let's add custom description for this.

extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -0,0 +1,175 @@
...
+            if (roles.length !== 1) {
+                log.error('expecting a single role in bucket replication ' +
+                'configuration when replicating to an external location', {
+                    method: 'MultipleBackendTask._setupRolesOnce',
+                    entry: entry.getLogInfo(),
+                    roles,
+                });
+                return cb(errors.BadRole);
+            }
+            if (roles[0] !== entryRoles[0]) {
```
rahulreddy: I know that we have implemented the same before, but I am conflicted on this one. For roles, if they are updated in the bucket's configuration we should just use the latest one. There is no need to penalize the user for switching the roles, especially since the crr backlog is async.

extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -0,0 +1,175 @@
...
+            Key: sourceEntry.getObjectKey(),
+            VersionId: sourceEntry.getEncodedVersionId(),
+            PartNumber: partNumber,
+        });
+        attachReqUids(sourceReq, log);
+        sourceReq.on('error', err => {
+            // eslint-disable-next-line no-param-reassign
+            err.origin = 'source';
+            if (err.statusCode === 404) {
+                return doneOnce(err);
```
rahulreddy: Log appropriately here.

extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -0,0 +1,175 @@
...
+            'configuration when replicating to an external location', {
+                method: 'MultipleBackendTask._setupRolesOnce',
+                entry: entry.getLogInfo(),
+                roles: entryRolesString,
+            });
+            return cb(errors.BadRole);
+        }
+        this.sourceRole = entryRoles[0];
+
+        this._setupSourceClients(this.sourceRole, log);
```
rahulreddy: I would recommend calling this explicitly, as this method is doing more than just `_setupRolesOnce`.

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -74,6 +74,68 @@
                 "payload": "Location"
             }
         },
+        "MultipleBackendPutObject": {
+            "http": {
+                "method": "PUT",
+                "requestUri": "/_/backbeat/multipleBackendPutObject/{Bucket}/{Key+}"
```
rahulreddy: Let's use lowercase for the route `/_/backbeat/multiplebackendput/{Bucket}/{Key+}`. Case-sensitive urls add confusion.

extensions/replication/queueProcessor/MultipleBackendTask.js
bennettbuchanan: Yes, I will have to make the part optional, since metadata-only operations will not have parts in the locations array.

extensions/replication/queueProcessor/MultipleBackendTask.js
bennettbuchanan: Do you mean moving this out of _setupRolesOnce?

rahulreddy: Yes.

extensions/replication/queueProcessor/MultipleBackendTask.js
bennettbuchanan: Updated.

extensions/replication/queueProcessor/MultipleBackendTask.js
bennettbuchanan: Updated.

lib/clients/backbeat-2017-07-01.api.json
bennettbuchanan: Updated.

rahulreddy:
@ironman-machine r+

PR #106, opened 2017-10-13
bf: disallow parallel queue populator batches

jonathan-gramain:
A regression introduced with abd4c7b0 disables the code that prevents
running multiple concurrent batches (when the previous batch is still
in progress). Fix the regression by reverting the relevant code.

Note: this regression can cause assert to be raised in generic queue
populator PR.


rahulreddy:
@ironman-machine r+

PR #107, opened 2017-10-13
target bootstrap port ignored

jonathan-gramain:
Fix use of target bootstrap during replication setup: use the port
set instead of default port 80.


rahulreddy:
@ironman-machine r+

PR #108, opened 2017-10-13
ft: serialize all ops on object

jonathan-gramain:
Ensure serialization of operations at object level instead of for
every version of the object, so that the ordering guarantees are valid
for AWS backend as well.

### DIFF
--- a/extensions/replication/queueProcessor/ReplicationTaskScheduler.js
+++ b/extensions/replication/queueProcessor/ReplicationTaskScheduler.js
@@ -2,21 +2,21 @@ const async = require('async');
 
 /**
  * Schedule replication tasks for replicated buckets, by ensuring that
- * all updates to the same version of an object are serialized
+ * all updates to the same object are serialized
 */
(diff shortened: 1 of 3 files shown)


extensions/replication/queueProcessor/ReplicationTaskScheduler.js
```diff
@@ -7,16 +7,16 @@ const async = require('async');
 class ReplicationTaskScheduler {
```
rahulreddy: Can you also update the description above 

 /**
  * Schedule replication tasks for replicated buckets, by ensuring that
  * all updates to the same ~~version of an~~ object are serialized
 */

jonathan-gramain: done

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r-

PR #111, opened 2017-10-18
FT: Add replication of delete markers

bennettbuchanan:


lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -136,6 +136,46 @@
                 }
             }
         },
+        "MultipleBackendDeleteObject": {
+            "http": {
+                "method": "PUT",
```
bennettbuchanan: Will update this route to use DELETE method and query param.

rahulreddy:
@ironman-machine r+

rahulreddy:
Build succeeded but CI failed to report http://ci.ironmann.io/gh/scality/Integration/15842

PR #112, opened 2017-10-18
FT: Add replication of putting object tags

bennettbuchanan:


rahulreddy:
Merging as we need this PR to be ready tomorrow. We will address any concerns with follow up PRs.

PR #113, opened 2017-10-18
ft: object updates to same kafka partition

jonathan-gramain:
Ensure all updates to an object go to the same Kafka partition for AWS
support. Previously only updates to the same version were guaranteed to
end in the same partition.
    
Use QueueEntry class as a helper to get the master key from the
versioned key and use the former as the Kafka key. We also benefit
from more sanity checking.


jonathan-gramain:
@ironman-machine try

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r-

PR #114, opened 2017-10-19
fix: disable timeouts on S3 SDK

rahulreddy:


rahulreddy:
@ironman-machine try

rahulreddy:
unrelated failure, 
@ironman-machine try

PR #116, opened 2017-10-20
bf: request logger for individual MPU parts

jonathan-gramain:
Spawn a new request logger for each part of an MPU, so that it does
not clutter logs when dumping and consume a lot of CPU. Keep the
parent ID in the logger to link to whole object logs.

Also log the part info when relevant.


rahulreddy:
@ironman-machine try

PR #118, opened 2017-10-23
Ft/one backbeat

LaurenSpiegel:
Opening for feedback on the opIndex.

The make "one populator" change is a hack for now but works.

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -269,11 +269,12 @@ class LogReader {
         });
     }
 
-    _processLogEntry(batchState, record, entry) {
+    _processLogEntry(batchState, record, entry, opIndex) {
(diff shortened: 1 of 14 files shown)


lib/queuePopulator/LogReader.js
```diff
@@ -289,12 +290,18 @@ class LogReader {
         }
 
         this._setEntryBatch(entriesToPublish);
-
+        // start at -1 since will increment when recieve data event
+        let recordNumber = batchState.logRes.info.start - 1;
```
jonathan-gramain: You can remove `batchState.`

jonathan-gramain:
As a reminder, we'll also need to transfer the original filtering of `PENDING` entries on the CRR queue processor side, otherwise he will likely complain loudly for all entries that are not PENDING.

PR #119, opened 2017-10-23
ft: support for aws-s3 replication endpoint

rahulreddy:
With this commit, bootstrapList expects the endpoint to have site/type or site/servers combination. Currently the only allowed type is aws_s3.

### DIFF
--- a/lib/config/configItems.joi.js
+++ b/lib/config/configItems.joi.js
@@ -7,12 +7,18 @@ const hostPortJoi = joi.object({
     port: joi.number().greater(0).required(),
 });
 
+const bootstrapServers = joi.object({
+    site: joi.string().required(),
+    servers: joi.array().items(joi.string()),
+});
+
+const bootstrapCloudBackend = joi.object({
+    site: joi.string().required(),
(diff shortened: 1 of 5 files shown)


lib/config/configItems.joi.js
```diff
@@ -11,6 +11,7 @@ const bootstrapListJoi = joi.array().items(
     joi.object({
         site: joi.string().required(),
         servers: joi.array().items(joi.string()),
+        type: joi.string().allow('aws_s3'),
```
jonathan-gramain: This does not prevent other values to be allowed (i.e. any string), you want to use `valid()` instead.

And IMO you should also check in the config parsing function that either `servers` or `type` is provided but not both. (or there may be more or less arcane ways to do this with `joi` directly).

jonathan-gramain: An additional style comment: the current `bootstrapListJoi` is CRR-specific, maybe we set the definition of the `servers` array as the `bootstrapListJoi` instead to allow re-use.

rahulreddy: You caught me being lazy 😄

PR #120, opened 2017-10-23
FIX: Update client JSON schema and ObjectMD usage

bennettbuchanan:


bennettbuchanan:
@ironman-machine try

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -136,10 +136,54 @@
                 }
             }
         },
-        "MultipleBackendPutObjectTagging": {
+        "MultipleBackendDeleteObject": {
+            "http": {
+                "method": "DELETE",
+                "requestUri": "/_/backbeat/multiplebackenddata/{Bucket}/{Key+}?operation=deleteobject"
```
jonathan-gramain: This should be the default operation on a DELETE with an object path so it should not be needed to provide the operation type :) but same comment, no need to spend too much time refactoring and we can still change it later if we want to, so it's good for me.

bennettbuchanan: I agree. I found that it would require multiple additional checks and routing alterations, so I've just added the query param to simplify that logic in S3 backbeat routes. Also, I was thinking we may want to add delete object tagging, so it could set up nicely for that.

bennettbuchanan:
E2E passed here: http://ci.ironmann.io/gh/scality/Integration/15938#build-parameters

PR #121, opened 2017-10-24
FT: Add delete object tagging replication to AWS

bennettbuchanan:


bennettbuchanan:
@ironman-machine try

extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -494,6 +496,53 @@ class MultipleBackendTask extends QueueProcessorTask {
...
+    _deleteObjectTagging(sourceEntry, destEntry, log, cb) {
+        this._retry({
+            actionDesc: 'delete object tagging',
+            entry: sourceEntry,
+            actionFunc: done => this._deleteObjectTaggingOnce(sourceEntry,
+                destEntry, log, done),
+            shouldRetryFunc: err => err.retryable,
+            onRetryFunc: err => {
+                if (err.origin === 'target') {
+                    this.destHosts.pickNextHost();
```
jonathan-gramain: There's no reason to round-robin the destination hosts here, because we're not using the destination hosts. We're just going to retry on the same unique source. Can you also remove this round-robin rotation for other commands that only contact the source? (i.e. all multiple backend commands?)

rahulreddy:
Arsenal branch is merged, you can drop the dropme commit.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

PR #122, opened 2017-10-25
implement "echo" replication mode

jonathan-gramain:
This mode processes the "create bucket" entries on the source to
    configures CRR automatically for this bucket through backbeat
    (including creating the destination account if it does not exist, then
    basically what does the replication.js setup script).
    
    The destination bucket has the same name than on the source, and the
    owner account has the same display name and email address (but a
    different user ID and canonical ID since on a different Vault).


jonathan-gramain:
Integration tests passed here: http://ci.ironmann.io/gh/scality/Integration/16266

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -63,24 +68,92 @@ class QueueProcessor {
...
-            this.sourceVault = new VaultClient(host, port);
+            const { host, port, adminPort, adminCredentialsFile }
+                      = sourceConfig.auth.vault;
+            this.vaultClientCache
+                .setHost('source:s3', host)
+                .setPort('source:s3', port);
+            if (adminCredentialsFile) {
+                this.vaultClientCache
+                    .setHost('source:admin', host)
+                    .setPort('source:admin', adminPort)
```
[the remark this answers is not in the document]
jonathan-gramain: Yes, but rather as an error condition: the configuration would be invalid if `adminCredentialsFile` is provided but `adminPort` is not. The credentials file is necessary for echo mode, which also requires to provide the port for the admin interface. But I agree it would be good to sanity check it and abort the process with a proper error if not specified, will do something about it.

extensions/replication/utils/SetupReplication.js
```diff
@@ -103,15 +98,31 @@ class SetupReplication {
...
         this._sourceBucket = source.bucket;
         this._targetBucket = target.bucket;
+        this._checkSanityEnabled = checkSanity || false;
+        this._skipSourceBucketCreation = skipSourceBucketCreation || false;
         this.destHosts = target.hosts;
         const destHost = this.destHosts.pickHost();
+        // XXX use target port through nginx gateway
+        // Hard-code port 8600 for now on IAM client while other vault
+        // clients go through the gateway (signature mismatch issue).
+        const destPort = (target.vault && target.vault.adminPort) || 8600;
```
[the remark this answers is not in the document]
jonathan-gramain: I think it works with javascript behavior of `||`, correct me if I'm wrong: when `target.vault.adminPort` is set (and not 0 or anything falsy), the expression will take its value and the part after `||` is not considered, but if it's not set or falsy, 8600 is the expression value instead, which matches what we want here, since port numbers cannot be 0 it should be good.

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine 
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
try

extensions/replication/utils/SetupReplication.js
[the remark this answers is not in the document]
philipyoo: ah ya you're right, it does set. Sorry I just assumed the && would return a boolean instead of the intended adminPort

extensions/replication/queueProcessor/QueueProcessor.js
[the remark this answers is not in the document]
jonathan-gramain: I added this sanity check and others in the config.joi.js file, so now `adminPort` should be guaranteed to be set here.

jonathan-gramain:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
try

jonathan-gramain:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
try

jonathan-gramain:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
try

jonathan-gramain:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
SCALITY_FEDERATION_BRANCH=ft/S3C-981-siteLevelReplication
try

jonathan-gramain:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
SCALITY_FEDERATION_BRANCH=ft/S3C-981-siteLevelReplication
try

extensions/replication/utils/BucketQueueEntry.js
```diff
@@ -0,0 +1,45 @@
...
+     * @param {string} key - entry key in database
+     */
+    constructor(key) {
+        const [ownerCanonicalID, bucketName] = key.split(splitter);
+        this._ownerCanonicalID = ownerCanonicalID;
+        this._bucket = bucketName;
+    }
+
+    clone() {
+        return new BucketQueueEntry(this._bucket);
```
philipyoo: Instead of `this._bucket`, don't we need to pass a replica of the key here?  Even passing just bucket name will set the clones `ownerCanonicalID` as the bucket name.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -132,13 +205,29 @@ class QueueProcessor {
...
-                new MultipleBackendTask(this) : new QueueProcessorTask(this),
-            entry: sourceEntry,
-        },
-        `${sourceEntry.getBucket()}/${sourceEntry.getObjectKey()}`,
-        done);
+        let task;
+        if (sourceEntry instanceof BucketQueueEntry) {
+            // FIXME support multiple destinations
+            if (this.destConfig.bootstrapList.length > 0 &&
+                this.destConfig.bootstrapList[0].echo) {
```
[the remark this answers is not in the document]
jonathan-gramain: For now we only support one destination site, hence we just check the first and only site declared in `bootstrapList` array (so at index 0).

*Update with latest rebase*: now with AWS support, we support only one **scality** destination, but there may be an additional AWS and/or azure etc. destination(s). Still the same applies for the unique scality destination if exists.

extensions/replication/tasks/EchoBucket.js
```diff
@@ -0,0 +1,203 @@
...
+        const { host, port } = this.destHosts.pickHost();
+        // if no nginx proxy is used, the client port is preset in the
+        // profile and the provided port is only used in the cache key
+        const destAdminVault = this.vaultClientCache.getClient('dest:admin',
+                                                               host, port);
+        const destS3Vault = this.vaultClientCache.getClient('dest:s3',
+                                                            host, port);
+
+        async.waterfall([
+            done => destS3Vault.getCanonicalIds(
```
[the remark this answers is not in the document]
jonathan-gramain: No, because Vault realms are different between source and destination, we need to fetch the destination canonical ID which is different than the source one, to set it properly in the destination metadata.

extensions/replication/utils/BucketQueueEntry.js
jonathan-gramain: You're right, good catch. Since this `clone()` function is never used I will just remove it instead.

extensions/replication/tasks/EchoBucket.js
```diff
@@ -0,0 +1,203 @@
...
+                        vault: this.destConfig.auth.vault,
+                        transport: this.destConfig.transport,
+                    },
+                    retryTimeoutS: this.repConfig.queueProcessor.retryTimeoutS,
+                    skipSourceBucketCreation: true,
+                    log,
+                });
+                setupReplication.setupReplication(done);
+            },
+            (info, done) => {
```
rahulreddy: Ideally this should go into the final callback where results are collected.

extensions/replication/utils/BucketQueueEntry.js
```diff
@@ -0,0 +1,41 @@
+const { splitter } = require('arsenal').constants;
+
+class BucketQueueEntry {
+
+    /**
+     * @constructor
+     * @param {string} key - entry key in database
+     */
+    constructor(key) {
+        const [ownerCanonicalID, bucketName] = key.split(splitter);
```
rahulreddy: Is key validated to be set always? What happens when key is set to null.

extensions/replication/utils/QueueEntry.js
```diff
@@ -1,119 +1,38 @@
...
+        return entry;
+    } catch (err) {
+        return { error: { message: 'malformed JSON in kafka entry',
+                          description: err.message } };
     }
 }
 
-module.exports = QueueEntry;
+module.exports = {
+    createFromKafkaEntry,
```
rahulreddy: You should either name this file `createFromKafkaEntry` and export default the method or keep it QueueEntry.

lib/clients/VaultClientCache.js
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * @param {string} profileName - name of profile to bind the
+     *   credentials to
+     * @param {string} filePath - path to credentials file. This file
+     *   must be a JSON file in the following format:
+     *   {"accessKey":"secretKey"}
+     * @return {VaultClientCache} this
+     */
+    loadAdminCredentialsFromFile(profileName, filePath) {
+        const adminCredsJSON = fs.readFileSync(filePath);
```
rahulreddy: This should happen in Config.json. Validate, parse and load into the config object.

lib/clients/VaultClientCache.js
```diff
@@ -0,0 +1,133 @@
...
+        const profile = this._profiles[profileName] || {};
+        const vaultHost = profile.host || host;
+        const vaultPort = profile.port || port;
+        if (!vaultHost || !vaultPort) {
+            return null;
+        }
+        const key = `${profileName}:${vaultHost}:${vaultPort}`;
+
+        if (this._vaultClients[key] === undefined) {
+            this._vaultClients[key] = new VaultClient(
```
rahulreddy: Minor naming suggestion. Across all projects we use `_vaultClients` in lowercase. I recommend keeping it likewise i.e `_vaultclients`.

lib/tasks/BackbeatTask.js
```diff
@@ -0,0 +1,75 @@
...
+                                 retryTotalMs: `${now - startTime}`,
+                             }, logFields || {}));
+                }
+                return done(...args);
+            }
+            if (!shouldRetryFunc(err)) {
+                return done(err);
+            }
+            if (onRetryFunc) {
+                onRetryFunc(err);
```
rahulreddy: Should this not be returned?

lib/tasks/BackbeatTask.js
```diff
@@ -0,0 +1,75 @@
...
+        const { actionDesc, logFields,
+                actionFunc, shouldRetryFunc, onRetryFunc, log } = args;
+        const backoffCtx = new BackOff(BACKOFF_PARAMS);
+        let nbRetries = 0;
+        const startTime = Date.now();
+        const self = this;
+
+        function _handleRes(...args) {
+            const err = args[0];
+            if (!err) {
```
rahulreddy: It's cleaner and consistent if you manage the error case first.

extensions/replication/queueProcessor/QueueProcessor.js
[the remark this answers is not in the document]
jonathan-gramain: Will add a check to ensure echo only works for scality target.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -64,25 +69,93 @@ class QueueProcessor {
...
             destConfig.bootstrapList.forEach(dest => {
                 if (Array.isArray(dest.servers)) {
                     this.destHosts =
                         new RoundRobin(dest.servers, { defaultPort: 80 });
+                    if (dest.echo) {
+                        if (process.env.BACKBEAT_ECHO_TEST_MODE === '1') {
+                            this.logger.info('starting in echo mode',
+                                             { testMode: true });
+                        } else {
+                            this.logger.info('starting in echo mode');
```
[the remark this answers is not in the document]
jonathan-gramain: No, it's in the QueueProcessor constructor, which is only instanciated once when the process starts.

lib/tasks/BackbeatTask.js
jonathan-gramain: no, this is just a custom helper set by the caller to do some side work, we still want to continue processing the failure.

extensions/replication/tasks/EchoBucket.js
```diff
@@ -0,0 +1,203 @@
...
+     *
+     * @constructor
+     * @param {QueueProcessor} qp - queue processor instance
+     */
+    constructor(qp) {
+        const qpState = qp.getStateVars();
+        super({
+            retryTimeoutS: qpState.repConfig.queueProcessor.retryTimeoutS,
+        });
+        Object.assign(this, qpState);
```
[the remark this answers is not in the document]
rahulreddy: My concern was that we lose visibility on what properties the class holds.

jonathan-gramain:
@ironman-machine try

rahulreddy:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

extensions/replication/tasks/EchoBucket.js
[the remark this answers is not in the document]
jonathan-gramain: OK, why not, I can set the needed properties from the state.

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

lib/tasks/BackbeatTask.js
```diff
@@ -0,0 +1,76 @@
...
+            }
+            if (nbRetries > 0) {
+                const now = Date.now();
+                log.info(`succeeded to ${actionDesc} after retries`,
+                         Object.assign({
+                             nbRetries,
+                             retryTotalMs: `${now - startTime}`,
+                         }, logFields || {}));
+            }
+            return done(...args);
```
philipyoo: Do we need to pass `null` as first arg here?

jonathan-gramain: It's good, we are in the success path so `args[0]` is `null` and we send the same status as first argument.

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
Integration tests passed in http://ci.ironmann.io/gh/scality/Integration/18070 (before latest rebase).

rahulreddy:
@ironman-machine r+

jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine r- r+

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine clean after=https://github.com/scality/Federation/pull/1188

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine clean r- after=https://github.com/scality/Federation/pull/1188

rahulreddy:
@ironman-machine clean r- 
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
after=https://github.com/scality/Federation/pull/1188

rahulreddy:
@ironman-machine clean r-
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
r+

PR #123, opened 2017-10-27
FT: Add replication to AWS option for setup tool

bennettbuchanan:


rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

bin/replication.js
```diff
@@ -48,6 +50,8 @@ function _createSetupReplication(command, options, log) {
             hosts: new RoundRobin(destination.bootstrapList[0].servers),
             transport: destination.transport,
         },
+        isTargetExternal,
```
jonathan-gramain: I suggest instead adding an `isExternal` option to the `target` config object because it's directly related to target configuration.

extensions/replication/utils/SetupReplication.js
```diff
@@ -417,14 +436,15 @@ class SetupReplication {
...
-    _enableReplication(roleArns, cb) {
+    _enableReplication(roleArns, siteName, cb) {
         const params = {
             Bucket: this._sourceBucket,
             ReplicationConfiguration: {
                 Role: roleArns,
                 Rules: [{
                     Destination: {
                         Bucket: `arn:aws:s3:::${this._targetBucket}`,
+                        StorageClass: siteName,
```
jonathan-gramain: I'm worried about the side effects here if `siteName` is undefined, depending on how the using code deals with own attributes that have undefined values. In doubt I would rather construct the `Destination` object separately only adding `StorageClass` if not undefined, but fine for me if you could validate that the replication config is correct with no `siteName`.

extensions/replication/utils/SetupReplication.js
```diff
@@ -103,46 +108,59 @@ class SetupReplication {
...
                 target.credentials),
         };
     }
 
     checkSanity(cb) {
         return async.waterfall([
             next => this._isValidBucket('source', next),
-            next => this._isValidBucket('target', next),
+            next => (this._isTargetExternal ? next() :
+                this._isValidBucket('target', next)),
```
jonathan-gramain: It would be great if sanity check functions `isValidBucket`, `isVersioningEnabled` etc. could also check external backends! I understand why it's skipped here because the implementation is different (has to go through the source with probably differing bits), but is there a plan to do this in the future? Similarly, we could also pre-create buckets and enable versioning :)

bin/replication.js
bennettbuchanan: Updated.

extensions/replication/utils/SetupReplication.js
bennettbuchanan: I did check that there are no side effects, but it definitely doesn't hurt to construct the object separately—that can only be a good thing. Updated.

extensions/replication/utils/SetupReplication.js
bennettbuchanan: We wanted to leave AWS buckets up to the user to handle, and not modify them in the script. I've documented that fact accordingly here: https://github.com/scality/backbeat/pull/128/files#diff-c86568b6c2391c54b2749733f7d7ceebR90.

bennettbuchanan:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #125, opened 2017-10-31
Ft/S3C-354 add metrics accumulation

philipyoo:
This PR adds metrics accumulation.

Metrics (ops and bytes) are acquired on population and processing of replication objects. These metrics are accumulated and stored into Kafka, and then will eventually be consumed and stored within Redis. This PR addresses the first part of accumulation and storing into Kafka.

### DIFF
--- a/extensions/replication/ReplicationQueuePopulator.js
+++ b/extensions/replication/ReplicationQueuePopulator.js
@@ -51,6 +51,8 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
         this.publish(this.repConfig.topic,
                      `${queueEntry.getBucket()}/${queueEntry.getObjectKey()}`,
                      JSON.stringify(entry));
+
+        this._incrementMetrics(entry.bucket, queueEntry.getBytesMetric());
     }
 }
 
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -11,7 +11,7 @@ const RoundRobin = require('arsenal').network.RoundRobin;
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
 const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
 const VaultClientCache = require('../../../lib/clients/VaultClientCache');
-const QueueEntry = require('../utils/QueueEntry');
+const QueueEntry = require('../../../lib/models/QueueEntry');
 const ReplicationTaskScheduler = require('../utils/ReplicationTaskScheduler');
 const ReplicateObject = require('../tasks/ReplicateObject');
 const MultipleBackendTask = require('../tasks/MultipleBackendTask');
@@ -20,7 +20,12 @@ const EchoBucket = require('../tasks/EchoBucket');
 const ObjectQueueEntry = require('../utils/ObjectQueueEntry');
 const BucketQueueEntry = require('../utils/BucketQueueEntry');
 
-const { proxyVaultPath, proxyIAMPath } = require('../constants');
+const {
+    proxyVaultPath,
+    proxyIAMPath,
+    metricsExtension,
+    metricsTypeProcessed,
+} = require('../constants');
 
 /**
 * Given that the largest object JSON from S3 is about 1.6 MB and adding some
@@ -40,7 +45,7 @@ class QueueProcessor extends EventEmitter {
      *
      * @constructor
      * @param {Object} zkConfig - zookeeper configuration object
-     * @param {string} zkConfig.connectionString - zookeeper connection string
+     * @param {String} zkConfig.connectionString - zookeeper connection string
      *   as "host:port[/chroot]"
      * @param {Object} sourceConfig - source S3 configuration
      * @param {Object} sourceConfig.s3 - s3 endpoint configuration object
@@ -57,8 +62,10 @@ class QueueProcessor extends EventEmitter {
      *   number of seconds before giving up retries of an entry
      *   replication
      * @param {String} site - site name
+     * @param {MetricsProducer} mProducer - instance of metrics producer
      */
-    constructor(zkConfig, sourceConfig, destConfig, repConfig, site) {
+    constructor(zkConfig, sourceConfig, destConfig, repConfig, site,
+    mProducer) {
         super();
         this.zkConfig = zkConfig;
         this.sourceConfig = sourceConfig;
@@ -70,6 +77,7 @@ class QueueProcessor extends EventEmitter {
         this.replicationStatusProducer = null;
         this._consumer = null;
         this.site = site;
+        this._mProducer = mProducer;
 
         this.echoMode = false;
 
@@ -249,6 +257,18 @@ class QueueProcessor extends EventEmitter {
                 });
                 this._consumer.on('error', () => {});
                 this._consumer.subscribe();
+
+                this._consumer.on('metrics', data => {
+                    // i.e. data = { my-bucket: { ops: 1, bytes: 124 } }
+                    this._mProducer.publishMetrics(data, metricsTypeProcessed,
+                        metricsExtension, err => {
+                            this.logger.trace('error occurred in publishing ' +
+                            'metrics', {
+                                error: err,
+                                method: 'QueueProcessor.start',
+                            });
+                        });
+                });
             }
             this.logger.info('queue processor is ready to consume ' +
                              'replication entries');
--- a/extensions/replication/queueProcessor/task.js
+++ b/extensions/replication/queueProcessor/task.js
@@ -3,11 +3,13 @@ const assert = require('assert');
 const werelogs = require('werelogs');
 
 const QueueProcessor = require('./QueueProcessor');
+const MetricsProducer = require('../../../lib/MetricsProducer');
 
 const config = require('../../../conf/Config');
 const zkConfig = config.zookeeper;
 const repConfig = config.extensions.replication;
 const sourceConfig = repConfig.source;
+const mConfig = config.metrics;
 
 const site = process.argv[2];
 assert(site, 'QueueProcessor task must be started with a site as argument');
@@ -19,11 +21,21 @@ assert(bootstrapList.length === 1, 'Invalid site argument. Site must match ' +
 
 const destConfig = Object.assign({}, repConfig.destination);
 destConfig.bootstrapList = bootstrapList;
-const queueProcessor = new QueueProcessor(zkConfig,
-                                          sourceConfig, destConfig,
-                                          repConfig, site);
 
+const log = new werelogs.Logger('Backbeat:QueueProcessor:task');
 werelogs.configure({ level: config.log.logLevel,
     dump: config.log.dumpLevel });
 
-queueProcessor.start();
+const metricsProducer = new MetricsProducer(zkConfig, mConfig);
+metricsProducer.setupProducer(err => {
+    if (err) {
+        log.error('error starting metrics producer for queue processor', {
+            error: err,
+            method: 'MetricsProducer::setupProducer',
+        });
+        return undefined;
+    }
+    const queueProcessor = new QueueProcessor(zkConfig, sourceConfig,
+        destConfig, repConfig, site, metricsProducer);
+    return queueProcessor.start();
+});
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -8,6 +8,10 @@ const BackbeatProducer = require('./BackbeatProducer');
 const Logger = require('werelogs').Logger;
 kafkaLogging.setLoggerProvider(new Logger('Consumer'));
 
+const QueueEntry = require('./models/QueueEntry');
+
+const CRR_TOPIC = require('../conf/Config').extensions.replication.topic;
+
 // controls the number of messages to process in parallel
 const CONCURRENCY_DEFAULT = 1;
 const CLIENT_ID = 'BackbeatConsumer';
@@ -60,6 +64,10 @@ class BackbeatConsumer extends EventEmitter {
         this._queueProcessor = queueProcessor;
         this._concurrency = concurrency;
         this._messagesConsumed = 0;
+
+        // metrics - consumption
+        this._metricsStore = {};
+
         this._consumer = new ConsumerGroup({
             host: this._zookeeperEndpoint,
             ssl,
@@ -100,6 +108,23 @@ class BackbeatConsumer extends EventEmitter {
                         offset,
                     });
                     this.emit('error', err, entry);
+                } else if (entry.topic === CRR_TOPIC) {
+                    const qEntry = QueueEntry.createFromKafkaEntry(entry);
+                    if (!qEntry.error && qEntry.getReducedLocations) {
+                        const locations = qEntry.getReducedLocations();
+                        const bytes = locations.reduce((sum, item) =>
+                            sum + item.size, 0);
+
+                        if (!this._metricsStore[qEntry.bucket]) {
+                            this._metricsStore[qEntry.bucket] = {
+                                ops: 1,
+                                bytes,
+                            };
+                        } else {
+                            this._metricsStore[qEntry.bucket].ops++;
+                            this._metricsStore[qEntry.bucket].bytes += bytes;
+                        }
+                    }
                 }
             });
         });
@@ -115,6 +140,10 @@ class BackbeatConsumer extends EventEmitter {
             }], () => {
                 this.emit('consumed', count);
                 this._messagesConsumed = 0;
+
+                this.emit('metrics', this._metricsStore);
+                this._metricsStore = {};
+
                 this._consumer.resume();
             });
         };
--- a/lib/MetricsProducer.js
+++ b/lib/MetricsProducer.js
@@ -1,0 +1,67 @@
+'use strict'; // eslint-disable-line strict
+
+const async = require('async');
+const { Logger } = require('werelogs');
+
+const BackbeatProducer = require('./BackbeatProducer');
+const MetricsModel = require('./models/MetricsModel');
+
+class MetricsProducer {
+    /**
+    * @constructor
+    * @param {Object} zkConfig - zookeeper configurations
+    * @param {Object} mConfig - metrics configurations
+    */
+    constructor(zkConfig, mConfig) {
+        this._zkConfig = zkConfig;
+        this._topic = mConfig.topic;
+
+        this._producer = null;
+        this._log = new Logger('MetricsProducer');
+    }
+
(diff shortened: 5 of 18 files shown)


lib/BackbeatConsumer.js
```diff
@@ -114,6 +144,12 @@ class BackbeatConsumer extends EventEmitter {
             }], () => {
                 this.emit('consumed', count);
                 this._messagesConsumed = 0;
+
+                this.emit('sendMetrics', this._metricsStore, () => {
```
jonathan-gramain: I don't like the idea of providing a callback to an event listener, because it's not how the event model fits as they are usually mutually exclusive: normally it should be safe to have 0 or more listeners to any event, here it does seem like the code assumes there will be exactly one listener. In the same vein, you could rename the event to simply `'metrics'`, because emitting events normally does not assume a specific action from the listener(s), rather provides a bit of information that the listener is free to do whatever he needs with.

I may be wrong but it seems you could simply reset the `this._metricsStore` object as soon as you emitted the `sendMetrics` event, because the original object will be left untouched for event listeners.

lib/queuePopulator/QueuePopulator.js
```diff
@@ -26,12 +26,15 @@ class QueuePopulator {
      *   configuration (mandatory if logSource is "bucket")
      * @param {Object} [qpConfig.dmd] - dmd source
      *   configuration (mandatory if logSource is "dmd")
+     * @param {Object} mConfig - metrics configuration object
+     * @param {string} mConfig.topic - metrics topic
```
bennettbuchanan: Minor: ~string~ String

lib/BackbeatConsumer.js
```diff
@@ -99,6 +105,30 @@ class BackbeatConsumer extends EventEmitter {
...
                     });
                     this.emit('error', err, entry);
+                } else {
+                    const qEntry = QueueEntry.createFromKafkaEntry(entry);
+                    if (typeof qEntry.getReducedLocations === 'function') {
+                        const locations = qEntry.getReducedLocations();
+                        const bytes = locations.reduce((sum, item) =>
+                            sum + item.size, 0);
+
+                        if (!this._metricsStore[qEntry.bucket]) {
```
jonathan-gramain: Same thing about putting the metrics update code in a separate method. Actually you could have this code in a new `QueueEntry.updateMetrics(metricsStore)` method for example, that would avoid duplicating the logic, since all needed knowledge is already in the queue entry. Ideally `metricsStore` could be managed by a separate class to embed the logic of counting operations per bucket in one code unit.

lib/MetricsProducer.js
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @param {object} extMetrics - an object where keys are all buckets for a
+     *   given extension and values are the metrics for the bucket
+     *   (i.e. { my-bucket: { ops: 1, bytes: 124 } } )
+     * @param {string} type - type of metric (queueud or processed)
+     * @param {string} ext - extension (i.e. 'crr')
+     * @param {function} cb - callback
+     * @return {undefined}
+     */
+    publishMetric(extMetrics, type, ext, cb) {
```
jonathan-gramain: rename to `publishMetrics`? Since there are metrics for each bucket.

extensions/replication/queueProcessor/task.js
```diff
@@ -3,18 +3,29 @@
...
 werelogs.configure({ level: config.log.logLevel,
     dump: config.log.dumpLevel });
 
-queueProcessor.start();
+const metricsProducer = new MetricsProducer(zkConfig, mConfig);
+metricsProducer.setupProducer(err => {
+    if (err) {
+        log.error('error starting metrics producer for queue processor', {
+            error: err,
+            method: 'QueueProcessor::task',
```
bennettbuchanan: `MetricsProducer::setupProducer`

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -23,9 +23,31 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
         if (queueEntry.getReplicationStatus() !== 'PENDING') {
             return;
         }
+
         this.publish(this.repConfig.topic,
                      `${queueEntry.getBucket()}/${queueEntry.getObjectKey()}`,
                      JSON.stringify(entry));
+
+        if (typeof queueEntry.getReducedLocations === 'function') {
```
jonathan-gramain: minor: you could simply test with `if (queueEntry.getReducedLocations) ...` (if it's there it ought to be a function)

lib/MetricsProducer.js
```diff
@@ -0,0 +1,63 @@
...
+        Object.keys(extMetrics).forEach(bucketName => {
```
bennettbuchanan: `forEach` is syncronous, when you would want to wait until all extensions were published before calling the initial callback.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -111,6 +115,12 @@ class QueueProcessor {
         consumer.on('error', () => {});
         consumer.subscribe();
 
+        consumer.on('sendMetrics', (data, cb) => {
+            // i.e. data = { my-bucket: { ops: 1, bytes: 124 } }
+            this._mProducer.publishMetric(data, METRICS_TYPE, METRICS_EXTENSION,
+                () => { cb(); });
```
bennettbuchanan: Pass `cb` here?

lib/models/MetricsModel.js
```diff
@@ -0,0 +1,37 @@
...
+     */
+
+    constructor(ops, bytes, extension, type, bucket) {
+        this._timestamp = Date.now();
+        this._ops = ops;
+        this._bytes = bytes;
+        this._extension = extension;
+        this._type = type;
+        this._bucket = bucket;
+        return this;
```
bennettbuchanan: Would be okay to not return anything in the constructor since the `new` keyword creates a new instance with `this` bound to the newly created object. So `new MetricsModel(...).serialize();` should still return the expected serialized object.

lib/BackbeatConsumer.js
```diff
@@ -99,6 +105,30 @@ class BackbeatConsumer extends EventEmitter {
...
+                                ops: 1,
+                                bytes,
+                            };
+                        } else {
+                            this._metricsStore[qEntry.bucket].ops++;
+                            this._metricsStore[qEntry.bucket].bytes += bytes;
+                        }
+                    } else {
+                        this._log.error('queue entry has no function '
+                        + ' getReducedLocations', {
```
bennettbuchanan: Minor: extra space in this message.

lib/queuePopulator/LogReader.js
```diff
@@ -365,7 +372,14 @@ class LogReader {
                 batchState.publishedEntries[topic] = topicEntries;
                 return done();
             });
-        }, done);
+        }, () => {
```
bennettbuchanan: Need a check for error first.

lib/queuePopulator/LogReader.js
```diff
@@ -365,7 +372,14 @@ class LogReader {
...
+            this._extensions.forEach(ext => {
```
bennettbuchanan: Should this `forEach` instead be asynchronous?

lib/queuePopulator/LogReader.js
```diff
@@ -365,7 +372,14 @@ class LogReader {
...
+                const extMetrics = ext.getMetrics();
+                this._mProducer.publishMetric(extMetrics, METRIC_TYPE,
+                    METRICS_EXTENSION, () => {});
```
bennettbuchanan: We don't want to handle the error here?

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -9,11 +9,13 @@ const RoundRobin = require('arsenal').network.RoundRobin;
 const VaultClient = require('vaultclient').Client;
 
 const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
-const QueueEntry = require('../utils/QueueEntry');
+const QueueEntry = require('../../../lib/models/QueueEntry');
 const ReplicationTaskScheduler = require('./ReplicationTaskScheduler');
 const QueueProcessorTask = require('./QueueProcessorTask');
 const MultipleBackendTask = require('./MultipleBackendTask');
 
+const METRICS_TYPE = 'processed';
```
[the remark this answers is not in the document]
philipyoo: Moved the `METRICS_EXTENSION` to the constants file 👍 

I use this same `METRICS_TYPE` in the populator side and the value is set to `queued`. In order to reduce confusion, I'm not too sure what would be best. 

For now, I'm thinking of renaming them to their respective types, i.e. `METRICS_QUEUED_TYPE` || `METRICS_PROCESSED_TYPE`?

lib/queuePopulator/QueuePopulatorExtension.js
```diff
@@ -73,6 +78,12 @@ class QueuePopulatorExtension {
     endBatch() {
         this._batch = null;
     }
+
+    getMetrics() {
+        const tempStore = Object.assign({}, this._metricsStore);
```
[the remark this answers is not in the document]
philipyoo: sorry I should've renamed this method better. This method, when called, should provide currently stored metrics and reset them after sending. Renaming this to `sendAndResetMetrics()`

lib/BackbeatConsumer.js
philipyoo: Great idea :)

lib/MetricsProducer.js
philipyoo: The `cb` is refactored out now and instead, if the producer returns an error, `log.trace` but continue the forEach loop :) Thanks!

lib/queuePopulator/LogReader.js
philipyoo: Ya thats a good idea 👍  this and in `MetricsProducer.publishMetrics` https://github.com/scality/backbeat/pull/125/files/aefeaf3118bfc397794c9255ca249a339f765670#diff-5062767101f20bda5a3063b13a8d46adR52
Both look like good opportunities for async calls

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -23,9 +23,23 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
         if (queueEntry.getReplicationStatus() !== 'PENDING') {
             return;
         }
+
         this.publish(this.repConfig.topic,
                      `${queueEntry.getBucket()}/${queueEntry.getObjectKey()}`,
                      JSON.stringify(entry));
+
+        if (queueEntry.getReducedLocations) {
```
jonathan-gramain: For info:

Although for now only entries to be replicated go through here, and they all should be of type `QueueEntry` which has `getReducedLocations` function, the pending PR for echo mode splits this class into ObjectQueueEntry and BucketQueueEntry, depending on whether the log entry is an object update or bucket creation.

In such case, instead of doing this metrics computation here, it can make more sense to have a `ObjectQueueEntry.getMetrics()` and possibly `BucketQueueEntry.getMetrics()` methods (or `updateMetrics()` methods if it's more appropriate), because at the end the metrics you want will depend on the type of log entry processed. So the logic of checking the presence of `getReducedLocations` may have to be changed then.

For now, let's leave as-is though.

lib/MetricsProducer.js
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * @param {Object} extMetrics - an object where keys are all buckets for a
+     *   given extension and values are the metrics for the bucket
+     *   (i.e. { my-bucket: { ops: 1, bytes: 124 } } )
+     * @param {String} type - type of metric (queueud or processed)
+     * @param {String} ext - extension (i.e. 'crr')
+     * @return {undefined}
+     */
+    publishMetrics(extMetrics, type, ext) {
```
jonathan-gramain: I would advise to add a `cb` callback argument when all metrics have been effectively published. It can be optional, if the caller wants fire&forget, but in many cases it can be useful to know this. Otherwise, if you prefer to leave as is, make sure you add a comment in the JSDoc stating that the actual processing is asynchronous, so the caller does not expect the metrics to be effectively published as soon as the function returns.

lib/queuePopulator/LogReader.js
```diff
@@ -365,7 +372,18 @@ class LogReader {
                 batchState.publishedEntries[topic] = topicEntries;
                 return done();
             });
-        }, done);
+        }, err => {
+            // TODO: On error, produce error metrics
+            if (err) {
+                return done(err);
+            }
+            return async.each(this._extensions, (ext, cb) => {
```
jonathan-gramain: Since you don't wait for a `publishMetrics` callback, the calls are synchronous here so you should just use a simple `for` loop, or `this._extensions.forEach(...)`.

lib/queuePopulator/QueuePopulator.js
```diff
@@ -51,18 +56,29 @@ class QueuePopulator {
      */
     open(cb) {
         this._loadExtensions();
-        this._setupZookeeper(err => {
+        async.series([
+            next => this._setupZookeeper(next),
+            next => process.nextTick(() => {
```
jonathan-gramain: You can avoid an unnecessary event loop iteration by integrating the `setupLogSources()` call into the next function in the series, just before calling `this._setupMetricsProducer()`, it will also slightly enhance clarity.

lib/queuePopulator/QueuePopulatorExtension.js
```diff
@@ -73,6 +78,34 @@ class QueuePopulatorExtension {
     endBatch() {
         this._batch = null;
     }
+
+    /**
+     * Get currently stored metrics and reset the counters
+     * @return {Object} metrics accumulated since last called
+     */
+    getAndResetMetrics() {
+        const tempStore = Object.assign({}, this._metricsStore);
```
jonathan-gramain: Here just do a simple assignment `tempStore = this._metricsStore`, no need to copy.

extensions/replication/queueProcessor/QueueProcessor.js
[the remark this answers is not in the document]
jonathan-gramain: Yes, sounds good

lib/queuePopulator/QueuePopulatorExtension.js
philipyoo: very cool, I thought I had to dupe the object because it would hold reference to the original object

lib/queuePopulator/QueuePopulator.js
philipyoo: Do you mean to add it to the series with `next => this._setupLogSources(next)`?
If so, I'll be adding a callback to `setupLogSources` which just gets invoked at the end of its function?

lib/MetricsProducer.js
philipyoo: I've added a cb to this, but the changes only currently log an error message. Especially for `LogReader`, I didn't want error from publishing metrics to stop the current batch from completing processing. Maybe it's not something I should be worried about? Not exactly sure

lib/queuePopulator/LogReader.js
philipyoo: With the above comment  (https://github.com/scality/backbeat/pull/125#discussion_r148858655), a callback has been added to the `publishMetrics` method 👍

lib/MetricsProducer.js
jonathan-gramain: Even logging at the end is a good reason to provide a callback to the function. I agree we should not impact the main batch if the metrics cannot be published. The main point here is that allowing a user to be notified when the operation completes does no harm an allows a few things to happen (e.g. debug logging at the end, in case we need to troubleshoot), even though the critical path does not need to wait on this callback.

It may also be used to avoid stacking operations when there is already one in progress (if publish metrics for some reason takes more time than the interval between two batches). In such case it can be cleaner to defer the next publish operation. Not that we have to do this now, just mentioning as a possible use case for a callback.

philipyoo:
@ironman-machine try

extensions/replication/ReplicationQueuePopulator.js
philipyoo: Just to note, I've added a method `getBytesMetric` to `ObjectQueueEntry`. For now, I am not collecting `BucketQueueEntry` as I'm unsure on how we want to display this metric. Do we just collect the number of bucket entries and display replicated buckets? Also, I believe this only applies to Echo

lib/BackbeatConsumer.js
```diff
@@ -100,6 +106,23 @@ class BackbeatConsumer extends EventEmitter {
                         offset,
                     });
                     this.emit('error', err, entry);
+                } else if (entry.topic === 'backbeat-replication') {
```
jonathan-gramain: can you use the value from the config (`config.extensions.replication.topic`)?

jonathan-gramain:
@ironman-machine try

lib/BackbeatConsumer.js
philipyoo: just a heads up for this, for now, I'll need to require the `Config` directly in this file. If it makes sense to, later on I can refactor by passing the crr topic directly to the consumer as a separate param

jonathan-gramain: OK for now, go for it.

conf/config.joi.js
```diff
@@ -60,6 +63,16 @@ const joiSchema = {
             },
         },
     },
+    redis: {
+        name: joi.string().required(),
+        password: joi.string().allow(''),
```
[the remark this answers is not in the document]
philipyoo: This password is set in Federation `group_vars/all` and shouldn't be used by the user directly. I think I'll remove this altogether from `config.json` and `config.joi.js`, and should set the value directly in `Config.js`.

jonathan-gramain:
@ironman-machine 
DEFAULT_BRANCH=rel/7.4
SCALITY_FEDERATION_BRANCH=FT/S3C-354-metrics-topic
try

jonathan-gramain:
@ironman-machine try

philipyoo:
@ironman-machine
SCALITY_FEDERATION_BRANCH=FT/S3C-354-metrics-topic
try

jonathan-gramain:
Integration passed here: http://ci.ironmann.io/gh/scality/Integration/19675

Requires the Federation PR to be merged first (https://github.com/scality/Federation/pull/1368)

philipyoo:
@ironman-machine try

jonathan-gramain:
@ironman-machine clean

jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine r-

jonathan-gramain:
@ironman-machine 
r+

PR #126, opened 2017-10-31
Ft/S3C-354 add metrics consumer

philipyoo:
This PR relies on https://github.com/scality/backbeat/pull/125

This PR is focused on consuming the metrics from Kafka and storing to Redis.

These Redis metric keys will be exposed by the Backbeat Server/API (https://github.com/scality/backbeat/pull/102) through to-be-built routes.

### DIFF
--- a/lib/MetricsConsumer.js
+++ b/lib/MetricsConsumer.js
@@ -1,0 +1,98 @@
+'use strict'; // eslint-disable-line strict
+
+const Logger = require('werelogs').Logger;
+const { RedisClient, StatsClient } = require('arsenal').metrics;
+
+const BackbeatConsumer = require('./BackbeatConsumer');
+const redisKeys = require('../extensions/replication/constants').redisKeys;
+
+// StatsClient constant defaults
+const INTERVAL = 300; // 5 minutes;
+const EXPIRY = 900; // 15 minutes
+
+// BackbeatConsumer constant defaults
+const CONSUMER_FETCH_MAX_BYTES = 5000020;
+const CONCURRENCY = 10;
+
+class MetricsConsumer {
+    /**
+     * @constructor
+     * @param {object} rConfig - redis configurations
+     * @param {string} rConfig.host - redis host
+     * @param {number} rConfig.port - redis port
+     * @param {object} mConfig - metrics configurations
+     * @param {string} mConfig.topic - metrics topic name
+     * @param {object} zkConfig - zookeeper configurations
+     * @param {string} zkConfig.connectionString - zookeeper connection string
+     *   as "host:port[/chroot]"
+     */
+    constructor(rConfig, mConfig, zkConfig) {
+        this.mConfig = mConfig;
+        this.zkConfig = zkConfig;
+
+        this.logger = new Logger('Backbeat:MetricsConsumer');
+
+        const redisClient = new RedisClient(rConfig, this.logger);
+        this._statsClient = new StatsClient(redisClient, INTERVAL,
+            EXPIRY);
+    }
+
+    start() {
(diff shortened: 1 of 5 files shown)


lib/MetricsConsumer.js
```diff
@@ -0,0 +1,90 @@
...
+     * @param {object} mConfig - metrics configurations
+     * @param {string} mConfig.topic - metrics topic name
+     * @param {object} zkConfig - zookeeper configurations
+     * @param {string} zkConfig.connectionString - zookeeper connection string
+     *   as "host:port[/chroot]"
+     */
+    constructor(rConfig, mConfig, zkConfig) {
+        this.mConfig = mConfig;
+        this.zkConfig = zkConfig;
+
```
bennettbuchanan: Could use validation of the config values to protect against anything unexpected.

lib/MetricsConsumer.js
```diff
@@ -0,0 +1,90 @@
...
+    start() {
+        const consumer = new BackbeatConsumer({
+            zookeeper: { connectionString: this.zkConfig.connectionString },
+            topic: this.mConfig.topic,
+            groupId: 'backbeat-metrics-group',
+            concurrency: CONCURRENCY,
+            queueProcessor: this.processKafkaEntry.bind(this),
+            fetchMaxBytes: CONSUMER_FETCH_MAX_BYTES,
+        });
+        consumer.on('error', () => {});
```
bennettbuchanan: Seems like `BackbeatConsumer` error event handling would be okay. How come changing this here?

philipyoo: I'm not entirely sure why emitted errors are consumed here. I followed the same as how `QueueProcessor` starts a BackbeatConsumer: https://github.com/scality/backbeat/blob/b0886dca99daca31855065f78602f806946f2960/extensions/replication/queueProcessor/QueueProcessor.js#L113

lib/MetricsConsumer.js
philipyoo: I feel like validating within the constructor seems a little too much because it basically would mean the wrong params were passed to the class. This could translate to any other class and method out there. I see classes like `BackbeatProducer` using joi validation, but it looks like it's mainly because all params are passed together within the same object.

lib/MetricsConsumer.js
jonathan-gramain: Since errors are logged in the BackbeatConsumer class, and if we don't have any specific processing to do, this should be good.

lib/MetricsConsumer.js
```diff
@@ -0,0 +1,90 @@
...
+        consumer.subscribe();
+
+        this.logger.info('metrics processor is ready to consume entries');
+    }
+
+    processKafkaEntry(kafkaEntry, done) {
+        const data = JSON.parse(kafkaEntry.value);
```
jonathan-gramain: Could be good to enclose in a `try/catch` or use `safeJsonParse` helper.

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

bennettbuchanan:
@ironman-machine 
DEFAULT_BRANCH=rel/7.4
try

bennettbuchanan:
@ironman-machine clean

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine clean

philipyoo:
@jonathan-gramain :( can you please re-open this for me?

PR #127, opened 2017-10-31
bf: don't overwrite part size in locations

jonathan-gramain:
When fetching locations, don't modify the original array of locations
when constructing the reduced locations array, to not induce
side-effects when reusing the original array (e.g to fetch the same
reduced locations again for another get/put cycle).

7.1 fix only, no need for forward port as master has a different
implementation and is not affected.


jonathan-gramain:
CI tests passed here: http://ci.ironmann.io/gh/scality/Integration/16632.

PR #128, opened 2017-11-03
FT: Add doc for replication to AWS setup

bennettbuchanan:
Add walkthrough documentation for setting up replication to AWS.

### DIFF
--- a/docs/crr-to-aws-s3.md
+++ b/docs/crr-to-aws-s3.md
@@ -2,7 +2,7 @@
 
 ## Installation
 
-AWS S3 endpoint (aws-us-east-1 for example) will be defined as a location
+AWS S3 endpoint (`awsbackend` for example) will be defined as a location
 constraint in the locationConstraint section of
 [CloudServer](https://github.com/scality/S3). The locationConstraint config has
 properties to set the bucket name, region, credentials for AWS S3.This enables
@@ -14,7 +14,7 @@ CloudServer
 
 ```
 env_replication_endpoints:
-  site: aws-us-east-1
+  site: awsbackend
   type: aws_s3
 ```
 
@@ -61,7 +61,7 @@ env_replication_endpoints:
   /_/backbeat MD request on source to update the properties replicationStatus to
   COMPLETED, x-amz-scal-version-id to the value received in the header.
 
-#### CloudServer PUT /_/backbeat Data for AWS route
+#### CloudServer PUT /\_/backbeat Data for AWS route
 
 * If the request has header `x-scal-replication-storage-type: aws_s3`, use the
   multipleDataBackend.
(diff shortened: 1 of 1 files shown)


docs/crr-to-aws-s3.md
```diff
@@ -75,3 +75,334 @@ env_replication_endpoints:
...
+    --bucket <target-bucket> \
+    --versioning-configuration Status=Enabled \
+    --profile <target-profile>
+    ```
+
+3. Generate account credentials for the source site using the command:
+
+    ```
+    ansible-playbook -i env/client-template/inventory \
+    tooling-playbooks/generate-account-accesskey.yml
```
jonathan-gramain: *`generate-account-access-key.yml`

docs/crr-to-aws-s3.md
```diff
@@ -75,3 +75,334 @@ env_replication_endpoints:
...
+    ```
+
+4. At the source, log in to the docker shell of a Backbeat container:
+
+    ```
+    docker exec -it scality-backbeat-1 bash
```
jonathan-gramain: On light-local it appears as `scality-backbeat-127.0.0.2-1`, not sure on a full stack deployment but it could be the same.

docs/crr-to-aws-s3.md
```diff
@@ -75,3 +75,334 @@ env_replication_endpoints:
...
+    ```
+
+    ```
+    aws s3api get-bucket-versioning \
+    --bucket <target-bucket> \
+    --profile <target-profile>
+    ```
+
+3. Establish a bucket policy to allow the Backbeat replication engine to
+    retrieve the appropriate information. Create the file _policy.json_ with the
```
jonathan-gramain: Not only retrieve, but also write data/metadata. Maybe rephrase to `to execute the required actions`?

docs/crr-to-aws-s3.md
```diff
@@ -75,3 +75,334 @@ env_replication_endpoints:
...
+                ]
+            },
+            {
+                "Effect": "Allow",
+                "Action": [
+                    "s3:ReplicateObject",
+                    "s3:ReplicateDelete",
+                    "s3:ReplicateTags"
+                ],
+                "Resource": "arn:aws:s3:::bennett-federationtest/*"
```
jonathan-gramain: For an example AWS bucket name, maybe use a more self-explanatory name like `bennett-aws-bucket`? Or stick with the generic `<target-bucket>`. Maybe same applies above for the source bucket.

docs/crr-to-aws-s3.md
```diff
@@ -75,3 +75,334 @@ env_replication_endpoints:
 
 * Backbeat updates the source object's metadata to store AWS' version id and set
   replication status as `COMPLETED`
+
+## Setting up Cross-Region Replication (CRR) to AWS
+
+This section contains steps to set up cross-region replication from Scality S3
+to Amazon Web Services S3 using the Backbeat replication engine. Two subsections
+are provided with different methods: the subsection "Basic Example" uses a
```
jonathan-gramain: The explanation about the subsections may go directly into their respective "Basic" and "Detailed" subsections introductory paragraph IMO. And instead of "Basic" and "Detailed", names like "Half-Automated" and "Manual" (for example) seem more appropriate to me.

jonathan-gramain (commented):
Great to see both the script version and manual version, so the user can see what's going on under the hood if he wants too :)

A few minor comments

docs/crr-to-aws-s3.md
bennettbuchanan: Nice-catch. Thanks!

docs/crr-to-aws-s3.md
bennettbuchanan: On a full stack deployment the name of the docker container is `scality-backbeat-1`. I will add a note here about the difference.

docs/crr-to-aws-s3.md
bennettbuchanan: Yes, this is meant to be `<target-bucket>`! That is left over from when I was testing :smile:.

docs/crr-to-aws-s3.md
bennettbuchanan: I like that idea! Updated.

docs/crr-to-aws-s3.md
bennettbuchanan: Agree—that is better wording. Updated.

philipyoo (approved):

docs/crr-to-aws-s3.md
```diff
@@ -75,3 +75,340 @@ env_replication_endpoints:
...
+    --profile <source-profile>
+    ```
+
+    ```
+    aws s3api get-bucket-versioning \
+    --bucket <target-bucket> \
+    --profile <target-profile>
+    ```
+
+3. Establish a bucket policy to allow the Backbeat replication engine to
```
rahulreddy: Let's not call it a bucket policy since bucket policies are a whole different game. `IAM resource policy` would be correct.
```
Establish an IAM resource policy on the source bucket to allow the Backbeat to
execute the required actions.
```

rahulreddy (changes requested):

docs/crr-to-aws-s3.md
```diff
@@ -75,3 +75,340 @@ env_replication_endpoints:
...
+    ```
+    [<source-profile>]
+    aws_access_key_id = <access-key>
+    aws_secret_access_key = <secret-key>
+    ```
+
+6. Set up replication on your buckets:
+
+    ```
+    node ~/replication/backbeat/bin/setupReplication.js setup \
```
bennettbuchanan: The path should be `bin/replication.js`

PR #132, opened 2017-11-10
FIX: Support multiple backend configuration

bennettbuchanan:
Allow omission of a cloud server site with
that includes the servers property

### DIFF
--- a/bin/replication.js
+++ b/bin/replication.js
@@ -35,7 +35,8 @@ function _createSetupReplication(command, options, log) {
               new SharedIniFileCredentials({ profile: sourceProfile });
     const targetCredentials =
               new SharedIniFileCredentials({ profile: targetProfile });
-
+    const destinationEndpoint =
+        destination.bootstrapList.find(dest => Array.isArray(dest.servers));
     return new SetupReplication({
         source: {
             bucket: sourceBucket,
@@ -47,7 +48,8 @@ function _createSetupReplication(command, options, log) {
         target: {
             bucket: targetBucket,
             credentials: targetCredentials,
-            hosts: new RoundRobin(destination.bootstrapList[0].servers),
+            hosts: targetIsExternal ?
+                undefined : new RoundRobin(destinationEndpoint.servers),
             transport: destination.transport,
             isExternal: targetIsExternal,
             siteName,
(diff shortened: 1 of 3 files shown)


bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -63,14 +64,13 @@ class QueueProcessor {
...
-        // FIXME support multiple destination sites
-        if (destConfig.bootstrapList.length > 0) {
-            this.destHosts =
-                new RoundRobin(destConfig.bootstrapList[0].servers,
-                               { defaultPort: 80 });
-        } else {
-            this.destHosts = null;
-        }
+        // FIXME support multiple cloud-server destination sites
+        destConfig.bootstrapList.forEach(dest => {
```
philipyoo: in the config.joi, `bootstrapList` is not a required field, so using the `forEach` here might cause a problem. Maybe we can guarantee by making `bootstrapList` a required field in the config.joi file? Are there scenarios where `bootstrapList` is not used?

Also using `forEach` here might not be best because that if we have multiple destination servers, we would be looping through all of them and setting a new `RoundRobin` each time

bennettbuchanan: I was going off the fact that we were already checking the length of `bootstrapList`. But I will add a conditional here to be sure.

I've included the 'FIXME' comment to refer to the second case, since we don't support multiple desintation servers anyway, I think it's okay for now to assume that the `bootstrapList` object with a `servers` property is the one that we want.

bennettbuchanan:
@ironman-machine try

bin/replication.js
```diff
@@ -47,7 +48,7 @@ function _createSetupReplication(command, options, log) {
         target: {
             bucket: targetBucket,
             credentials: targetCredentials,
-            hosts: new RoundRobin(destination.bootstrapList[0].servers),
+            hosts: new RoundRobin(destinationEndpoint.servers),
```
philipyoo: if destination is `aws_s3`, this could be a problem since there is no `servers` attribute on it

This may require changes here and also in class `SetupReplication` since it relies on `RoundRobin` objects for the `hosts` attribute

bennettbuchanan: You're right, pushed an update.

bennettbuchanan:
@ironman-machine try

rahulreddy:
Jira issue for reference https://scality.atlassian.net/browse/S3C-1071

PR #133, opened 2017-11-11
FIX: Update all error origins for AWS task

bennettbuchanan:
On retry during an error we do not want to use the round robin `pickNextHost` method since `destHosts` is not set in a configuration to replicate to AWS.

See: https://github.com/scality/backbeat/blob/master/extensions/replication/queueProcessor/QueueProcessorTask.js#L157


bennettbuchanan:
@ironman-machine try

PR #134, opened 2017-11-11
FIX: Send additional headers to replication to AWS

bennettbuchanan:


bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E failed due to flaky Utapi test:
```
Utapi tests Utapi: USERS:: Basic UserPolicy BasicUserPolicy:: Attaching a policy to Bart that grant him access to getMetrics should allow Bart to access bucketMetrics
```

Rerunning now.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

rahulreddy:
Got retried here http://ci.ironmann.io/gh/scality/Integration/17000

extensions/replication/utils/QueueEntry.js
```diff
@@ -73,6 +73,20 @@ class QueueEntry extends ObjectMD {
...
 
+    getUserMetadata() {
+        const metaHeaders = {};
+        const data = this.getValue();
+        Object.keys(data).forEach(key => {
+            if (key.startsWith('x-amz-meta-')) {
+                metaHeaders[key] = data[key];
+            }
+        });
+        if (metaHeaders !== {}) {
```
rahulreddy: This conditional will always be true.

```js
> x = {}
{}
> x === {}
false
> x !== {}
true
```
Best thing is to use `Object.keys.length > 0`

bennettbuchanan: Updated to `if (Object.keys(metaHeaders).length > 0)`

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E failed due to what appears to be a flaky multiple backend test, as this particular test was observed previously to both pass and fail on the same branch.

Rerunning here: http://ci.ironmann.io/gh/scality/Integration/17007.

bennettbuchanan:
E2E passed on the retry here: http://ci.ironmann.io/gh/scality/Integration/17007

PR #135, opened 2017-11-17
FIX: Source should get destination version for AWS

bennettbuchanan:
Depends on https://github.com/scality/S3/pull/1010.

Object tagging operations that are made before the source object has completed replication do not have the `datastoreVersionId` in the Kafka entry. 

Because all operations for the multiple backend are serialized, when any tagging operation is being replicated we can guarantee that the source object's metadata will have been updated at this point. Hence we send the source bucket and source version ID for cloud server to retrieve the information before making the request to an external location. Also, we can eliminate a GET request for the source object's tags because they are already available in the Kafka entry.

### DIFF
--- a/extensions/replication/queueProcessor/MultipleBackendTask.js
+++ b/extensions/replication/queueProcessor/MultipleBackendTask.js
@@ -380,112 +380,47 @@ class MultipleBackendTask extends QueueProcessorTask {
         });
     }
 
-    _getAndPutObjectTagging(sourceEntry, destEntry, log, cb) {
+    _putObjectTagging(sourceEntry, destEntry, log, cb) {
         this._retry({
             actionDesc: 'send object tagging XML data',
             entry: sourceEntry,
-            actionFunc: done => this._getAndPutObjectTaggingOnce(
-               sourceEntry, destEntry, log, done),
+            actionFunc: done => this._putObjectTaggingOnce(
+                sourceEntry, destEntry, log, done),
             shouldRetryFunc: err => err.retryable,
             log,
         }, cb);
     }
 
-    _getAndPutObjectTaggingOnce(sourceEntry, destEntry, log, cb) {
+    _putObjectTaggingOnce(sourceEntry, destEntry, log, cb) {
         const doneOnce = jsutil.once(cb);
         log.debug('replicating object tags', {
             entry: sourceEntry.getLogInfo(),
         });
-        const sourceReq = this.S3source.getObjectTagging({
-            Bucket: sourceEntry.getBucket(),
-            Key: sourceEntry.getObjectKey(),
-            VersionId: sourceEntry.getEncodedVersionId(),
-        });
-        attachReqUids(sourceReq, log);
-        sourceReq.on('error', err => {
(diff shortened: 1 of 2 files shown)


extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -380,112 +380,48 @@ class MultipleBackendTask extends QueueProcessorTask {
...
-                    peer: this.sourceConfig.s3,
+        attachReqUids(destReq, log);
+        return destReq.send((err, data) => {
+            if (err) {
+                // eslint-disable-next-line no-param-reassign
+                err.origin = 'source';
+                log.error('an error occurred putting object tagging to S3', {
+                    method: 'MultipleBackendTask._putObjectTaggingOnce',
+                    entry: destEntry.getLogInfo(),
+                    origin: 'target',
```
electrachong: is it intentional that the origin for the error and here differ?

bennettbuchanan: It is doing nothing here so I've removed it. 

The `err.origin` is set to 'source' throughout this task so that in a retry it never enters [this kind of condition](https://github.com/scality/backbeat/blob/master/extensions/replication/queueProcessor/QueueProcessorTask.js#L156) (as this class is inherited). For the log though, the error should come from AWS so I've left it at as 'target'. It looks like all these could removed as the check is `err.origin === 'target'`.

bennettbuchanan:
@ironman-machine 
SCALITY_S3_BRANCH=fix/S3C-1100/get-destination-version-id-from-metadata
try

bennettbuchanan:
@ironman-machine
SCALITY_S3_BRANCH=
try

extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -380,112 +380,48 @@ class MultipleBackendTask extends QueueProcessorTask {
...
-                    origin: 'source',
-                    peer: this.sourceConfig.s3,
+        attachReqUids(destReq, log);
+        return destReq.send((err, data) => {
+            if (err) {
+                log.error('an error occurred putting object tagging to S3', {
+                    method: 'MultipleBackendTask._putObjectTaggingOnce',
+                    entry: destEntry.getLogInfo(),
+                    origin: 'target',
+                    peer: this.destBackbeatHost,
```
jonathan-gramain: I think there's no `peer` to log here.

bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine
SCALITY_S3_BRANCH=fix/S3C-1100/get-destination-version-id-from-metadata
try

rahulreddy:
@ironman-machine clean after=https://github.com/scality/S3/pull/1010

bennettbuchanan:
@ironman-machine try

rahulreddy:
No need to try. I will r+

rahulreddy:
@ironman-machine clean r+

PR #137, opened 2017-11-17
bf: remove pause() call on Kafka errors

jonathan-gramain:
It's not necessary to pause the consumer when the Kafka client returns
errors, and it may block the consumption for no good reason, so better
remove it.


rahulreddy:
@ironman-machine r+

rahulreddy:
http://ci.ironmann.io/gh/scality/Integration/17139

rahulreddy:
@ironman-machine r- r+

PR #138, opened 2017-11-20
Fix/dependencies

ThibaultRiviere:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "1.0.0",
```
DavidPineauScality: I would leave that to @jonathan-gramain to see if we wants to bump the version but I doubt it.

rahulreddy:
Since this is not used as an npm module, we should be fine without having to bump the version.

ThibaultRiviere:
@rahulreddy i didn't dump the version here. And having the good branching for the dependencies is not a bad thing here.

PR #139, opened 2017-11-22
Ft/S3C-1001 add metric routes

philipyoo:
This PR to be merged after https://github.com/scality/backbeat/pull/153
Need to drop a commit before merging. Redis configs defined in
https://github.com/scality/backbeat/pull/125 needed for this PR.

This PR is focused on exposing metrics from Redis to API routes defined.

Example output for the "getAllMetrics" route:

`
{"backlog":{"description":"Number of incomplete replication operations (count) and number of incomplete MB transferred (size)","results":{"count":0,"size":"0.00"}},"completions":{"description":"Number of completed replication operations (count) and number of MB transferred (size) in the last 900 seconds","results":{"count":0,"size":"0.00"}},"throughput":{"description":"Current throughput for replication operations in ops/sec (count) and MB/sec (size)","results":{"count":"0.00","size":"0.00"}}}
`

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -6,10 +6,8 @@ const zookeeper = require('node-zookeeper-client');
 const { errors } = require('arsenal');
 
 const BackbeatProducer = require('../BackbeatProducer');
+const Healthcheck = require('./Healthcheck');
 const routes = require('./routes');
-
-// In-Sync Replicas
-const ISRS = 3;
 
 /**
  * Class representing Backbeat API endpoints and internals
@@ -35,6 +33,7 @@ class BackbeatAPI {
         this._crrProducer = null;
         this._crrStatusProducer = null;
         this._metricProducer = null;
+        this._healthcheck = null;
         this._zkClient = null;
     }
 
@@ -61,33 +60,13 @@ class BackbeatAPI {
             && this._crrStatusProducer.isReady();
     }
 
-    _getConnectionDetails() {
-        return {
-            zookeeper: {
-                status: this._zkClient.getState().name === 'SYNC_CONNECTED' ?
-                    'ok' : 'error',
-                details: this._zkClient.getState(),
-            },
-            kafkaProducer: {
-                status: this._checkProducersReady() ? 'ok' : 'error',
-            },
-        };
-    }
-
     /**
-     * Checks health of in-sync replicas
-     * @param {object} md - topic metadata object
-     * @return {boolean} true if ISR health is ok
+     * Check if Zookeeper and Producer are connected
+     * @return {boolean} true/false
      */
-    _checkISRHealth(md) {
-        // eslint-disable-next-line consistent-return
-        const keys = Object.keys(md);
-        for (let i = 0; i < keys.length; i++) {
-            if (md[keys[i]].isr && md[keys[i]].isr.length !== ISRS) {
-                return 'error';
-            }
-        }
-        return 'ok';
+    isConnected() {
+        return this._zkClient.getState().name === 'SYNC_CONNECTED'
+            && this._checkProducersReady();
     }
 
     /**
@@ -96,47 +75,20 @@ class BackbeatAPI {
      * @return {undefined}
      */
     healthcheck(cb) {
-        const client = this._crrProducer.getKafkaClient();
-
-        // TODO: refactor by calling specific topic
-        client.loadMetadataForTopics([], (err, res) => {
+        return this._healthcheck.getHealthcheck((err, data) => {
             if (err) {
-                this._logger.error('error getting healthcheck');
-                return cb(errors.InternalError
-                    .customizeDescription('error getting healthcheck metadata' +
-                    ' for topics'));
+                this._logger.error('error getting healthcheck', err);
+                return cb(errors.InternalError);
             }
-            const response = res.map(i => (Object.assign({}, i)));
-            const connections = {};
-            const topicMD = {};
-            response.forEach((obj, idx) => {
-                if (obj.metadata && obj.metadata[this._repConfig.topic]) {
-                    let copy;
-                    try {
-                        copy = JSON.parse(JSON.stringify(obj.metadata[
-                            this._repConfig.topic]));
-                        topicMD.metadata = copy;
-                        response.splice(idx, 1);
-                    } catch (e) {
-                        this._logger.error('error getting topic metadata');
-                    }
-                }
-            });
-            response.push(topicMD);
-
-            if (topicMD.metadata) {
-                connections.isrHealth = this._checkISRHealth(topicMD.metadata);
-            }
-
-            Object.assign(connections, this._getConnectionDetails());
-            response.push({
-                internalConnections: connections,
-            });
-
-            return cb(null, response);
+            return cb(null, data);
         });
     }
 
+    /**
+     * Setup internals
+     * @param {function} cb - callback(error)
+     * @return {undefined}
+     */
     setupInternals(cb) {
         async.parallel([
             done => this._setZookeeper(done),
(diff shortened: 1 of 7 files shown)


lib/api/BackbeatAPI.js
```diff
@@ -113,6 +135,114 @@ class BackbeatAPI {
         });
     }
 
+    /**
+     * Get replication backlog in bytes and count
+     * @param {function} cb -callback(error, data)
+     * @return {undefined}
+     */
+    backlog(cb) {
+        async.series([
```
jonathan-gramain: It should be fine to query them in parallel, to get that little gain in latency.

lib/api/BackbeatAPI.js
```diff
@@ -113,6 +135,114 @@ class BackbeatAPI {
...
+        });
+    }
+
+    /**
+     * Get completed replicated stats by bytes and count
+     * @param {function} cb - callback(error, data)
+     * @return {undefined}
+     */
+    completions(cb) {
+        async.series([
```
jonathan-gramain: Same for series -> parallel

lib/api/BackbeatAPI.js
```diff
@@ -113,6 +135,114 @@ class BackbeatAPI {
...
+            next => this._queryStats(OPS_DONE, next),
+            next => this._queryStats(BYTES_DONE, next),
+        ], (err, res) => {
+            if (err) {
+                this._logger.trace('error getting replication completions');
+                return cb(errors.InternalError);
+            }
+            const data = {
+                completions: {
+                    details: 'Replciations completed in bytes and count in ' +
```
jonathan-gramain: typo: Replications

I would also rephrase to something more explicit for what is count and bytes, so something like

Number of completed replication operations (count) and number of bytes transferred (bytes) in the last ... seconds

lib/api/BackbeatAPI.js
```diff
@@ -113,6 +135,114 @@ class BackbeatAPI {
...
+     * @param {function} cb = callback(error, data)
+     * @return {undefined}
+     */
+    bandwidth(cb) {
+        this._queryStats(BYTES_DONE, (err, res) => {
+            if (err) {
+                this._logger.trace('error getting bandwidth');
+                return cb(errors.InternalError);
+            }
+            const bandwidth = Math.floor(
```
jonathan-gramain: Why double `Math.floor`?

Actually you could provide some decimal digits since MB/s is kind of rough precision (especially if you CRR to a distant public cloud), though 2 or 3 more decimals should be enough. So you could just get rid of Math.floor and compute with a simple division, or use `toFixed()` if you wish but outputting a plain full precision floating point value should be fine.

lib/api/BackbeatAPI.js
```diff
@@ -113,6 +135,114 @@ class BackbeatAPI {
...
+                Math.floor(res.requests / res.sampleDuration) / 1000
+            );
+
+            const data = {
+                bandwidth: {
+                    details: 'CRR bandwidth in MBs/sec',
```
jonathan-gramain: `MB/s`

lib/api/BackbeatAPI.js
```diff
@@ -113,6 +135,114 @@ class BackbeatAPI {
...
+                    details: 'CRR throughput in ops/sec',
+                    value: throughput,
+                },
+            };
+            return cb(null, data);
+        });
+    }
+
+    /**
+     * Get current bandwidth (MBs/sec)
```
jonathan-gramain: `MB/s` or `MB/sec`

lib/api/BackbeatAPI.js
```diff
@@ -113,6 +135,114 @@ class BackbeatAPI {
...
+            next => this._queryStats(OPS, next),
+            next => this._queryStats(BYTES, next),
+        ], (err, res) => {
+            if (err) {
+                this._logger.trace('error occurred getting backlog');
+                return cb(errors.InternalError);
+            }
+            const data = {
+                backlog: {
+                    details: 'Replication backlog in bytes and count',
```
jonathan-gramain: See comment below for rephrasing more explicitly.

Personally I'm not fond of "details", I would rather use "description" which is more descriptive (sorry :sheep: )

As a side note, in the ring we ran into issues because it has so many configuration values and statistics and each one of them carries a lengthy description, leading to many superflous bytes transmitted for every statistics polling that it became an issue on its own. But for now it should not be a problem for us, and it's still good to have the description close to the data transmitted, if it eventually becomes an issue we can rethink this at a later time.

lib/api/BackbeatAPI.js
philipyoo: I've switched to converting total bytes in MBs to keep things uniform. Also, instead of naming the property "bytes" or "MBs", I've just switched to using "size"

Lmk if that sounds good, or if you prefer different property names 👍

lib/api/routes.js
```diff
@@ -7,4 +7,16 @@ module.exports = [
     {
         path: '/_/healthcheck',
     },
+    {
+        path: '/_/backlog',
```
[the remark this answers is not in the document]
philipyoo: With the current setup, I wasn't sure how best to setup a prepended "metrics" route. So currently my `routes.js` would look something like:
`{ path: '/_/metrics/backlog',  method: 'getBacklog' }`

Now instead of relying on the path and parsing it to determine which method to call, I will now match by path names and I use a predefined method for calling the right method

philipyoo:
I can't seem to figure out how to specify the PR https://github.com/scality/Arsenal/pull/369 in `package.json`. I see certain modules installing master branch of Arsenal, so might be just overriding my specified branch?

I believe tests should work with the PR https://github.com/scality/Arsenal/pull/369

philipyoo:
@ironman-machine
SCALITY_ARSENAL_BRANCH=ft/statsclient-add-incrbyEx
try

lib/api/BackbeatAPI.js
```diff
@@ -259,6 +259,43 @@ class BackbeatAPI {
...
 
+    /**
+     * Get all metrics
+     * @param {object} details - route details from lib/api/routes.js
+     * @param {function} cb = callback(error, data)
+     * @param {array} data - optional field providing already fetched data
+     *   in order of dataPoints mentioned for each route in lib/api/routes.js
+     * @return {undefined}
+     */
+    getAll(details, cb, data) {
```
jonathan-gramain: getAllMetrics would be a slightly better name

conf/Config.js
```diff
@@ -53,6 +53,12 @@ class Config {
             this.healthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
         }
 
+        // TODO: Drop this before merge
```
jonathan-gramain: Something to do here? :)

philipyoo: haha yup, this needs to be dropped before merging as it is a placeholder for redis configs (defined in metrics accumulator PR#125)

lib/api/BackbeatAPI.js
```diff
@@ -4,11 +4,17 @@ const async = require('async');
...
+const { RedisClient, StatsClient } = require('arsenal').metrics;
 
 const BackbeatProducer = require('../BackbeatProducer');
 const Healthcheck = require('./Healthcheck');
 const routes = require('./routes');
 
+// StatsClient constant defaults
+// TODO: This should be moved to constants file
+const INTERVAL = 300; // 5 minutes
+const EXPIRY = 900; // 15 minutes
```
jonathan-gramain: It would be good to have those two values configurable through config.json, but let's stick with this for now.

rahulreddy:
@ironman-machine clean try

jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine r- r+

PR #140, opened 2017-11-28
bf: allow retries on errors in CredentialsManager.refresh()

jonathan-gramain:
Set the 'retryable' field in the errors generated from
CredentialsManager.refresh() when relevant, to be in sync with AWS SDK
behavior, which is also how the calling code of the SDK checks whether
it should retry the request.

The criteria are when the error is an InternalError or
ServiceUnavailble we retry, otherwise we don't.


rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

PR #143, opened 2017-12-05
rf: rename bin/replication.js to bin/setupReplication.js

jonathan-gramain:
It's better to have the action verb as part of the script name that
tells what it's doing.


jonathan-gramain:
@ironman-machine r+

rahulreddy:
I would not recommend this.
`node bin/setupReplication setup` looks redundant and we also have another command that this tool supports `validate`. And `node bin/setupReplication validate` doesn't make sense. I recommend keeping it as is and the fix the docs.

jonathan-gramain:
@rahulreddy I tend to agree with you on most part, but I'm thinking that `replication.js` is not telling enough about what it is (i.e. a tool to help with a replication setup). But I guess there's no ideal name and we have to use one. (btw `node bin/setupReplication validate` does not disturb me as it can be seen as validating the setup previously done).

This being said I don't think it's worth arguing too much for a tool name, I'm good sticking with what it is and fixing the doc.

PR #149, opened 2017-12-07
ft: support specifying partition in entries

philipyoo:
To support deep healthcheck, where I want to produce to all
partitions, allow for specifying a partition within an entry.

### DIFF
--- a/lib/BackbeatProducer.js
+++ b/lib/BackbeatProducer.js
@@ -128,17 +128,23 @@ class BackbeatProducer extends EventEmitter {
                 cb(errors.InternalError);
             });
         }
-        const payload = this._partition === undefined ?
-                  entries.map(item => ({
-                      topic: this._topic,
-                      messages: new KeyedMessage(item.key, item.message),
-                      key: item.key,
-                  })) :
-                  entries.map(item => ({
-                      topic: this._topic,
-                      messages: item.message,
-                      partition: this._partition,
-                  }));
+
+        const payload = entries.map(item => {
+            if (item.partition !== undefined || this._partition !== undefined) {
+                // specifiying partition takes precedence
+                return {
+                    topic: this._topic,
+                    messages: item.message,
+                    partition: item.partition || this._partition,
+                };
+            }
+            return {
+                topic: this._topic,
+                messages: new KeyedMessage(item.key, item.message),
+                key: item.key,
+            };
+        });
+
         this._client.refreshMetadata([this._topic], () =>
             this._producer.send(payload, err => {
                 if (err) {


lib/BackbeatProducer.js
```diff
@@ -128,17 +128,28 @@ class BackbeatProducer extends EventEmitter {
...
-                  }));
+
+        const payload = entries.map(item => {
+            if (this._partition === undefined && item.partition === undefined) {
+                return {
+                    topic: this._topic,
+                    messages: new KeyedMessage(item.key, item.message),
+                    key: item.key,
+                };
+            } else if (item.partition && this._partition === undefined) {
```
jonathan-gramain: You can get rid of the `else`.

Minor style comment: the tests would be simpler and more understandable if you use the following sequence instead:

```
if (this._partition) { return ...}
if (item.partition) { return ...}
return {...}
```

jonathan-gramain (dismissed):

lib/BackbeatProducer.js
```diff
@@ -128,17 +128,29 @@ class BackbeatProducer extends EventEmitter {
...
-                      key: item.key,
-                  })) :
-                  entries.map(item => ({
-                      topic: this._topic,
-                      messages: item.message,
-                      partition: this._partition,
-                  }));
+
+        const payload = entries.map(item => {
+            if (item.partition) {
```
rahulreddy: This needs to be `if (item.partition !== undefined)` since partition can be `0`.

lib/BackbeatProducer.js
```diff
@@ -128,17 +128,29 @@ class BackbeatProducer extends EventEmitter {
...
+                return {
+                    topic: this._topic,
+                    messages: item.message,
+                    partition: item.partition,
+                };
+            }
+            if (this._partition) {
```
rahulreddy: Same here `if (this._partition !== undefined)`

rahulreddy (changes requested):

philipyoo:
Sorry, one additional field I'd like to add for entries that have explicit partition declared. Added PR: 79de399

I realize when parsing entries in `QueueProcessor.processKafkaEntry`, I want to catch entries specific to deep healthcheck. By setting a `key` property for these entries, I can catch and ignore deep healthcheck entries.

jonathan-gramain:
@philipyoo I had a similar issue with testing to inject artificial messages and ensure the consumer would receive them, while not disturbing the regular backbeat consumer too much. For this I added a "bootstrapId" field to the JSON message, and backbeat consumer skips those entries (actually sends back an error but keeps going with next entries).

In case of healthchecks, we do not want to error but simply skip them in regular processing. I'm thinking `key` can be confused with other fields with the same name (object key, kafka key...), hence maybe just setting a `"healthcheck": true` or `"healthcheckKey": "foo"` field can do the trick, but make sure that it's clearly visible somehow that the entry comes from the healthcheck to ease troubleshooting.

philipyoo:
As @jonathan-gramain commented, rather than catching deep healthcheck entries by the key property, instead putting the identifier within the message property and filtering within `QueueEntry.createFromKafkaEntry`

philipyoo:
Implementation has changed with switch to node-rdkafka. Plan is now to produce to Kafka directly using the same topic and consumer group.

PR #152, opened 2017-12-14
rf: single producer for all clusters

philipyoo:
Multiple producers were being created for each cluster.
Rather, instantiating a single producer per topic
would be more efficient.

Also, I realized I created a producer for a metric topic only.
Instead, I should be checking the health of producers of all
topics, including metrics topic.

Should be merged after https://github.com/scality/backbeat/pull/102

### DIFF
--- a/lib/api/BackbeatServer.js
+++ b/lib/api/BackbeatServer.js
@@ -1,0 +1,225 @@
+'use strict'; // eslint-disable-line strict
+
+const http = require('http');
+
+const { Clustering, errors, ipCheck } = require('arsenal');
+
+const BackbeatRequest = require('./BackbeatRequest');
+const BackbeatAPI = require('./BackbeatAPI');
+
+const WORKERS = 4;
+
+class BackbeatServer {
+    /**
+     * @constructor
+     * @param {Worker} [worker=null] - Track the worker using cluster
+     * @param {object} config - configurations for server setup
+     * @param {werelogs.Logger} logger - Logger object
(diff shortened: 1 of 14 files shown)


lib/api/BackbeatServer.js
```diff
@@ -203,9 +205,40 @@ function run(config, Logger) {
     const logger = new Logger('BackbeatServer');
     const apiLogger = new Logger('BackbeatAPI');
     const cluster = new Clustering(WORKERS, logger);
-    const backbeatAPI = new BackbeatAPI(config, apiLogger);
+    let backbeatAPI;
 
-    backbeatAPI.setupInternals(err => {
```
jonathan-gramain: This was done before the workers were started (done below by `cluster.start()`), so I don't think there were multiple producers/consumers created in the previous code?

jonathan-gramain: And, even if we spawn a consumer per worker, Kafka advertises that maintaining a consumer is very cheap on server side, I don't know about the consumer client but it should be relatively low footprint for what it has to do, not sure this overhead is a real concern.

As discussed offline, I don't know if there is a need to have workers at all, unless there are CPU intensive tasks -
 @rahulreddy should we keep the cluster workers in your opinion?

philipyoo: Ah sorry I see now. For now, I will close this PR and revise the original healthcheck PR for checking all current BB topics 👍 

For the workers, I talked with @rahulreddy offline and he suggested we keep it to 2 workers for now in the case where something should happen to one of the workers. Switching to a single http server should be fairly quick if needed as well

PR #153, opened 2017-12-14
rf: move healthcheck

philipyoo:
Move healthcheck functions to its own class and add 
`BackbeatAPI.setup` for handling all internal setup for 
the api.

This is mainly in preparation for deep healthcheck where
I want to isolate all healthcheck related methods to its
own class

Should be merged after https://github.com/scality/backbeat/pull/152

PR #154, opened 2017-12-15
Ft/S3C-979 add deep healthcheck

philipyoo:
The goal is to use BackbeatProducer and BackbeatConsumer 
using the same replication topic, sending a message to every 
available partition, "consuming" the message, and returning 
a status for each partition.

Current PR only shows if a partition is working or not 
(i.e. `ok` || `error`) and also shows elapsed time to show how 
long the healthcheck took. 

Currently, a deep healthcheck max time for checking is set 
to 1 minute and interval checks of 500 milliseconds. These 
are hard-coded as a constant and can be changed to 
requested value.
Also, deep healthchecks can only be handled in parallel
depending on number of clusters. This is a TODO fix.

Need to drop https://github.com/scality/backbeat/commit/dfd85f5e675fc350650de62bd05ed66c5950c67e
and should be merged after https://github.com/scality/backbeat/pull/153

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -307,7 +307,13 @@ class QueueProcessor extends EventEmitter {
      * @return {undefined}
      */
     processKafkaEntry(kafkaEntry, done) {
+        // if (kafkaEntry.key && kafkaEntry.key.startsWith('deephealthcheck')) {
+        //     return process.nextTick(() => done());
+        // }
         const sourceEntry = QueueEntry.createFromKafkaEntry(kafkaEntry);
+        if (sourceEntry.healthcheck) {
+            return process.nextTick(() => done());
+        }
         if (sourceEntry.error) {
             this.logger.error('error processing source entry',
                               { error: sourceEntry.error });
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -307,9 +307,6 @@ class QueueProcessor extends EventEmitter {
      * @return {undefined}
      */
     processKafkaEntry(kafkaEntry, done) {
-        // if (kafkaEntry.key && kafkaEntry.key.startsWith('deephealthcheck')) {
-        //     return process.nextTick(() => done());
-        // }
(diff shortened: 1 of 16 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -287,7 +287,13 @@ class QueueProcessor extends EventEmitter {
      * @return {undefined}
      */
     processKafkaEntry(kafkaEntry, done) {
+        // if (kafkaEntry.key && kafkaEntry.key.startsWith('deephealthcheck')) {
```
bennettbuchanan: I guess this is meant to be removed?

lib/api/Healthcheck.js
```diff
@@ -101,6 +115,174 @@ class Healthcheck {
...
+                this._probe.setupStore(partitions);
+                this._createEntries(id, partitions, done);
+            },
+            (entries, done) => this._crrProducer.send(entries, done),
+            done => this._createConsumer('deephealthcheck-group', done),
+        ], (err, consumer) => {
+            if (err || consumer === undefined) {
+                const error = {
+                    method: 'Healthcheck.getDeepHealthcheck',
+                    error: err,
```
bennettbuchanan: May be good to add a custom error if `consumer` is undefined. Perhaps something alluding to the fact that the BackbeatConsumer was not created somehow. Then again, I think this case may ever happen because it is an instantiated class.

lib/api/Healthcheck.js
```diff
@@ -101,6 +115,174 @@ class Healthcheck {
...
+                interval: DH_INTERVAL,
+                errorFilter: err => err === 'retry',
+            },
+            this._waitAndCheck.bind(this, start),
+            (err, elapsed) => {
+                if (err) {
+                    // replace all undefined values as 'error'
+                    this._probe.setStoreErrors();
+                }
+                this._probe.setStoreData('timeElapsed', elapsed);
```
bennettbuchanan: Looks like this needs to be in an `else` block.

philipyoo: I want to set this property no matter success or failure. Just thought to add a time elapsed indicator since the time it takes for a request to finish is unknown 👍

PR #155, opened 2017-12-16
queue replication status updates in kafka

jonathan-gramain:
Use a separate kafka queue to publish updates of replication status,
instead of updating it directly.

The immediate need for this is to support multiple replication
targets, such that updates consumed from kafka become serialized per
object. An additional benefit is that it decouples replication
operation from status update, leading to more flexible handling of
failures (updates in kafka queue can always be retried at a later time
without impacting the main replication tasks).

It will incur additional storage and I/O cost but it should be
acceptable considering that kafka I/O overhead is small.

Changes:

- A new processor task is run with "npm run
  replication_status_processor", to process replication status updates
  on the source.

- The queue_processor task has been modified to publish updates to
  kafka (whether completed or failed), keeping the same format than
  for original replication entries.

- function tests (ft_test) now rely on local kafka for the status
  queue

Please review individual commits:
- https://github.com/scality/backbeat/pull/155/commits/d7d2a3ae90e41cfaf1e09b68f4b60bfee47ffe32
- https://github.com/scality/backbeat/pull/155/commits/69b3a6ab3e2e1cd1e432a18e4b2e1d33c51364c9
- https://github.com/scality/backbeat/pull/155/commits/3daae3213df06e113a8895ba2ab7f18c8d0a22bb


jonathan-gramain:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=ft/S3C-1146-CRRCompletionStatusInNewKafkaTopic
SCALITY_INTEGRATION_BRANCH=ft/S3C-981-siteLevelReplication
try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@philipyoo I think you're right, we should not pre-create it since it's done by Federation, I don't remember why I added this originally but I see no reason.

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine clean try

rahulreddy:
@ironman-machine
SCALITY_FEDERATION_BRANCH=ft/S3C-1146-CRRCompletionStatusInNewKafkaTopic
try

lib/BackbeatConsumer.js
```diff
@@ -4,6 +4,7 @@ const kafkaLogging = require('kafka-node/logging');
 const async = require('async');
 const joi = require('joi');
 
+const BackbeatProducer = require('./BackbeatProducer');
```
rahulreddy: It would be ideal to keep `BackbeatProducer` and `BackbeatConsumer` separate, without requiring them in each other. We can make sure they pass tests independently. For bootstrap, since it's only for testing, you can pass producer object as a parameter.
Adding this comment here so that we can cleanup sometime.

rahulreddy:
@ironman-machine clean
after=https://github.com/scality/Federation/pull/1286
r+

rahulreddy:
@ironman-machine clean r- r+

PR #157, opened 2017-12-27
Eve CI backport

ssalaues:


eve/workers/unit_and_feature_tests/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM spotify/kafka
```
LaurenSpiegel: why spotify kafka instead of the one we're now using in zenko?

eve/main.yml
```diff
@@ -0,0 +1,33 @@
...
+          command: npm run --silent lint_md
+      - ShellCommand:
+          name: run static analysis tools on code
+          command: npm run --silent lint
+      - ShellCommand:
+          name: run unit tests
+          command: npm test
+      - ShellCommand:
+          name: run feature tests
+          command: npm run ft_test
```
LaurenSpiegel: what about the 'behavior' tests that are in the package.json ("bh_test": "mocha --recursive tests/behavior",)

LaurenSpiegel (dismissed):

eve/workers/unit_and_feature_tests/Dockerfile
ssalaues: Good point, when this was originally made, we didn't have those but I'll test it out with that.

rahulreddy: I remember having this conversation with Rayenne at that time. His goal was to port whatever we were doing in circle.yml at that time and then do the switch to our kafka later. I think setting it up with our kafka would require pulling in Zookeeper alpine, getting the quorum work, setting up topics etc. I recommend doing that as a different PR.

eve/main.yml
ssalaues: We were matching the circle enviroment so I'm not sure why those aren't included. @rahulreddy Should we add in the behavior tests into the CI?

rahulreddy:
@ssalaues Can  you do git commit -x so that it references your original commit.

eve/main.yml
rahulreddy: Yes we should.

eve/workers/unit_and_feature_tests/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "aws-sdk": "2.147.0",
+    "backo": "^1.1.0",
+    "bucketclient": "scality/bucketclient",
+    "commander": "^2.11.0",
+    "eslint": "^2.4.0",
+    "eslint-config-airbnb": "^6.0.0",
+    "eslint-config-scality": "scality/Guidelines",
+    "eslint-plugin-react": "^4.2.3",
+    "joi": "^10.6",
+    "kafka-node": "^1.6.0",
```
rahulreddy: Can you match the version of kafka-node

rahulreddy (dismissed):

eve/workers/unit_and_feature_tests/package.json
ssalaues: I swear I changed that but I guess I lost it when I squashed. I'll fix it.

jonathan-gramain:
Per @tcarmet DM, backbeat 7 CI is OK so we should be able to close this PR.

PR #158, opened 2017-12-28
bf: send dest req only on success

philipyoo:
Before sending destination request, ensure that source
data is properly received. Although errors were being logged,
they were not returned and instead the request was still
sent and status on destination was still set to REPLICA.
The error that appeared in integration was 'socket hang up'

### DIFF
--- a/extensions/replication/queueProcessor/MultipleBackendTask.js
+++ b/extensions/replication/queueProcessor/MultipleBackendTask.js
@@ -155,20 +155,22 @@ class MultipleBackendTask extends QueueProcessorTask {
             Body: incomingMsg,
         });
         attachReqUids(destReq, log);
-        return destReq.send((err, data) => {
-            if (err) {
-                // eslint-disable-next-line no-param-reassign
-                err.origin = 'source';
-                log.error('an error occurred on putting MPU part to S3', {
-                    method: 'MultipleBackendTask._getAndPutMPUPartOnce',
-                    entry: destEntry.getLogInfo(),
-                    origin: 'target',
-                    peer: this.destBackbeatHost,
-                    error: err.message,
-                });
-                return doneOnce(err);
-            }
-            return doneOnce(null, data);
+        incomingMsg.once('data', () => {
+            destReq.send((err, data) => {
+                if (err) {
+                    // eslint-disable-next-line no-param-reassign
+                    err.origin = 'source';
+                    log.error('an error occurred on putting MPU part to S3', {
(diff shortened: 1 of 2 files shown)


philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

extensions/replication/queueProcessor/MultipleBackendTask.js
```diff
@@ -362,21 +364,23 @@ class MultipleBackendTask extends QueueProcessorTask {
...
-                    entry: destEntry.getLogInfo(),
-                    origin: 'target',
-                    peer: this.destBackbeatHost,
-                    error: err.message,
-                });
-                return doneOnce(err);
-            }
-            sourceEntry.setReplicationDataStoreVersionId(data.versionId);
-            return doneOnce(null, data);
+        incomingMsg.once('readable', () => {
```
rahulreddy: `readable` event could be emitted multiple times. Were you not able to hook onto the `success` event?

philipyoo: ya seemed like the `success` event was not being emitted. It looks like `incomingMsg` is an instance of `stream.PassThrough`.

I was thinking the `readable` event would be emitted because there was no error and instead data is available. Since `destReq.Body` accepts a stream, it was ok to `send` by that point? Purely assumptions though :) What do you think?

bennettbuchanan: Could you use the 'data' event to ensure that we can read from the stream?
http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#createReadStream-property

philipyoo: @bennettbuchanan I can try it out 👍

philipyoo:
@ironman-machine try

extensions/replication/queueProcessor/MultipleBackendTask.js
rahulreddy: `readable` with `once` handler should be sufficient. If you add a listener for `data` and ignore the chunk, the chunk is lost.

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

PR #159, opened 2017-12-28
rf: cleanup setup replication script

philipyoo:
Refactor/cleanup setup replication script by:
- Using version specified in `package.json`
- Reusing the `option` flags using a loop
- Add newline after any `outputHelp`


philipyoo:
rebased only- wish it didn't drop approvals   ಠ_ಠ

PR #163, opened 2018-01-09
FT: Add default field to replication endpoint

bennettbuchanan:
Allow a default field in a replication endpoint object.


bennettbuchanan:
@ironman-machine try

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r-

PR #164, opened 2018-01-09
ft: support for get and put metadata blob

rahulreddy:
With the introduction of one to many crr, there will be a status update for each backend.
Since there will be multiple requests to the source to update the replication status, it's important
to get the latest copy of the metadata blob as previous updates would render the copy it's holding stale.

### DIFF
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -12,6 +12,7 @@ const AccountCredentials =
           require('../../../lib/credentials/AccountCredentials');
 const RoleCredentials =
           require('../../../lib/credentials/RoleCredentials');
+const ObjectQueueEntry = require('../utils/ObjectQueueEntry');
 
 const MPU_CONC_LIMIT = 10;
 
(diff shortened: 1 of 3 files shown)


extensions/replication/tasks/ReplicateObject.js
```diff
@@ -12,6 +12,7 @@ const AccountCredentials =
           require('../../../lib/credentials/AccountCredentials');
 const RoleCredentials =
           require('../../../lib/credentials/RoleCredentials');
+const ObjectQueueEntry = require('../utils//ObjectQueueEntry');
```
philipyoo: extra slash

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -132,29 +133,101 @@ class ReplicateObject extends BackbeatTask {
...
+
+    _getMetadataOnce(entry, log, cb) {
+        log.debug('getting metadata', {
+            where: 'source',
+            entry: entry.getLogInfo(),
+            method: 'ReplicateObject._getMetadataOnce',
+        });
+        const cbOnce = jsutil.once(cb);
+
+        const req = this.backbeatDest.getMetadata({
```
bennettbuchanan: I think this is meant to be `this.backbeatSource`?

rahulreddy: Ah right, good catch!

rahulreddy:
Integration passed here http://ci.ironmann.io/gh/scality/Integration/18648

PR #166, opened 2018-01-13
doc: lifecycle design doc

jonathan-gramain:
First version of the lifecycle detailed design, please review and comment or suggest changes.

### DIFF
--- a/docs/lifecycle.md
+++ b/docs/lifecycle.md
@@ -1,0 +1,309 @@
+# Object Lifecycle Management
+
+## OVERVIEW
+
+The purpose is to apply various lifecycle actions to objects after a
+certain time has passed since object creation, per lifecycle rules as
+specified in AWS S3 APIs.
+
+### EXPIRATION
+
+We support expiration of versioned or non-versioned objects, when a
+number of days have passed since their creation.
+
+One important use case is to allow automatic deletion of older
+versions of versioned objects to reclaim storage space.
+
+## DESIGN
+
+The implementation of lifecycle in S3 connector comprises several
+components which, by working together provide lifecycle
+functionalities.
+
+Most components rely on Kafka and Zookeeper to pass state to each
+other.
+
+### Workflow Overview
+
+Here's an example workflow that expires version *v1* of object *obj*
+from bucket *foobucket*:
+
+```
+ __________________
+/                  \
+|   S3 client      |
+\__________________/
+       |
+       | PUT /foobucket?lifecycle
+       |
+ ______v___________
+/                  \         ___________
+|                  |        /           \
+| S3 Connector     |--------|  METADATA |
+|                  |        |    log    |
+|                  |        \___________/
+\__________________/              |
+    ^    ^                        |
+    |    |                        | entry: "PUT /foobucket?lifecycle"
+    |    |                        |
+    |    |                 _______v_________                 ___________
+    |    |                /                 \  foobucket    /           \
+    |    |                | Queue Populator |-------------->|           |
+    |    |                \_________________/               | ZOOKEEPER |
+    |    |                                   foobucket      |           |
+    |    |                              /-------------------\___________/
+    |    |                 _____________v___
+    |    |                /                 \
+    |    |                |   Conductor     | foobucket@null ___________
+    |    |                \_________________/-------------->/           \
+    |    |                                                  |           |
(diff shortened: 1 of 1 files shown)


docs/lifecycle.md
```diff
@@ -0,0 +1,278 @@
...
+  (e.g. if "delete", delete the target object version on S3 connector)
+
+### Message formats
+
+#### Bucket tasks
+
+A message in *backbeat-lifecycle-bucket-tasks* topic represents a
+listing task for one bucket, it has the following format:
+
+```json
```
bennettbuchanan: Maybe remove JSON formatting here since it doesn't like those optional keys.

docs/lifecycle.md
```diff
@@ -0,0 +1,278 @@
...
+certain time has passed since object creation, per lifecycle rules as
+specified in AWS S3 APIs.
+
+### EXPIRATION
+
+We support expiration of versioned or non-versioned objects, when a
+number of days have passed since their creation.
+
+One important use case is to allow automatic deletion of older
+versions of versioned objects to reclaim storage space.
```
LaureVergeron: Will there be an option for versions expiration (without deleting the object itself)? For example, one updates a versioned object every day, and wants to get rid of versions older than three months? I believe there's a field usecase for that, but I'm unsure whether this has been discussed or not (it might simply be possible by applying the policy to each version with its id?)

docs/lifecycle.md
```diff
@@ -0,0 +1,278 @@
...
+#### Bucket tasks
+
+A message in *backbeat-lifecycle-bucket-tasks* topic represents a
+listing task for one bucket, it has the following format:
+
+```json
+{
+    "bucket": "bucketname",
+    "owner": "ownerID"
+    [, "prefix": "someprefix"
```
ThibaultRiviere: I think you can remove the first `,`

docs/lifecycle.md
```diff
@@ -0,0 +1,278 @@
...
+```json
+{
+    "action": "actionname",
+    "target": {
+        "owner": "ownerID",
+        "bucket": "bucketname",
+        "key": "objectkey",
+        "version": "objectversion"
+    }
+    [, "details": {...} ]
```
ThibaultRiviere: same here

docs/lifecycle.md
```diff
@@ -0,0 +1,278 @@
...
+processes items from the *backbeat-lifecycle-bucket-tasks* topic. For
+each of them, it does the following:
+
+* Generate or use cached credentials to be able to query the target
+  bucket (through admin-backbeat.json?)
+* Get the lifecycle configuration from S3 connector of the target
+  bucket (GET /foobucket?lifecycle)
+  * If no lifecycle configuration is associated to the bucket, skip
+    the rest of bucket processing
+* List the object versions in the bucket, starting from the
```
philipyoo: Just to clarify the steps here:
- list every object in the bucket (which defaults to a max of 1000 listings)
- If exceeds 1000 object listings, re-queue to Kafka w/ `StartAfter` to get next object listings
- For each object, if versioned, list the objects versions (defaults to a max of 1000 listings again).
- If versioned and exceeds 1000 versions, need to re-queue to Kafka w/ `KeyMarker` and `VersionIdMarker`.
- If not versioned, just go ahead and queue up the delete request for the object to the consumer. If versioned, queue up the delete request to the consumer for every version of the object.

Pre-filter step needed to determine if the Kafka entry is a bucket or a versioned object w/ marker. I'm not sure how difficult this step will be at this moment, but the "Bucket tasks" Kafka entry might need to have some sort of indicator. Maybe just based on if the entry holds properties like `KeyMarker` or `StartAfter`?

docs/lifecycle.md
jonathan-gramain: What you describe is how it's going to work, actually, and it's the first use case we implement lifecycle for, to delete older versions that CRR has created on the backup site. Maybe the description is confusing, let me know if you think of a better formulation.

The API is going to support what AWS supports for expiration: https://docs.aws.amazon.com/AmazonS3/latest/dev/intro-lifecycle-rules.html

The implementation will be based (for now at least) on a cron-style task that wakes up periodically (say one or more times per day), lists buckets to get their objects and their versions, match against bucket lifecycle configurations, and take appropriate actions when there is a match.

PR #169, opened 2018-01-16
bf: fix concurrency issue with log reader

jonathan-gramain:
Call QueuePopulatorExtension.setBatch() every time new entries are received,
because multiple readers can be intermingled sending their respective data,
and although the LogReader class is managing a single reader, the extensions
`batch` field is global to all readers, so it has to be set every time.

Note that this fix is only required with streamed log reader (i.e. with https://github.com/scality/bucketclient/pull/126 and https://github.com/scality/Arsenal/pull/379)


jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine r+

PR #173, opened 2018-01-31
ft: update config with lifecycle params

jonathan-gramain:
Add config params related to lifecycle components: kafka topics, conductor,
producer, consumer, and rule-specific config (expiration...).

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -6,6 +6,19 @@
     "kafka": {
         "hosts": "127.0.0.1:9092"
     },
+    "s3": {
+        "host": "127.0.0.1",
+        "port": 8000
+    },
+    "auth": {
+        "type": "account",
+        "account": "bart",
+        "vault": {
+            "host": "127.0.0.1",
+            "port": 8500,
+            "adminPort": 8600
+        }
+    },
     "queuePopulator": {
         "cronRule": "*/5 * * * * *",
         "batchMaxRead": 10000,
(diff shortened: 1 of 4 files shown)


conf/config.json
```diff
@@ -6,6 +6,19 @@
     "kafka": {
         "hosts": "127.0.0.1:9092"
     },
+    "s3": {
```
dora-korpar: I believe we still need a `transport` key. Not sure if you want to set that globally, too

jonathan-gramain:
@ironman-machine r+

conf/config.json
jonathan-gramain: Yes, it's part of the config, but the default is "http" so it should be fine. You can access the transport key from the config object, it will be always populated by the joi parser even if not set in config.json (with the default value in such case).

jonathan-gramain: And this config.json is not used for deployments with Federation, since we generate a config from a template (see PR https://github.com/scality/Federation/pull/1315)

jonathan-gramain:
@ironman-machine r-

jonathan-gramain:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=ft/S3C-1154-lifecycleConfigUpdate
try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
Integration test passed in http://ci.ironmann.io/gh/scality/Integration/19139, merging manually in feature branch

PR #174, opened 2018-02-01
Disable caching node modules

ssalaues:
Disabled the caching as there is not way to easily clear the docker cache (which keeps the npm cache).

Moved the package.json back to the root as it is not needed in the docker context with these changes.


jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine r- r+

PR #180, opened 2018-02-07
FT: Add lifecycle producer

bennettbuchanan:


extensions/replication/queueProcessor/task.js
```diff
@@ -1,6 +1,5 @@
 'use strict'; // eslint-disable-line
 const assert = require('assert');
-const werelogs = require('werelogs');
```
jonathan-gramain: Was this change intended to be committed?

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,256 @@
...
+     * @param {number} authConfig.vault.adminPort - vault admin port
+     * @param {String} transport - http or https
+     */
+    constructor(zkConfig, lcConfig, s3Config, authConfig, transport) {
+        this._log = new Logger('Backbeat:LifecycleProducer');
+        this._zkConfig = zkConfig;
+        this._lcConfig = lcConfig;
+        this._s3Client = new AWS.S3({
+            endpoint: `${transport}://${s3Config.host}:${s3Config.port}`,
+            credentials: new AccountCredentials(authConfig, this._log),
```
jonathan-gramain: When using vault, we will have to use the admin credentials in admin-backbeat.json to create new access keys for the account owning the bucket, and cache them in memory, like echo feature is doing (it also creates accounts on the destination but we do not want this for lifecycle). See how the `VaultClientCache` class is used to do this.

A corollary is that we'll need different S3 clients for each account, each time providing them the proper account access/secret key for the object/bucket to process. For replication, we create new clients for every object to replicate, sounds a bit overkill but does not seem to cause major CPU/memory issues in practice so I think it's acceptable.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,256 @@
...
+            this._log.error('could not parse bucket entry', { error });
+            return process.nextTick(() => cb(error));
+        }
+        return this._s3Client.getBucketLifecycleConfiguration({
+            Bucket: value.bucket,
+        }, (err, config) => {
+            if (err) {
+                this._log.error('error getting bucket lifecycle configuration',
+                    { error: err });
+                return process.nextTick(() => cb(err));
```
jonathan-gramain: since we're already in an async context (answer from `getBucketLifecycleConfiguration) we can get rid of the `process.nextTick`

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,256 @@
...
+            }
+            if (!this._shouldProcessConfig(config)) {
+                return process.nextTick(cb);
```
jonathan-gramain: same here

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,256 @@
...
+class LifecycleProducer {
+    /**
+     * Constructor of LifecycleProducer
+     * @param {Object} zkConfig - zookeeper config
+     * @param {Object} lcConfig - lifecycle config
+     * @param {Object} s3Config - s3 config
+     * @param {String} s3Config.host - host ip
+     * @param {String} s3Config.port - port
+     * @param {Object} authConfig - auth config
+     * @param {String} authConfig.account - account name
```
jonathan-gramain: Technically, only one of either `account` or `vault` sections is mandatory, so you can specify both as optional.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * Send entry to the object task topic
+     * @param {[type]} entry - The Kafka entry to send to the topic
+     * @param {Function} cb - The callback to call
+     * @return {undefined}
+     */
+    sendObjectEntry(entry, cb) {
+        const strEntry = JSON.stringify(entry);
+        this._objectProducer.send([strEntry], cb);
```
jonathan-gramain: I think this will not work, because the `send` method expects an array of *objects*, where each object has a `message` attribute and optionally a `key` attribute as well. So normally the proper way would be to construct an array of objects with a `message` attribute which is the result of `JSON.stringify`, if I'm not mistaken.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * Send entry back to bucket task topic
+     * @param {[type]} entry - The Kafka entry to send to the topic
+     * @param {Function} cb - The callback to call
+     * @return {undefined}
+     */
+    sendBucketEntry(entry, cb) {
+        const str = JSON.stringify(entry);
+        this._bucketProducer.send([str], cb);
```
jonathan-gramain: same issue here

extensions/replication/queueProcessor/task.js
bennettbuchanan: Yes, I noticed it was not used in the file. I guess maybe it is outside the scope of this PR though.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
philipyoo: So just making sure, but would it then be something like `[ { message: strEntry } ]`

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
bennettbuchanan: Updated.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
bennettbuchanan: Updated.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
bennettbuchanan: Updated.

extensions/replication/queueProcessor/task.js
bennettbuchanan: Removed these changes.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
jonathan-gramain: yes

bennettbuchanan: Updated.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
bennettbuchanan: Updated.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
bennettbuchanan: I've updated this to generate a new key for each unique canonical ID (using cache if the canonical ID has been seen before). I then create a new S3 client for each `getBucketLifecycle` request using the access/secret keys.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,377 @@
...
+        consumer.subscribe();
+    }
+
+    /**
+     * Set up the vault client cache for making requests to vault.
+     * @return {undefined}
+     */
+    _setupVaultClientCache() {
+        const { vault } = this._authConfig;
+        const { host, port, adminPort, adminCredentialsFile } = vault;
```
philipyoo: just a quick note, `adminCredentialsFile` will need to be added to config files and a joi required field

bennettbuchanan: Was relying on this addition: https://github.com/scality/backbeat/blob/ft/S3C-1154-lifecycle/conf/config.joi.js#L31 and the error thrown if we cannot correctly parse credentials from the file. I guess we should also check for the existence of the `adminCredentialsFile` field, or else change `.optional()` to `.required()` in the joi validation.

bennettbuchanan:
@LaurenSpiegel brought up a good point that this class handles more than just setting up producers for Kafka—it also sets up consumers, does a listing and filtering of the lifecycle config, and manages passing the config to the lifecycle task. 

What do you all think about updating the name of this component to "LifecycleProcessor" to follow the pattern of the replication extension? This differs from the original design doc so wanted to get everyone's input. @jonathan-gramain @dora-korpar @philipyoo

dora-korpar:
That makes sense to me

philipyoo:
Yea that sounds good to me 👍

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,378 @@
...
+     * @param {Function} cb - The callback to call
+     * @return {undefined}
+     */
+    _processBucketEntry(entry, cb) {
+        const { error, result } = safeJsonParse(entry.value);
+        if (error) {
+            this._log.error('could not parse bucket entry', { error });
+            return process.nextTick(() => cb(error));
+        }
+        const { bucket, owner } = result.value;
```
jonathan-gramain: I think the correct deconstruction is `const { bucket, owner } = result;`, because the kafka message object has no "value" field but "bucket" and "owner" are top-level fields of the message object.

jonathan-gramain:
OK for name change to lifecycle processor, don't forget to update also the config.json section (and will have to do it in Federation PR as well).

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
jonathan-gramain: This exception occurred:
```
/home/scality/backbeat/extensions/lifecycle/lifecycleProducer/LifecycleProducer.js:177
        const { bucket, owner } = result.value;
                                        ^

TypeError: Cannot match against 'undefined' or 'null'.
    at LifecycleProducer._processBucketEntry (/home/scality/backbeat/extensions/lifecycle/lifecycleProducer/LifecycleProducer.js:177:41)
    at /home/scality/backbeat/node_modules/async/dist/async.js:4082:9
    at Object.process (/home/scality/backbeat/node_modules/async/dist/async.js:2330:17)
    at /home/scality/backbeat/node_modules/async/dist/async.js:2238:19
    at Immediate.<anonymous> (/home/scality/backbeat/node_modules/async/dist/async.js:119:16)
    at runCallback (timers.js:649:20)
    at tryOnImmediate (timers.js:622:5)
    at processImmediate [as _immediateCallback] (timers.js:594:5)
```

We should both fix the case, and also make the parsing of messages more resilient (i.e. do not crash but log an error and keep processing the next messages).

bennettbuchanan: I had seen this structure when publishing to a topic with a Kafka producer and needed to deconstruct this way. I will validate this one way or another as I'm adding tests.

bennettbuchanan: Okay, I will update accordingly.

philipyoo: I think Bennett added a good try/catch for json parsing. For now, with the fix, I also check to make sure `bucket` and `owner` exist. If not, log error and return

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -173,7 +173,15 @@ class LifecycleProducer {
             this._log.error('could not parse bucket entry', { error });
             return process.nextTick(() => cb(error));
         }
-        const { bucket, owner } = result.value;
+        const { bucket, owner } = result;
+        if (!bucket || !owner) {
+            this._log.error('kafka bucket entry missing required fields', {
```
jonathan-gramain: :+1:

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,386 @@
...
+        // The task scheduler for processing lifecycle tasks concurrently.
+        this._internalTaskScheduler = async.queue((ctx, cb) => {
+            const { task, rules, value } = ctx;
+            return task.processBucketEntry(rules, value, cb);
+        }, this._lcConfig.producer.concurrency);
+
+        // Listen for errors from any task being processed.
+        this._internalTaskScheduler.drain(err => {
+            if (err) {
+                this._log.error('error occured when during task processing', {
```
dora-korpar: trivial, but typo including `when during`. should use either one or the other

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
```diff
@@ -0,0 +1,386 @@
...
+        return {
+            sendBucketEntry: this.sendBucketEntry,
+            sendObjectEntry: this.sendObjectEntry,
+            enabledRules: this._lcConfig.rules,
+            log: this._log,
+        };
+    }
+
+    /**
+     * Return an S3 client instance using the given accound credentials.
```
dora-korpar: ~~acound~~ -> account

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
bennettbuchanan: Updated.

extensions/lifecycle/lifecycleProducer/LifecycleProducer.js
bennettbuchanan: Updated.

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine

jonathan-gramain:
@ironman-machine 
DEFAULT_BRANCH=rel/7.4
try

jonathan-gramain:
@ironman-machine 
DEFAULT_BRANCH=rel/7.4
try

jonathan-gramain:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=ft/S3C-1154-lifecycleDeployment
DEFAULT_BRANCH=rel/7.4
try

jonathan-gramain:
@ironman-machine
DEFAULT_BRANCH=rel/7.4
try

PR #182, opened 2018-02-08
ft: add lifecycle object consumer

dora-korpar:
Last part of lifecycle implementation. Should be merged last


extensions/lifecycle/lifecycleConsumer/LifecycleConsumer.js
```diff
@@ -0,0 +1,101 @@
...
+        try {
+            entryData = JSON.parse(kafkaEntry.value);
+        } catch (err) {
+            this.logger.error('error processing lifecycle object entry',
+                { error: err });
+            return process.nextTick(() => done(errors.InternalError));
+        }
+        const task = new LifecycleObjectTask();
+        const q = async.queue(task.processQueueEntry);
+        return q.push(entryData, done);
```
bennettbuchanan: Wondering if it worth including calling the `drain` method on the queue to log any errors from `LifecycleObjectTask.processQueueEntry`.

extensions/lifecycle/lifecycleConsumer/task.js
```diff
@@ -0,0 +1,18 @@
...
+const config = require('../../../conf/Config');
+const zkConfig = config.zookeeper;
+const lcConfig = config.extensions.lifecycle;
+const s3Config = config.s3;
+const authConfig = config.auth;
+
+const lifecycleConsumer = new LifecycleConsumer(
+    zkConfig, lcConfig, s3Config, authConfig);
+
+werelogs.configure({ level: config.log.logLevel,
```
bennettbuchanan: I noticed this pattern also in the queue processor task file. I may be mistaken but I don't think configuring `werelogs` like this without using the logger instance does anything.

extensions/lifecycle/tasks/LifecycleObjectTask.js
```diff
@@ -0,0 +1,95 @@
...
+            retryTimeoutS: qpState.repConfig.queueProcessor.retryTimeoutS,
+        });
+        Object.assign(this, qpState);
+
+        this.s3Credentials = null;
+        this.s3Client = null;
+    }
+
+    _setupClients(log, done) {
+        this.s3Credentials = new AccountCredentials(this.authConfig, log);
```
bennettbuchanan: `this.s3Credentials` is not used elsewhere in the class, so may be easier to just assign to a constant variable.

extensions/lifecycle/tasks/LifecycleObjectTask.js
```diff
@@ -0,0 +1,95 @@
...
+            endpoint: `${transport}://${s3.host}:${s3.port}`,
+            credentials: this.s3Credentials,
+            sslEnabled: transport === 'https',
+            s3ForcePathStyle: true,
+            signatureVersion: 'v4',
+            httpOptions: { agent: this.httpAgent, timeout: 0 },
+            maxRetries: 0,
+        });
+
+        return done();
```
bennettbuchanan: No need to use a callback in this method.

extensions/lifecycle/lifecycleConsumer/LifecycleConsumer.js
```diff
@@ -0,0 +1,101 @@
...
+        let entryData;
+        try {
+            entryData = JSON.parse(kafkaEntry.value);
+        } catch (err) {
+            this.logger.error('error processing lifecycle object entry',
+                { error: err });
+            return process.nextTick(() => done(errors.InternalError));
+        }
+        const task = new LifecycleObjectTask();
+        const q = async.queue(task.processQueueEntry);
```
bennettbuchanan: `processQueueEntry` => `_processQueueEntry`.

extensions/lifecycle/tasks/LifecycleObjectTask.js
```diff
@@ -0,0 +1,95 @@
...
+
+    _executeAction(entry, log, done) {
+        const action = entry.action;
+        const { bucket, key, version } = entry.target;
+        const details = entry.details;
+        const reqParams = { Bucket: bucket, Key: key };
+        // default action is 'delete'
+        let reqMethod = 'deleteObject';
+
+        if (version) {
```
bennettbuchanan: If `version` is `undefined`, we can just set it in the `reqParams` object above and then we can remove this if block.

extensions/lifecycle/tasks/LifecycleObjectTask.js
```diff
@@ -0,0 +1,95 @@
...
+            reqParams.VersionId = version;
+        }
+        if (action === 'deleteMPU') {
+            reqParams.UploadId = details.uploadId;
```
bennettbuchanan: `details.uploadId` => `details.UploadId`. However, I think probably the lifecycle task should instead stick with the lowercase convention for entry key names.

extensions/lifecycle/tasks/LifecycleObjectTask.js
```diff
@@ -0,0 +1,95 @@
...
+        }
+        const req = this.s3Client[reqMethod](reqParams);
+        attachReqUids(req, log);
+        req.on('error', err => {
+            log.error(`an error occurred on ${reqMethod} from S3`,
+                { method: 'LifecycleObjectTask._executeAction',
+                    error: err.message,
+                    httpStatus: err.statusCode });
+            return done(err);
+        });
```
bennettbuchanan: No sending the request here with the callback?

extensions/lifecycle/tasks/LifecycleObjectTask.js
```diff
@@ -0,0 +1,95 @@
...
+
+class LifecycleObjectTask extends BackbeatTask {
+    /**
+     * Process a lifecycle object entry
+     *
+     * @constructor
+     * @param {QueueProcessor} qp - queue processor instance
+     */
+    constructor(qp) {
+        const qpState = qp.getStateVars();
```
philipyoo: will need to define `getStateVars` in your ObjectConsumer class

extensions/lifecycle/tasks/LifecycleObjectTask.js
```diff
@@ -0,0 +1,95 @@
...
+        super({
+            retryTimeoutS: qpState.repConfig.queueProcessor.retryTimeoutS,
```
philipyoo: Might be separating this to its own process, so use instead `lifecycle.retryTimeoutS` depending on how it is defined in `qpState`

extensions/lifecycle/tasks/LifecycleObjectTask.js
philipyoo: Does this need to use `VaultClientCache`?

bennettbuchanan: Yes, it will need to make the lifecycle operation requests with an S3 client using the cached access/secret keys for the bucket owner. I've pushed the caching/key generation step in lifecycle producer now, so we can leverage the vaultclient cache here. Let me know if any changes are needed in the producer.

extensions/lifecycle/lifecycleConsumer/task.js
jonathan-gramain: This has been changed recently (a few months ago still) in werelogs, this is now the common way of setting the log level globally for the process.

extensions/lifecycle/lifecycleConsumer/LifecycleConsumer.js
```diff
@@ -0,0 +1,111 @@
...
+                connectionString: this.zkConfig.connectionString,
+            },
+            topic: this.lcConfig.objectTasksTopic,
+            groupId: this.lcConfig.consumer.groupId,
+            concurrency: this.lcConfig.consumer.concurrency,
+            queueProcessor: this.processKafkaEntry.bind(this),
+        });
+        this._consumer.on('error', () => {});
+        this._consumer.subscribe();
+        this.logger.info('queue processor is ready to consume ' +
```
jonathan-gramain: The message should be adapted for lifecycle. I recommend:

```
lifecycle consumer successfully started
```

To match the startup message of other lifecycle components.

jonathan-gramain: My concern is that we use queue processor for the CRR process, so it may be confused with the "queue processor" of lifecycle.

jonathan-gramain:
We need to implement role credentials management (through vault) with the RoleCredentials class.

jonathan-gramain:
@ironman-machine
DEFAULT_BRANCH=rel/7.4
try

PR #183, opened 2018-02-08
ft: add lifecycle producer task

philipyoo:
Lifecycle Tasks handles the AWS operations for validating and matching rules. These tasks are processed via the Lifecycle Producer. 
Please refer to the Lifecycle design docs to view the LifecycleProducer/LifecycleTask operations.

To be added and in the works are adding functional tests for specifically LifecycleTask class

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -1,0 +1,769 @@
+'use strict'; // eslint-disable-line
+
+const async = require('async');
+
+const attachReqUids = require('../../replication/utils/attachReqUids');
+const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
+
+const RETRYTIMEOUTS = 300;
+
+// Default max AWS limit is 1000 for both list objects and list object versions
+const MAX_KEYS = 1000;
+// concurrency mainly used in async calls
+const CONCURRENCY_DEFAULT = 10;
+
+class LifecycleTask extends BackbeatTask {
+    /**
+     * Process a list of versioned or non-versioned objects
+     *
+     * @constructor
+     * @param {LifecycleProducer} lp - lifecycle producer instance
+     */
+    constructor(lp) {
+        const lpState = lp.getStateVars();
+        super({ retryTimeoutS: RETRYTIMEOUTS });
+        Object.assign(this, lpState);
+        this._publishedNextListingTask = false;
+    }
+
+    /**
+     * Handles non-versioned objects
+     * @param {object} bucketData - bucket data
+     * @param {object} params - params for AWS S3 `listObjects`
+     * @param {Logger.newRequestLogger} log - logger object
+     * @param {function} done - callback(error, data)
+     * @return {undefined}
+     */
+    _getObjectList(bucketData, params, log, done) {
+        const req = this.s3target.listObjects(params);
+        attachReqUids(req, log);
+        async.waterfall([
+            next => req.send((err, data) => {
+                if (err) {
+                    log.error('error listing bucket objects', {
+                        error: err,
+                        bucket: params.Bucket,
+                    });
+                    return next(err);
+                }
+                return next(null, data);
+            }),
+            (data, next) => {
+                if (data.IsTruncated) {
+                    // re-queue to Kafka topic bucketTasksTopic
+                    // with bucket name and `data.marker`
+                    let marker = data.NextMarker;
+                    if (!marker && data.Contents.length > 0) {
+                        marker = data.Contents[data.Contents.length - 1].Key;
+                    }
+
+                    const entry = Object.assign({}, bucketData, {
+                        details: { marker },
+                    });
+                    return this.sendBucketEntry(entry, err => {
+                        if (!err) {
+                            log.debug(
+                                'sent kafka entry for bucket consumption', {
+                                    method: 'LifecycleTask._getObjectList',
+                                });
+                            this._publishedNextListingTask = true;
+                        }
+                        return next(null, data);
+                    });
+                }
+                return process.nextTick(() => next(null, data));
+            },
+        ], (err, data) => {
+            if (err) {
+                return done(err);
+            }
+            return done(null, { Contents: data.Contents });
+        });
+    }
+
+    /**
+     * Handles versioned objects (both enabled and suspended)
+     * @param {object} bucketData - bucket data
+     * @param {object} params - params for AWS S3 `listObjectVersions`
+     * @param {Logger.newRequestLogger} log - logger object
+     * @param {function} done - callback(error, data)
+     * @return {undefined}
+     */
+    _getObjectVersions(bucketData, params, log, done) {
+        const req = this.s3target.listObjectVersions(params);
+        attachReqUids(req, log);
+        async.waterfall([
+            next => req.send((err, data) => {
+                if (err) {
+                    log.error('error listing versioned bucket objects', {
+                        error: err,
+                        bucket: params.Bucket,
+                    });
+                    return next(err);
+                }
+                return next(null, data);
+            }),
+            (data, next) => {
+                if (data.IsTruncated) {
+                    // NOTE: this might only be needed if a LC rule defines
+                    //   a rule for processing NoncurrentVersion actions
+                    const entry = Object.assign({}, bucketData, {
+                        details: {
+                            keyMarker: data.NextKeyMarker,
+                            versionIdMarker: data.NextVersionIdMarker,
+                        },
+                    });
+                    return this.sendBucketEntry(entry, err => {
+                        if (!err) {
+                            log.debug(
+                                'sent kafka entry for bucket consumption', {
+                                    method: 'LifecycleTask._getObjectVersions',
+                                });
+                            this._publishedNextListingTask = true;
+                        }
+                        return next(null, data);
+                    });
+                }
+                return process.nextTick(() => next(null, data));
+            },
+        ], (err, data) => {
+            if (err) {
+                return done(err);
+            }
+            return done(null, {
+                Versions: data.Versions,
+                DeleteMarkers: data.DeleteMarkers,
+            });
+        });
+    }
+
+    /**
+     * Handles incomplete multipart uploads
+     * @param {object} bucketData - bucket data
+     * @param {array} bucketLCRules - array of bucket lifecycle rules
+     * @param {object} params - params for AWS S3 `listObjectVersions`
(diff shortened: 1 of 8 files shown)


extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -0,0 +1,661 @@
...
+            // reduce filteredRules to only get earliest dates
+            return done(null, this._getApplicableRules(filterRules));
+        });
+    }
+
+    _compareContents(bucketData, lcRules, contents, log, done) {
+        if (!contents) {
+            return done();
+        }
+        return async.each(contents, (obj, cb) => {
```
jonathan-gramain: Since contents can hold a fair number of elements (up to 1000 or maybe more?) it's preferable to use `eachLimit` to limit the number of concurrent requests (say, 10 should be good enough, we can refine later).

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -0,0 +1,661 @@
...
+                    bucketData, obj, applicableRules, log, next),
+            ], err => cb(err));
+        }, done);
+    }
+
+    _compareVersions(bucketData, lcRules, versions, log, done) {
+        if (!versions) {
+            return done();
+        }
+        return async.each(versions, (version, cb) => {
```
jonathan-gramain: same, use `eachLimit`

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -0,0 +1,661 @@
...
+        }, done);
+    }
+
+    // _compareDeleteMarkers(bucket, rules, deleteMarkers, versions, log, cb) {
+    //      // will be completed later
+    // }
+
+    _compareWithLCRules(bucketData, bucketLCRules, data, log, done) {
+        async.parallel([
+            next => this._compareContents(bucketData, bucketLCRules,
```
jonathan-gramain: `next` in the context of parallel requests should rather be called `done`

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -0,0 +1,661 @@
...
+
+        // NOTE:
+        //  With the addition of mpu's, there was no easy way to handle
+        //  how to process entries w/ markers. Will need to rethink this and
+        //  refactor
+
+        // get all objects whether versioned or non-versioned
+        async.parallel([
+            cb => {
+                if (bucketData.versionIdMarker ||
```
jonathan-gramain: Are `versionIdMarker` and `uploadIdMarker` mutually exclusive? If so, you can get rid of `async.parallel` and pass them the `done` callback directly.

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -0,0 +1,340 @@
...
+            };
+            const item2 = {
+                Key: 'cat-test',
+                LastModified: currentDate,
+            };
+            const objTags = { TagSet: [] };
+
+            const res1 = lct._filterRules(mBucketRules, item1, objTags);
+            assert.strictEqual(res1.length, 3);
+            const expRes1 = getRuleIDs(mBucketRules.filter(rule =>
```
jonathan-gramain: In this test, it looks like you're just testing the test functions, is it what you want? As `res1` is not checked (same for `res2` below).

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -0,0 +1,340 @@
...
+                    .addTag('tag3-1', 'val3').build(),
+            ];
+            const item = {
+                Key: 'example-item',
+                LastModified: currentDate,
+            };
+            const objTags1 = { TagSet: [{ Key: 'tag1', Value: 'val1' }] };
+            const res1 = lct._filterRules(mBucketRules, item, objTags1);
+            assert.strictEqual(res1.length, 2);
+            const expRes1 = getRuleIDs(mBucketRules.filter(rule =>
```
jonathan-gramain: ditto, did you intend to test `res1`?

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -0,0 +1,661 @@
...
+            }
+            if (rule.Filter && rule.Filter.And) {
+                // multiple tags or prefix w/ tag(s)
+                const prefix = rule.Filter.And.Prefix;
+                if (prefix && !item.Key.startsWith(prefix)) {
+                    return false;
+                }
+                // must have equal amount of tags because tags are applied with
+                // AND operand
+                if (rule.Filter.And.Tags.length !== objTags.TagSet.length) {
```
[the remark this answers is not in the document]
philipyoo: within `And` we are guaranteed to have a `Tags` array with at least one tag, so this should be ok

tests/unit/lifecycle/LifecycleTask.spec.js
philipyoo: oops my mistake, ty meant to test against `res1`

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -0,0 +1,661 @@
...
+                        method: 'LifecycleTask._getObjectList',
+                    });
+                });
+            }
+            if (data.Contents.length === 0) {
+                // if bucket is empty, just skip
+                log.debug('no objects in bucket', {
+                    bucket: params.Bucket,
+                });
+                return done();
```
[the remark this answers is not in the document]
philipyoo: Since within `_compareWithLCRules`, `data.Contents` will not be defined, the next step `_compareContents` will be skipped since I just check if the param is defined or not. I know this is a bit confusing and am hoping to refactor all this in the coming weeks, but for now, I think I should not set `data.Contents` to anything

extensions/lifecycle/tasks/LifecycleTask.js
philipyoo: Yes, they are mutually exclusive, but there is the case when we receive a new bucket task that has none of these markers defined, in which case, I want to go ahead and process both mpu's and (non-version && version). When paginating through markers, a bucket entry will have only one marker defined, in which case, we process for that one marker type and ignore others

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -0,0 +1,661 @@
...
+     *   token marker for non-versioned buckets
+     * @param {function} done - callback(error, data)
+     * @return {undefined}
+     */
+    processBucketEntry(bucketLCRules, bucketData, done) {
+        const log = this.log.newRequestLogger();
+
+        log.debug('processing bucket entry', {
+            bucket: bucketData.bucket,
+            owner: bucketData.ownerId,
```
bennettbuchanan: Based on the design docs, and what I have in the producer, `bucketData` should have an `owner` property rather than `ownerId`. Either way works for me, but we should just settle on one.

jonathan-gramain:
@ironman-machine 
DEFAULT_BRANCH=rel/7.4
try

philipyoo:
@jonathan-gramain lgtm 👍

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -125,33 +149,46 @@ class LifecycleTask extends BackbeatTask {
...
-                        log.debug('sent kafka entry for bucket consumption', {
-                            method: 'LifecycleTask._getMPUs',
-                        });
-                        this._publishedNextListingTask = true;
-                    }
-                });
-            }
             this._compareMPUUploads(bucketData, bucketLCRules,
-                data.Uploads, log);
+                                    data.Uploads, log);
```
dora-korpar: trivial, but I think this indentation style is really detrimental to reading the code. it's not worth changing rn, but maybe we should decide on a standard to improve readability/consistency for future backbeat contributors :)

jonathan-gramain: This was an involuntary change that I should not have committed (my beloved editor e\*\*\*s indents this way when pressing TAB), though personally speaking I tend to prefer this style, but point taken, I'm down for one or the other depending on which one everyone is most comfortable with :wink:

jonathan-gramain:
Integration test passed in http://ci.ironmann.io/gh/scality/Integration/19892.

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -0,0 +1,234 @@
...
+                { Key: 'tag1', Value: 'val1' },
+                { Key: 'tag3', Value: 'val1' },
+            ] };
+            const res3 = lct._filterRules(mBucketRules, item, objTags3);
+            assert.strictEqual(res3.length, 1);
+            assert.deepStrictEqual(getRuleIDs(res3), ['task-7']);
+        });
+    });
+
+    // describe('_filterRules for listObjectVersions versions',
```
bennettbuchanan: Leftover code, looks like.

tests/utils/Rule.js
```diff
@@ -0,0 +1,113 @@
...
+            if (this.tags.length) {
+                filter.Tag = this.tags[0];
+            }
+        }
+        rule.Filter = filter;
+
+        return rule;
+    }
+
+    addID(id) {
```
bennettbuchanan: Minor, but maybe worth noting: I've seen setters/getters in this repo named with 'set', so instead this method would be `setID`.

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -0,0 +1,234 @@
...
+
+            const res = lct._filterRules(mBucketRules, item, objTags);
+            const expected = mBucketRules.filter(rule =>
+                rule.Status === 'Enabled');
+            assert.deepStrictEqual(getRuleIDs(res), getRuleIDs(expected));
+        });
+
+        it('should filter out unmatched prefixes', () => {
+            const mBucketRules = [
+                new Rule().addID('task-1').addPrefix('atask').build(),
```
bennettbuchanan: I really like this declarative style!

tests/utils/Rule.js
```diff
@@ -0,0 +1,113 @@
...
+            Key: key,
+            Value: value,
+        });
+        return this;
+    }
+
+    /**
+    * Expiration
+    * @param {string} prop - Property must be defined in `validProps`
+    * @param {integer|boolean} value - integer for `Date` or `Days`, or
```
bennettbuchanan: The type JSDoc uses for `@param` is `number` instead of `integer`. This would be helpful if we ever do want
 to generate docs for our code base. 🙏

PR #186, opened 2018-02-22
non-keyed partitioner in BackbeatProducer

jonathan-gramain:
Allow the possibility to use non-keyed partitioning without specifying
a partition number, so that the default partitioning scheme is used
to spread messages across partitions without providing a key.

### DIFF
--- a/lib/BackbeatProducer.js
+++ b/lib/BackbeatProducer.js
@@ -16,12 +16,17 @@ const SNAPPY_COMPRESSION = 2; // eslint-disable-line
 
 // time in ms. to wait for acks from Kafka
 const ACK_TIMEOUT = 100;
+// default (dumb) kafka partitioning - always goes to partition 0
+// const DEFAULT_PARTITIONER = 0;
+// decent choice to spread messages across partitions, but not as fair
+// as cyclic partitioner
+// const RANDOM_PARTITIONER = 1;
+// the cyclic partitioner guarantees fair dispersion across partitions
+const CYCLIC_PARTITIONER = 2;
 // use keyed-message mechanism for partitioning by default. This lets Kafka
 // choose a partition based on the key and ensures entries with the same key
 // ends up in the same partition
 const KEYED_PARTITIONER = 3;
-// default kafka partitioning
-const KAFKA_PARTITIONER = 0;
 // waits for an ack for messages
 const REQUIRE_ACKS = 1;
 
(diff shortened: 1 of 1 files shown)


rahulreddy:
@philipyoo  is right. I remember him having issues during Metrics.

jonathan-gramain:
@philipyoo I'm not sure I get what you mean. If we have a keyed partitioning (i.e. KEYED_PARTITIONER type), we're supposed to always go through `KeyedMessage` construction, unless we have `this._partition !== undefined` in which case we don't use keyed partitioning, which we do not want because we specify a unique partition to publish to.

What I added there is the other way, the possibility to not use keyed partitioning if not wanted, so that we write to multiple partitions.

Maybe I'm missing something?

jonathan-gramain:
I'm counting on the fact that `item.key` should be ignored if we're not using keyed partitioning, so I did not change this code (to be cleaner we could also have a third case that does not provide `item.key`, a `KeyedPartitioner` object nor a `partition` for default partitioning)

philipyoo:
Ah I see, thank you for the explanation!

jonathan-gramain:
updated, can you have another look?

jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine 
DEFAULT_BRANCH=rel/7.4
r+

jonathan-gramain:
@ironman-machine r- r+

jonathan-gramain:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=zenko/bbp2-rc1
try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine 
try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=rel/7.4
SCALITY_METADATA_BRANCH=rel/7.4
try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine 
try

jonathan-gramain:
@ironman-machine 
try

jonathan-gramain:
@ironman-machine 
try

jonathan-gramain:
Integration passed in http://ci.ironmann.io/gh/scality/Integration/19847

PR #187, opened 2018-02-24
Ft/ MongoDB oplogs

JianqinWang:
JIRA ticket: https://scality.atlassian.net/browse/ZENKO-15
https://github.com/scality/Arsenal/pull/425 needs to be merged first.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,13 @@ FROM node:6-slim
 
 WORKDIR /usr/src/app
 
+RUN apt-get update \
+    && apt-get install -y jq python git build-essential --no-install-recommends
+
 # Keep the .git directory in order to properly report version
 COPY . /usr/src/app
 
-RUN apt-get update \
-    && apt-get install -y jq python git build-essential --no-install-recommends \
-    && npm install --production \
+RUN npm install --production \
     && apt-get autoremove --purge -y python git build-essential \
     && rm -rf /var/lib/apt/lists/* \
     && npm cache clear \
--- a/conf/config.json
+++ b/conf/config.json
@@ -10,7 +10,7 @@
         "cronRule": "*/5 * * * * *",
         "batchMaxRead": 10000,
         "zookeeperPath": "/queue-populator",
-        "logSource": "dmd",
+        "logSource": "mongo",
         "bucketd": {
             "host": "127.0.0.1",
             "port": 9000
@@ -18,6 +18,14 @@
         "dmd": {
             "host": "127.0.0.1",
             "port": 9990
+        },
+        "mongo": {
+            "replicaSetHosts":
+                "localhost:27017,localhost:27018,localhost:27019",
+            "writeConcern": "majority",
+            "replicaSet": "rs0",
+            "readPreference": "primary",
+            "database": "metadata"
         }
     },
     "extensions": {
@@ -26,11 +34,11 @@
                 "transport": "http",
                 "s3": {
                     "host": "127.0.0.1",
-                    "port": 8000
+                    "port": 8001
                 },
                 "auth": {
-                    "type": "account",
-                    "account": "bart",
+                    "type": "service",
+                    "account": "service-replication",
                     "vault": {
                         "host": "127.0.0.1",
                         "port": 8500,
@@ -41,7 +49,7 @@
             "destination": {
                 "transport": "http",
                 "bootstrapList": [
-                    { "site": "sf", "servers": ["127.0.0.1:9443"],
+                    { "site": "zenko", "servers": ["localhost:8001"],
                       "echo": false },
                     { "site": "us-east-1", "type": "aws_s3" }
                 ],
@@ -51,8 +59,8 @@
                     "ca": "ssl/ca.crt"
                 },
                 "auth": {
-                    "type": "account",
-                    "account": "lisa"
+                    "type": "service",
+                    "account": "service-replication"
                 }
             },
             "topic": "backbeat-replication",
--- a/extensions/replication/queueProcessor/task.js
+++ b/extensions/replication/queueProcessor/task.js
@@ -28,14 +28,15 @@ const queueProcessor = new QueueProcessor(zkConfig,
 werelogs.configure({ level: config.log.logLevel,
     dump: config.log.dumpLevel });
 
+const logger = new werelogs.Logger('queueProcessorInit');
 function initAndStart() {
     initManagement(error => {
         if (error) {
-            console.error('could not load managment db', error);
+            logger.error('could not load managment db', error);
             setTimeout(initAndStart, 5000);
             return;
         }
-        console.log('management init done');
+        logger.info('management init done');
         queueProcessor.start();
         // setInterval(initManagement, 5000, error => {
         //     if (error) {
--- a/lib/credentials/AccountCredentials.js
+++ b/lib/credentials/AccountCredentials.js
@@ -2,13 +2,11 @@ const assert = require('assert');
 const AWS = require('aws-sdk');
 
 const errors = require('arsenal').errors;
-const management = require('../../../lib/management');
+const management = require('../../lib/management');
 
 class AccountCredentials extends AWS.Credentials {
-    constructor(authConfig, log) {
-        this._log = log;
-        this.initAuthData();
-        const accountInfo = this.getAuthData().accounts.find(
+    constructor(authConfig, getAuthDataCb, log) {
+        const accountInfo = getAuthDataCb().accounts.find(
             account => account.name === authConfig.account);
         if (accountInfo === undefined) {
             throw Error(`No such account registered: ${authConfig.account}`);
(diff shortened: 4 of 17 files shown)


lib/management.js
```diff
@@ -3,12 +3,44 @@ const forge = require('node-forge');
...
-    host: config.queuePopulator.dmd.host,
-    port: config.queuePopulator.dmd.port,
-});
+const Logger = require('werelogs').Logger;
+
+const logger = new Logger('mongoclient');
+
+const serviceBucket = 'PENSIEVE';
+
+let setup = false;
```
bennettbuchanan: I think it could be clearer to define this in the function `initManagement` where it is used and redefined.

extensions/replication/queueProcessor/task.js
```diff
@@ -35,7 +35,6 @@ function initAndStart() {
             setTimeout(initAndStart, 5000);
             return;
         }
-        console.log('management init done');
         queueProcessor.start();
         // setInterval(initManagement, 5000, error => {
```
bennettbuchanan: Since moving the `setInterval` to `initManagement` maybe we can go ahead and remove this comment?

lib/management.js
```diff
@@ -245,64 +290,72 @@ function saveServiceCredentials(conf, auth) {
...
+        return async.waterfall([
+            cb => {
+                if (!setup) {
+                    setup = true;
+                    metadata = new Metadata('mongodb', params, logger);
+                    return metadata.setup(() => cb());
+                }
+                return process.nextTick(cb());
+            },
+            cb => {
```
bennettbuchanan: Minor: for these steps in the waterfall we could save some lines by not using the braces (as done below https://github.com/scality/backbeat/pull/187/files#diff-c529f19cba932673dfc25817365dec76R334).

lib/management.js
```diff
@@ -3,12 +3,44 @@ const forge = require('node-forge');
...
-
-const MetadataFileClient = arsenal.storage.metadata.MetadataFileClient;
-const mdClient = new MetadataFileClient({
-    host: config.queuePopulator.dmd.host,
-    port: config.queuePopulator.dmd.port,
-});
+const Logger = require('werelogs').Logger;
+
+const logger = new Logger('mongoclient');
+
```
bennettbuchanan: It could be useful to add a configuation to werelogs so we can manage the log level from the config.json, e.g.:

```js
werelogs.configure({ 
    level: config.log.logLevel,
    dump: config.log.dumpLevel,
});
```

conf/config.json
```diff
@@ -70,7 +78,7 @@
         }
     },
     "log": {
-        "logLevel": "info",
+        "logLevel": "trace",
```
bennettbuchanan: Do we want to set the log level to trace?

lib/credentials/AccountCredentials.js
```diff
@@ -66,14 +72,15 @@ class StaticFileAccountCredentials extends AccountCredentials {
...
         assert.strictEqual(authConfig.type, 'service');
-        super(authConfig, log);
+        const serviceAuth = management.getLatestServiceAccountCredentials();
+        super(authConfig, log, serviceAuth);
     }
 
-    initAuthData() {
-        this._authdata = management.getLatestServiceAccountCredentials();
-    }
+    // initAuthData() {
```
bennettbuchanan: Is this okay to be removed?

lib/management.js
```diff
@@ -148,59 +183,65 @@ function applyReplicationState(conf, currentStateSerialized, db, cb) {
...
                 putReplication(stream, err => {
                     if (err) {
                         return cb(err);
                     }
                     return streamConfigured(stream, cb);
                 });
             } else {
-                console.log('stream has expected state, doing nothing', stream.streamId, stream.enabled);
+                logger.debug('stream has expected state, doing nothing' +
+                    `${stream.streamId}, ${stream.enabled}`);
```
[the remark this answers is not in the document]
JianqinWang: I asked around and I think people told me that it's ok to not include an object/second arg for the logger - I'm also just converting the previous console.logs, so I don't want to add too much unnecessary info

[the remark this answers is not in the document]
bennettbuchanan: Oh yes, this looks fine. I had seen the comma without realizing it was part of the template string.

Dockerfile
```diff
@@ -2,12 +2,13 @@ FROM node:6-slim
...
+RUN apt-get update \
+    && apt-get install -y jq python git build-essential --no-install-recommends
+
 # Keep the .git directory in order to properly report version
 COPY . /usr/src/app
 
-RUN apt-get update \
-    && apt-get install -y jq python git build-essential --no-install-recommends \
-    && npm install --production \
+RUN npm install --production \
```
rahulreddy: What is the reason for this change? Doesn't this add an extra layer to the image?

lib/management.js
```diff
@@ -3,12 +3,46 @@ const forge = require('node-forge');
...
 
-const MetadataFileClient = arsenal.storage.metadata.MetadataFileClient;
-const mdClient = new MetadataFileClient({
-    host: config.queuePopulator.dmd.host,
-    port: config.queuePopulator.dmd.port,
+werelogs.configure({
+    level: config.log.logLevel,
+    dump: config.log.dumpLevel,
 });
+const logger = new werelogs.Logger('mongoclient');
```
rahulreddy: Since this is for all backends (Metadata, Mongo), we should make it more general `MetadataManager` or `Management`, I am ok with either.

lib/management.js
```diff
@@ -3,12 +3,46 @@ const forge = require('node-forge');
...
+
+const serviceBucket = 'PENSIEVE';
+
+const Metadata = arsenal.storage.metadata.MetadataWrapper;
+let metadata;
+
+const params = {
```
rahulreddy: I am assuming that we are planning on cleaning this up later.

lib/management.js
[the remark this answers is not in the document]
rahulreddy: We wrote this down in logging guidelines so that everyone writes logs the same way https://github.com/scality/Guidelines/blob/master/CONTRIBUTING.md#examples

lib/management.js
```diff
@@ -148,59 +188,66 @@ function applyReplicationState(conf, currentStateSerialized, db, cb) {
...
+                logger.debug(`enabling new stream ${stream.streamId}`);
                 putReplication(stream, err => {
                     if (err) {
                         return cb(err);
                     }
                     return streamConfigured(stream, cb);
                 });
             } else {
-                console.log('new disabled stream, doing nothing', stream.streamId);
+                logger.log('new disabled stream, doing nothing' +
```
rahulreddy: `logger.debug` you mean?

lib/credentials/AccountCredentials.js
```diff
@@ -2,14 +2,20 @@ const assert = require('assert');
...
+
+const authdata = require('../../conf/authdata.json');
 
 class AccountCredentials extends AWS.Credentials {
-    constructor(authConfig, log) {
-        this._log = log;
-        this.initAuthData();
-        const accountInfo = this.getAuthData().accounts.find(
-            account => account.name === authConfig.account);
+    constructor(authConfig, log, serviceAuth) {
```
RachedBenMustapha: Why move the provisioned service account logic here? it makes `ProvisionedServiceAccountCredentials` basically a passthrough while the base class now has knowledge of several auth types

lib/management.js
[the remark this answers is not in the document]
RachedBenMustapha: Agreed, since you are just converting it's fine, but for werelogs logs the message should not have "content", content should go in an attached object to be searchable.

lib/management.js
RachedBenMustapha: This probably crashes

lib/queuePopulator/LogReader.js
```diff
@@ -252,6 +252,7 @@ class LogReader {
             readOptions.limit = params.maxRead;
         }
         this.log.debug('reading records', { readOptions });
+        // need to implement this method for mongo!!
```
RachedBenMustapha: what happens if it's not implemented?

lib/management.js
```diff
@@ -245,64 +297,72 @@ function saveServiceCredentials(conf, auth) {
...
+            cb => {
+                if (!setup) {
+                    setup = true;
+                    metadata = new Metadata('mongodb', params, bucketclient,
+                        logger);
+                    return metadata.setup(() => cb());
+                }
+                return process.nextTick(cb);
+            },
+            cb => metadata.listObject(serviceBucket, {}, logger, () => cb()),
```
RachedBenMustapha: Why list objects here?

lib/management.js
```diff
@@ -245,64 +297,72 @@ function saveServiceCredentials(conf, auth) {
...
+            cb => {
+                metadata.getObjectMD(serviceBucket,
+                'configuration/overlay-version', {}, logger, (err, res) =>
+                    cb(err, res));
+            },
+            (version, cb) => {
```
RachedBenMustapha: could be written as `(version, cb) => loadOverlayVersion(metadata, version, cb)`, as this waterfall is becoming busy

Dockerfile
JianqinWang: I think @ssalaues mentioned that this would speed up the docker build process?

ssalaues: Yeah it's a faster docker build because the apt update and apt install doesn't ever change so the the builds can use that cached layer. It's not at all critical to the PR but helped speed up the feedback loop during build and testing.

RachedBenMustapha: Yeah extra layers actually are a good thing when placed strategically, not sure why they have a bad rep ;)

lib/queuePopulator/LogReader.js
JianqinWang: Mmmmm I think this was a residual comment before the completion of this: https://github.com/scality/Arsenal/blob/master/lib/storage/metadata/mongoclient/LogConsumer.js
:sweat_smile:

JianqinWang:
@ironman-machine try

PR #188, opened 2018-02-26
FT: Update docs for replication to public cloud

bennettbuchanan:


JianqinWang (dismissed):
Thanks!

docs/crr-to-public-cloud.md
```diff
@@ -0,0 +1,123 @@
+# CRR from CloudServer to public cloud - Workflow
+
+## Installation
+
+An endpoint ('awsbackend', for example) will be defined in the [location
+configuration](https://github.com/scality/S3/blob/master/locationConfig.json) of
```
LaureVergeron: location configuration file?

docs/crr-to-public-cloud.md
```diff
@@ -0,0 +1,123 @@
...
+[CloudServer](https://github.com/scality/S3). Each location constraint in the
+configuration file has properties needed for performing operations for the
+location. For example, if replicating to AWS S3, the properties will include the
+bucket name, region, and credentials for AWS S3. This enables the multiple
+backend feature in CloudServer that Backbeat uses for CRR to a public cloud.
```
LaureVergeron: You could add a link somewhere in there to the S3 doc, to point them to the full info :) : http://s3-server.readthedocs.io/en/latest/USING_PUBLIC_CLOUDS/

docs/crr-to-public-cloud.md
```diff
@@ -0,0 +1,123 @@
...
+
+Replication endpoints are defined in the config for both Backbeat and
+CloudServer in Federation's [group_vars/all](https://github.com/scality/
```
LaureVergeron: Federation is EE only... do we have something to point open source users to? if ot's the paragraph below, can we add templates, like for Federation?

docs/crr-to-public-cloud.md
```diff
@@ -0,0 +1,123 @@
...
+bucket with replication enabled.
+
+![design](/res/queue-processor-sequence-diagram.png)
+
+#### CloudServer: PUT Bucket Replication Request
+
+* CloudServer receives the request and checks the storage class set in the
+  bucket replication configuration (see the
+  `StorageClass` property [here](https://docs.aws.amazon.com/
+  AWSJavaScriptSDK/latest/AWS/S3.html#putBucketReplication-property)).
```
LaureVergeron: We should specify that storage class is different for us and AWS: we don't support specifying the storage class in the AWS sense for replication workflow (Standard infrquent access, for example, which ppl tend to choose as it is cheaper)

LaureVergeron (commented):
Thanks for providing this, super neat! Just a few suggestions :)

PR #190, opened 2018-02-28
ft: add metrics consumer

philipyoo:
This PR is focused on consuming the metrics from Kafka and storing to Redis.

These Redis metric keys will be exposed by the Backbeat Server/API (#102) through to-be-built routes.

### DIFF
--- a/lib/MetricsConsumer.js
+++ b/lib/MetricsConsumer.js
@@ -1,0 +1,101 @@
+'use strict'; // eslint-disable-line strict
+
+const Logger = require('werelogs').Logger;
+const { RedisClient, StatsClient } = require('arsenal').metrics;
+
+const BackbeatConsumer = require('./BackbeatConsumer');
+const redisKeys = require('../extensions/replication/constants').redisKeys;
+
+// StatsClient constant defaults
+const INTERVAL = 300; // 5 minutes;
+const EXPIRY = 900; // 15 minutes
+
+// BackbeatConsumer constant defaults
+const CONSUMER_FETCH_MAX_BYTES = 5000020;
+const CONCURRENCY = 10;
+
+class MetricsConsumer {
+    /**
+     * @constructor
+     * @param {object} rConfig - redis configurations
+     * @param {string} rConfig.host - redis host
+     * @param {number} rConfig.port - redis port
+     * @param {object} mConfig - metrics configurations
+     * @param {string} mConfig.topic - metrics topic name
+     * @param {object} zkConfig - zookeeper configurations
+     * @param {string} zkConfig.connectionString - zookeeper connection string
+     *   as "host:port[/chroot]"
+     */
+    constructor(rConfig, mConfig, zkConfig) {
+        this.mConfig = mConfig;
+        this.zkConfig = zkConfig;
+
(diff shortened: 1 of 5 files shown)


philipyoo:
@ironman-machine
try

philipyoo:
@ironman-machine
try

philipyoo:
@ironman-machine clean try

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine reset

philipyoo:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=rel/7.4
try

lib/MetricsConsumer.js
```diff
@@ -0,0 +1,98 @@
...
+        */
+        if (data.type === 'processed') {
+            this._sendRequest(redisKeys.opsDone, data.ops);
+            this._sendRequest(redisKeys.bytesDone, data.bytes);
+        } else if (data.type === 'queued') {
+            this._sendRequest(redisKeys.ops, data.ops);
+            this._sendRequest(redisKeys.bytes, data.bytes);
+        } else {
+            // unknown type
+            this.logger.error('unknown type field encountered in metrics '
```
rahulreddy: I have a suggestion for the logger. I recommend using a new request logger for every entry. The problem is that module logger lives and keeps accumulating logs forever and dumps the entire logs once a logger.error is called. Having request logger and ending it after the request is processed ensures that logger object doesn't grow forever, it also helps identifying a specific entry on why it failed with the reqUid.

lib/MetricsConsumer.js
```diff
@@ -0,0 +1,98 @@
...
+    start() {
+        const consumer = new BackbeatConsumer({
+            zookeeper: { connectionString: this.zkConfig.connectionString },
+            topic: this.mConfig.topic,
+            groupId: 'backbeat-metrics-group',
+            concurrency: CONCURRENCY,
+            queueProcessor: this.processKafkaEntry.bind(this),
+            fetchMaxBytes: CONSUMER_FETCH_MAX_BYTES,
+        });
+        consumer.on('error', () => {});
```
rahulreddy: No need to log the errors here?

philipyoo: Jonathan previously mentioned errors are logged in the `BackbeatConsumer` class so no need to log here unless expanding on the cb

lib/MetricsConsumer.js
```diff
@@ -53,11 +53,12 @@ class MetricsConsumer {
     }
 
     processKafkaEntry(kafkaEntry, done) {
+        const log = this.logger.newRequestLogger();
         let data;
         try {
             data = JSON.parse(kafkaEntry.value);
         } catch (err) {
-            this.logger.error('error processing metrics entry', {
+            log.end().error('error processing metrics entry', {
```
rahulreddy: `log.end()` should be called always for all requests. You can use it this way
```js
try {
    data = JSON.parse(kafkaEntry.value);
} catch (err) {
   log.error('error processing metrics entry', {
       method: 'MetricsConsumer.processKafkaEntry',
                 error: err,
    });
    log.end();
    return done();
}


```

rahulreddy:
@ironman-machine clean try

PR #193, opened 2018-03-05
Ft/S3C-1148 multi cloud metrics

philipyoo:
This PR needs rebasing once the refactored routes PR is merged and forward ported to `rel/7.4-beta`

Support separate cloud metrics by specifying site name

### DIFF
--- a/extensions/replication/ReplicationQueuePopulator.js
+++ b/extensions/replication/ReplicationQueuePopulator.js
@@ -56,8 +56,12 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
                      `${queueEntry.getBucket()}/${queueEntry.getObjectKey()}`,
                      JSON.stringify(entry));
 
-        this._incrementMetrics(queueEntry.getSite(),
-            queueEntry.getContentLength());
+        const repSites = queueEntry.getReplicationInfo().backends;
+
+        // for one-to-many
+        repSites.filter(entry => entry.status === 'PENDING').forEach(site => {
+            this._incrementMetrics(site, queueEntry.getContentLength());
+        });
     }
 }
 
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -266,7 +266,7 @@ class BackbeatConsumer extends EventEmitter {
                 entry: { topic, partition, offset, key, timestamp },
             });
             this.emit('error', err, entry);
-        } else if (entry.topic === CRR_TOPIC) {
+        } else if (entry.topic === CRR_TOPIC && this._topic === CRR_TOPIC) {
             const qEntry = QueueEntry.createFromKafkaEntry(entry);
             if (!qEntry.error && qEntry instanceof ObjectQueueEntry) {
                 const bytes = qEntry.getContentLength();
@@ -426,6 +426,34 @@ class BackbeatConsumer extends EventEmitter {
      */
     getMetadata(params, cb) {
         this._consumer.getMetadata(params, cb);
+    }
+
+    _incrementMetrics(entry) {
+        const qEntry = QueueEntry.createFromKafkaEntry(entry);
+        if (!qEntry.error && qEntry instanceof ObjectQueueEntry) {
+            const bytes = qEntry.getContentLength();
+
+            const repSites = qEntry.getReplicationInfo().backends;
+            const sites = repSites.reduce((store, entry) => {
+                if (entry.status === 'PENDING') {
+                    store.push(entry.site);
+                }
+                return store;
+            }, []);
+
+            // for one-to-many
+            sites.forEach(site => {
+                if (!this._metricsStore[site]) {
+                    this._metricsStore[site] = {
+                        ops: 1,
+                        bytes,
+                    };
+                } else {
+                    this._metricsStore[site].ops++;
+                    this._metricsStore[site].bytes += bytes;
+                }
+            });
+        }
     }
 
     /**
--- a/lib/MetricsConsumer.js
+++ b/lib/MetricsConsumer.js
@@ -76,12 +76,13 @@ class MetricsConsumer {
                 type: 'processed'
             }
         */
+        const site = data.site;
         if (data.type === 'processed') {
-            this._sendRequest(redisKeys.opsDone, data.ops);
-            this._sendRequest(redisKeys.bytesDone, data.bytes);
+            this._sendRequest(`${site}:${redisKeys.opsDone}`, data.ops);
+            this._sendRequest(`${site}:${redisKeys.bytesDone}`, data.bytes);
         } else if (data.type === 'queued') {
-            this._sendRequest(redisKeys.ops, data.ops);
-            this._sendRequest(redisKeys.bytes, data.bytes);
+            this._sendRequest(`${site}:${redisKeys.ops}`, data.ops);
+            this._sendRequest(`${site}:${redisKeys.bytes}`, data.bytes);
         } else {
             // unknown type
             log.error('unknown type field encountered in metrics '
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -40,6 +40,9 @@ class BackbeatAPI {
         this._redisConfig = config.redis;
         this._logger = logger;
 
+        this._validSites = this._repConfig.destination.bootstrapList.map(
+            item => item.site);
+
         this._crrProducer = null;
         this._crrStatusProducer = null;
         this._metricProducer = null;
@@ -55,8 +58,9 @@ class BackbeatAPI {
             this._internalStart = Date.now();
         }
 
-        const redisClient = new RedisClient(this._redisConfig, this._logger);
-        this._statsClient = new StatsModel(redisClient, INTERVAL, EXPIRY);
+        this._redisClient = new RedisClient(this._redisConfig, this._logger);
+        this._statsClient = new StatsModel(this._redisClient, INTERVAL,
+            EXPIRY);
     }
 
     /**
@@ -142,15 +146,6 @@ class BackbeatAPI {
     }
 
     /**
-     * Check if Zookeeper and Producer are connected
-     * @return {boolean} true/false
-     */
-    isConnected() {
-        return this._zkClient.getState().name === 'SYNC_CONNECTED'
-            && this._checkProducersReady();
-    }
-
-    /**
      * Get Kafka healthcheck
      * @param {object} details - route details from lib/api/routes.js
      * @param {function} cb - callback(error, data)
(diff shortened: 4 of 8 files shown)


lib/BackbeatConsumer.js
```diff
@@ -115,14 +116,32 @@ class BackbeatConsumer extends EventEmitter {
...
+                        const repInfo = qEntry.getReplicationInfo();
+                        let site;
+                        if (repInfo.storageType) {
+                            // known external cloud (i.e. aws_s3 || azure)
+                            site = repInfo.storageType;
+                        } else if (repInfo.storageClass) {
+                            site = repInfo.storageClass;
+                        } else {
+                            const id = uuid().replace(/-/g, '');
+                            site = `bb-unknown-type-${id}`;
```
philipyoo: Just to note, I don't ever expect an unknown type to happen. It is here more as a precautionary measure and we will see the logs and may be able to debug from there.

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -52,7 +52,24 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
                      `${queueEntry.getBucket()}/${queueEntry.getObjectKey()}`,
                      JSON.stringify(entry));
 
-        this._incrementMetrics(entry.bucket, queueEntry.getBytesMetric());
```
jonathan-gramain: It seems we don't provide per-bucket metrics anymore, is it expected?

lib/BackbeatConsumer.js
```diff
@@ -115,14 +115,31 @@ class BackbeatConsumer extends EventEmitter {
                         const bytes = locations.reduce((sum, item) =>
                             sum + item.size, 0);
 
-                        if (!this._metricsStore[qEntry.bucket]) {
-                            this._metricsStore[qEntry.bucket] = {
+                        const repInfo = qEntry.getReplicationInfo();
```
jonathan-gramain: This event handler is already quite long and with lots of indentation levels. Can you try putting the metrics code into a separate method? It should help clarity.

lib/api/BackbeatAPI.js
```diff
@@ -281,12 +310,47 @@ class BackbeatAPI {
...
      * @param {function} cb - callback(err, res)
      * @return {undefined}
      */
-    _queryStats(ops, cb) {
+    _queryStats(ops, site, cb) {
         return async.map(ops, (op, done) => {
-            this._statsClient.getStats(this._logger, op, done);
+            const queryString = `${op}:${site}`;
+            // Query all sites if given wildcard
+            if (queryString.indexOf('*') >= 0) {
```
jonathan-gramain: Why not checking `site === '*'` directly?

lib/api/BackbeatAPI.js
```diff
@@ -281,12 +310,47 @@ class BackbeatAPI {
...
+                            type: errors.InternalError,
+                            method: 'BackbeatAPI._queryStats',
+                        });
+                    }
+                    const allKeys = res.map(key => {
+                        const arr = key.split(':');
+                        // Remove the "requests:<timestamp>" and process
+                        return arr.slice(0, arr.length - 2).join(':');
+                    });
+                    const reducedKeys = [...new Set(allKeys)];
```
jonathan-gramain: I like this construct to uniquify an array :+1: 

Though is the `scan` function prone to returning duplicate keys in `res`?

lib/api/BackbeatServer.js
```diff
@@ -30,18 +30,20 @@ class BackbeatServer {
     /**
      * Last log message to send per incoming request
      * @param {werelogs.newRequestLogger} logger - request logger
+     * @param {object} data - if error, can carry a specific error message
      * @param {object} req - request object
      * @param {object} res - response object
      * @return {undefined}
      */
-    _logRequestEnd(logger, req, res) {
+    _logRequestEnd(logger, data, req, res) {
```
jonathan-gramain: Since `data` is optional and `req` and `res` are mandatory params, it would be less intrusive and more appropriate to put it as last argument, and `data` looks somewhat too generic for a variable name here.

lib/api/BackbeatServer.js
```diff
@@ -120,8 +122,19 @@ class BackbeatServer {
         && this._areConditionsOk(bbRequest)) {
             bbRequest.setStatusCode(200);
 
-            const routeDetails = routes.find(route =>
-                route.path === bbRequest.getRoute());
+            const route = bbRequest.getRoute();
+            let routeDetails = routes.find(r => r.path === route);
+            if (!routeDetails) {
+                // a valid route with an optional param
+                const subRoute = route.substr(0, route.lastIndexOf('/'));
```
jonathan-gramain: Do you mean parent route instead of sub-route?

lib/MetricsConsumer.js
```diff
@@ -74,12 +74,13 @@ class MetricsConsumer {
                 type: 'processed'
             }
         */
+        const site = data.site;
         if (data.type === 'processed') {
-            this._sendRequest(redisKeys.opsDone, data.ops);
-            this._sendRequest(redisKeys.bytesDone, data.bytes);
+            this._sendRequest(`${redisKeys.opsDone}:${site}`, data.ops);
```
jonathan-gramain: Following the API route organization, I would rather put the `site` argument in front in the key.

lib/api/BackbeatServer.js
```diff
@@ -120,8 +122,19 @@ class BackbeatServer {
...
+                routeDetails = routes.find(r => r.path === subRoute);
+
+                const site = route.substr(route.lastIndexOf('/') + 1);
```
jonathan-gramain: I think routes would be better organized (and also less prone to errors/name conflicts) in the following way (I'm not sure of the current exact semantics, so feel free to adapt to what is currently there, but I want to show the general idea).

```
/_/metrics/crr/all/backlog
/_/metrics/crr/all/throughput
...
/_/metrics/crr/aws_s3/throughput
```

Later we could also provide additional grouping capability like:
```
/_/metrics/by_site/aws_s3     // would provide all metrics related to aws_s3.
```

The first level names should be reserved as namespace, with a fixed number that we control in the implementation, avoiding the need to hammer in with some pain the new metrics later for various other components. Since no one has deployed or tooled this API yet, I think it's a good time to do this change.

extensions/replication/ReplicationQueuePopulator.js
philipyoo: Previously, I had collected metrics by bucket, but was never used. Because of time constraints, I decided to go ahead and swap that out for `site`. 

Also, important to note is we want to keep redis keys as low as possible. Each new layer adds exponentially more keys. Adding sites parameters increased keys by the number of sites :/

lib/api/BackbeatAPI.js
philipyoo: Yes there is a possibility `scan` will return duplicate keys

lib/api/BackbeatServer.js
philipyoo: hmm, for now i'll name it `desc`. I want this field to be a more informative message, so I added this field :)

lib/api/BackbeatServer.js
philipyoo: I'll be reflecting these changes in a new PR 👍 . For now, I'll build a new branch off this one so it will be easy to rebase later

philipyoo:
Consumption metrics is not accumulating properly, at least in my test environment. For one-to-many replication, I'm experiencing `opsdone` and `bytesdone` accumulating only for "philz-site". The "aws_s3" site is not accumulating for `opsdone` and `bytesdone`.

I checked group-ids and looks fine. Watching `kafka-console-consumer`, I only see a single entry, but the replication to multiple sites is working properly.

philipyoo:
@ironman-machine try

lib/BackbeatConsumer.js
```diff
@@ -161,6 +147,34 @@ class BackbeatConsumer extends EventEmitter {
         return this;
     }
 
+    _incrementMetrics(entry) {
+        const qEntry = QueueEntry.createFromKafkaEntry(entry);
+        if (!qEntry.error) {
```
bennettbuchanan: Looks like we should include a check that this an instance of `ObjectQueueEntry`, as it can sometimes be a `BucketQueueEntry` instance, in which case there is no method `getContentLength`.

philipyoo:
@ironman-machine try

lib/BackbeatConsumer.js
```diff
@@ -149,7 +149,7 @@ class BackbeatConsumer extends EventEmitter {
 
     _incrementMetrics(entry) {
         const qEntry = QueueEntry.createFromKafkaEntry(entry);
-        if (!qEntry.error) {
+        if (!qEntry.error && qEntry.constructor.name === 'ObjectQueueEntry') {
```
jonathan-gramain: a more standard way to do the same (plus supports inheritance) is `qEntry instanceof ObjectQueueEntry`

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -52,7 +52,18 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+        const repSites = queueEntry.getReplicationInfo().backends;
+        const sites = repSites.reduce((store, entry) => {
```
jonathan-gramain: A cleaner way for this is to use `Array.filter` method:

```
const sites = repSites.filter(entry => entry.status === 'PENDING')
```

Or you may also simply add this check in the below `forEach` loop directly.

lib/BackbeatConsumer.js
philipyoo: Ah cool, this way seemed a bit hacky to me so ty, will change this

lib/BackbeatConsumer.js
```diff
@@ -161,6 +147,34 @@ class BackbeatConsumer extends EventEmitter {
         return this;
     }
 
+    _incrementMetrics(entry) {
+        const qEntry = QueueEntry.createFromKafkaEntry(entry);
+        if (!qEntry.error && qEntry.constructor.name === 'ObjectQueueEntry') {
```
jonathan-gramain: Commented in the commit itself, reporting here for clarity: you can do `qEntry instanceof ObjectQueueEntry`, which also supports inheritance from ObjectQueueEntry.

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=true
try

PR #196, opened 2018-03-09
bf: statsmodel intervals

philipyoo:
StatsClient normalized time by seconds. Since we want to use
intervals by a default of 5 minutes (or configurable to be higher),
we want to normalize time by minutes.

Please only review 877feceb10d9f16cd38b1b4454cfa10d288b048b

### DIFF
--- a/lib/models/StatsModel.js
+++ b/lib/models/StatsModel.js
@@ -1,0 +1,22 @@
+const { StatsClient } = require('arsenal').metrics;
+
+/**
+ * @class StatsModel
+ *
+ * @classdesc Extend and overwrite how timestamps are normalized by minutes
+ * rather than by seconds
+ */
+class StatsModel extends StatsClient {
+
+    /**
+    * normalize to the nearest interval
+    * @param {object} d - Date instance
+    * @return {number} timestamp - normalized to the nearest interval
+    */
+    _normalizeTimestamp(d) {
+        const m = d.getMinutes();
+        return d.setMinutes(m - m % (Math.floor(this._interval / 60)), 0, 0);
+    }
+}
+
+module.exports = StatsModel;
--- a/lib/MetricsConsumer.js
+++ b/lib/MetricsConsumer.js
@@ -1,8 +1,9 @@
 'use strict'; // eslint-disable-line strict
 
 const Logger = require('werelogs').Logger;
-const { RedisClient, StatsClient } = require('arsenal').metrics;
+const { RedisClient } = require('arsenal').metrics;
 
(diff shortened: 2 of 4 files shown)


lib/models/StatsModel.js
```diff
@@ -0,0 +1,22 @@
...
+class StatsModel extends StatsClient {
+
+    /**
+    * normalize to the nearest interval
+    * @param {object} d - Date instance
+    * @return {number} timestamp - normalized to the nearest interval
+    */
+    _normalizeTimestamp(d) {
+        const m = d.getMinutes();
+        return d.setMinutes(m - m % (Math.floor(this._interval / 60)), 0, 0);
```
jonathan-gramain: Ideally we would make sure `this._interval` is greater than 60 to avoid doing a modulo by 0. No big deal though as it's just a validation issue.

lib/models/StatsModel.js
```diff
@@ -0,0 +1,22 @@
...
+ * rather than by seconds
+ */
+class StatsModel extends StatsClient {
+
+    /**
+    * normalize to the nearest interval
+    * @param {object} d - Date instance
+    * @return {number} timestamp - normalized to the nearest interval
+    */
+    _normalizeTimestamp(d) {
```
jonathan-gramain: Why is this a private member? I don't see it used in the class itself so I expect it to be called from outside?

philipyoo: It is used in the `StatsClient` class, and since other components are reliant on the class, it seemed best to override this helper function locally for our project rather than creating or extending the functionality of `StatsClient`. Utapi relied on this method to normalize timestamps by seconds, but for backbeat, we want it in minutes.

Also, important to note that expiry and intervals should be set by fixed minutes. So for bloomberg release, it is important for configurations of these variables to be handled properly. Maybe it will have to be by fixed minutes? Not too sure at the moment

lib/models/StatsModel.js
philipyoo: Ya I agree, I will need to think of a better way for validating the configuration options of interval and expiry to guarantee that the user will set these variables properly or handle the edge cases. For now, interval and expiry are always fixed.

lib/models/StatsModel.js
jonathan-gramain: I'm lacking some context to answer this, but thanks for the explanation for the overriden function.

philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #198, opened 2018-03-10
remove broken design with unlockBucket messages

jonathan-gramain:
The issue with the current design is that we only send one
    'unlockBucket' message after the bucket has been listed entirely, but
    nothing guarantees that previous listing tasks executed by other
    producers will not be longer to proceed, effectively voiding the
    purpose of knowing when all objects of this bucket have been
    processed.
    
    One case where this is very likely to happen is when a specific range
    of objects in the bucket have to be expired at once (other than the
    last listed range), because it will take longer to process objects for
    that particular range than it will be to finish listing the bucket and
    processing the expirations for last range. Moreover this is a common
    case in production (e.g. expiring a sub-directory).
    
    An alternative implementation based on monitoring the backlog of kafka
    to apply backpressure will be done instead.

### DIFF
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -2,7 +2,6 @@
 
 const async = require('async');
 const schedule = require('node-schedule');
-const zookeeper = require('node-zookeeper-client');
 
 const Logger = require('werelogs').Logger;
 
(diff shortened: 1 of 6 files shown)


jonathan-gramain:
@ironman-machine 
SCALITY_METADATA_BRANCH=zenko/bbp2-rc2
DEFAULT_BRANCH=rel/7.4
try

jonathan-gramain:
@ironman-machine
SCALITY_METADATA_BRANCH=rel/7.4
DEFAULT_BRANCH=rel/7.4
try

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=fwd/7.4/ft/S3C-1154-lifecycle
try

PR #209, opened 2018-03-15
bf: rework backpressure

jonathan-gramain:
* bf: add backlog metrics to BackbeatConsumer

Now BackbeatConsumer will publish metrics periodically in zookeeper
about the current backlog: consumer offset and latest topic
offset. This will be used to check how much is the backlog (or if
there is any at all) and throttle processing if needed.

* bf: backpressure for lifecycle

In order to avoid accumulating lifecycle tasks and diverging when the
system cannot process everything in the allocated time, add a
mechanism to skip processing cycles when the previous cycle has not
finished yet.

Kafka consumers in LifecycleConsumer and LifecycleProducer now publish
to zookeeper, for their consumed topic partitions, their current
committed offset and the latest topic partition offset.

The conductor checks those when he's about to start a new cycle: if
they are not equal (i.e. if the backlog is non-null), it means there
are still things being processed, so the conductor does not trigger a
new cycle and waits for the next one.

### DIFF
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -2,7 +2,9 @@
 
 const async = require('async');
 const schedule = require('node-schedule');
+const zookeeper = require('node-zookeeper-client');
 
+const errors = require('arsenal').errors;
 const Logger = require('werelogs').Logger;
 
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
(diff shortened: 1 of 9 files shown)


lib/BackbeatConsumer.js
```diff
@@ -343,14 +484,31 @@ class BackbeatConsumer extends EventEmitter {
...
+                    this._consumer.unsubscribe();
+                    this._consumer.disconnect();
+                    this._consumer.on('disconnected', () => next());
+                } else {
+                    process.nextTick(next);
+                }
+            },
+            next => {
+                if (this._publishBacklogMetricsTimer) {
+                    clearInterval(this._publishBacklogMetricsTimer);
```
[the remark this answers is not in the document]
jonathan-gramain: indeed, thanks for catching this :+1:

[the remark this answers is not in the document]
jonathan-gramain: fixed

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -130,6 +154,100 @@ class LifecycleConductor {
         ], () => {});
     }
 
+    _controlBacklog(done) {
+        if (!this.lcConfig.conductor.backlogControl.enabled ||
+            !this.lcConfig.backlogMetrics) {
+            return process.nextTick(done);
```
philipyoo: Since `_controlBacklog` is called within an `async.waterfall`, would returning `done()` here work as well?

jonathan-gramain: Would work indeed, though I think it's better to stay safe, i.e. be consistent in the scope of a single function (in case this function would later be called from a sync context).

jonathan-gramain: And performance here is a secondary concern.

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine 
RUN_BACKBEAT_CRR_TESTS=1
r+

jonathan-gramain:
Integration passed in http://ci.ironmann.io/gh/scality/Integration/20529

jonathan-gramain:
@ironman-machine r-

PR #210, opened 2018-03-16
bf: disable removal of zookeeper bucket node

jonathan-gramain:
Rationale:

From the raft log, we observe that bucket operations come from
multiple raft sessions (observed bucket metadata for one bucket from
sessions 1, 2 and 3), where the original expectation was that
operations on one bucket came from only one raft session at a time.

This means that we have no guarantee in which order we process
operations on a single bucket, since raft sessions are processed
independently by multiple queue populators. Because of time
constraints on delivery we disable the removal of bucket nodes for
now, but we should consider re-enabling removal with a correct method
later.


jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine r-

jonathan-gramain:
We're going to change the solution for this (see ticket description in https://scality.atlassian.net/browse/S3C-1327).

PR #211, opened 2018-03-19
properly shutdown lifecycle processes

jonathan-gramain:
Catch SIGTERM in producer and consumer processes and call close() functions.


jonathan-gramain:
@philipyoo Not specific to node-rdkafka, as closing also flushes the consumer commit offsets, otherwise the component once restarted will resume from where it last committed, which could be far off in time, hence redoing a bunch of operations. With auto-commit on (for lifecycle only) it should not be too far off though.

For other components, yes it's best to make sure they also shut down properly. This is done for lifecycle components, for CRR processes it should already be ok I think.

jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine r-

PR #212, opened 2018-03-20
transient source design doc

jonathan-gramain:


docs/transient-crr-source.md
```diff
@@ -0,0 +1,193 @@
...
+refused until the transient source used space diminishes below a
+certain threshold. This will be done as a separate feature though.
+
+## Rejected Options
+
+There were discussions about implementing a caching layer on top of
+S3, where all incoming data would be duplicated for a certain period
+of time. This option has been rejected because of its complexity and
+the short time frame available, and because we found a satisfactory
+simpler alternative.
```
vrancurel, line 225: could you indicate here (in rejected options) why we don't have a transient: true property on a location to indicate a transient source but rather have a transient_source: location in every target location.

jonathan-gramain: updated

vrancurel (dismissed):

docs/transient-crr-source.md
```diff
@@ -0,0 +1,177 @@
+# Transient CRR Source
+
+## Description
+
+The goal of this enhancement is to save the cost of reading data from
```
bennettbuchanan: I think we can use the word "avoid" instead of "save" here so it's not interpreted as if there is still some cloud charge involved in the read.

docs/transient-crr-source.md
```diff
@@ -0,0 +1,177 @@
...
+
+Once the status is globally *COMPLETED* for an object (i.e. all CRR
+targets are up-to-date), if the object location is transient, before
+writing the source metadata through the backbeat route, the
+replication status processor also updates the location of the object
+to be the default configured CRR target location, if it is one of the
+object's CRR targets, otherwise behavior TBD.
+
+If the update succeeds, it then publishes a message to the
+`backbeat-gc` queue to delete the transient **data**. The messages
```
bennettbuchanan: What is the significance of bolding "data"?

docs/transient-crr-source.md
```diff
@@ -0,0 +1,177 @@
...
+should contain a list of data locations to delete, in a similar format
+than in the original object's locations.
```
bennettbuchanan: ~than in~ as

docs/transient-crr-source.md
```diff
@@ -0,0 +1,177 @@
...
+
+This means that backbeat must be enabled to use transient source.
+
+## Operational Considerations
+
+### Link with CRR
+
+This feature only works when buckets are configured with at least one
+replication target, and likely makes real sense when there are at
+least two targets.
```
bennettbuchanan: It seems like we should actually avoid using a transient location if just replicating to a sole target public cloud as it would be a superfluous use of resources. Maybe it's not worth complicating though.

docs/transient-crr-source.md
```diff
@@ -0,0 +1,177 @@
...
+  being a transient source, then the put status is returned as usual,
+
+* CRR processors will transfer the object asynchronously to all CRR
+  targets set through replication API (also as usual),
+
+* finally, once all CRR targets have been processed, the object
+  location will be altered to be the preferred (default) CRR target
+  location, and the transient data will be deleted asynchronously by a
+  garbage collection process.
+
```
bennettbuchanan: Maybe add some explanation about what happens in the case that one of the backends fails. Do we still want to clean up the transient data? Perhaps in that case we can accept any future read from the updated location?

bennettbuchanan (dismissed):
Cool! 👍

philipyoo (dismissed):
sounds like a cool idea, I like it 👍

docs/transient-crr-source.md
jonathan-gramain: That is to stress that the object existence (metadata) is not touched. But if it raises questions I will remove the bolding.

docs/transient-crr-source.md
jonathan-gramain: As it is implemented today, it expresses functionally as "the location is transient as long as you have some (one or more) replication targets", where replacing by "as long as you have two or more replication targets" could make the behavior more confusing.

Also there may be use cases where having a transient source doing CRR to a single public cloud could make sense, e.g. if you have a very low local storage latency and you want to benefit from this for fast local writes, and have background parallel writes to the cloud through CRR.

jonathan-gramain: In this respect I could remove the "likely makes real sense..." part.

bennettbuchanan (dismissed):

PR #216, opened 2018-03-20
FT: Add replication retry documentation

bennettbuchanan:
"I can accept failure, everyone fails at something. But I can't accept not trying."

― Michael Jordan


docs/replication-retry.md
```diff
@@ -0,0 +1,134 @@
...
+        bucket: <bucket>,
+        key: <key>,
+        versionId: <versionId>,
+        site: <site>
+    }]
+    ```
+
+* POST `/_/crr/retry`
+
+    This POST request retries a failed operation.
```
jonathan-gramain: You may also support a list input compatible with the result of the GET (making it simple to retry all failed operations at once).

docs/replication-retry.md
```diff
@@ -0,0 +1,134 @@
...
+        versionId: <versionId>,
+        site: <site>
+    }
+    ```
+
+    Response:
+
+    ```sh
+    {
+        site: <site>,
```
jonathan-gramain: Did you mean to also output the original params like `bucket`, `key` and `versionId` here?

jonathan-gramain (dismissed):
Looks good! Just a few comments around the API.

docs/replication-retry.md
```diff
@@ -0,0 +1,134 @@
...
+* The retry of a specific failed operation.
+
+## Design
+
+A RESTful API will expose methods for users to list and retry failed operations.
+See [Explanations](#explanations) for more detail.
+
+## Definition of API
+
+* GET `/_/crr/list/failed/all`
```
jonathan-gramain: maybe `/list` is superfluous, i.e. `/_/crr/failed/...` can be simpler and organized enough for the purpose.

docs/replication-retry.md
```diff
@@ -0,0 +1,134 @@
...
+    ```sh
+    [{
+        bucket: <bucket>,
+        key: <key>,
+        versionId: <versionId>,
+        site: <site>
+    }]
+    ```
+
+* POST `/_/crr/retry`
```
jonathan-gramain: For consistency with the GET, maybe use the same prefix: `/_/crr/failed/retry` or `/_/crr/failed?op=retry`

docs/replication-retry.md
jonathan-gramain: The idea behind is that a GET already carries the notion of retrieval, so if the GET occurs on a "directory-like" structure it implies we want a listing of this structure.

jonathan-gramain: So for getting all failed entries, you could also support directly `GET /_/crr/failed`

docs/replication-retry.md
jonathan-gramain: The second form stays compatible with using a GET scheme like `/_/crr/failed/bucket/object/versionId`

docs/replication-retry.md
bennettbuchanan: I figured this would be redundant, but with the change to support multiple retries in the POST request, it makes sense to include those params. I will do so, and update the response to be a list of retries as well since it could be multiple retries.

docs/replication-retry.md
bennettbuchanan: Good point—this simplifies the route. I will update this.

docs/replication-retry.md
bennettbuchanan: Will update to `/_/crr/failed/retry`.

docs/replication-retry.md
bennettbuchanan: Great idea!

bennettbuchanan (commented):
Thanks for the helpful feedback!

jonathan-gramain (dismissed):

docs/replication-retry.md
```diff
@@ -0,0 +1,137 @@
...
+* The retry of a specific failed operation.
+
+## Design
+
+A RESTful API will expose methods for users to list and retry failed operations.
+See [Explanations](#explanations) for more detail.
+
+## Definition of API
+
+* GET `/_/crr/failed/all`
```
philipyoo: For PR#200, I refactored the routes to follow:
`/_/metrics/<extension>/<site>/<specific-metric>`

Can these routes be edited to follow the same design? Maybe something like:
`/_/retry/crr/all`
`/_/retry/crr/all/<bucket>/<key>/<versionId>`

docs/replication-retry.md
```diff
@@ -0,0 +1,137 @@
...
+        versionId: <versionId>,
+        site: <site>
+    }]
+    ```
+
+* POST `/_/crr/failed/retry`
+
+    This POST request retries a failed operation.
+
+    Request Body:
```
philipyoo: Does this request only retry specified objects? Maybe also add an option for retrying everything that has failed?

docs/replication-retry.md
```diff
@@ -0,0 +1,137 @@
+# Replication Retry
```
bennettbuchanan: Change references of "Replication" to "Cross Region Replication (CRR)".

docs/replication-retry.md
rahulreddy: I think `/all` can be dropped as it's implied already by the route.

docs/replication-retry.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Explanations
+
+### Redis
+
+The replication status processor sets a Redis key for any backend with a FAILED
+status with the following key schema:
+
+```sh
+backbeat:crr:failed:<bucket>:<key>:<versionId>:<site>
```
rahulreddy: Have you evaluated Lists data structure instead of keys? How are you planning on searching the key pattern, using `keys <prefix>*` or `srange`? You may want to do some sizing and performance tests comparing these before choosing the data structure. Putting a large number of keys using test scripts would help.

docs/replication-retry.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+The value of the Redis key is the object's metadata at the time of failure.
+
+### Listing
+
+When a GET request is received for the listing route (e.g.,
+`/_/crr/failed/all`), all Redis keys beginning with `backbeat:crr:failed` will
```
rahulreddy: Please performance test the listing call (see above for more explanation).

docs/replication-retry.md
bennettbuchanan: Updated.

docs/replication-retry.md
bennettbuchanan: Okay, as discussed I will leave the scheme like this and accommodate the difference in the route parser.

docs/replication-retry.md
bennettbuchanan: I agree that would be a helpful feature, and I will aim to add this route. If it ends up adding too much for the delivery for March 30th, I suppose we can settle on using the response from the `/_/crr/failed/all` as the POST body.

docs/replication-retry.md
bennettbuchanan: I would like to use a Redis hash where the hash key will be `backbeat:crr:failed`. The hash fields then would be `<bucket>:<key>:<versionId>:<site>`, the field value being the object's metadata. Both `GET` and `HGET` have a constant time complexity O(1), but we would only have one key using a hash which is appealing to me. I am not considering Lists because I want a data structure with a key-value schema.

With a hash, we can store up to (2^32 - 1) field-value pairs, and probably if we reach that limit then we have a greater problem on our hands. Also, as per Redis docs, "small hashes (i.e., a few elements with small values) are encoded in special way in memory that make them very memory efficient"—I think this is relevant since the number of failures will likely (ideally) be small at any given point.

philipyoo (dismissed):

docs/replication-retry.md
bennettbuchanan: For some more numbers using keys with mock object metadata: 
* The serialized length of a hash with 1,000,000 keys is 989887235.
* The serialized length of a string key is 1055 or (1,000,000 * 1055) = 1055000000.

bennettbuchanan:
Some additions:

* Set expiry of hash fields to seven days—which should be configurable.
* Response should be format:

```
{
    "Versions": [
        {
            "LastModified": "2018-04-13T21:03:24.940Z",
            "VersionId": "null",
            "ETag": "\"9fc9d606912030dca86582ed62595cf7\"",
            "StorageClass": "STANDARD",
            "Key": "bar",
            "Owner": {
                "DisplayName": "retryuser",
                "ID": "1d79449d2b781e585051d057ecf99cad1898120cda80b7b2a94e3fe99d8cad82"
            },
            "IsLatest": true,
            "Size": 12
        },
[...]
```

Where `IsLatest` is always `false`, and `StorageClass` is the failed backend.

POST request should take the same format, and give a listing in the format of a HEAD object request, which includes the replication status.

docs/replication-retry.md
```diff
@@ -59,24 +91,32 @@ See [Explanations](#explanations) for more detail.
...
-        versionId: <versionId>,
-        site: <site>
-    }]
+    {
+        Versions: [{
+            Bucket: <bucket>,
+            Key: <key>,
+            VersionId: <version-id>,
+            StorageClass: <site>,
+            Size: <size>,
```
jonathan-gramain: Size and LastModified fields should not be mandatory in the request, so better not document them (if the user provides the actual listing result from the GET it should work though, but a user may decide to create its own list from another data source as long as he complies with the API).

Also I guess in the description "This POST request retries a failed operation." we should say "retries a set of failed operations" instead.

docs/replication-retry.md
```diff
@@ -18,21 +18,47 @@ See [Explanations](#explanations) for more detail.
...
+        IsTruncated: false,
+        Versions: [{
+            Bucket: <bucket>,
+            Key: <key>,
+            VersionId: <version-id>,
+            StorageClass: <site>,
+            Size: <size>,
+            LastModified: <last-modified>,
+        }],
+        MaxKeys: 100,
```
jonathan-gramain: Is 100 the default value? I would recommend to use the same default than S3 default (1000 AFAIK) for the principle of least surprise and also because it may help with large listings.

docs/replication-retry.md
bennettbuchanan: I agree. The implementation handles it as such. I will update this.

docs/replication-retry.md
bennettbuchanan: Updated so that Redis will return the result of a count of 1000 from the hash. This could end up being 1000 entries or slightly more, so I've removed the `MaxKeys` value for now.

jonathan-gramain (dismissed):

docs/replication-retry.md
```diff
@@ -0,0 +1,217 @@
...
+* The retry of a specific failed operation.
+
+## Design
+
+A RESTful API will expose methods for users to list and retry failed operations.
+See [Explanations](#explanations) for more detail.
+
+## Definition of API
+
+* GET `/_/backbeat/api/_/crr/failed?marker=marker`
```
jonathan-gramain, line 21: This route and the following should be: `/_/backbeat/api/crr/failed?...`

PR #217, opened 2018-03-21
Bf/S3C-1341 fix noncurrentversionexpiration

philipyoo:
Includes separate commits for functional tests (expiration and ncve)

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -9,7 +9,7 @@ const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 const RETRYTIMEOUTS = 300;
 
 // Default max AWS limit is 1000 for both list objects and list object versions
-const MAX_KEYS = 1000;
+const MAX_KEYS = process.env.TEST_SWITCH ? 3 : 1000;
 // concurrency mainly used in async calls
 const CONCURRENCY_DEFAULT = 10;
 
@@ -530,12 +530,20 @@ class LifecycleTask extends BackbeatTask {
             async.waterfall([
                 next => this._getRules(bucketData, lcRules, obj, log, next),
                 (applicableRules, next) => {
-                    if (versioned) {
-                        return this._compareVersion(bucketData, obj,
-                            applicableRules, log, next);
+                    let rules = applicableRules;
+                    // Hijack for testing
+                    // Idea is to set any "Days" rule to `Days - 1`
+                    const testIsOn = process.env.TEST_SWITCH === '1';
+                    if (testIsOn) {
+                        rules = this._adjustRulesForTesting(rules);
                     }
-                    return this._compareObject(bucketData, obj,
-                        applicableRules, log, next);
+
+                    if (versioned) {
+                        return this._compareVersion(bucketData, obj, rules, log,
+                            next);
+                    }
+                    return this._compareObject(bucketData, obj, rules, log,
+                        next);
                 },
             ], cb);
         }, done);
@@ -636,7 +644,7 @@ class LifecycleTask extends BackbeatTask {
                         }
                     });
                 }
-                if (!alreadySent && rules.Expiration.Days &&
+                if (!alreadySent && rules.Expiration.Days !== undefined &&
                 daysSinceInitiated >= rules.Expiration.Days) {
                     alreadySent = true;
                     const entry = {
@@ -664,6 +672,37 @@ class LifecycleTask extends BackbeatTask {
 
             return done();
         });
+    }
+
+    /**
+     * Only to be used when testing (when process.env.TEST_SWITCH).
+     * The idea is to adjust any "Days" or "NoncurrentDays" rules so that rules
+     * set with 1 day should expire, but any days set with 2+ days will not.
+     * Since Days/NoncurrentDays cannot be set to 0, this is a way to set the
+     * rule and test methods are working as intended.
+     * @param {object} rules - applicable rules
+     * @return {object} adjusted rules object
+     */
+    _adjustRulesForTesting(rules) {
+        /* eslint-disable no-param-reassign */
+        if (rules.Expiration &&
+        rules.Expiration.Days) {
+            rules.Expiration.Days--;
+        }
+        const ncve = 'NoncurrentVersionExpiration';
+        const ncd = 'NoncurrentDays';
+        if (rules[ncve] &&
+        rules[ncve][ncd]) {
+            rules[ncve][ncd]--;
+        }
+        const aimu = 'AbortIncompleteMultipartUpload';
+        const dai = 'DaysAfterInitiation';
+        if (rules[aimu] &&
+        rules[aimu][dai]) {
+            rules[aimu][dai]--;
+        }
+        /* eslint-enable no-param-reassign */
+        return rules;
     }
 
     /**
@@ -698,7 +737,8 @@ class LifecycleTask extends BackbeatTask {
         const daysSinceInitiated = this._findDaysSince(new Date(staleDate));
         const ncve = 'NoncurrentVersionExpiration';
         const ncd = 'NoncurrentDays';
-        const doesNCVExpirationRuleApply = (rules[ncve] && rules[ncve][ncd] &&
+        const doesNCVExpirationRuleApply = (rules[ncve] &&
+            rules[ncve][ncd] !== undefined &&
             daysSinceInitiated >= rules[ncve][ncd]);
         if (doesNCVExpirationRuleApply) {
             const entry = {
@@ -753,7 +793,7 @@ class LifecycleTask extends BackbeatTask {
                 new Date(upload.Initiated));
             const abortRule = aRules.AbortIncompleteMultipartUpload;
             const doesAbortRuleApply = (abortRule &&
-                abortRule.DaysAfterInitiation &&
+                abortRule.DaysAfterInitiation !== undefined &&
                 daysSinceInitiated >= abortRule.DaysAfterInitiation);
             if (doesAbortRuleApply) {
                 log.debug('send mpu upload for aborting', {
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -1,6 +1,7 @@
 'use strict'; // eslint-disable-line
 
 const async = require('async');
+const { errors } = require('arsenal');
 
 const attachReqUids = require('../../replication/utils/attachReqUids');
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
@@ -14,7 +15,8 @@ const CONCURRENCY_DEFAULT = 10;
 
 class LifecycleTask extends BackbeatTask {
     /**
-     * Process a list of versioned or non-versioned objects
+     * Processes Kafka Bucket entries and determines if specific Lifecycle
+     * actions apply to an object, version of an object, or MPU.
      *
      * @constructor
      * @param {LifecycleProducer} lp - lifecycle producer instance
@@ -28,18 +30,27 @@ class LifecycleTask extends BackbeatTask {
     /**
      * Handles non-versioned objects
      * @param {object} bucketData - bucket data
-     * @param {object} params - params for AWS S3 `listObjects`
+     * @param {array} bucketLCRules - array of bucket lifecycle rules
      * @param {Logger.newRequestLogger} log - logger object
      * @param {function} done - callback(error, data)
      * @return {undefined}
      */
-    _getObjectList(bucketData, params, log, done) {
+    _getObjectList(bucketData, bucketLCRules, log, done) {
+        const params = {
+            Bucket: bucketData.target.bucket,
+            MaxKeys: MAX_KEYS,
+        };
+        if (bucketData.details.marker) {
+            params.Marker = bucketData.details.marker;
+        }
+
         const req = this.s3target.listObjects(params);
         attachReqUids(req, log);
         async.waterfall([
             next => req.send((err, data) => {
                 if (err) {
                     log.error('error listing bucket objects', {
+                        method: 'LifecycleTask._getObjectList',
                         error: err,
                         bucket: params.Bucket,
                     });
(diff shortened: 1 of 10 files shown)


extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -102,36 +108,44 @@ class LifecycleTask extends BackbeatTask {
                 return next(null, data);
             }),
             (data, next) => {
+                // sorted newest to oldest
+                // NOTE: parentDate can be undefined
+                let parentDate = prevDate;
```
jonathan-gramain: Rename to `staleDate` or `noncurrentDate`? Seems like what it represents when used to set a field in `sortedVersions` is the date of the nearest newer version, where a parent would be an ancestor (older) version, just a semantic issue but may be clearer using another naming than parent.

philipyoo: Ah good call, I agree, use of parent does sound like ancestor

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -554,59 +568,52 @@ class LifecycleTask extends BackbeatTask {
...
+                        if (version.StorageClass && version.Size &&
+                        version.ETag) {
+                            return this._compareObject(bucketData, version,
+                                applicableRules, log, next);
                         }
                         return next();
-                    });
+                    }
+
+                    const now = Date.now();
```
jonathan-gramain: May be better organized putting this code into a separate method like `applyNoncurrentRules()`

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -554,59 +568,52 @@ class LifecycleTask extends BackbeatTask {
...
-                                    log.debug('sent object entry for ' +
-                                    'consumption', {
-                                        method: 'LifecycleTask.' +
-                                            '_compareVersions',
-                                        entry,
-                                    });
-                                }
-                            });
+                        // `_compareObject` should not apply for delete markers
+                        if (version.StorageClass && version.Size &&
```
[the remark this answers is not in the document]
philipyoo: Oops, this check is no longer needed as we are ignoring delete markers.

But, it is important to note that ignoring delete markers means staleDate is now compared between only versions and delete markers are not considered. I think it's important to consider them because delete markers are versions themselves.

extensions/lifecycle/tasks/LifecycleTask.js
philipyoo: Sounds good to me 👍 
Possible chance of refactoring this logic as well since there are multiple areas where the dates are compared. Will need to check again to make sure

philipyoo:
Use of `listObjectVersions` will list all versions for all keys within the bucket. These are then ordered by the key name, and then their last modified times.

Steps to fix: (in process now)
- If versioned, `listObjects` and for each.. If the Kafka entry specifies a `details.Key`, don't need to do this operation and move to next step.
- `listObjectVersions` specifying a Prefix to the params, and adding an additional conditional, making sure the key matches exactly. If `details.keyMarker` and `details.versionIdMarker`, then use in params.
- If `IsTruncated`, add back to kafka using same process defined in the initial commits of this PR, but now specifying additional `details.Key`

extensions/lifecycle/tasks/LifecycleTask.js
jonathan-gramain: Sounds good to factorize, just make sure there's enough test coverage in and around this function.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -71,67 +79,94 @@ class LifecycleTask extends BackbeatTask {
...
+                            if (version.Key === key) {
+                                const copy = Object.assign({}, version,
+                                    { staleDate });
+                                staleDate = version.LastModified;
+                                store.push(copy);
+                            }
+                            return store;
+                        }, []);
+
+                    if (data.IsTruncated) {
```
jonathan-gramain: I think this applies to the current `listObjectVersions` of this particular object, I doubt you want to spawn a new bucket entry in case this particular listing is truncated (but you probably want to keep going and aggregate the rest of the listing as a single big list).

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -379,7 +414,7 @@ class LifecycleTask extends BackbeatTask {
      * @param {function} done - callback(error, data)
      * @return {undefined}
      */
-    _compareObject(bucketData, obj, rules, log, done) {
+    _compareAnObject(bucketData, obj, rules, log, done) {
```
jonathan-gramain: Since `_compareObject` is already singular it can be understood as applying to a single object, I don't think it adds clarity to rename to `_compareAnObject`

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -531,85 +566,76 @@ class LifecycleTask extends BackbeatTask {
...
+    /**
+     * Compare a version to most applicable rules
+     * @param {object} bucketData - bucket data
+     * @param {object} version - single version from `_getObjectVersions`
+     * @param {object} rules - most applicable rules from `_getApplicableRules`
+     * @param {Logger.newRequestLogger} log - logger object
+     * @param {function} done - callback(error, data)
+     * @return {undefined}
+     */
+    _compareAVersion(bucketData, version, rules, log, done) {
```
jonathan-gramain: same, you can rename instead to `_compareVersion` (though I like the idea of separating a single version processing in its own function).

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -721,30 +728,33 @@ class LifecycleTask extends BackbeatTask {
...
+                                bucketData.details.prevDate, log, next);
                         }
                         // handle non-versioned stuff
-                        if (bucketData.details.marker) {
-                            params.Marker =
-                                bucketData.details.marker;
-                        }
-                        return this._getObjectList(bucketData, params, log,
-                            next);
+                        return this._compareNonversions(bucketData,
```
jonathan-gramain: Where is this function defined?

extensions/lifecycle/tasks/LifecycleTask.js
philipyoo: I was worried this would potentially take a long time to complete as it would list versions for all keys of a bucket. Instead, I was thinking if `IsTruncated`, I get the `LastModified` date of the last version of the listing and pass this as `prevDate` to the bucket entry. 

On the next listing for that given key, even if the previous listings have all been expired already, I still have the `prevDate` to get the stale date for the next version in the listing.

Maybe I'm worrying about something that I don't need to worry about, but would love your input on this and can adjust and change as needed. 👍

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -190,55 +205,131 @@ class LifecycleTask extends BackbeatTask {
...
     /**
-     * For all incomplete MPU uploads, compare with rules (based only on prefix
-     * because no tags exist for incomplete MPU), and apply
-     * AbortIncompleteMultipartUpload rule (if any) on each upload
+     * Helper method to merge and sort Versions and DeleteMarkers lists
+     * @param {array} versions - versions list
+     * @param {array} deleteMarkers - delete markers list
+     * @return {array} merge sorted array
+     */
+    _mergeSortVersionsLists(versions, deleteMarkers) {
```
jonathan-gramain: minor: `_mergeSortedVersionsAndDeleteMarkers` could be more accurate (`sorted` instead of `sort` because we're not technically doing a merge sort, which is a special type of sort, but we're merging sorted lists together).

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -190,55 +205,131 @@ class LifecycleTask extends BackbeatTask {
...
+                }
+            } else if (versions[vIdx].IsLatest) {
+                // 3. if both above match,
+                //    find if one `IsLatest` and take preference
+                sortedList.push(versions[vIdx++]);
+            } else if (deleteMarkers[dmIdx].IsLatest) {
+                sortedList.push(deleteMarkers[dmIdx++]);
+            } else {
+                // 4. None of above? Check VersionId
+                const isVVersionIdGreater = (versions[vIdx].VersionId >
```
jonathan-gramain: As said above, please double-check if version lexicographic comparison order reflects correctly the order in which you want them in the merged array. May be correct as is but worth checking.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -59,78 +69,82 @@ class LifecycleTask extends BackbeatTask {
...
+        this._listVersions(bucketData, paramDetails, log, (err, data) => {
             if (err) {
+                // error already logged at source
                 return done(err);
             }
-            return done(null, {
-                Versions: data.Versions,
-                DeleteMarkers: data.DeleteMarkers,
-            });
+            const list = this._mergeSortVersionsLists(data.Versions,
```
jonathan-gramain: No big deal here, but variable names could be a bit more representative of what they are, like `allVersions`, `allVersionsWithStaleDate` etc. (just examples, feel free to choose appropriate names for the purpose).

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -59,78 +69,82 @@ class LifecycleTask extends BackbeatTask {
...
+                    if (!err) {
+                        log.debug('sent kafka entry for bucket ' +
+                        'consumption', {
+                            method: 'LifecycleTask._getObjectVersions',
+                        });
+                    }
+                });
+            }
+
+            // if not versions to process, skip further processing for this
```
jonathan-gramain: typo `if no versions`

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -190,55 +205,131 @@ class LifecycleTask extends BackbeatTask {
...
+     * @param {string} [bucketDetails.keyMarker] - previous listing key name
+     * @param {string} [bucketDetails.prevDate] - previous listing LastModified
+     * @param {array} list - list of sorted versions and delete markers
+     * @return {array} an updated array of Versions only with applied staleDate
+     */
+    _applyVersionStaleDate(bucketDetails, list) {
+        const appliedList = [];
+
+        for (let i = 0; i < list.length; i++) {
+            // manually check for properties only Versions have and
```
jonathan-gramain: I'm not sure you want to ignore delete markers here. If my understanding is correct, delete markers also are candidates for noncurrent version expiration rules. `ExpiredObjectDeleteMarker` only applies to current delete markers with zero non-current versions for the object, which is not the case when delete markers are noncurrent or co-exist with other versions (or other noncurrent delete markers).

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -492,130 +668,57 @@ class LifecycleTask extends BackbeatTask {
...
-            return done(null, this._getApplicableRules(filterRules));
-        });
-    }
-
-    _compareContents(bucketData, lcRules, contents, log, done) {
-        if (!contents) {
+        const staleDate = version.staleDate;
+        if (!staleDate) {
+            // NOTE: this should never happen. Logging here for debug purposes
+            log.debug('missing staleDate on the version', {
```
jonathan-gramain: Maybe `log.error` if this is not supposed to happen?

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -261,4 +283,198 @@ describe('lifecycle task helper methods', () => {
...
+    describe('_applyVersionStaleDate', () => {
+        const defaultVersions = {
+            ETag: '"6f5902ac237024bdd0c176cb93063dc4"',
+            Size: 0,
+            StorageClass: 'STANDARD',
+        };
+
+        it('should apply a staleDate property on each version on a list ' +
+        'of versions', () => {
+            const list = [
```
jonathan-gramain: Suggesting to add the second object key in the same list (`obj-2`), so that you can check that `staleDate` applies independently to each object.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -190,55 +205,131 @@ class LifecycleTask extends BackbeatTask {
...
+            } else if (versions[vIdx].Key !== deleteMarkers[dmIdx].Key) {
+                // 1. by Key name
+                const isVKeyEarlier = (versions[vIdx].Key <
+                    deleteMarkers[dmIdx].Key);
+                if (isVKeyEarlier) {
+                    sortedList.push(versions[vIdx++]);
+                } else {
+                    sortedList.push(deleteMarkers[dmIdx++]);
+                }
+            } else if (versions[vIdx].LastModified !==
```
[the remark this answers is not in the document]
jonathan-gramain: To precise my thoughts, you still need to consider `isLatest` so that you don't expire current versions with the `NoncurrentVersionExpiration` rule, but it should not be necessary to make it a sorting criteria.

tests/unit/lifecycle/LifecycleTask.spec.js
philipyoo: good suggestion, actually caught a bug :)

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -210,49 +212,36 @@ class LifecycleTask extends BackbeatTask {
...
             if (versions[vIdx] === undefined) {
+                // versions list is empty, just need to merge remaining DM's
                 sortedList.push(deleteMarkers[dmIdx++]);
             } else if (deleteMarkers[dmIdx] === undefined) {
+                // DM's list is empty, just need to merge remaining versions
                 sortedList.push(versions[vIdx++]);
             } else if (versions[vIdx].Key !== deleteMarkers[dmIdx].Key) {
-                // 1. by Key name
+                // 1. by Key name, alphabetical order sorted by ascii values
                 const isVKeyEarlier = (versions[vIdx].Key <
```
jonathan-gramain: Maybe to reduce confusion you could rename this var to `isVKeyNewer`

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -210,49 +212,36 @@ class LifecycleTask extends BackbeatTask {
...
-                // 3. if both above match,
-                //    find if one `IsLatest` and take preference
-                sortedList.push(versions[vIdx++]);
-            } else if (deleteMarkers[dmIdx].IsLatest) {
-                sortedList.push(deleteMarkers[dmIdx++]);
             } else {
-                // 4. None of above? Check VersionId
-                const isVVersionIdGreater = (versions[vIdx].VersionId >
+                // 2. by VersionId, lower means newer
+                const isVersionVIdGreater = (versions[vIdx].VersionId >
```
jonathan-gramain: Similar here, although the var is quite explicit, for consistency it may be renamed to `isVersionVidOlder` (or reversed logic with `isVersionVidNewer` to keep the same logic than above) - which makes the comment at line 241 unnecessary.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -492,130 +655,58 @@ class LifecycleTask extends BackbeatTask {
...
-            return done();
+        const staleDate = version.staleDate;
+        if (!staleDate) {
+            // NOTE: this should never happen. Logging here for debug purposes
+            log.error('missing staleDate on the version', {
+                method: 'LifecycleTask._compareVersion',
+                bucket: bucketData.target.bucket,
+                versionId: version.VersionId,
+            });
+            return done(true);
```
jonathan-gramain: I think it's best to return an internal error object (possibly with a customized description, although it's more of a sanity check than anything else) rather than a boolean as the error.

philipyoo:
One more check, when calling `getRules`, for DeleteMarkers, there are no tags, so need to handle this differently

philipyoo:
It's also important to note that in a versioning enabled or versioning suspended bucket, the `IsLatest` versions of all keys within that bucket are not expired in any case. This is because the check requires the check for rule `ExpiredObjectDeleteMarker`.

I think with the reorganization of the logic here, implementing `ExpiredObjectDeleteMarker` is very do-able.

jonathan-gramain:
We have identified potential issues with handling of `null` versions that will be fixed in subsequent PRs.

PR #218, opened 2018-03-22
bf: lifecycle bucket tagging match

philipyoo:
Bucket tags should be included in the object tags but does
not have to have all the objects tags.

For example:
If a bucket rule has no tags/prefixes, this rule should apply
to all objects in the bucket.

If a bucket rule has rule 'key1/value1', the object should
also have AT LEAST 'key1/value1'. The object tags can have
many tags in total, but must only guarantee any of those
tags matches the bucket tags to qualify for the lifecycle
rule defined on the bucket rule.


philipyoo:
Re-opening with Jira tag

PR #221, opened 2018-03-23
Ft: add lifecycle task expire functional tests

philipyoo:
Initial commit for LifecycleTask tests. Only tests expiration rule and will add tests for each
rule separately.

Solely tests the LifecycleTask class by counting the number of times a bucket or object Kafka
entry is made and tracks each object entry name.

### DIFF
--- a/tests/functional/lifecycle/LifecycleTask.js
+++ b/tests/functional/lifecycle/LifecycleTask.js
@@ -1,13 +1,22 @@
 const assert = require('assert');
-const uuid = require('uuid/v4');
-const util = require('util');
+const async = require('async');
+const AWS = require('aws-sdk');
 
-const { errors } = require('arsenal');
 const Logger = require('werelogs').Logger;
 
 const LifecycleTask = require('../../../extensions/lifecycle' +
     '/tasks/LifecycleTask');
 const Rule = require('../../utils/Rule');
+const testConfig = require('../../config.json');
+
+const S3 = AWS.S3;
+const s3config = {
+    endpoint: `${testConfig.s3.transport}://` +
+        `${testConfig.s3.host}:${testConfig.s3.port}`,
+    s3ForcePathStyle: true,
+    credentials: new AWS.Credentials(testConfig.s3.accessKey,
+                                     testConfig.s3.secretKey),
+};
 
 /*
     NOTE:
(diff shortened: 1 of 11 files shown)


tests/functional/lifecycle/LifecycleTask.js
```diff
@@ -0,0 +1,770 @@
...
+            return cb(null, res);
+        }
+
+        let markerIdx = 0;
+        let maxIdx = params.MaxKeys;
+        if (params.KeyMarker) {
+            markerIdx = bucketVersions.findIndex(v =>
+                v.KeyMarker === params.KeyMarker &&
+                v.VersionId === params.VersionIdMarker);
+            if (markerIdx === -1) {
```
philipyoo: Just to note, these errors here were for myself to validate that I correctly implemented the mock functions. I can remove these if requested 👍

philipyoo:
Sorry for any that are currently reviewing. I am planning on removing versioning and mpu functions for now and will add them back in as I add tests for them.

tests/functional/lifecycle/LifecycleTask.js
```diff
@@ -0,0 +1,770 @@
...
+PAST.setDate(PAST.getDate() - 5);
+// 5 days after currentDate
+const FUTURE = new Date(CURRENT);
+FUTURE.setDate(FUTURE.getDate() + 5);
+
+const OWNER = 'testOwner';
+
+// NOTE: Since all S3 calls are only 'GET's, I decided to just go for mock
+//  S3 that defines all buckets and their conditions
+class S3Mock {
```
jonathan-gramain: As a general comment, mocking is most useful when testing corner cases or unusual cases (like error injection). When it comes to emulating a real S3 behavior, it can be a better test to use the actual S3 process through functional tests, since creating mocks involves the risk of testing components which do not reflect the behavior of the real ones, as well as increasing complexity of writing and maintaining the mock code itself.

philipyoo: I think as I was writing this, I realized how bulky it was getting and I also was thinking it'd be hard to maintain and extend :/

Will look into using S3 👍

philipyoo:
closing this for now, will open again when all functional tests are ready :)

PR #227, opened 2018-03-28
Forward/orbit

RachedBenMustapha:


extensions/replication/replicationStatusProcessor/task.js
```diff
@@ -9,10 +9,24 @@ const kafkaConfig = config.kafka;
...
 werelogs.configure({ level: config.log.logLevel,
-                     dump: config.log.dumpLevel });
+     dump: config.log.dumpLevel });
 
-replicationStatusProcessor.start();
+const logger = new werelogs.Logger('replicationStatusProcessorInit');
+function initAndStart() {
+    initManagement(error => {
+        if (error) {
+            logger.error('could not load managment db', error);
```
jonathan-gramain: typo `management`

lib/management.js
```diff
@@ -0,0 +1,424 @@
...
+
+const Metadata = arsenal.storage.metadata.MetadataWrapper;
+let metadata;
+
+const params = {
+    bucketdBootstrap: ['localhost'],
+    bucketdLog: null,
+    https: null,
+    metadataClient: {
+        host: config.queuePopulator.dmd.host,
```
jonathan-gramain: This may need rework since dmd is not used anymore (e.g. if `config.queuePopulator.dmd` section does not exist it does not start).

lib/management.js
```diff
@@ -0,0 +1,424 @@
...
+            config.setBootstrapList(locationsWithReplicationBackend);
+        } catch (error) {
+            return cb(error);
+        }
+        overlayVersion = conf.version;
+    }
+    return process.nextTick(cb, null, conf);
+}
+
+function initManagement(done) {
```
jonathan-gramain: It could be necessary to have a REMOTE_MANAGEMENT_DISABLED env var similar to S3 one in Federation deployments.

lib/management.js
```diff
@@ -0,0 +1,424 @@
...
+                    metadata, cb);
+            }),
+        ], done);
+    }
+
+    iterate(err => {
+        if (err) {
+            return done(err);
+        }
+        setInterval(iterate, 5000, err => {
```
jonathan-gramain: refresh interval could be set in a constant or via the config file.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -256,13 +255,17 @@ class QueueProcessor extends EventEmitter {
...
                 // i.e. data = { my-site: { ops: 1, bytes: 124 } }
-                const filteredData = Object.keys(data).filter(key =>
-                    key === this.site).reduce((store, k) => {
-                        // eslint-disable-next-line no-param-reassign
-                        store[k] = data[this.site];
-                        return store;
-                    }, {});
-                this._mProducer.publishMetrics(filteredData,
+
+                // NOTE: No need to filter data since we are accepting
```
nicolas2bert: @jonathan-gramain I removed this filter. Let me know if you have any concerns :)

lib/management.js
```diff
@@ -0,0 +1,430 @@
...
+});
+const logger = new werelogs.Logger('mdManagement');
+
+const serviceBucket = 'PENSIEVE';
+const refreshInterval = 5000;
+
+const Metadata = arsenal.storage.metadata.MetadataWrapper;
+let metadata;
+
+const params = {
```
jonathan-gramain: I think you'll need to condition this also to `REMOVE_MANAGEMENT_DISABLE` var, so that `config.queuePopulator.dmd` is not accessed if the var is set.

jonathan-gramain: Or maybe declare it inside `initManagement`?

lib/queuePopulator/LogReader.js
```diff
@@ -355,8 +358,17 @@ class LogReader {
...
             return done();
         }
-        batchState.nextLogOffset =
-            logRes.info.start + logStats.nbLogRecordsRead;
+
+        // Hopefully there is a better way to find out if dmd, file, or MongoDB
+        if (this.logConsumer.mongoUrl) {
+            // We need to store both the timestamp and uniq Oplog id for MongoDB
+            // oplogs, so we concat the string to be '${timestamp}|${uniqID}'
+            batchState.nextLogOffset =
```
jonathan-gramain: in `_readLogOffset` we use `Number.parseInt` to fetch the saved log offset. I'm thinking this can cause issues when reading the log offset in this format.

nicolas2bert: Should we not be using `data.toString()` instead of `Number.parseInt(data, 10)` in `_readLogOffset`?

_NOTE_: we use  `this.logOffset.toString();` in `_writeLogOffset()`?

Also  `Number.parseInt("6544059656791130000|-8572070413137960701", 10)` returns `6544059656791130000` so, for MongoDB, it seems that we are reading the entire logOffset because of `Number.parseInt`. @jonathan-gramain Do you know any reason why we are using `Number.parseInt` and no `.toString();`

jonathan-gramain: As long as we don't have to increment/decrement the value, but only forward the value as-is from the source, it should be good to use `toString`. I don't recall the exact reason why we're using `parseInt` originally but there may be none particular.

Note that it should also be possible to use a serialized JSON object instead of a custom separator, and use `JSON.parse()` in place of `Number.parseInt`, it would be more extensible and self-documented, and we should be backward-compatible since a serialized integer is a valid JSON object.

nicolas2bert:
@jonathan-gramain `JSON.parse('6544059656791130000|-8572070413137960701')` fails returning `Unexpected token | in JSON at position 19` so I will be using `toString()` for now :)

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: This change lgtm. The filter was originally added so we don't duplicate metrics for every QueueProcessor that receives the metrics event. Since we only use a single QueueProcessor for all sites now, I think this change looks good. Thanks!

nicolas2bert:
@ironman-machine
SCALITY_FEDERATION_BRANCH=fx/disable-management
try

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: Then, can you remove the commented code? Thanks!

rahulreddy:
@ironman-machine 
clean try

nicolas2bert:
@ironman-machine
SCALITY_FEDERATION_BRANCH=master
try

nicolas2bert:
E2E passed: http://ci.ironmann.io/gh/scality/Integration/21254

PR #228, opened 2018-03-29
transient source

jonathan-gramain:
transient source support

Mainly two important changes, the rest is cosmetic:

- the queue processor matching the default CRR location replaces the
  `nonTransientLocation` attribute by the actual CRR location array of
  the object. The `nonTransientLocation` attribute is originally set
  by cloud-server for objects on transient locations.

- the replication status processor, at the time it processes the last
  CRR successful completion, replaces the original object location by
  the CRR target one stored in `nonTransientLocation`, and removes the
  latter attribute. This way the object becomes readable natively on
  the default CRR external target, then the transient data can be
  removed (not implemented yet).

### DIFF
--- a/extensions/replication/utils/ObjectQueueEntry.js
+++ b/extensions/replication/utils/ObjectQueueEntry.js
@@ -9,22 +9,6 @@ function _extractVersionedBaseKey(key) {
     return '';
 }
 
-function _getGlobalReplicationStatus(data) {
-    // Check the global status relative to the other backends
-    if (Array.isArray(data.replicationInfo.backends)) {
-        const statuses = data.replicationInfo.backends.map(backend =>
-            backend.status);
-        // If any site replication failed, set the global status to FAILED.
-        if (statuses.includes('FAILED')) {
-            return 'FAILED';
-        }
-        if (statuses.includes('PENDING')) {
-            return 'PROCESSING';
-        }
-    }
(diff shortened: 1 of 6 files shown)


extensions/replication/utils/ObjectQueueEntry.js
```diff
@@ -114,16 +98,36 @@ class ObjectQueueEntry extends ObjectMD {
...
-        newEntry.setReplicationStatus('REPLICA');
+        newEntry
+            .setBucket(this.getReplicationTargetBucket())
+            .setReplicationSiteStatus(site, 'REPLICA')
+            .setReplicationStatus('REPLICA')
+            .setNonTransientLocation(undefined);
         return newEntry;
     }
 
+    _getGlobalReplicationStatus() {
```
bennettbuchanan: This LGTM, but why move the method here? I think it's helpful to have the `to...Entry` methods one after the next in this class 😸.

jonathan-gramain: I like when methods are close to where they are used, but no worry I can move it somewhere up in the class. The main point being making the function a class member :)

jonathan-gramain: done

jonathan-gramain:
POC only

PR #229, opened 2018-03-29
FX: forward/orbit branch

nicolas2bert:


yarn.lock
```diff
@@ -9,10 +9,6 @@ JSONStream@^1.0.0:
     jsonparse "^1.2.0"
     through ">=2.2.7 <3"
 
-abbrev@1:
```
jonathan-gramain: Shouldn't we remove this file from the source tree?

extensions/replication/queueProcessor/task.js
```diff
@@ -35,18 +35,14 @@ metricsProducer.setupProducer(err => {
             log.info('management init done');
 
             const replicationEndpoints = config.getReplicationEndpoints();
-            const bootstrapList = replicationEndpoints.bootstrapList
-                .filter(item => item.site === site);
```
bennettbuchanan: I see there is no site assigned to the queue processor. Will this work for one-to-many replication? Just wondering about the intention here.

extensions/replication/queueProcessor/task.js
```diff
@@ -35,18 +35,14 @@ metricsProducer.setupProducer(err => {
...
-            assert(bootstrapList.length === 1, 'Invalid site argument. Site must match ' +
-                'one of the replication endpoints defined');
 
             const destConfig = Object.assign({}, repConfig.destination);
-            destConfig.bootstrapList = bootstrapList;
+            destConfig.bootstrapList = replicationEndpoints;
```
jonathan-gramain: I may be wrong, but this looks like a drastic change. Are you sure it does not break existing CRR (zenko and S3 connector), i.e. the required changes have already been applied? Otherwise with this change, it could be worth renaming the variable as well to `replicationEndpoints`.

extensions/replication/queueProcessor/task.js
nicolas2bert: In Orbit we didn't want to spawn a queue_processor process for each single site. With this change we should be able to use a unique queue_processor for any sites.

bennettbuchanan: Okay, as long as there is no need to do one-to-many replication that should be fine. Otherwise it will only replicate one of the destinations, I believe.

yarn.lock
RachedBenMustapha: We should actually keep `.lock` files for reproducibility as it ensures the exact revisions of all dependencies installed (provided we install with `--frozen-lockfile`), no surprises.

jonathan-gramain: ok, I'm a yarn noob :rofl:

PR #230, opened 2018-03-29
fix: improve lifecycle prefix unit test

philipyoo:
Most aws prefix examples show use of "/". Prefix tests, when filtering,
should filter based on the object key vs prefix defined in a bucket rule.

### DIFF
--- a/tests/unit/lifecycle/LifecycleTask.spec.js
+++ b/tests/unit/lifecycle/LifecycleTask.spec.js
@@ -61,17 +61,17 @@ describe('lifecycle task helper methods', () => {
 
         it('should filter out unmatched prefixes', () => {
             const mBucketRules = [
-                new Rule().addID('task-1').addPrefix('atask').build(),
-                new Rule().addID('task-2').addPrefix('atasker').build(),
+                new Rule().addID('task-1').addPrefix('atask/').build(),
+                new Rule().addID('task-2').addPrefix('atasker/').build(),
                 new Rule().addID('task-3').addPrefix('cat-').build(),
-                new Rule().addID('task-4').addPrefix('xatask').build(),
-                new Rule().addID('task-5').addPrefix('atasr').build(),
-                new Rule().addID('task-6').addPrefix('Atask').build(),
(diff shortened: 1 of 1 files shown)


tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -82,9 +82,16 @@ describe('lifecycle task helper methods', () => {
...
             assert.strictEqual(res1.length, 3);
-            const expRes1 = getRuleIDs(mBucketRules.filter(rule =>
-                (rule.Filter.Prefix && rule.Filter.Prefix.startsWith('atask'))
-                || !rule.Filter.Prefix));
+            const expRes1 = getRuleIDs(mBucketRules.filter(rule => {
+                if (rule.Filter.Prefix &&
+                item1.Key.startsWith(rule.Filter.Prefix)) {
+                    return true;
+                }
+                if (!rule.Filter.Prefix) {
```
jonathan-gramain: Don't you want this rule check above the previous one? if checking for empty string it should work though, but if undefined it can cause issues.

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -82,9 +82,16 @@ describe('lifecycle task helper methods', () => {
...
             const res1 = lct._filterRules(mBucketRules, item1, objTags);
             assert.strictEqual(res1.length, 3);
-            const expRes1 = getRuleIDs(mBucketRules.filter(rule =>
-                (rule.Filter.Prefix && rule.Filter.Prefix.startsWith('atask'))
-                || !rule.Filter.Prefix));
+            const expRes1 = getRuleIDs(mBucketRules.filter(rule => {
+                if (!rule.Filter.Prefix) {
+                    return true;
+                }
+                if (rule.Filter.Prefix &&
```
jonathan-gramain: Now you can remove this first check, since test at https://github.com/scality/backbeat/pull/230/files#diff-70184adf6b658fa3260d32a93bb664dbR86 guarantees it already.

PR #234, opened 2018-04-02
Fix/sandbox deployment

RachedBenMustapha:
Adds features to ease deployment in orbit sandboxes:

* Add `dockerize` to wait for dependency services and hard-fail if they are unavailable after a timeout
* Make redis host/port configurable
* Allow configuring healthcheck clients


RachedBenMustapha:
Merging to feature branch

PR #236, opened 2018-04-02
ft: add metrics unit tests for route checks

philipyoo:
Targeting `rel/7.4`, to be forward ported to beta and updated for the changes there (if any), then to be forward ported to master.

### DIFF
--- a/tests/unit/api/BackbeatAPI.spec.js
+++ b/tests/unit/api/BackbeatAPI.spec.js
@@ -1,0 +1,48 @@
+const assert = require('assert');
+
+const BackbeatAPI = require('../../../lib/api/BackbeatAPI');
+const BackbeatRequest = require('../../../lib/api/BackbeatRequest');
+const config = require('../../config.json');
+
+const fakeLogger = {
+    trace: () => {},
+    error: () => {},
+    info: () => {},
+    debug: () => {},
+};
+
+describe('BackbeatAPI', () => {
+    let bbapi;
+
+    before(() => {
+        bbapi = new BackbeatAPI(config, fakeLogger, { timer: true });
+    });
+
+    it('should validate routes', () => {
+        // valid routes
+        [
+            '/_/metrics/crr/all',
+            '/_/healthcheck',
+            '/_/metrics/crr/all/backlog',
+            '/_/metrics/crr/all/completions',
+            '/_/metrics/crr/all/throughput',
+        ].forEach(path => {
+            const req = new BackbeatRequest({ url: path });
+
+            assert.ok(bbapi.isValidRoute(req));
+        });
+
+        // invalid routes
+        [
+            '/_/invalid/crr/all',
+            '/_/metrics/ext/all',
+            '/_/metrics/crr/test',
+            '/_/metrics/crr/all/backlo',
+            '/_/metrics/crr/all/completionss',
+        ].forEach(path => {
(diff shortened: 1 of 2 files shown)


tests/unit/api/BackbeatAPI.spec.js
```diff
@@ -0,0 +1,52 @@
...
+describe('BackbeatAPI', () => {
+    let bbapi;
+
+    before(() => {
+        bbapi = new BackbeatAPI(config, fakeLogger, { timer: true });
+    });
+
+    it('should validate routes', () => {
+        // valid routes
+        const valid1 = new BackbeatRequest({ url: '/_/metrics/crr/all' });
```
jonathan-gramain: For a more compact and less redundant code, instead of having multiple variables, you could have one array for the valid requests, and one for the invalid ones, and loop over the array items to validate with `assert.ok` or `assert.equal`

tests/unit/api/BackbeatRequest.spec.js
```diff
@@ -0,0 +1,41 @@
+const assert = require('assert');
+
+const BackbeatRequest = require('../../../lib/api/BackbeatRequest');
+
+function setRouteDetails(bbRequest, route) {
+    bbRequest.setRoute(route);
+    bbRequest._parseRoute();
```
jonathan-gramain: Instead of calling this private method here, shouldn't it be called automatically from `getRouteDetails()` (only if needed, i.e if not done before)?

tests/unit/api/BackbeatRequest.spec.js
```diff
@@ -0,0 +1,41 @@
...
+        const results = setRouteDetails(bbRequest, '/_/healthcheck');
+
+        assert.deepStrictEqual(results, {});
+    });
+
+    it('should parse routes and store internally as route details', () => {
+        const results = setRouteDetails(bbRequest, '/_/metrics/crr/all');
+        const routeDetails = bbRequest.getRouteDetails();
+
+        assert.strictEqual(results.category, routeDetails.category);
```
jonathan-gramain: I don't get the intent of these checks, since `results` comes from a call to `bbRequest.getRouteDetails()` already. It checks that results from `getRouteDetails` are consistent but that does not sound like very useful to check, did you intend some other test instead? If you expect some specific info in those fields you might check them directly against what you expect.

tests/unit/api/BackbeatRequest.spec.js
```diff
@@ -0,0 +1,41 @@
...
+        assert.strictEqual(results.extension, routeDetails.extension);
+        assert.strictEqual(results.site, routeDetails.site);
+        assert.strictEqual(results.metric, routeDetails.metric);
+        assert.strictEqual(routeDetails.metric, undefined);
+
+        const results2 = setRouteDetails(bbRequest,
+            '/_/metrics/crr/test/backlog');
+        const routeDetails2 = bbRequest.getRouteDetails();
+        assert.strictEqual(results2.site, routeDetails2.site);
```
jonathan-gramain: same here

tests/unit/api/BackbeatRequest.spec.js
philipyoo: Ah, especially for unit tests, this would look confusing. Rewriting this to use a new instance of BackbeatRequest for each test

PR #240, opened 2018-04-09
ft: add LifecycleTask functional tests expiration

philipyoo:
Only includes tests for non-versioned bucket expiration rule.

PR #241, opened 2018-04-09
ft: add LifecycleTask functional tests versioned

philipyoo:
Only includes tests for versioned bucket expiration rule
(NoncurrentVersionExpiration)

To be merged after (or into) https://github.com/scality/backbeat/pull/240

### DIFF
--- a/tests/functional/lifecycle/LifecycleTask.js
+++ b/tests/functional/lifecycle/LifecycleTask.js
@@ -53,13 +53,15 @@ class S3Helper {
         this.bucket = undefined;
 
         this._scenario = [
-            // 0. create objects: non-versioned, no pagination
+            // 0. all unique object names, no pagination
+            //    Useful for: non-versioned
             {
                 keyNames: ['object-1', 'object-2', 'object-3'],
                 tags: ['key1=value1', 'key1=value1&key2=value2', 'key2=value2'],
 
             },
-            // 1. create objects: non-versioned, pagination, prefix
+            // 1. all unique object names, pagination, prefix
+            //    Useful for: non-versioned
             {
                 keyNames: [
                     'test/obj-1', 'obj-2', 'test/obj-3',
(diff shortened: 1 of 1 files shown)


tests/functional/lifecycle/LifecycleTask.js
```diff
@@ -174,6 +248,13 @@ class S3Helper {
                 if (!this._scenario[scenarioNumber]) {
                     return next(new Error('incorrect scenario number'));
                 }
+                // TODO: refactor
+                if (scenarioNumber === 2 || scenarioNumber === 3) {
```
jonathan-gramain: Instead of numbering scenarios, you might just as well create separate functions to prepare each scenario, called from ad-hoc places in test code. Or if you want to benefit from some factorization and genericity, you could pass a params object instead of scenario number, including the scenario params that apply to the particular test.

philipyoo: I've decided to keep use of `scenarioNumber` because I think it'd be nice to reuse names and tags for any scenario to build with, and since the only differences were whether to create a non-versioned bucket, or versioned bucket, or versioned bucket w/ delete markers, I now pass the function name I want to the buildScenario method. 

Depending on how tests turn out for other rules, I'm sure there will be opportunities to refactor these tests further. If future tests don't really utilize re-use of scenarios, I'll update and pass in the list of names directly like you mentioned

tests/functional/lifecycle/LifecycleTask.js
```diff
@@ -345,7 +422,7 @@ describe('lifecycle task functional tests', () => {
                 next => s3Helper.setAndCreateBucket('test-bucket', next),
                 next => s3Helper.buildScenario([
                     new Rule().addExpiration('Date', FUTURE).build(),
-                ], 0, next),
+                ], 'createObjects', 0, next),
```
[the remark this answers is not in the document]
philipyoo: Refactored 👍 One thing though is the scenario number. I can make it more explicit by creating separate functions, but I like the idea of someone being able to just create objects and their tags by adding it to an array and reusing the names defined in the array for any given test

PR #245, opened 2018-04-10
Fwdport: 7.4 to 7.4 beta

philipyoo:


jonathan-gramain:
As discussed, we'll wait for 7.4GA and forward-port then (unless a critical 7.4 fix is required in the meantime).

PR #247, opened 2018-04-13
Ft: lifecycletask tests abortmpu

philipyoo:
Separating and contains a temporary bug fix for `listMultipartUploads`

Small bug where when specifying KeyMarker & UploadIdMarker, the listing will include that previously listed upload.

i.e. 
// list once with max keys = 2
[upload1, upload2] && truncated = true
// list again using NextKeyMarker and NextUploadIdMarker from previously listing
[upload2, upload3]
^ Re-lists upload2

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -202,6 +202,17 @@ class LifecycleTask extends BackbeatTask {
                 return done(err);
             }
 
+            // TODO:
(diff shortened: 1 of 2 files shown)


jonathan-gramain:
As discussed, let's keep the test as if they were using a bug-free API, and let's open a ticket to fix the S3 bug. We can keep this PR around until the bug is fixed, otherwise the workaround will become an annoying bug whenever it gets fixed on S3.

Good catch btw!

PR #248, opened 2018-04-13
Zenko-21: FT: New Route for Prometheus Client

anurag4DSB:
Added new route for Prometheus client: /_/monitoring/metrics

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -11,6 +11,7 @@ const BackbeatProducer = require('../BackbeatProducer');
 const Healthcheck = require('./Healthcheck');
 const routes = require('./routes');
 const { redisKeys } = require('../../extensions/replication/constants');
+const client = require('prom-client');
 
 // StatsClient constant defaults
 // TODO: This should be moved to constants file
@@ -77,6 +78,9 @@ class BackbeatAPI {
         const route = bbRequest.getRoute();
         // first validate healthcheck routes
         if (route.substring(3) === 'healthcheck') {
+            return true;
+        }
+        if (route.substring(3) === 'monitoring/metrics') {
             return true;
         }
 
@@ -367,6 +371,11 @@ class BackbeatAPI {
                 return cb(null, store);
             });
         });
+    }
+
+    monitoringHandler(details, cb) {
+        const promMetrics = client.register.metrics();
+        return cb(null, promMetrics);
     }
 
     /**
--- a/lib/api/routes.js
+++ b/lib/api/routes.js
@@ -77,7 +77,7 @@ module.exports = [
         httpMethod: 'GET',
         category: 'monitoring',
         type: 'monitoring',
-        extensions: { crr: [...allSites, 'all'] },
+        extensions: {},
         method: 'monitoringHandler',
     }
 ];
--- a/lib/api/routes.js
+++ b/lib/api/routes.js
@@ -73,4 +73,11 @@ module.exports = [
         extensions: { crr: ['failed'] },
         method: 'retryFailedCRR',
     },
+    {
+        httpMethod: 'GET',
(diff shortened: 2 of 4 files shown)


lib/api/BackbeatAPI.js
```diff
@@ -11,6 +11,7 @@ const BackbeatProducer = require('../BackbeatProducer');
 const Healthcheck = require('./Healthcheck');
 const routes = require('./routes');
 const { redisKeys } = require('../../extensions/replication/constants');
+const client = require('prom-client');
```
jonathan-gramain: I would suggest renaming the variable to `promClient` because the context is global to backbeat API.

If prometheus server is not available (and nobody is accessing the `/monitoring` route), can you confirm this does not cause trouble?

A corollary is, how does it behave when `/monitoring` is accessed but the prometheus server is not available? I would expect a kind of HTTP error to be returned.

lib/api/routes.js
```diff
@@ -73,4 +73,11 @@ module.exports = [
         extensions: { crr: ['failed'] },
         method: 'retryFailedCRR',
     },
+    {
+        httpMethod: 'GET',
+        category: 'monitoring',
+        type: 'monitoring',
```
jonathan-gramain: do you need the `type` field? If not I would rather leave it unset for now until we have a good use for it.

lib/api/BackbeatAPI.js
```diff
@@ -79,6 +80,9 @@ class BackbeatAPI {
         if (route.substring(3) === 'healthcheck') {
             return true;
         }
+        if (route.substring(3) === 'monitoring/metrics') {
```
jonathan-gramain: On one hand, I understand that we want to separate prometheus metrics from raw CRR metrics, since prometheus has its own logic of handling it, on the other hand, I can find confusing for the user to see the `/_/metrics` and `/_/monitoring/metrics` routes. Is the end goal to expose everything through prometheus and get rid of other `/_/metrics/xxx` routes? Or are they meant to co-exist?

Maybe a `/_/metrics/monitoring` route could make sense then? Not sure, just raising the question, it's completely open to discussion.

It may not be very important but when the product is getting complex with more and more routes it can get messy quickly if we don't plan for some good organization, and it becomes harder to change once applications start using the routes.

lib/api/routes.js
philipyoo: ya, maybe we can leave it empty as we aren't using it. I think the focus was to keep it as uniform with other routes as possible. Just to note also for healthcheck, we do not use the `type` field, so I could also set that to empty as well in a refactor later

lib/api/BackbeatAPI.js
philipyoo: While discussing with Anurag, I thought it is definitely possible to have prometheus track all metrics since it does the same operations of incrementing the 4 data points we collect. I'm not sure of performance, but a great thing about the promclient was throughput can be represented more accurately rather than the current representation through an average of a period of time (15 minutes by default).

Didn't look too far into promclient but supporting site level can be potentially hard to implement. Not sure if we can specify a site name to each increment.

For now, the promclient will have to sit separate, so maybe to address your concerns on routes, we can change the route name for the prometheus metrics.

lib/api/BackbeatAPI.js
anurag4DSB: I will change the name of the variable to promClient.
If prometheus server is not available (and nobody is accessing the /monitoring route), can you confirm this does not cause trouble? -> This will not cause any problem.
A corollary is, how does it behave when /monitoring is accessed but the prometheus server is not available? -> We just get an error with code: <Error><Code>InvalidURI

anurag4DSB:
Cannot merge this. As per instructions, all pull requests will be redirected to master.

PR #250, opened 2018-04-15
Zenko-21: FT: New Route for Prometheus Client

anurag4DSB:
Added new route for Prometheus client: /_/monitoring/metrics

Signed-off-by: anurag4DSB <anurag.213@gmail.com>


lib/api/BackbeatAPI.js
```diff
@@ -370,6 +374,17 @@ class BackbeatAPI {
         });
     }
 
+    /**
+     * Collects metrics from Prometheus variables and register for response
+     * @param {*} details - route details from lib/api/routes.js
```
philipyoo: Instead of using the asterisk in the braces, can you switch to using `Object` and `Function` respectively for these params to keep uniform within the code?

PR #252, opened 2018-04-16
FT: Use ranged gets for GCP MPUs > 1024 parts

bennettbuchanan:
If the source object is being replicated to GCP and it is composed of more than 1024 parts, use ranged requests to translate the MPU from N > 1024 parts to N <= parts.

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -9,6 +9,7 @@ const ReplicateObject = require('./ReplicateObject');
 const attachReqUids = require('../utils/attachReqUids');
 
 const MPU_CONC_LIMIT = 10;
+const MPU_GCP_MAX_PARTS = 1024;
 
 class MultipleBackendTask extends ReplicateObject {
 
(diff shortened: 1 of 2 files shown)


extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -108,16 +109,129 @@ class MultipleBackendTask extends ReplicateObject {
...
+            log,
+        }, cb);
+    }
+
+    /**
+     * Calculate the byte size of the given range.
+     * @param {String} range - The range request string (e.g., 'bytes=0-1').
+     * @return {Number} The number of bytes that the range represents.
+     */
+    _getRangeSize(range) {
```
jonathan-gramain: Instead of manipulating HTTP range strings, it could be simpler to pass around range JS objects (eg. with `start` and `end` attributes) to avoid any unnecessary parsing, only when preparing the HTTP GET request will we create the range header. Not talking about performance here but rather extra code complexity.

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -229,6 +369,70 @@ class MultipleBackendTask extends ReplicateObject {
...
+     * range count does not exceed 1024 elements.
+     * @param {Number} contentLength - The content length of the whole object
+     * @return {Array} The array of byte ranges.
+     */
+    _getRanges(contentLength) {
+        const ranges = [];
+        let start = 0;
+        let end = 0;
+        for (let i = 0; i < MPU_GCP_MAX_PARTS; i++) {
+            end = start + Math.ceil(contentLength / MPU_GCP_MAX_PARTS) - 1;
```
jonathan-gramain: No big deal, but `Math.ceil(contentLength / MPU_GCP_MAX_PARTS)` may be precomputed to save some CPU cycles.

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -229,6 +369,70 @@ class MultipleBackendTask extends ReplicateObject {
...
+     * Get byte ranges for an object of the given content length, such that the
+     * range count does not exceed 1024 elements.
+     * @param {Number} contentLength - The content length of the whole object
+     * @return {Array} The array of byte ranges.
+     */
+    _getRanges(contentLength) {
+        const ranges = [];
+        let start = 0;
+        let end = 0;
+        for (let i = 0; i < MPU_GCP_MAX_PARTS; i++) {
```
jonathan-gramain: Since you rely on the condition `end >= contentLength - 1` in the loop to break from it (which is fine for me), you may just do an infinite loop `while` condition instead, simplifying the reasoning.

Then with the same logic, for code clarity you could also rewrite the loop as a `while (end < contentLength - 1)` and adding the last part if not empty after the loop.

jonathan-gramain (dismissed):
Some minor comments, otherwise looks good!

extensions/replication/tasks/MultipleBackendTask.js
jonathan-gramain: Now that I think about it, it may be more natural to round the part size to the nearest power of two (because computers are generally more efficient dealing with those, may make a small difference for GET/PUT while possibly reducing fragmentation of parts when copying the objects around).

jonathan-gramain (dismissed):

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -229,6 +357,71 @@ class MultipleBackendTask extends ReplicateObject {
...
 
+    /**
+     * Get byte ranges for an object of the given content length, such that the
+     * range count does not exceed 1024 elements.
+     * @param {Number} contentLen - The content length of the whole object
+     * @return {Array} The array of byte ranges.
+     */
+    _getRanges(contentLen) {
+        const pow = Math.pow(2, Math.ceil(Math.log(contentLen) / Math.log(2)));
+        const range = Math.ceil(pow / MPU_GCP_MAX_PARTS);
```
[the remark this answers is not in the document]
bennettbuchanan: This creates ranges that have sizes of powers of 2.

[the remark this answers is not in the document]
jonathan-gramain: Right, since `MPU_GCP_MAX_PARTS` is also a power of two. That's ok then.

jonathan-gramain (dismissed):

jonathan-gramain (dismissed):

jonathan-gramain (approved):

alexanderchan-scality (approved):

PR #253, opened 2018-04-17
bf: handle null case versions

philipyoo:
Addresses null versions from versioning suspended buckets

As a last resort, compare LastModified times. If these
match, default to always pushing versions first over
delete markers. This is a bug that will need to be
fixed.

Bug tracking: S3C-1418
As known and requiring a fix

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -215,7 +215,9 @@ class LifecycleTask extends BackbeatTask {
      */
     _mergeSortedVersionsAndDeleteMarkers(versions, deleteMarkers) {
         const sortedList = [];
+        // Version index counter
         let vIdx = 0;
+        // Delete Marker index counter
         let dmIdx = 0;
 
         while (vIdx < versions.length || dmIdx < deleteMarkers.length) {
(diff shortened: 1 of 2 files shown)


tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -299,29 +299,29 @@ describe('lifecycle task helper methods', () => {
                 {
                     Key: 'obj-1',
                     VersionId:
-                    '834373636323233323831333639393939393952473030312020363035',
+                    '834373636323233323831333639393939393952473030312020363030',
```
philipyoo: Changed `VersionId` so the version is clearly based by the last few numbers in the id rather than a number in the middle of the id as well as at the end. Easier to manage and see

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -236,12 +236,26 @@ class LifecycleTask extends BackbeatTask {
...
+                    && deleteMarkers[dmIdx].VersionId !== 'null');
                 const isVersionVidNewer = (versions[vIdx].VersionId <
                     deleteMarkers[dmIdx].VersionId);
-                if (isVersionVidNewer) {
+                if (doesVersionIdExist && isVersionVidNewer) {
                     sortedList.push(versions[vIdx++]);
-                } else {
+                } else if (doesVersionIdExist) {
+                    sortedList.push(deleteMarkers[dmIdx++]);
+                } else if (versions[vIdx].LastModified >
```
jonathan-gramain: I'm not 100% sure we can use lexicographic comparison for dates, as it obviously depends on the format. I would feel safer to convert them to `Date` objects first, or add a comment stating it's on purpose and safe to do so.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -236,12 +236,26 @@ class LifecycleTask extends BackbeatTask {
                 }
             } else {
                 // 2. by VersionId, lower number means newer
+                //    NOTE: VersionId may be null
+                const doesVersionIdExist = (versions[vIdx].VersionId !== 'null'
```
jonathan-gramain: Maybe I'm being nitpicky, but the variable name does not sound so helpful, since a version ID does exist somewhere even though there's a null version on the other side (and we can argue that null is also a version ID), why not calling it e.g. `noNullVersion`?

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -353,6 +353,47 @@ describe('lifecycle task helper methods', () => {
...
+                    LastModified: '2018-04-04T23:16:46.000Z',
+                },
+            ];
+            const dms = [
+                {
+                    Key: 'obj-1',
+                    VersionId:
+                    '834373636323233323831343639393939393952473030312020363035',
+                    IsLatest: false,
+                    LastModified: '2018-04-04T23:16:46.000Z',
```
jonathan-gramain: You may differentiate the last modified date from the null version's one to improve coverage (otherwise you're testing the "random" case which is mildly useful).

Also, you may add one more noncurrent delete marker that has an older timestamp than null version, the other with a more recent timestamp, and expect the null version to be stored in-between the two delete markers.

jonathan-gramain:
Is it still on hold? I think after a rebase and an extra approval we can get this fix merged.

philipyoo:
@jonathan-gramain oops forgot to remove, ty 👍

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -236,12 +236,26 @@ class LifecycleTask extends BackbeatTask {
                 }
             } else {
                 // 2. by VersionId, lower number means newer
+                //    NOTE: VersionId may be null
+                const noNullVersion = (versions[vIdx].VersionId !== 'null'
```
rahulreddy: Double negation alert!
It's easier to read
```js
const nullVersion = (versions[vIdx].VersionId === 'null' || deleteMarkers[dmIdx].VersionId === 'null');

if (nullVersion) {
 // logic here
} else if (isVersionVidNewer) {
// logic here, automatically handles !nullVersion and isVersionVidNewer
} else {
// logic here for !nullVersion
}
```

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -236,12 +236,26 @@ class LifecycleTask extends BackbeatTask {
                 }
             } else {
                 // 2. by VersionId, lower number means newer
+                //    NOTE: VersionId may be null
+                const nullVersion = (versions[vIdx].VersionId === 'null'
+                    || deleteMarkers[dmIdx].VersionId === 'null');
                 const isVersionVidNewer = (versions[vIdx].VersionId <
                     deleteMarkers[dmIdx].VersionId);
-                if (isVersionVidNewer) {
+                if (!nullVersion && isVersionVidNewer) {
```
jonathan-gramain: You can actually group the null version logic in a `if (nullVersion) {}` block, as Rahul suggested, even though there will be an extra nesting level of `if` it makes it easier to read and understand overall.

PR #259, opened 2018-04-19
bf: ZENKO-252 Use source bucket for multiple backend replication

bennettbuchanan:
Sets the source bucket (instead of the destination bucket) as the bucket name when replicating to a public cloud. This allows S3's multiple backend to prefix the key properly when a location constraint has `bucketMatch` set to `false`. For example, the bucket name will be passed as context to S3's multiple backend here: https://github.com/scality/S3/blob/master/lib/routes/routeBackbeat.js#L352.

### DIFF
--- a/extensions/replication/utils/ObjectQueueEntry.js
+++ b/extensions/replication/utils/ObjectQueueEntry.js
@@ -120,6 +120,19 @@ class ObjectQueueEntry extends ObjectMD {
         return newEntry;
     }
 
+    /**
+     * Set the destination entry to have a REPLICA status and the bucket as the
+     * source bucket (used for context in CloudServers's multiple backend).
+     * @param {String} site - The replication site given in the configuration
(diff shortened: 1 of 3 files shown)


extensions/replication/utils/ObjectQueueEntry.js
```diff
@@ -120,6 +120,20 @@ class ObjectQueueEntry extends ObjectMD {
         return newEntry;
     }
 
+    /**
+     * Set the destination entry to have a REPLICA status and the bucket as the
+     * source bucket (used for contextual puts in CloudServers's multple
```
philipyoo: small type: multiple

bennettbuchanan:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=1
try

bennettbuchanan:
@ironman-machine try

rahulreddy:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=1
try

PR #262, opened 2018-04-23
ft: LifecycleTask tests for AbortMPU

philipyoo:
A change was added to MPU params to guarantee that if the bucket entry contains `keyMarker`, an `uploadIdMarker` must be there, and vice versa.

Change added on L#922

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -202,6 +202,7 @@ class LifecycleTask extends BackbeatTask {
             if (err) {
                 return done(err);
             }
+
             this._compareMPUUploads(bucketData, bucketLCRules,
                 data.Uploads, log);
             return done();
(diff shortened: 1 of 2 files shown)


extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -909,7 +919,8 @@ class LifecycleTask extends BackbeatTask {
                     MaxUploads: MAX_KEYS,
                     KeyMarker: bucketData.details.uploadIdMarker &&
                         bucketData.details.keyMarker,
-                    UploadIdMarker: bucketData.details.uploadIdMarker,
+                    UploadIdMarker: bucketData.details.keyMarker &&
```
jonathan-gramain: Is this a bugfix? Maybe it should be reflected in the PR title/description then

philipyoo: I added just in-case since both the key marker and upload id marker are required. I'll edit description to take note of the change and reason for why

PR #264, opened 2018-04-23
bf: ZENKO-271 Set message.max.bytes in producer

bennettbuchanan:
Allow the node-rdkafka producer to push a message when performing a 10,000 part MPU.


lib/BackbeatProducer.js
```diff
@@ -38,20 +47,23 @@ class BackbeatProducer extends EventEmitter {
             topic: joi.string().required(),
             keyedPartitioner: joi.boolean().default(true),
             pollIntervalMs: joi.number().default(2000),
+            fetchMaxBytes: joi.number(),
```
jonathan-gramain: Change the parameter name too?

PR #267, opened 2018-04-23
bf: ZENKO-147 Check site when setting failed hash

bennettbuchanan:
Now that we have a single replication status processor so that operations occur sequentially, the `replicationInfo` backends may contain a 'FAILED' status from a previous operation. This change ensures that the failure hash field is set only once.

I will aim to get https://github.com/scality/Integration/pull/660 merged to avoid this kind of unexpected problem going forward.


bennettbuchanan:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=true
try

PR #268, opened 2018-04-24
bf: S3C-1427 use expired interval to average out throughput

philipyoo:
When an interval of data in Redis expires, throughput will
abruptly reduce. Given the data we collect, we can only
calculate the average throughput. To ease the erratic
decrease on expiration of an interval, instead, get the
average of elapsed time for the newest interval and
remaining time of the interval multiplied against the
average of the just-expired interval.

In Redis, we need to save an extra interval to reference
the just-expired data.


philipyoo:
Also, this should not be added to master. Instead, changes should be added to Arsenal master branch with support for multiple sites

philipyoo:
Yup, I think prometheus has a lot to offer in terms of functionality as well, so should be a good refactor for crr metrics, and possibly metrics for other extensions in backbeat 👍

jonathan-gramain:
@ironman-machine 
RUN_BACKBEAT_CRR_TESTS=1
r+

philipyoo:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=1
try

philipyoo:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=1
try

jonathan-gramain:
Integration succeeded in http://ci.ironmann.io/gh/scality/Integration/21601

PR #272, opened 2018-04-27
rf: ZENKO-147 Refactor retry CRR API

bennettbuchanan:
* Add pagination
* Model closer to getObjectVersions API

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -80,6 +80,20 @@ class BackbeatAPI {
     }
 
     /**
+     * Validate possible query strings.
+     * @param {BackbeatRequest} bbRequest - Relevant data about request
+     * @return {Object|null} - The error object or `null` if no error
+     */
+    validateQuery(bbRequest) {
+        const { marker } = bbRequest.getRouteDetails();
+        if (marker !== undefined && (marker === '' || isNaN(marker))) {
+            return errors.InvalidQueryParameter
+                .customizeDescription('marker must be a number');
+        }
+        return null;
+    }
+
+    /**
      * Check if incoming request is valid
      * @param {BackbeatRequest} bbRequest - holds relevant data about request
      * @return {boolean} true/false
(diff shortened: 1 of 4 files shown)


bennettbuchanan:
@ironman-machine 
RUN_BACKBEAT_CRR_TESTS=true
try

philipyoo:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=z/1.0
SCALITY_FEDERATION_BRANCH=bf/test-gittag
try

tests/functional/api/BackbeatServer.js
```diff
@@ -460,11 +485,31 @@ describe('Backbeat Server', () => {
...
+                    const scanCount = 1000;
+                    let marker = 0;
+                    const sites = [];
+                    async.timesSeries(keyCount / scanCount, (i, next) =>
+                        getRequest(`/_/crr/failed?marker=${marker}`,
+                        (err, res) => {
+                            assert.ifError(err);
+                            res.Versions.forEach(version => {
+                                // Ensure we have no duplicate results.
+                                assert(!sites.includes(version.StorageClass));
```
jonathan-gramain: Not a strong concern, but the check has linear complexity, so the whole process is O(n^2), with 10000 elements it may slow down the test quite a bit. An easy fix is to use a `Set` object and use `add()`, `has()` and `size` on it.

bennettbuchanan: Thanks for that tip, I was actually not aware of `Set`. I've updated as such.

philipyoo:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=z/1.0
SCALITY_FEDERATION_BRANCH=bf/mongodb-version-tagging
try

tests/functional/api/BackbeatServer.js
```diff
@@ -341,27 +357,27 @@ describe('Backbeat Server', () => {
             ['a'],
             [{
                 // Missing bucket property.
```
jonathan-gramain: Comments to be updated to reflect new property names (capitalized, StorageClass)

lib/api/BackbeatAPI.js
```diff
@@ -86,7 +86,7 @@ class BackbeatAPI {
      */
     validateQuery(bbRequest) {
         const { marker } = bbRequest.getRouteDetails();
-        if (marker !== undefined && (marker === '' || isNaN(marker))) {
+        if (marker !== undefined && Number.isNaN(Number.parseInt(marker, 10))) {
```
jonathan-gramain: Playing around with `isNaN` and `Number.parseInt`, looks like validation was stronger in the previous implementation, e.g. `parseInt('3abc')` returns 3, while `isNaN('3abc')` returns true.

bennettbuchanan: Nice catch, I did not now about this case. I'll stick with the global `isNaN`.

PR #274, opened 2018-04-30
rf: ZENKO-285 Single test for range calculation

bennettbuchanan:


jonathan-gramain (dismissed):
:+1:

rahulreddy (commented):
Isn't the earlier setup the right way of writing tests? With this approach, you are testing multiple cases in one test. This would go against our `a test should test only one thing` approach.

jonathan-gramain:
In this particular case I think it's fine because it's essentially testing one thing (with 10000 different inputs though), but having 10000 separate tests is overkill and clutters the output making it hard to get the result from other tests, so I suggested this change to Bennett.

jonathan-gramain:
I've also observed the complete unit tests run to take more than 30 seconds to complete and show a status after the last test has been executed, I'm strongly suspecting this change will significantly reduce this time.

rahulreddy (dismissed):

jonathan-gramain (approved):

PR #278, opened 2018-05-02
ft: S3C-1398 Use retry CRR topic

bennettbuchanan:
Decouple logic around making calls to Redis and updating object metadata. Allows for usage of `ioredis`'s offline queue during cases where the connection to Redis is faulty. To that end, this PR alters the retry feature to do the following:
* Push failed entries to a kafka "retry" topic.
* Set Redis keys during processing of "retry" kafka entry.

Depends on https://github.com/scality/Federation/pull/1514 for creating the topic.

### DIFF
--- a/extensions/replication/failedCRR/FailedCRRConsumer.js
+++ b/extensions/replication/failedCRR/FailedCRRConsumer.js
@@ -1,0 +1,130 @@
+'use strict'; // eslint-disable-line strict
+
+const Logger = require('werelogs').Logger;
+const redisClient = require('../../replication/utils/getRedisClient')();
+
+const FailedCRRProducer = require('./FailedCRRProducer');
+const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
+const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
+const redisKeys = require('../constants').redisKeys;
+const config = require('../../../conf/Config');
+
+// BackbeatConsumer constant defaults
+const CONSUMER_FETCH_MAX_BYTES = 5000020;
+const CONCURRENCY = 10;
+
+class FailedCRRConsumer {
+    /**
+     * Create the retry consumer.
+     */
+    constructor() {
+        this._kafkaConfig = config.kafka;
+        this._topic = config.extensions.replication.replicationFailedTopic;
+        this.logger = new Logger('Backbeat:FailedCRRConsumer');
+        this._failedCRRProducer = new FailedCRRProducer(this.kafkaConfig);
+        this._backbeatTask = new BackbeatTask();
+    }
+
+    /**
+     * Start the retry consumer by subscribing to the retry kafka topic. Setup
+     * the failed CRR producer for pushing any failed redis operations back to
+     * the queue.
+     * @param {Function} cb - The callback function
+     * @return {undefined}
+     */
+    start(cb) {
+        const consumer = new BackbeatConsumer({
+            kafka: { hosts: this._kafkaConfig.hosts },
+            topic: this._topic,
+            groupId: 'backbeat-retry-group',
+            concurrency: CONCURRENCY,
+            queueProcessor: this.processKafkaEntry.bind(this),
+            fetchMaxBytes: CONSUMER_FETCH_MAX_BYTES,
+        });
+        consumer.on('error', () => {});
+        consumer.on('ready', () => {
+            consumer.subscribe();
+            this.logger.info('retry consumer is ready to consume entries');
+        });
+        return this._failedCRRProducer.setupProducer(err => {
+            if (err) {
+                this.logger.error('could not setup producer', {
+                    method: 'FailedCRRConsumer.processKafkaEntry',
+                    error: err,
+                });
+                return cb(err);
+            }
+            return cb();
+        });
+    }
+
+    /**
+     * Process an entry from the retry topic, and set the data in a Redis hash.
+     * @param {Object} kafkaEntry - The entry from the retry topic
+     * @param {function} cb - The callback function
+     * @return {undefined}
+     */
+    processKafkaEntry(kafkaEntry, cb) {
+        const log = this.logger.newRequestLogger();
+        let data;
+        try {
+            data = JSON.parse(kafkaEntry.value);
+        } catch (err) {
+            log.error('error processing retry entry', {
+                method: 'FailedCRRConsumer.processKafkaEntry',
+                error: err,
+            });
+            log.end();
+            return cb();
+        }
+        return this._setRedisHash(data, kafkaEntry, log, cb);
+    }
+
+    /**
+     * Attempt to set the Redis hash, using an exponential backoff should the
+     * key set fail. If the backoff time is exceeded, push the entry back into
+     * the retry entry topic for a later reattempt.
+     * @param {Object} data - The field and value for the Redis hash
+     * @param {Object} kafkaEntry - The entry from the retry topic
+     * @param {Werelogs} log - The werelogs logger
+     * @param {Function} cb - The callback function
+     * @return {undefined}
+     */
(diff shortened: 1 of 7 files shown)


jonathan-gramain:
@bennettbuchanan thanks for reporting this, looking at it, it could be a real bug, since a kafka consumer could throw when trying to commit its offset if it's not connected to kafka, we should guard this in a try/catch block IMO.

On the tests side, not sure why the chain of callbacks ends up in the consumer side directly from the producer side, but it may be possible to fix by serializing the creation of the BackbeatProducer after the BackbeatConsumer class has emitted the 'ready' event (so that it will be connected at this stage).

lib/queuePopulator/QueuePopulator.js
```diff
@@ -109,6 +114,19 @@ class QueuePopulator {
         this._mProducer.setupProducer(cb);
     }
 
+    /**
+     * Set up the consumer and producer for retrying failed CRR operations.
+     * @param {Function} cb - The callback function
+     * @return {undefined}
+     */
+    _setupRetryClients(cb) {
+        this._retryConsumer = new RetryConsumer(this.kafkaConfig);
```
jonathan-gramain: Let's try not to stretch the role of queue populator beyond its metadata ingestion scope.

The producer is fine in the replication status processor IMO because that's where the info is readily available to be published.

The consumer should be part of its own component updating redis for separation of concerns.

bennettbuchanan: I'm not sure I follow what separating the start of the consumer it into its own component would look like. Should it be a backbeat "extension" with its own task? Maybe a good place is https://github.com/scality/backbeat/blob/master/extensions/replication/queueProcessor/task.js#L49. In any case, I noticed the metrics consumer was being started when opening the queue populator so added it here.

jonathan-gramain: The idea is to follow a philosophy where each backbeat internal component should be runnable as an independent entity. All entities that belong to the same service (e.g. replication) will go under a single kub service. The general idea as I understand it is to make it more manageable, pluggable on-demand and scalable independently from each other.

In this case, I think it makes sense to create a new component for consuming the retry queue and updating redis. In that case, indeed there should be an entrypoint that starts the process (task.js or service.js), and a set of classes doing the job (e.g. in tasks/ for the processing triggered by consumption of a single kafka entry).

I think it was an overlook (or a shortcut) to put the metrics consumer in the queue populator, it should only contain the producer, and the consumer should be a separate process.

lib/FailedCRRProducer.js
```diff
@@ -0,0 +1,61 @@
...
+        });
+    }
+
+    /**
+     * Publish the given message to the retry Kafka topic.
+     * @param {String} message - The message to publish
+     * @param {Function} cb - The callback function
+     * @return {undefined}
+     */
+    publishRetryEntry(message, cb) {
```
jonathan-gramain: To be renamed as well to `publishFailedCRREntry`

lib/FailedCRRProducer.js
```diff
@@ -0,0 +1,61 @@
+'use strict'; // eslint-disable-line strict
```
jonathan-gramain: Since the producer and consumer classes are specific to replication service, I think they'd be better located somewhere in extensions/replication/ (maybe creating an extra failedCRR/ dir there can make sense).

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -126,38 +125,32 @@ class ReplicationStatusProcessor {
...
-        });
+        const backend = backends.find(b =>
+            b.status === 'FAILED' && b.site === queueEntry.getSite());
+        if (backend) {
+            const bucket = queueEntry.getBucket();
+            const key = queueEntry.getObjectKey();
+            const versionId = queueEntry.getEncodedVersionId();
+            const { site } = backend;
+            const message = {
+                field: `${bucket}:${key}:${versionId}:${site}`,
```
bennettbuchanan: Since moving away from Redis hash, `field` should be updated to `key`.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -126,38 +125,32 @@ class ReplicationStatusProcessor {
...
+                value: Buffer.from(kafkaEntry.value).toString(),
```
bennettbuchanan: Since moving away from storing object metadata in Redis, we can eliminate this.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -126,38 +125,32 @@ class ReplicationStatusProcessor {
     }
 
     /**
-     * Set the Redis hash key for each failed backend.
+     * Push any failed entry to the "retry" topic.
```
jonathan-gramain: should be `to the "failed" topic`

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
bennettbuchanan: Will update in subsequent PRs.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
bennettbuchanan: Will update in subsequent PRs.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
bennettbuchanan: Will update in a subsequent PR so that we can maintain functionality when merging this one.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
bennettbuchanan: Will update in a subsequent PR so that we can maintain functionality when merging this one.

extensions/replication/utils/getRedisClient.js
```diff
@@ -9,10 +9,7 @@ werelogs.configure({
 });
 
 function getRedisClient() {
-    const redisConfig = Object.assign({}, config.redis, {
-        enableOfflineQueue: false,
```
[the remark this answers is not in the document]
bennettbuchanan: This seems like a good option as we can reuse the `retry` method from the `BackbeatTask` class.

extensions/replication/failedCRR/FailedCRRConsumer.js
```diff
@@ -60,13 +64,62 @@ class FailedCRRConsumer {
...
+    _setRedisHash(data, kafkaEntry, log, cb) {
+        this._backbeatTask.retry({
+            actionDesc: 'set redis key',
+            logFields: {},
+            actionFunc: done => this._setRedisHashOnce(data, log, done),
+            shouldRetryFunc: err => err.retryable,
+            log,
+        }, err => {
+            if (err && err.retryable === true) {
+                return this._failedCRRProducer.setupProducer(err => {
```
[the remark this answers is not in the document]
bennettbuchanan: I had reservations as well, but didn't want to use a callback in the constructor. I'll move it to a separate method and call it when initializing the class.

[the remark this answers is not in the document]
jonathan-gramain: You can have a `setup` or `init` async method called separately from the constructor.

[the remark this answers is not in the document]
bennettbuchanan: Updated to set up when starting the consumer.

PR #283, opened 2018-05-08
poc: queue processor per site

philipyoo:
This is a POC and opening to test

Differences from previous implementation of a QueueProcessor per site is that instead of taking a site as an argument and calling the npm script per site, instead, based on the config file, instantiate a new QueueProcessor instance per site after waiting for the `initManagement`

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -1,7 +1,6 @@
 'use strict'; // eslint-disable-line
 
 const http = require('http');
-const async = require('async');
 const { EventEmitter } = require('events');
 
 const Logger = require('werelogs').Logger;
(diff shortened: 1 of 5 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -311,25 +319,16 @@ class QueueProcessor extends EventEmitter {
...
-                    task = new ReplicateObject(this, site);
-                }
-                this.logger.debug('source entry is being pushed',
-                  { entry: sourceEntry.getLogInfo() });
-                return this.taskScheduler.push({ task, entry: sourceEntry },
-                                               sourceEntry.getCanonicalKey(),
-                                               cb);
-            }, err => done(err));
+        } else if (sourceEntry instanceof ObjectQueueEntry &&
+        sourceEntry.getReplicationStorageClass().includes(this.site)) {
```
jonathan-gramain: You will need to keep the conversion to an array first, to avoid side-effects of substring matching (e.g. `cloud` location would also match `cloud1` with string matching)

philipyoo:
Re-opening with jira tag

PR #284, opened 2018-05-08
ft: ZENKO-147 delete hash field on retry

bennettbuchanan:
:shipit:

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -310,6 +310,38 @@ class BackbeatAPI {
     }
 
     /**
+     * Delete the failed CRR Redis hash field.
+     * @param {String} field - The field in the hash to delete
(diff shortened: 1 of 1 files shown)


lib/api/BackbeatAPI.js
```diff
@@ -309,6 +309,38 @@ class BackbeatAPI {
...
 
+    /**
+     * Delete the failed CRR Redis hash field.
+     * @param {String} field - The field in the hash to delete
+     * @param {Function} cb - The callback function
+     * @return {undefined}
+     */
+    _deleteFailedCRRField(field, cb) {
+        const cmds = ['hdel', redisKeys.failedCRR, field];
+        return this._redisClient.batch([cmds], (err, res) => {
```
philipyoo: just making sure, using the square brackets around `cmds` will make this a 2d array

PR #286, opened 2018-05-10
ft: ZENKO-147 Use Redis keys instead of hash

bennettbuchanan:
We want to be able to expire entries from failed CRR after a configurable amount of time (default is 24 hours). Unfortunately Redis does not support expiry of hash data structure fields, so this PR moves the design to store failed CRR as keys. 

Alternatives considered:
* Setting the hash fields with a timestamp and then retroactively removing hashes at certain time intervals.
* Setting hash keys with a timestamp, one for each expiry time interval, then retroactively deleting old hashes.

I considered this the simplest solution because it leverages Redis' built-in expiry and we obviate the use of a cron job to delete old keys.

This PR also fixes a problem where when >= 1000 keys are stored, the listing of specific version failures does not iterate beyond the first count. The solution is to use a recursive scan listing.

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -6,6 +6,8 @@ const { hostPortJoi, bootstrapListJoi, adminCredsJoi } =
 const transportJoi = joi.alternatives().try('http', 'https')
     .default('http');
 
+const CRR_FAILURE_EXPIRY = 24 * 60 * 60; // Expire Redis keys after 24 hours.
+
 const joiSchema = {
     source: {
         transport: transportJoi,
@@ -52,6 +54,8 @@ const joiSchema = {
     replicationStatusTopic: joi.string().required(),
     monitorReplicationFailures: joi.boolean().default(true),
     replicationFailedTopic: joi.string().required(),
+    monitorReplicationFailureExpiryTimeS:
+        joi.number().default(CRR_FAILURE_EXPIRY),
     queueProcessor: {
         groupId: joi.string().required(),
         retryTimeoutS: joi.number().default(300),
--- a/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
+++ b/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
@@ -14,6 +14,7 @@ const ReplicationTaskScheduler = require('../utils/ReplicationTaskScheduler');
 const UpdateReplicationStatus = require('../tasks/UpdateReplicationStatus');
 const QueueEntry = require('../../../lib/models/QueueEntry');
 const ObjectQueueEntry = require('../utils/ObjectQueueEntry');
+const { redisKeys } = require('../constants');
 
 /**
  * @class ReplicationStatusProcessor
@@ -145,7 +146,8 @@ class ReplicationStatusProcessor {
             const versionId = queueEntry.getEncodedVersionId();
             const { site } = backend;
             const message = {
-                field: `${bucket}:${key}:${versionId}:${site}`,
+                key: `${redisKeys.failedCRR}:` +
+                    `${bucket}:${key}:${versionId}:${site}`,
                 value: Buffer.from(kafkaEntry.value).toString(),
             };
             return this._FailedCRRProducer
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -12,6 +12,7 @@ const QueueEntry = require('../../lib/models/QueueEntry');
 const Healthcheck = require('./Healthcheck');
 const routes = require('./routes');
 const { redisKeys } = require('../../extensions/replication/constants');
+const getFailedCRRKey = require('../util/getFailedCRRKey');
 const monitoringClient = require('../clients/monitoringHandler').client;
 
 // StatsClient constant defaults
@@ -216,10 +217,11 @@ class BackbeatAPI {
     /**
      * Builds the failed CRR response.
      * @param {String} cursor - The Redis HSCAN cursor
-     * @param {Array} hashes - The collection of Redis hashes for the iteration
-     * @return {Object} - The response object
+     * @param {Array} keys - The collection of Redis keys for the iteration
+     * @param {Function} cb - The callback function
+     * @return {undefined}
      */
-    _getFailedCRRResponse(cursor, hashes) {
+    _getFailedCRRResponse(cursor, keys, cb) {
         const response = {
             IsTruncated: Number.parseInt(cursor, 10) !== 0,
             Versions: [],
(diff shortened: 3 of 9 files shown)


extensions/replication/ReplicationConfigValidator.js
```diff
@@ -50,6 +52,7 @@ const joiSchema = {
     },
     topic: joi.string().required(),
     replicationStatusTopic: joi.string().required(),
+    monitorReplicationFailureExpiry: joi.number().default(CRR_FAILURE_EXPIRY),
```
jonathan-gramain: I suggest adding a `S` at the end of the param name to tell we expect a number of seconds (which also implicitly tells it's a time interval which is not obvious otherwise, or may be renamed to `replicationFailureEntryExpiryTimeS` - not a fan of `monitor` in the name, I find it a bit confusing).

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -138,20 +138,25 @@ class ReplicationStatusProcessor {
...
-                const field = `${bucket}:${key}:${versionId}:${site}`;
-                const value = JSON.parse(kafkaEntry.value);
-                fields.push(field, JSON.stringify(value));
+                const key = `${redisKeys.failedCRR}:` +
+                    `${bucket}:${objectKey}:${versionId}:${site}`;
+                const value = JSON.stringify(JSON.parse(kafkaEntry.value));
+                const expiry = this.repConfig.monitorReplicationFailureExpiry;
+                cmds.push(['set', key, value, 'EX', expiry]);
             }
             return undefined;
```
jonathan-gramain: That line looks unnecessary

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -138,20 +138,25 @@ class ReplicationStatusProcessor {
...
+        const cmds = [];
         backends.forEach(backend => {
             const { status, site } = backend;
             if (status === 'FAILED' && site === queueEntry.getSite()) {
-                const field = `${bucket}:${key}:${versionId}:${site}`;
-                const value = JSON.parse(kafkaEntry.value);
-                fields.push(field, JSON.stringify(value));
+                const key = `${redisKeys.failedCRR}:` +
+                    `${bucket}:${objectKey}:${versionId}:${site}`;
+                const value = JSON.stringify(JSON.parse(kafkaEntry.value));
```
jonathan-gramain: Is it needed to copy the `kafkaEntry.value` field? It's no big issue, just noticed it could be useless, in doubt we can leave it.

lib/api/BackbeatAPI.js
```diff
@@ -214,57 +214,85 @@ class BackbeatAPI {
...
                 return cb(cmdErr);
             }
-            const [cursor, hashes] = collection;
-            return cb(null, this._getFailedCRRResponse(cursor, hashes));
+            const [cursor, keys] = collection;
+            allKeys.push(...keys);
+            if (allKeys.length >= 1000 || Number.parseInt(cursor, 10) === 0) {
+                return cb(null, cursor, allKeys);
+            }
+            return this._scanAllKeys(pattern, cursor, allKeys, cb);
```
jonathan-gramain, line 281: I'm not familiar with redis API, but if asked for 1000 keys, may it return you less than this even if there are 1000+ keys to return? In such case it looks correct to call `_scanAllKeys` again, otherwise it seems we could just return the results in the callback in all cases.

lib/api/BackbeatAPI.js
```diff
@@ -378,11 +395,12 @@ class BackbeatAPI {
                     LastModified: queueEntry.getLastModified(),
                     ReplicationStatus: 'PENDING',
                 });
-                const field = `${Bucket}:${Key}:${VersionId}:${StorageClass}`;
-                return this._deleteFailedCRRField(field, err => {
+                const key = `${redisKeys.failedCRR}:` +
```
jonathan-gramain: It would be ideal if the construction of the key happened in a shared helper function

tests/functional/api/BackbeatServer.js
```diff
@@ -30,19 +30,23 @@ function getUrl(options, path) {
...
-    const cmds = fields.map(field => {
+function setKey(redisClient, keys, cb) {
+    const cmds = keys.map(key => {
         // eslint-disable-next-line no-unused-vars
-        const [bucket, key, versionId, site] = field.split(':');
-        const value = getKafkaEntry(bucket, key, site);
-        return ['hset', REDIS_KEY_FAILED_CRR, field, value];
+        const [bucket, objectKey, versionId, site] = key.split(':');
+        const value = getKafkaEntry(bucket, objectKey, site);
+        return ['set', `${REDIS_KEY_FAILED_CRR}:${key}`, value];
```
jonathan-gramain: Is `REDIS_KEY_FAILED_CRR` the same than the `redisKeys.failedCRR` constant used elsewhere? In which case better use the same one consistently and remove the other.

tests/functional/api/BackbeatServer.js
```diff
@@ -458,11 +462,12 @@ describe('Backbeat Server', () => {
                         // We cannot guarantee order because it depends on how
                         // Redis fetches the keys.
                         keys.forEach(k => {
+                            // eslint-disable-next-line no-unused-vars
                             const [bucket, key, versionId, site] = k.split(':');
                             assert(res.Versions.some(o => (
                                 o.Bucket === bucket &&
                                 o.Key === key &&
-                                o.VersionId === versionId &&
+                                o.VersionId === testVersionId &&
```
jonathan-gramain, line 470: Why is it needed (or better) to change this test?

extensions/replication/ReplicationConfigValidator.js
bennettbuchanan: Okay, I was following the config field `monitorReplicationFailure` which is forthcoming. I might suggest updating to `monitorReplicationFailureExpiryTimeS` so it's clear these fields reference the same functionality, but I'm not especially wedded to any naming scheme.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
bennettbuchanan: The subsequent PR will reduce this to just storing the source role to get the failed object's metadata.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
bennettbuchanan: I believe this is to make the linter happy. 😄

lib/api/BackbeatAPI.js
bennettbuchanan: There isn't a guarantee on the number of keys being returned by Redis during the SCAN operation. (I've added a point about this in the [Operational Considerations](https://github.com/scality/backbeat/blob/065434d2b9f18c3a49d0937e49ca816947b17694/docs/replication-retry.md#operational-considerations) section of the doc for this feature.) We do have a guarantee that the scan has completed when the returned cursor is `0`, so in such a case we can return whatever results are there even if less than 1000.

tests/functional/api/BackbeatServer.js
bennettbuchanan: Because the response gets the version ID using the object MD instead of the key name (see change [here](https://github.com/scality/backbeat/pull/286/files/d4fa76fad604b5b98fffbe6760dd4a20da0eab34#diff-7012bb0401146916f9385c1c205fe3dcR243). I think it's better because it actually checks that the returned value is the correct version ID. I will go ahead an update the key names to use `testVersionId ` just to be consistent.

extensions/replication/ReplicationConfigValidator.js
jonathan-gramain: I may have written this comment originally before we decided to use `monitorReplication...` naming for the conf vars, and left it as pending for some time, so I'm good keeping consistency across config.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
jonathan-gramain: The inner `forEach` function does not seem to return anything elsewhere, so I'm suspecting this is a left-over from a previous version that required it, but I may be wrong.

lib/api/BackbeatAPI.js
```diff
@@ -214,57 +214,85 @@ class BackbeatAPI {
...
             }
             const [cmdErr, collection] = res[0];
             if (cmdErr) {
                 return cb(cmdErr);
             }
-            const [cursor, hashes] = collection;
-            return cb(null, this._getFailedCRRResponse(cursor, hashes));
+            const [cursor, keys] = collection;
+            allKeys.push(...keys);
+            if (allKeys.length >= 1000 || Number.parseInt(cursor, 10) === 0) {
```
philipyoo, line 278: `allKeys.length >= 1000` 
If we get 1000 keys initially, won't this return before getting next set of keys? Or was this intentional?

philipyoo (dismissed):

lib/api/BackbeatAPI.js
bennettbuchanan: Yes, in effect it allows for a paginated response. I wanted to limit the API response listing to ~1000 keys to model closer to the default for version listings in S3. We cannot guarantee the exact number because Redis does not make a guarantee for the number of keys returned during the scan. So if it's 1000 or more, the API will include a `NextMarker` value for subsequent listings.

extensions/replication/ReplicationConfigValidator.js
bennettbuchanan: `monitorReplicationFailureExpiryTimeS` it is!

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
bennettbuchanan: This was removed during the previous refactor 🍾, so either way!

tests/functional/api/BackbeatServer.js
bennettbuchanan: This is used because the test suite is not run with the `TEST_SWITCH` environment var. I updated the constant variable name to `TEST_REDIS_KEY_FAILED_CRR` to make that more apparent.

bennettbuchanan: Maybe we want to update that in the CI? I'm fine either way.

philipyoo (approved):

jonathan-gramain (approved):

PR #288, opened 2018-05-10
bf: ZENKO-343 adapt lifecycle qp extension for mongo

jonathan-gramain:
Mongo log have a different structure than raft logs, for this reason
we adapt the queue populator lifecycle extension filter to correctly
process bucket updates from mongo.


rahulreddy:
@philipyoo It's only for this Zenko branch. We need to ensure that this change is not propagated to S3C's release branches. Ultimately we need a uniform way to access the logs, but we can address that later.

PR #290, opened 2018-05-13
bf: ZENKO-351 allow healthchecks even if clients not ok

philipyoo:
All routes had preliminary checks made. One of these checks
were to check if conditions were ok. This check defeats the
purpose of a healthcheck, so the check needs to be eased for
healthcheck routes only

cherry-picked from rel/7.4: https://github.com/scality/backbeat/pull/289


philipyoo:
The forward port from 7.4 has also resolved this 👍

PR #291, opened 2018-05-14
ft: add multiple site setup script

philipyoo:
StorageClass uses a comma-separated list of target site names.
Role Arn is just the source role Arn or if there is a destination zenko site, role arn is a comma separated source role Arn and target Zenko role Arn

### DIFF
--- a/bin/replication.js
+++ b/bin/replication.js
@@ -22,7 +22,12 @@ function _createSetupReplication(command, options, log) {
     const sourceProfile = options.sourceProfile;
     const targetProfile = options.targetProfile;
     const siteName = options.siteName;
-    const targetIsExternal = siteName !== undefined;
+
+    let targetIsExternal = siteName !== undefined;
+    if (command === 'multi-site-setup' && targetProfile) {
+        // need to create target zenko site bucket and role arn
+        targetIsExternal = false;
+    }
 
     // Required options
     if (!sourceBucket || !targetBucket ||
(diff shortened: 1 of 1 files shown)


bin/replication.js
```diff
@@ -105,6 +105,49 @@ program.version(version);
         });
 });
 
+program
+    .command('multi-setup')
+    .description('NOTE: Only works for external sites')
+    .option('--source-bucket <name>', '[required] source bucket name')
+    .option('--source-profile <name>',
+            '[required] source aws/credentials profile')
+    .option('--target-bucket <name>', '[required] target bucket names')
```
bennettbuchanan: Maybe make this optional in case the target is a Scality destination.

bin/replication.js
```diff
@@ -105,6 +105,49 @@ program.version(version);
         });
 });
 
+program
+    .command('multi-setup')
+    .description('NOTE: Only works for external sites')
```
bennettbuchanan: Why does it only work for external sites?

bennettbuchanan (dismissed):
Thanks for this. Left a couple things to consider but otherwise LGTM! :shipit:

bin/replication.js
```diff
@@ -105,6 +105,49 @@ program.version(version);
         });
 });
 
+program
+    .command('multi-setup')
```
bennettbuchanan: Perhaps 'multi-site-setup' would be more descriptive.

bin/replication.js
philipyoo: I wasn't sure how to include another Scality site in a one-to-many configuration.
If I were to include a Scality site, would I go through the same process of a one-to-one setup to another Scality site, setting the arn to `sourceArn,targetArn` but also setting StorageClass to the target external site names (excluding the site name for the target Scality site)?

bennettbuchanan: Exactly! You would include the Scality replication endpoint name as one of the comma-separated storage classes. Then the role would be a comma-separated pair of role ARNs (i.e., `sourceArn,targetArn`) and the destination bucket ARN would be the bucket in that Scality instance to replicate to.

bin/replication.js
```diff
@@ -105,6 +110,73 @@ program.version(version);
...
+        const targetProfile = options.targetProfile;
+        const multiSites = options.multiSites;
+
+        const destinationEndpoint = config.getBootstrapList()
+            .find(dest => Array.isArray(dest.servers));
+
+        const hasIncorrectMultiSites = !multiSites || multiSites.split(',')
+            .some(site => site.length === 0);
+        const hasValidOptionals = (targetBucket || targetProfile) ?
+            (!(!targetBucket || !targetProfile) && (!hasIncorrectMultiSites &&
```
philipyoo: if `targetBucket` or `targetProfile` is undefined, using an `&&` here will return an undefined value

philipyoo:
Sorry, will reopen with a jira tag

PR #293, opened 2018-05-16
rf: ZENKO-353 Avoid storing all objectMD in Redis

bennettbuchanan:
Thanks for your review! 🚀

Make a request to cloud server to get the failed source object's metadata instead of storing it in Redis.

Previously, a Redis key storing all object metadata consumed ~2625 bytes of memory. With this change the key will consume ~115 bytes by only storing the IAM role ARN (which is used to authenticate the GET metadata request to cloud server). 

If a failure occurs each second of the 24 hours (the default key expiry time) this change saves ~216MB per day. Extrapolated to many days and multiple failures per second (in a 1-M scenario) the memory usage saved could be on the order of GBs.

For this purpose, I’ve extracted some common code from `UpdateReplicationStatus` for creating a backbeat client and getting metadata from the source.


lib/api/BackbeatAPI.js
```diff
@@ -211,60 +217,129 @@ class BackbeatAPI {
...
+                if (role === null) {
+                    return next();
+                }
+                const entry = new ObjectFailureEntry(key, role);
+                return this._getObjectQueueEntry(entry, (err, queueEntry) => {
+                    if (err) {
+                        // If we cannot retrieve the source object, then we no
+                        // longer can consider it a failed operation. Delete the
+                        // key used to monitor the failure.
+                        return this._deleteFailedCRRKey(key, next);
```
bennettbuchanan, line 334: Do we still want to include the object in the GET response even if we can’t retry it due to a missing source object? My sense is that if the source object is deleted, we should just remove the key from Redis as it's no longer representing an existing object with a FAILED status. It depends on how you view the API's monitoring functionality though—i.e. is it tracking _all_ previous failures or _retryable_ failures. I'm open to either view, but my sense  is the latter 😄.

jonathan-gramain: If the version does not exist anymore, I think there's no point in tracking it anymore, so indeed we should remove it from redis on getting a 404 from cloudserver IMHO.

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -1,14 +1,9 @@
...
 const ObjectQueueEntry = require('../../replication/utils/ObjectQueueEntry');
-const BackbeatClient = require('../../../lib/clients/BackbeatClient');
-const { attachReqUids } = require('../../../lib/clients/utils');
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
-const { getAccountCredentials } =
-          require('../../../lib/credentials/AccountCredentials');
-const RoleCredentials =
-          require('../../../lib/credentials/RoleCredentials');
+const SetupBackbeatClient =
+    require('../../replication/utils/SetupBackbeatClient');
```
jonathan-gramain: Can be shortened to `../utils/SetupBackbeatClient`

extensions/replication/utils/SetupBackbeatClient.js
```diff
@@ -0,0 +1,175 @@
...
+            const { host, port } = this.sourceConfig.auth.vault;
+            this.vaultclientCache
+                .setHost('source:s3', host)
+                .setPort('source:s3', port);
+        }
+    }
+
+    _createCredentials(log) {
+        const authConfig = this.sourceConfig.auth;
+        if (authConfig.type === 'account') {
```
jonathan-gramain: The original code now uses `getAccountCredentials()` helper, it should still be used.

extensions/replication/utils/SetupBackbeatClient.js
```diff
@@ -0,0 +1,175 @@
...
+    _putMetadataOnce(entry, log, cb) {
+        log.debug('putting metadata',
+                  { where: 'source', entry: entry.getLogInfo(),
+                    replicationStatus: entry.getReplicationStatus() });
+
+        // sends extra header x-scal-replication-content to the target
+        // if it's a metadata operation only
+        const mdBlob = entry.getSerialized();
+        const req = this.backbeatSource.putMetadata({
+            Bucket: entry.getBucket(),
```
jonathan-gramain: I'm all for factoring the code that sets up the backbeat client and handles GET/PUT of metadata with retries etc., but I would rather not use polymorphism on the `entry` object to support this, as it makes things unnecessarily error-prone (one has to understand and maintain the interface on the `entry` object required to execute the actions, say, for `getBucket()` etc.).

A more direct approach to this is to have the helper method ask for the specific params it needs to do the job (`bucket`, `objectKey` etc.) either through a list of parameters or through a `params` object. Then each user gets and provides its own parameters as he wishes (e.g. by calling `entry.getBucket()` etc) before calling the helper method, which decouples the caller logic from the callee logic.

extensions/replication/utils/SetupBackbeatClient.js
```diff
@@ -0,0 +1,175 @@
...
+const BackbeatClient = require('../../../lib/clients/BackbeatClient');
+const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
+const {
+    StaticFileAccountCredentials,
+    ProvisionedServiceAccountCredentials,
+} = require('../../../lib/credentials/AccountCredentials');
+const { attachReqUids } = require('../../../lib/clients/utils');
+const RoleCredentials = require('../../../lib/credentials/RoleCredentials');
+
+class SetupBackbeatClient extends BackbeatTask {
```
jonathan-gramain: Since this is a helper class that wraps the low-level backbeat client with its own functions, it should have a name reflecting this, could be called `BackbeatMetadataProxy` instead (for example, suggestions welcome!)

extensions/replication/utils/SetupBackbeatClient.js
bennettbuchanan: That makes sense. I've updated the `getMetadata` helpers since the `putMetadata` isn't affected in the same way by polymorphism.

bennettbuchanan:
Closing in favor of a branch named properly for eve integration: https://github.com/scality/backbeat/pull/344

PR #295, opened 2018-05-17
ft: ZENKO-300 Add environment config for lifecycle

tmacro:
This is needed for the lifecycle helm chart


ssalaues:
Rebased off z/1.0

PR #296, opened 2018-05-21
bf: crash when unable to connect with kafka

giacomoguiulfo:
When Kafka is unavailable, the backbeat queue populator will hang and not connect to it even if it becomes available later on. These changes will make the backbeat queue populator throw an exception if it's unable to connect to Kafka after a certain amount of time.

In an orchestrated environment, the orchestration tool can recognize this crash and restart the failed container, thus allowing it to try to connect again to Kafka (which might be available later on).

### DIFF
--- a/lib/BackbeatProducer.js
+++ b/lib/BackbeatProducer.js
@@ -68,7 +68,14 @@ class BackbeatProducer extends EventEmitter {
             'request.timeout.ms': ACK_TIMEOUT,
         });
         this._ready = false;
-        this._producer.connect();
+        this._producer.connect({ timeout: 30000 }, () => {
+            const opts = { topic: 'backbeat-sanitycheck', timeout: 10000 };
+            this._producer.getMetadata(opts, err => {
+                if (err) {
+                    this.emit('error', err);
+                }
+            });
+        });
         this._producer.on('ready', () => {
             this._ready = true;
             this.emit('ready');
--- a/lib/queuePopulator/QueuePopulator.js
+++ b/lib/queuePopulator/QueuePopulator.js
@@ -69,7 +69,15 @@ class QueuePopulator {
     open(cb) {
         this._loadExtensions();
         async.series([
-            next => this._setupMetricsClients(next),
+            next => this._setupMetricsClients(err => {
+                if (err) {
+                    this.log.error('error setting up metrics client', {
+                        method: 'QueuePopulator.open',
+                        error: err,
+                    });
+                }
+                return next(err);
+            }),
             next => this._setupExtensions(err => {
                 if (err) {
                     this.log.error(


lib/BackbeatProducer.js
```diff
@@ -68,7 +68,14 @@ class BackbeatProducer extends EventEmitter {
             'request.timeout.ms': ACK_TIMEOUT,
         });
         this._ready = false;
-        this._producer.connect();
+        this._producer.connect({ timeout: 30000 }, () => {
+            const opts = { topic: 'librdtesting-01', timeout: 10000 };
```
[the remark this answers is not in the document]
giacomoguiulfo: Changed requseted topic name.

lib/BackbeatProducer.js
```diff
@@ -68,7 +68,14 @@ class BackbeatProducer extends EventEmitter {
             'request.timeout.ms': ACK_TIMEOUT,
         });
         this._ready = false;
-        this._producer.connect();
+        this._producer.connect({ timeout: 30000 }, () => {
+            const opts = { topic: 'backbeat-sanitycheck', timeout: 10000 };
```
jonathan-gramain: Maybe you could get metadata from `this._topic` instead of using a dummy topic?

lib/BackbeatProducer.js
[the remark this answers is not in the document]
giacomoguiulfo: According to the node-rdkafka's README, ```"when fetching metadata for a specific topic, if a topic reference does not exist, one is created using the default config".``` So I think there shouldn't not be any problem with ```auto.creation.topic.enable```. I tested with the Zenko k8s stack and it worked as expected.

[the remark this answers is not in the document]
rahulreddy: I think when auto creation is disabled, Kafka would return an error saying its unable to create a topic which is a false negative.

lib/BackbeatProducer.js
jonathan-gramain: I think calling `getMetadata` on a topic creates it if does not exist, which could possibly be avoided by using an existing topic.

rahulreddy: This would create the `backbeat-replication` topic with some partition, replica defaults apart from what's specified in the config, which is not ideal.

philipyoo: From the library docs: 
Note: using a metadataOptions.topic parameter has a potential side-effect. A Topic object will be created, if it did not exist yet, with default options and it will be cached by librdkafka. A subsequent call to create the topic object with specific options (e.g. acks) will return the previous instance and the specific options will be silently ignored. To avoid this side effect, the topic object can be created with the expected options before requesting metadata, or the metadata request can be performed for all topics (by omitting metadataOptions.topic).

Should we reserve a topic name for the specific purpose of checking if kafka is available? I think crashing is useful as well for non-kubernetes use instead of hanging

lib/BackbeatProducer.js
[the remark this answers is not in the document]
giacomoguiulfo: When ```auto creation``` is disabled, topics aren't created. But backbeat keeps running normally.

lib/BackbeatProducer.js
giacomoguiulfo: Good Idea!

jonathan-gramain: @rahulreddy I don't think we have code to pre-create topics for now in kube stack, but when we do it it should not be an issue anymore. And I'm thinking we need this setup code in order to create topics with the number of partitions and replicas that we want?

jonathan-gramain: Unless we let kube do the topic setup at the same time it attempts to start other components (that seems to be the philosophy of it?) in which case indeed it becomes an issue to let backbeat auto-create topics.

giacomoguiulfo: @jonathan-gramain Using ```this._topic``` also makes CI fail for some tests, unlike using the ```backbeat-sanitycheck``` topic.

giacomoguiulfo:
Reverted topic to be ```'backbeat-sanitycheck'``` again.

philipyoo:
just a rebase and the tests should pass 👍

PR #298, opened 2018-05-24
bf: ZENKO-408 no zookeeper provisioning for conductor

jonathan-gramain:
Disable zookeeper provisioning for lifecycle conductors, as kubernetes is able to spawn the required number of instances (i.e. 1) and respawn on another machine whenever one goes down. The zookeeper provisioning was used for about the same purpose.


extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -63,7 +61,6 @@ class LifecycleConductor {
         this._zkClient = null;
         this._started = false;
         this._isActive = false;
```
philipyoo: can also remove this variable i think

jonathan-gramain: thanks, just removed

PR #299, opened 2018-05-24
ft: ZENKO-404 service account support for lifecycle

jonathan-gramain:
Support a service account 'service-lifecycle' for lifecycle feature
    
* move code related to replication management from lib/management.js
  to extensions/replication/management.js

* add optional callback function 'applyState' that modules can use to
  periodically apply and save their state to management db

* change authentication support for lifecycle to use configuration in
  extensions.lifecycle.auth config, to be able to configure the service
  account name

* update docker-entrypoint.sh with missing S3_HOST and S3_PORT env var
  check

* update docker-entrypoint.sh to add log level support (LOG_LEVEL)

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -10,15 +10,6 @@
         "host": "127.0.0.1",
         "port": 8000
     },
-    "auth": {
-        "type": "account",
-        "account": "bart",
-        "vault": {
-            "host": "127.0.0.1",
-            "port": 8500,
-            "adminPort": 8600
-        }
-    },
     "queuePopulator": {
         "cronRule": "*/5 * * * * *",
         "batchMaxRead": 10000,
@@ -93,6 +84,15 @@
             }
         },
         "lifecycle": {
+            "auth": {
+                "type": "service",
+                "account": "service-lifecycle",
+                "vault": {
+                    "host": "127.0.0.1",
+                    "port": 8500,
+                    "adminPort": 8600
+                }
+            },
             "zookeeperPath": "/lifecycle",
             "bucketTasksTopic": "backbeat-lifecycle-bucket-tasks",
             "objectTasksTopic": "backbeat-lifecycle-object-tasks",
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,6 +6,15 @@ set -e
 # modifying config.json
 JQ_FILTERS_CONFIG="."
 
+if [[ "$LOG_LEVEL" ]]; then
+    if [[ "$LOG_LEVEL" == "info" || "$LOG_LEVEL" == "debug" || "$LOG_LEVEL" == "trace" ]]; then
+        JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .log.logLevel=\"$LOG_LEVEL\""
+        echo "Log level has been modified to $LOG_LEVEL"
+    else
+        echo "The log level you provided is incorrect (info/debug/trace)"
+    fi
+fi
+
 if [[ "$ZOOKEEPER_AUTO_CREATE_NAMESPACE" ]]; then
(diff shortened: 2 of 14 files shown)


conf/config.json
```diff
@@ -90,6 +81,15 @@
             }
         },
         "lifecycle": {
+            "auth": {
+                "type": "service",
+                "account": "service-lifecycle",
```
rachedbenmustapha, line 89: `service-lifecycle` is actually the `accountType` attribute, not really the account

tmacro (dismissed):

docker-entrypoint.sh
```diff
@@ -47,6 +56,14 @@ if [[ "$MONGODB_DATABASE" ]]; then
    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .queuePopulator.mongo.database=\"$MONGODB_DATABASE\""
 fi
 
+if [[ "$S3_HOST" ]]; then
+    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .s3.host=\"$S3_HOST\""
```
rachedbenmustapha, line 60: Is this used as a fallback if extensions are missing s3 host/port?

rachedbenmustapha (dismissed):
nice cleanup 👍

conf/config.json
jonathan-gramain: oh ok, as multiple such account can coexist with different canonical IDs? I re-used the terminology we used for the "auth" config objects.

jonathan-gramain: with my latest commit I also removed the incrementing account ID so if such case can occur they would all have the same account ID in the array (but I doubt this can occur in practice?)

docker-entrypoint.sh
jonathan-gramain: My understanding is that to allow extensions to access an S3 endpoint they need those two environment variables, which get translated into config.json. We introduced this `s3` config object for lifecycle. Is there any existing environment that is set automatically? I think NicolasT referred to kube setting some env var to other services automatically in another PR which could be used here.

conf/config.json
jonathan-gramain: Maybe in case we re-generate a new service account because one is compromised, there could be a time where two of them co-exist before we remove the older one?

rachedbenmustapha: In case a key is compromised Orbit can regenerate the key without changing the account canonical id.

I think it's safe to assume you'll have at most one account of type `service-lifecycle`

rachedbenmustapha (dismissed):

docker-entrypoint.sh
ssalaues: All kubernetes pods get injected with environment variables that have the updated IP and port information for all available services which I believe is what NicolasT was referring to and could potentially be used 🙂

rachedbenmustapha: Wouldn't a simple `docker run` impose the weird var name format? It seems fine as an intermediary mapping step to inject say `CACHE_SERVICE_HOST` into `REDIS_HOST` (fictional names) in the pod template, but a name like `CACHE_SERVICE_HOST` is kind of dictated by the service names defined at runtime outside of the pod, whereas these variables in the entrypoint are defined at build time.

rachedbenmustapha (dismissed):

docker-entrypoint.sh
ssalaues: Yeah exactly, they would be dictated by the potentially different service names and release names so we wouldn't be able to hard code the values. My point really being that I think the idea in general to be good, it would require to rework the entire entrypoint script (at the very least) to take something like release name as a parameter and assume the default services names for Zenko K8s specific context which seems like a lot of work for not really much benefit.

rachedbenmustapha (approved):

PR #300, opened 2018-05-24
bf: ZENKO-359 instantiate a QueueProcessor per site

philipyoo:
We originally intended to revert to having a queue processor per site.
This way, we can process entries much quicker for every site we are
replicating to rather than queueing all entries to a single queue
processor.

By switching to queue processors per site, the status bug behavior
no longer remains (at least for the few tests done manually). 

In a one-to-many replication setup, if a failed status occurs on a destination
site (i.e. destination bucket doesn't exist), and if the failed destination bucket
is created, the original entry can be re-processed. When processing an entry,
and an error occurs, the replication operation will be retried for a certain
period of time before putting a FAILED status on the entry destination site.
Using a single queue processor meant that the entry could be read again for
the next site within the storage class, and again be retried. This lead to a retry
on the FAILED site and overwriting any COMPLETED sites version id.

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -1,7 +1,6 @@
 'use strict'; // eslint-disable-line
 
 const http = require('http');
-const async = require('async');
 const { EventEmitter } = require('events');
 
 const Logger = require('werelogs').Logger;
@@ -56,9 +55,10 @@ class QueueProcessor extends EventEmitter {
      *   number of seconds before giving up retries of an entry
      *   replication
      * @param {MetricsProducer} mProducer - instance of metrics producer
+     * @param {String} site - site name
      */
     constructor(zkConfig, kafkaConfig, sourceConfig, destConfig, repConfig,
-        mProducer) {
+        mProducer, site) {
         super();
         this.kafkaConfig = kafkaConfig;
         this.sourceConfig = sourceConfig;
@@ -70,10 +70,12 @@ class QueueProcessor extends EventEmitter {
         this.replicationStatusProducer = null;
         this._consumer = null;
         this._mProducer = mProducer;
+        this.site = site;
 
         this.echoMode = false;
 
-        this.logger = new Logger('Backbeat:Replication:QueueProcessor');
+        this.logger = new Logger(
+            `Backbeat:Replication:QueueProcessor:${this.site}`);
 
         // global variables
         // TODO: for SSL support, create HTTPS agents instead
@@ -210,6 +212,7 @@ class QueueProcessor extends EventEmitter {
             accountCredsCache: this.accountCredsCache,
             replicationStatusProducer: this.replicationStatusProducer,
             logger: this.logger,
+            site: this.site,
         };
     }
 
@@ -238,7 +241,7 @@ class QueueProcessor extends EventEmitter {
                 return undefined;
             }
             const groupId =
-                `${this.repConfig.queueProcessor.groupId}`;
+                `${this.repConfig.queueProcessor.groupId}-${this.site}`;
             this._consumer = new BackbeatConsumer({
                 kafka: { hosts: this.kafkaConfig.hosts },
                 topic: this.repConfig.topic,
@@ -255,7 +258,9 @@ class QueueProcessor extends EventEmitter {
             });
             this._consumer.on('metrics', data => {
                 // i.e. data = { my-site: { ops: 1, bytes: 124 } }
-                this._mProducer.publishMetrics(data,
+                const filteredData = {};
+                filteredData[this.site] = data[this.site];
+                this._mProducer.publishMetrics(filteredData,
                     metricsTypeProcessed, metricsExtension, err => {
                         this.logger.trace('error occurred in publishing ' +
                             'metrics', {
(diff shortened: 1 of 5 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -315,21 +318,17 @@ class QueueProcessor extends EventEmitter {
...
+            if (sites.includes(this.site)) {
                 const replicationEndpoint = this.destConfig.bootstrapList
-                    .find(endpoint => endpoint.site === site);
-                if (replicationEndpoint
-                    && replicationBackends.includes(replicationEndpoint.type)) {
-                    task = new MultipleBackendTask(this, site);
+                    .find(endpoint => endpoint.site === this.site);
+                if (replicationEndpoint &&
+                replicationBackends.includes(replicationEndpoint.type)) {
+                    task = new MultipleBackendTask(this);
```
jonathan-gramain, line 329: No need to adapt `MultipleBackendTask` constructor? Because I don't see the change in this PR.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -255,7 +258,7 @@ class QueueProcessor extends EventEmitter {
             });
             this._consumer.on('metrics', data => {
                 // i.e. data = { my-site: { ops: 1, bytes: 124 } }
-                this._mProducer.publishMetrics(data,
+                this._mProducer.publishMetrics(data[this.site],
```
jonathan-gramain: Did you change the API of `publishMetrics`? It seems to me you changed the structure of metrics passed by accessing `[this.site]` key, and I would think that each specialized processor would push metrics the same way but will only provide metrics for the site they're responsible for as a consequence.

philipyoo: The constructor change should be reflected in `ReplicateObject` and site is passed through the `getStateVars`

extensions/replication/queueProcessor/task.js
```diff
@@ -44,9 +45,13 @@ metricsProducer.setupProducer(err => {
                 destConfig.bootstrapList = config.getBootstrapList();
             });
 
-            const queueProcessor = new QueueProcessor(zkConfig, kafkaConfig,
-                sourceConfig, destConfig, repConfig, metricsProducer);
-            queueProcessor.start();
+            // Start QueueProcessor for each site
+            const siteNames = bootstrapList.map(i => i.site);
+            siteNames.forEach(site => {
+                const queueProcessor = new QueueProcessor(zkConfig, kafkaConfig,
```
rahulreddy, line 57: You cannot define a constant in a loop.

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: The constructor change should be reflected in `ReplicateObject` and site is passed through the `getStateVars`

extensions/replication/queueProcessor/task.js
philipyoo: shouldn't `const` be used here? We are only using within the scope of the loop and using `let` will throw a linter error since the variable is never redefined

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: Right, there is no explicit constructor in `MultipleBackendTask`

jonathan-gramain (dismissed):

extensions/replication/queueProcessor/task.js
rahulreddy: Ah, it's never used outside the loop. I missed that.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -255,7 +258,9 @@ class QueueProcessor extends EventEmitter {
             });
             this._consumer.on('metrics', data => {
                 // i.e. data = { my-site: { ops: 1, bytes: 124 } }
-                this._mProducer.publishMetrics(data,
+                const filteredData = {};
```
bennettbuchanan, line 261: Is there a need for this object? E.g. why not just send `data[this.site]` to `publichMetrics`?

extensions/replication/queueProcessor/QueueProcessor.js
bennettbuchanan: In that case, doesn't the `getStateVars` method have to be updated to send the site?

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: I think he intends to send an object containing `data[this.site]` as a key, hence the new object.

philipyoo: yup, i'm trying to keep the data in the form of `{ site-name: { ... } }`

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: yup, passed as var name `site` from `getStateVars`

extensions/replication/queueProcessor/QueueProcessor.js
bennettbuchanan: Oh I see, it's an object that contains only the relevant key, hence it is filtered. Thanks!

bennettbuchanan (dismissed):

tmacro (dismissed):

PR #301, opened 2018-05-29
ft: ZENKO-453 add multiple site setup script

philipyoo:
StorageClass uses a comma-separated list of target site names.
Role Arn is just the source role Arn or if there is a destination zenko site, role arn is a comma separated source role Arn and target Zenko role Arn

### DIFF
--- a/bin/replication.js
+++ b/bin/replication.js
@@ -22,7 +22,12 @@ function _createSetupReplication(command, options, log) {
     const sourceProfile = options.sourceProfile;
     const targetProfile = options.targetProfile;
     const siteName = options.siteName;
-    const targetIsExternal = siteName !== undefined;
+
+    let targetIsExternal = siteName !== undefined;
+    if (command === 'multi-site-setup' && targetProfile) {
+        // need to create target zenko site bucket and role arn
+        targetIsExternal = false;
+    }
 
     // Required options
     if (!sourceBucket || !targetBucket ||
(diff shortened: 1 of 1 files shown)


bin/replication.js
```diff
@@ -105,6 +110,73 @@ program.version(version);
...
+        const targetProfile = options.targetProfile;
+        const multiSites = options.multiSites;
+
+        const destinationEndpoint = config.getBootstrapList()
+            .find(dest => Array.isArray(dest.servers));
+
+        const hasIncorrectMultiSites = !multiSites || multiSites.split(',')
+            .some(site => site.length === 0);
+        const hasValidOptionals = (targetBucket || targetProfile) ?
+            (!(!targetBucket || !targetProfile) && (!hasIncorrectMultiSites &&
```
philipyoo, line 141: if `targetBucket` or `targetProfile` is undefined, using an `(targetBucket && targetProfile)...` here will return an undefined value, which is the reason for the validation made here

bin/replication.js
```diff
@@ -105,6 +110,73 @@ program.version(version);
...
 });
 
+program
+    .command('multi-site-setup')
+    .description('if a Zenko site is included as a replication endpoint for a' +
+        ' one-to-many setup, "target-bucket" and "target-profile" are needed')
+    .option('--source-bucket <name>', '[required] source bucket name')
+    .option('--source-profile <name>',
+            '[required] source aws/credentials profile')
+    .option('--target-bucket <name>', '[optional] necessary if adding a zenko' +
```
bennettbuchanan: Following the other descriptions maybe this should read as '[optional] target bucket name (necessary if adding a Zenko site as a replication endpoint)

bin/replication.js
```diff
@@ -105,6 +110,73 @@ program.version(version);
...
+            ' site as a replication endpoint')
+    .option('--target-profile <name>', '[optional] target aws/credentials ' +
+            'profile (necessary if adding a zenko site as a replication ' +
```
bennettbuchanan: I guess we should be consistent in the capitalization of 'Zenko'.

bin/replication.js
```diff
@@ -105,6 +110,73 @@ program.version(version);
...
+            'endpoint)')
+    .option('--multi-sites <site1,site2,...>', '[required] comma separated ' +
+            'list of sites"')
```
bennettbuchanan: There is an extra `"`.

bin/replication.js
```diff
@@ -105,6 +110,73 @@ program.version(version);
...
+        const multiSites = options.multiSites;
+
+        const destinationEndpoint = config.getBootstrapList()
+            .find(dest => Array.isArray(dest.servers));
+
+        const hasIncorrectMultiSites = !multiSites || multiSites.split(',')
+            .some(site => site.length === 0);
+        const hasValidOptionals = (targetBucket || targetProfile) ?
+            (!(!targetBucket || !targetProfile) && (!hasIncorrectMultiSites &&
+            multiSites.split(',').indexOf(destinationEndpoint.site) !== -1))
```
bennettbuchanan: I think using `.includes` on the array to check for the site is a bit cleaner, and this works great as well 😄.

bennettbuchanan (dismissed):
Nice! This will be very useful 😸. Left some minor comments here.

bin/replication.js
philipyoo: nice, I like the use of `includes` 👍

bin/replication.js
```diff
@@ -105,6 +110,72 @@ program.version(version);
...
+        let targetBucket = options.targetBucket;
+        const targetProfile = options.targetProfile;
+        const multiSites = options.multiSites;
+
+        const destinationEndpoint = config.getBootstrapList()
+            .find(dest => Array.isArray(dest.servers));
+
+        const hasIncorrectMultiSites = !multiSites || multiSites.split(',')
+            .some(site => site.length === 0);
+        const hasValidOptionals = (targetBucket || targetProfile) ?
```
jonathan-gramain, line 140: Suggestions to improve clarity:

- avoid the use of `?:` operator: when the check gets complex enough, prefer a regular `if..else` construct

- logically speaking, i.e. when used in a boolean context, I think `!(!targetBucket || !targetProfile)` is equivalent to `targetBucket && targetProfile`

- avoid double negation when checking: change var to `hasCorrectMultiSites`

PR #307, opened 2018-06-06
bf: crash status processor when unable to initialize

giacope:
When Kafka is unavailable, the backbeat replication status processor will hang and not connect to it even if it becomes available later on. These changes will make the process exit if it's unable to connect to Kafka after a certain amount of time.

In an orchestrated environment, the orchestration tool can recognize this crash and restart the failed container, thus allowing it to try to connect again to Kafka (which might be available later on).

This PR is very similar to #296. However, this fix is on the backbeat consumer code (and one of its extensions). Consequently, any extension/program that the instanciates a Backbeat Consumer object should have this behavior (not only the replication status processor).

### DIFF
--- a/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
+++ b/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
@@ -94,6 +94,7 @@ class ReplicationStatusProcessor {
      * @return {undefined}
      */
     start(options, cb) {
+        let consumerReady = false;
         this._FailedCRRProducer = new FailedCRRProducer(this.kafkaConfig);
         this._consumer = new BackbeatConsumer({
             kafka: { hosts: this.kafkaConfig.hosts },
@@ -104,8 +105,14 @@ class ReplicationStatusProcessor {
             queueProcessor: this.processKafkaEntry.bind(this),
             bootstrap: options && options.bootstrap,
         });
-        this._consumer.on('error', () => {});
+        this._consumer.on('error', () => {
+            if (!consumerReady) {
+                this.logger.fatal('error starting a backbeat consumer');
+                process.exit(1);
+            }
+        });
         this._consumer.on('ready', () => {
+            consumerReady = true;
             this.logger.info('replication status processor is ready to ' +
                              'consume replication status entries');
             this._consumer.subscribe();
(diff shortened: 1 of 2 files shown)


JianqinWang (dismissed):

dora-korpar (dismissed):

alexanderchan-scality (dismissed):

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -104,7 +104,10 @@ class ReplicationStatusProcessor {
             queueProcessor: this.processKafkaEntry.bind(this),
             bootstrap: options && options.bootstrap,
         });
-        this._consumer.on('error', () => {});
+        this._consumer.on('error', () => {
+            this.logger.fatal('error starting a backbeat consumer');
```
jonathan-gramain: I'm concerned that we would exit the process also after it has correctly connected to Kafka, in case errors are raised. Not that it's necessarily a bad thing, since those processes should be stateless, but kafka client should handle reconnection automatically in such case, so we may want to refine e.g. by only exiting if `!this._consumerReady`.

jonathan-gramain (dismissed):

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -104,7 +104,12 @@ class ReplicationStatusProcessor {
             queueProcessor: this.processKafkaEntry.bind(this),
             bootstrap: options && options.bootstrap,
         });
-        this._consumer.on('error', () => {});
+        this._consumer.on('error', () => {
+            if (!this._consumer._consumerReady) {
```
giacope: @jonathan-gramain Is it better like this?

jonathan-gramain: Actually I think it's not a good idea to access the private members directly.

Can you instead use a local variable that you update when the `'ready'` event has been emitted?

giacope: Done 👍

giacope:
Bert-e?

PR #311, opened 2018-06-07
bugfix: ZENKO-331 use correct http verb and refactor routing

philipyoo:
Fix HTTP Verb methods for api routes
Refactor route parser and re-organize the server functional tests

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -98,79 +98,102 @@ class BackbeatAPI {
     }
 
     /**
-     * Check if incoming request is valid
+     * Find valid route from list of routes defined in Arsenal backbeat routes
+     * Sets the matched route on the BackbeatRequest object with possible
+     * extra properties.
      * @param {BackbeatRequest} bbRequest - holds relevant data about request
-     * @return {boolean} true/false
+     * @return {Object|null} - The error object or `null` if no error
      */
-    isValidRoute(bbRequest) {
+    findValidRoute(bbRequest) {
+        /* eslint-disable no-param-reassign */
         const rDetails = bbRequest.getRouteDetails();
-        if (!rDetails) {
-            return false;
-        }
         const route = bbRequest.getRoute();
-        // first validate healthcheck routes
-        if (route.substring(3) === 'healthcheck') {
-            return true;
-        }
-        if (route.substring(3) === 'monitoring/metrics') {
-            return true;
-        }
+        const addKeys = {};
 
-        /*
-            {
-                category: 'metrics',
-                extension: 'crr',
-                site: 'my-site-name',
-                metric: 'backlog', (optional)
+        // first validate healthcheck routes or prom routes since they do not
+        // have rDetails set
+        if (route === 'healthcheck') {
+            if (bbRequest.getHTTPMethod() !== 'GET') {
+                return errors.MethodNotAllowed;
             }
-        */
-        // check metric routes
-        // Are there any routes with matching extension?
-        const extensions = routes.reduce((store, r) => {
-            if (r.extensions[rDetails.extension] &&
-                r.extensions[rDetails.extension].includes(rDetails.status)) {
-                store.push(Object.keys(r.extensions));
-            } else if (rDetails.category === 'metrics') {
-                store.push(Object.keys(r.extensions));
+            // TODO: this logic changes with addition of deep healthcheck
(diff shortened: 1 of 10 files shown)


tests/functional/api/BackbeatServer.js
```diff
@@ -88,8 +88,35 @@ function getRequest(path, done) {
     req.end();
 }
 
-describe('Backbeat Server', () => {
-    describe('healthcheck route', () => {
+describe('Backbeat Server', function f() {
```
philipyoo: The tests here have only been re-ordered to fit inside their describe details. Other than that, timeouts were added to all blocks

rahulreddy: Why don't you enforce timeouts using mocha.opts so set the same timeout globally?

philipyoo: I'm not sure about adding a global timeout, but I can set it for this file from the package.json script since the script only targets this file

lib/api/BackbeatAPI.js
```diff
@@ -98,79 +98,105 @@ class BackbeatAPI {
...
         const route = bbRequest.getRoute();
-        // first validate healthcheck routes
+        const addKeys = {};
+
+        // first validate healthcheck routes or prom routes since they do not
+        // have rDetails set
         if (route === 'healthcheck') {
-            return true;
+            if (bbRequest.getHTTPVerb() !== 'GET') {
+                return errors.MethodNotAllowed
```
jonathan-gramain: I would recommend not customizing the message here, MethodNotAllowed standard message should be fine and will be less confusing.

lib/api/BackbeatAPI.js
```diff
@@ -98,79 +98,105 @@ class BackbeatAPI {
...
+            bbRequest.setMatchedRoute(healthcheckRoute);
+            return null;
         }
         if (route === 'monitoring/metrics') {
-            return true;
+            if (bbRequest.getHTTPVerb() !== 'GET') {
+                return errors.MethodNotAllowed
+                    .customizeDescription('invalid http verb');
+            }
+            const promRoute = routes.filter(r =>
```
jonathan-gramain: Not sure I understand this correctly - If there's only one route under `monitoring` category, it may still be safer to also match the subroute (`metrics`), to avoid unforeseen issues if we add other `monitoring` routes in the future.

lib/api/BackbeatAPI.js
```diff
@@ -98,79 +98,105 @@ class BackbeatAPI {
...
-            if (r.extensions[rDetails.extension] &&
-                r.extensions[rDetails.extension].includes(rDetails.status)) {
-                store.push(Object.keys(r.extensions));
-            } else if (rDetails.category === 'metrics') {
-                store.push(Object.keys(r.extensions));
+        // All routes at this point have extensions property in use
+        // Match http verb and match extension name
+        let filteredRoutes = routes.filter(r => (
+            bbRequest.getHTTPVerb() === r.httpMethod &&
+            Object.keys(r.extensions).includes(rDetails.extension)
```
jonathan-gramain: A more straightforward and efficient way is to write `&& r.extensions[rDetails.extension]`

lib/api/BackbeatRequest.js
```diff
@@ -79,6 +85,14 @@ class BackbeatRequest {
         return this._routeDetails;
     }
 
+    /**
+     * Get the request http verb method
+     * @return {string} request http verb
+     */
+    getHTTPVerb() {
```
jonathan-gramain: It would be better to stick to the RFC terminology and call this `getHTTPMethod` instead.

lib/api/BackbeatAPI.js
philipyoo: Sounds good, I wasn't sure about this route but agree that it'd be good to parse it by each path segment. I'll assume there are only 2 path segments: monitoring category, and type. I want to slightly change the route details for this then in Arsenal, so will add a small change there as well just so it fits in better

jonathan-gramain (commented):
looks good, waiting for squash to validate (otherwise bert-e may pull the rug out and merge)

bennettbuchanan (dismissed):

bennettbuchanan (dismissed):

bennettbuchanan (approved):

jonathan-gramain (approved):

PR #312, opened 2018-06-07
ft: ZENKO-235 manual crr pause resume

philipyoo:
Pause/Resume CRR by pausing all CRR subscribed consumers

Please note that producers are never stopped. This means the user can
still add CRR requests, but they won't be processed if the consumers
have been paused.
Consumers are kept alive. If a crr consumer was paused and it is resumed,
it will start to replicate everything from its last consumed offset.

### DIFF
--- a/docs/crr-pause-resume.md
+++ b/docs/crr-pause-resume.md
@@ -1,0 +1,76 @@
+# Cross-Region Replication (CRR) Manual Pause and Resume
+
+## Description
+
+This feature offers a way for users to manually pause and resume cross-region
+replication (CRR) operations by storage locations.
+
+## Design
+
+A RESTful API will expose methods for users to pause and resume cross-region
+replication operations.
+
+Redis offers a pub/sub function. We will utilize this function to propagate
+requests to all active CRR Kafka Consumers on all nodes with backbeat
+containers setup for replication.
+
+We want to pause and resume the CRR service at the lowest level (in our case,
+pause and resume all Kafka Consumers subscribed to the CRR topic). We want to
+perform these actions at the lowest level in order to stop processing any
+replication entries that might have already been populated by Kafka but have yet
+to be consumed and queued for replication. Any entries that have already been
+consumed by the Kafka Consumer and are being processed for replication will
+continue to finish replication and will not be paused.
+
+The API will have a Redis instance publishing messages to a specific channel.
+All CRR Kafka Consumers will subscribe to this channel and complete any given
+valid requests.
+
+When a Kafka Consumer pauses, the Consumer is still kept alive and maintains
+any internal state, including offset. The Consumer will no longer be
+subscribed to the CRR topic, so will no longer try consuming any entries.
+When the paused Consumer is resumed, it will again resume consuming entries
+from its last offset.
+
+## Definition of API
+
+* POST `/_/crr/pause`
+
+    This POST request is to manually pause the cross-region replication service
+    for all locations configured as destination replication endpoints.
+
+    Response:
+    ```sh
+    {}
+    ```
+
+* POST `/_/crr/pause/<location-name>`
+
+    This POST request is to manually pause the cross-region replication service
+    for a specified location configured as a destination replication endpoint.
+
+    Response:
+    ```sh
+    {}
+    ```
+
+* POST `/_/crr/resume`
+
+    This POST request is to manually resume the cross-region replication
+    service for all locations configured as destination replication endpoints.
+
+    Response:
+    ```sh
+    {}
+    ```
+
+* POST `/_/crr/resume/<location-name>`
+
+    This POST request is to manually resume the cross-region replication
+    service for a specified location configured as a destination replication
+    endpoint.
+
+    Response:
+    ```sh
+    {}
+    ```
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -2,6 +2,7 @@
 
 const http = require('http');
 const { EventEmitter } = require('events');
+const Redis = require('ioredis');
 
 const Logger = require('werelogs').Logger;
 
@@ -54,11 +55,12 @@ class QueueProcessor extends EventEmitter {
      * @param {String} repConfig.queueProcessor.retryTimeoutS -
      *   number of seconds before giving up retries of an entry
      *   replication
+     * @param {Object} redisConfig - redis configuration
      * @param {MetricsProducer} mProducer - instance of metrics producer
      * @param {String} site - site name
      */
     constructor(zkConfig, kafkaConfig, sourceConfig, destConfig, repConfig,
-        mProducer, site) {
+        redisConfig, mProducer, site) {
         super();
         this.kafkaConfig = kafkaConfig;
         this.sourceConfig = sourceConfig;
@@ -83,6 +85,7 @@ class QueueProcessor extends EventEmitter {
         this.destHTTPAgent = new http.Agent({ keepAlive: true });
 
         this._setupVaultclientCache();
+        this._setupRedis(redisConfig);
 
         // FIXME support multiple scality destination sites
         if (Array.isArray(destConfig.bootstrapList)) {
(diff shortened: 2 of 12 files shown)


philipyoo:
commitlint doesn't like square brackets :(

docs/crr-pause-resume.md
```diff
@@ -0,0 +1,54 @@
...
+When the paused Consumer is resumed, it will again resume consuming entries
+from its last offset.
+
+## Definition of API
+
+* POST `/_/crr/pause/<site>`
+
+    This POST request is to manually pause the cross-region replication service.
+    You may specify a site name, or if you would like to pause replication for
+    all sites, specify "all" as the site name.
```
bennettbuchanan: My feeling is that it would make sense to consider `/_/crr/pause/all` a separate route, and so should be another bullet here. The same would apply for `/_/crr/resume/all`. Alternatively, I think it may be simpler to have the route `/_/crr/pause` pause all sites. Maybe this design makes more sense for how the routing is implemented though?

docs/crr-pause-resume.md
```diff
@@ -0,0 +1,54 @@
...
+
+    Response:
+    ```sh
+    {}
```
bennettbuchanan: It might be helpful UX to include some kind of confirmation in the response. I'm imagining something like:

```
{
    aws-backend: PAUSED,
}
```

lib/api/BackbeatAPI.js
```diff
@@ -518,6 +526,48 @@ class BackbeatAPI {
         return { reqBody };
     }
 
+    /**
+     * Pause all CRR operations
```
bennettbuchanan: Minor: not necessarily 'all'.

lib/api/BackbeatAPI.js
```diff
@@ -518,6 +526,48 @@ class BackbeatAPI {
...
+        }
+        sites.map(s => `${this._crrTopic}-${s}`).forEach(channel => {
+            this._redisPublisher.publish(channel, 'pauseService');
+        });
+        this._logger.info('replication service paused');
+        return cb(null, {});
+    }
+
+    /**
+     * Resume all CRR operations
```
bennettbuchanan: Minor: not necessarily 'all'.

docs/crr-pause-resume.md
```diff
@@ -0,0 +1,54 @@
...
+replication operations.
+
+Redis offers a pub/sub function. We will utilize this function to propagate
+requests to all active CRR Kafka Consumers on all nodes with backbeat
+containers setup for replication.
+
+We want to pause and resume the CRR service at the lowest level (in our case,
+pause and resume all Kafka Consumers subscribed to the CRR topic). We want to
+perform these actions at the lowest level in order to stop processing any
+current batch jobs.
```
bennettbuchanan: Does this stop jobs that are currently in process of being replicated (i.e. multipart uploads).

bennettbuchanan (commented):
Nice! 😸

docs/crr-pause-resume.md
philipyoo: No, once the consumer "consumes" the job, the job has already been sent to the queue for completing the task. So I'm afraid it won't stop the currently processing jobs. If the job hasn't yet been sent to the internal Kafka consumer `consume` method, then the job shouldn't process or should pause

bennettbuchanan: That seems fine to me 😄. I just thought the wording here made it sound as such. Particularly 'stop processing any current batch jobs'.

docs/crr-pause-resume.md
philipyoo: i'm a little hesitant in returning the status right away because when a user sends a request to pause or resume, I don't confirm that it was actually set. Since i'm using Redis to communicate between the processes, if the subscriber fails in receiving the message, there is no way for the publisher to know that unless it checks crr status (which will be opened in a separate pr)

Instead, I'm hoping a user using this feature will perform the pause/resume action and will send a separate request to check status. That way, the info provided to the user will be more accurate

docs/crr-pause-resume.md
philipyoo: ah good to know, I'll update the wording here

docs/crr-pause-resume.md
philipyoo: sounds good, I'm leaning towards `/_/crr/pause` to indicate all sites, and if the user wants a specific site, they can specify it as the last part `/_/crr/pause/<site>`. Editing docs to reflect and show both variations

docs/crr-pause-resume.md
```diff
@@ -0,0 +1,76 @@
...
+subscribed to the CRR topic, so will no longer try consuming any entries.
+When the paused Consumer is resumed, it will again resume consuming entries
+from its last offset.
+
+## Definition of API
+
+* POST `/_/crr/pause`
+
+    This POST request is to manually pause the cross-region replication service
+    for all sites configured as destination replication endpoints.
```
jonathan-gramain: We should avoid the term "site", let's use "storage locations" instead (or just "locations" when the context is not ambiguous) throughout the doc. In the code it can still be "site" for now.

lib/BackbeatConsumer.js
```diff
@@ -511,10 +514,38 @@ class BackbeatConsumer extends EventEmitter {
...
     /**
-    * force commit the current offset and close the client connection
-    * @param {callback} cb - callback to invoke
-    * @return {undefined}
-    */
+     * pause the kafka consumer
+     * @param {string} site - name of site
+     * @return {undefined}
+     */
+    pauseService(site) {
```
jonathan-gramain: In this class there's no notion of service, just a consumer. So the function names can be simply "pause" and "resume".

lib/BackbeatConsumer.js
```diff
@@ -511,10 +514,38 @@ class BackbeatConsumer extends EventEmitter {
...
+        // Use of KafkaConsumer#pause did not work. Using alternative
+        // of unsubscribe/subscribe
+        if (this._consumerReady) {
+            this._consumer.unsubscribe();
+            this._consumerReady = false;
+            this._log.info(`paused replication for site ${site}`);
```
jonathan-gramain: Similarly, we're not bound to replication service here. We may log something here (preferably debug) but this log would go into the service code (QueueProcessor) instead.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -200,6 +205,30 @@ class QueueProcessor extends EventEmitter {
...
+        this._redis.subscribe(channelName, err => {
+            if (err) {
+                this.logger.fatal('queue processor failed to subscribe to ' +
+                                  `crr redis channel for site ${this.site}`,
+                                  { method: 'QueueProcessor.constructor',
+                                    error: err });
+                process.exit(1);
+            }
+            this._redis.on('message', (channel, message) => {
+                if (channel === channelName && this._consumer[message]) {
```
jonathan-gramain: I guess you cannot do the economy here of having to dispatch explicitly `"pauseService" -> consumer.pause()` and `"resumeService" -> consumer.resume()`, as those are part of the implementation logic (i.e. to pause my service, I have to pause that consumer, but I also may want later to add some more pausing logic or feedback, like logging).

If you prefer you can have the `pauseService` and `resumeService` methods defined in the `QueueProcessor` class and dispatch by accessing `this[message]` but I feel this will be more confusing than explicit method calls, and also you still have to check for validity of the message.

lib/api/BackbeatAPI.js
```diff
@@ -527,6 +535,48 @@ class BackbeatAPI {
...
+     * @return {undefined}
+     */
+    pauseCRRService(details, body, cb) {
+        let sites;
+        if (details.site === 'all') {
+            sites = details.extensions.crr.filter(s => s !== 'all');
+        } else {
+            sites = [details.site];
+        }
+        sites.map(s => `${this._crrTopic}-${s}`).forEach(channel => {
```
jonathan-gramain: No need to use `map` and create a temporary array, just do `sites.forEach(site => {...})`

lib/api/BackbeatRequest.js
```diff
@@ -33,18 +33,25 @@ class BackbeatRequest {
...
-    _parseRetryRoutes(parts, query) {
-        this._routeDetails.extension = parts[0];
-        this._routeDetails.status = parts[1];
-        this._routeDetails.bucket = parts[2];
-        this._routeDetails.key = parts[3];
-        this._routeDetails.versionId = parts[4];
-        this._routeDetails.marker = querystring.parse(query).marker;
+    _parseCRRRoutes(parts, query) {
+        if (parts[1] && parts[1] === 'failed') {
+            this._routeDetails.extension = parts[0];
```
jonathan-gramain, line 43: As a future refactor, we should only parse path components up to what allows to dispatch to the proper route, and defer the responsability of parsing the rest of the path components to the specific route implementation, as it's going to become more spaghetti-style each time we add a new route.

jonathan-gramain (changes requested):

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: If adding more pausing logic, I think implementing a separate `pauseService` or `resumeService` method in QueueProcessor would make sense because we will not only be stopping the consumer but performing other actions as well.

I guess if we wanted to handle messages explicitly, I can match the message by the desired or expected commands using a case/switch or if/else statements? Then these can use the appropriate methods in QueueProcessor to then handle the actions necessary (i.e. logging, unsubscribing consumer, etc)

jonathan-gramain: That, or a mapping object that maps the message action names with the appropriate handler method, either works! I tend to prefer slightly the mapping object which feels cleaner in the code.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -211,24 +211,42 @@ class QueueProcessor extends EventEmitter {
...
             }
-            this._redis.on('message', (channel, message) => {
-                if (channel === channelName && this._consumer[message]) {
-                    // should only validate valid functions that can be invoked
-                    // from the redis pub/sub
-                    const validCmds = ['pauseService', 'resumeService'];
-                    if (validCmds.includes(message)) {
-                        this._consumer[message](this.site);
-                    }
+            this._redis.on('message', (channel, action) => {
```
jonathan-gramain: Sorry to iterate again on this as it comes to my mind only now, but I think it would be cleaner to be more future-proof and embed the action into a JSON object like `{"action":"pauseService"}`, the reason being that there are high chances we add more commands that also require some parameters to be passed along, and it will be easier to handle if all message are JSON objects than having to check whether the message is a raw string or a JSON object.

jonathan-gramain:
@philipyoo Can you add the ticket number to the commit messages?

jonathan-gramain (approved):

bennettbuchanan (approved):

PR #316, opened 2018-06-11
ft/ZENKO-476: Add build image step to pre-merge

anurag4DSB:
Changes:

- Add a docker build step in premerge to check if the images are build properly
- Created a new step called post-merge. This step will only be triggered when there a pull requests is perged into development/* branches. 

EVE: tests passed - https://eve.devsca.com/github/scality/backbeat/#/builders/5/builds/2650

TODO: Once we get the credentials from the Bert-e team we need to update the post merge step.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -4,16 +4,29 @@ version: 0.2
 branches:
   feature/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*, ft/*, bf/*, chore/*, docs/*:
     stage: "pre-merge"
+  development/*:
+    stage: "post-merge"
+
 
 stages:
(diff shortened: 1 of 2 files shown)


eve/main.yml
```diff
@@ -4,16 +4,29 @@ version: 0.2
...
-    worker:
+    worker: &pod
+      type: kube_pod
+      path: eve/workers/pod.yml
+    steps:
+    - TriggerStages:
+        name: trigger all the tests
+        stage_names:
+        - run-tests
+        - docker-build
```
ssalaues, line 21: I thought we we only doing the docker-build post-merge and not during pre-merge

PR #326, opened 2018-06-20
feature: ZENKO-315 Allow CRR without versioning

bennettbuchanan:


extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -47,6 +44,10 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
         if (sanityCheckRes) {
             return;
         }
+        // ALlow a non-versioned object if being replicated from an NFS bucket.
+        if (!isMasterKey(entry.key) && !queueEntry.getReplicationIsNFS()) {
```
jonathan-gramain: I think the logic should be `if (isMasterKey(entry.key) && !queueEntry.getReplicationIsNFS()) return;`

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -150,7 +150,11 @@ class ReplicationStatusProcessor {
         if (backend) {
             const bucket = queueEntry.getBucket();
             const key = queueEntry.getObjectKey();
-            const versionId = queueEntry.getEncodedVersionId();
+            // If the original CRR was on a bucket without versioning enabled
+            // (i.e. an NFS bucket), maintain the Redis key schema by using
+            // and empty string.
+            const versionId = queueEntry.getVersionId() ?
```
jonathan-gramain: Since you're now doing the `getVersionId()` check every time to get the encoded version, why not implementing this directly into `getEncodedVersionId()`.

tests/functional/api/BackbeatServer.js
```diff
@@ -573,12 +573,6 @@ describe('Backbeat Server', () => {
                 VersionId: 'c',
                 StorageClass: 'd',
             }],
-            [{
```
jonathan-gramain, line 576: I think it would make sense to keep this check when on non-NFS bucket, and possibly add a check on NFS bucket expecting success without versionId.

jonathan-gramain (changes requested):

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
bennettbuchanan: Good idea—I wasn't sure if `getEncodedVersionId` was being used elsewhere but it looks like it was added for Backbeat.

extensions/replication/ReplicationQueuePopulator.js
bennettbuchanan: Good catch!

tests/functional/api/BackbeatServer.js
bennettbuchanan: I agree, although the API cannot be aware of whether it is an retrying an NFS bucket as it just pushes the request information to a kafka topic and responds that CRR is PENDING.

jonathan-gramain:
@bert-e wait

jonathan-gramain (dismissed):

lib/api/BackbeatAPI.js
```diff
@@ -419,12 +424,15 @@ class BackbeatAPI {
                 }
                 const Bucket = queueEntry.getBucket();
                 const Key = queueEntry.getObjectKey();
-                const VersionId = queueEntry.getEncodedVersionId();
+                // If the original CRR was on a bucket without `versioning`
+                // enabled (i.e. an NFS bucket), maintain the Redis key schema
+                // by using and empty string.
+                const VersionId = queueEntry.getEncodedVersionId() || '';
```
jonathan-gramain: A minor cleanup suggestion, it could be more direct to do the fallback to `''` only when calling `getFailedCRRKey` as that's where it is required for Redis.

jonathan-gramain (dismissed):

JianqinWang (dismissed):

jonathan-gramain (approved):

JianqinWang (approved):

PR #331, opened 2018-06-25
ft: ZENKO-583 get CRR status

philipyoo:
Check if CRR is enabled for a given destination replication location via the API.
Save CRR status state within zookeeper. If pod goes down, read and resume from previous state

### DIFF
--- a/docs/crr-pause-resume.md
+++ b/docs/crr-pause-resume.md
@@ -11,7 +11,7 @@ A RESTful API will expose methods for users to pause and resume cross-region
 replication operations.
 
 Redis offers a pub/sub function. We will utilize this function to propagate
-requests to all active CRR Kafka Consumers on all nodes with backbeat
+requests to all active CRR Kafka Consumers on all nodes that have backbeat
 containers setup for replication.
 
 We want to pause and resume the CRR service at the lowest level (in our case,
@@ -23,16 +23,44 @@ consumed by the Kafka Consumer and are being processed for replication will
 continue to finish replication and will not be paused.
 
 The API will have a Redis instance publishing messages to a specific channel.
-All CRR Kafka Consumers will subscribe to this channel and complete any given
-valid requests.
+QueueProcessors will subscribe to this channel, and on receiving a request
+to pause or resume CRR, will notify all of their BackbeatConsumers to perform
+the action if applicable. If an action occurred, the QueueProcessor will receive
+an update on the current status of each Consumer. Based on the global status of
+a location, the status will be updated in Zookeeper if a change has occurred.
 
-When a Kafka Consumer pauses, the Consumer is still kept alive and maintains
-any internal state, including offset. The Consumer will no longer be
-subscribed to the CRR topic, so will no longer try consuming any entries.
-When the paused Consumer is resumed, it will again resume consuming entries
-from its last offset.
+It is important to note, when a Consumer pauses, the Consumer process is still
+kept alive and maintains any internal state, including offset. The Consumer will
+no longer be subscribed to the CRR topic, so will no longer try consuming any
+entries. When the paused Consumer is resumed, it will again resume consuming
+entries from its last offset.
 
 ## Definition of API
+
+* GET `/_/crr/status`
+
+    This GET request checks if cross-region replication is enabled or not for
+    all locations configured as destination replication endpoints.
+
+    Response:
+    ```json
+    {
+        "location1": "disabled",
+        "location2": "enabled"
+    }
+    ```
+
+* GET `/_/crr/status/<location-name>`
+
+    This GET request checks if cross-region replication is enabled or not for
+    a specified location configured as a destination replication endpoint.
+
+    Response:
+    ```json
+    {
+        "<location-name>": "enabled"
+    }
+    ```
 
 * POST `/_/crr/pause`
 
@@ -40,7 +68,7 @@ from its last offset.
     for all locations configured as destination replication endpoints.
 
     Response:
-    ```sh
+    ```json
     {}
     ```
 
@@ -50,7 +78,7 @@ from its last offset.
     for a specified location configured as a destination replication endpoint.
 
     Response:
-    ```sh
+    ```json
     {}
     ```
 
@@ -60,7 +88,7 @@ from its last offset.
     service for all locations configured as destination replication endpoints.
 
     Response:
-    ```sh
+    ```json
     {}
     ```
 
@@ -71,6 +99,6 @@ from its last offset.
     endpoint.
 
     Response:
-    ```sh
+    ```json
     {}
     ```
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -1,5 +1,6 @@
 'use strict'; // eslint-disable-line
 
+const async = require('async');
 const http = require('http');
 const { EventEmitter } = require('events');
 const Redis = require('ioredis');
@@ -21,6 +22,9 @@ const EchoBucket = require('../tasks/EchoBucket');
 const ObjectQueueEntry = require('../utils/ObjectQueueEntry');
 const BucketQueueEntry = require('../utils/BucketQueueEntry');
 
+const { zookeeperReplicationNamespace } = require('../constants');
+const ZK_CRR_STATE_PATH = '/state';
+
 const {
     proxyVaultPath,
     proxyIAMPath,
@@ -37,9 +41,9 @@ class QueueProcessor extends EventEmitter {
      * entries to a target S3 endpoint.
      *
      * @constructor
-     * @param {Object} zkConfig - zookeeper configuration object
+     * @param {node-zookeeper-client.Client} zkClient - zookeeper client
      * @param {Object} kafkaConfig - kafka configuration object
-     * @param {string} kafkaConfig.hosts - list of kafka brokers
+     * @param {String} kafkaConfig.hosts - list of kafka brokers
      *   as "host:port[,host:port...]"
      * @param {Object} sourceConfig - source S3 configuration
      * @param {Object} sourceConfig.s3 - s3 endpoint configuration object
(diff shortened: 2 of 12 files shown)


docs/crr-pause-resume.md
```diff
@@ -23,17 +23,45 @@ consumed by the Kafka Consumer and are being processed for replication will
...
-All CRR Kafka Consumers will subscribe to this channel and complete any given
-valid requests.
-
-When a Kafka Consumer pauses, the Consumer is still kept alive and maintains
-any internal state, including offset. The Consumer will no longer be
-subscribed to the CRR topic, so will no longer try consuming any entries.
-When the paused Consumer is resumed, it will again resume consuming entries
-from its last offset.
+QueueProcessors will subscribe to this channel, and on receiving a request
+to pause or resume CRR, will notify all of its BackbeatConsumers to perform the
```
jonathan-gramain: `all of their BackbeatConsumers`

docs/crr-pause-resume.md
```diff
@@ -23,17 +23,45 @@ consumed by the Kafka Consumer and are being processed for replication will
...
 
 ## Definition of API
 
+* GET `/_/crr/status`
+
+    This GET request checks if cross-region replication is enabled or not for
+    all sites configured as destination replication endpoints.
+
+    Response:
+    ```sh
```
jonathan-gramain: I think you should use `json` not `sh`.

docs/crr-pause-resume.md
```diff
@@ -23,17 +23,45 @@ consumed by the Kafka Consumer and are being processed for replication will
...
 
 ## Definition of API
 
+* GET `/_/crr/status`
+
+    This GET request checks if cross-region replication is enabled or not for
+    all locations configured as destination replication endpoints.
+
+    Response:
+    ```sh
```
jonathan-gramain: I think you should use `json` not `sh`.

docs/crr-pause-resume.md
```diff
@@ -23,17 +23,45 @@ consumed by the Kafka Consumer and are being processed for replication will
...
+    {
+        location1: "disabled",
```
jonathan-gramain: Since the output shall be JSON, let's add double-quotes to the keys like `"location1"`

docs/crr-pause-resume.md
```diff
@@ -23,17 +23,45 @@ consumed by the Kafka Consumer and are being processed for replication will
...
+
+* GET `/_/crr/status/<location-name>`
+
+    This GET request checks if cross-region replication is enabled or not for
+    a specified location configured as a destination replication endpoint.
+
+    Response:
+    ```sh
+    {
+        location: "enabled",
```
jonathan-gramain: May be better to stick with the name that appears in the syntax line: `"<location-name>": "enabled"`

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -252,9 +263,36 @@ class QueueProcessor extends EventEmitter {
...
 
+    _getStatusZkPath() {
+        return `${zookeeperReplicationNamespace}${ZK_CRR_STATUS_PATH}/` +
+            `${this.site}`;
+    }
+
+    /**
+     * Update Kafka consumer status in zookeeper for this site-defined
+     * QueueProcessor
+     * @param {boolean} status - true if consumer is enabled
```
jonathan-gramain: Better store actual meaningful status names instead of using booleans that only have implicit meaning (in the sense that `true` is not a meaningful status for an observer).

For example, `active` / `paused` would work. Or we can also store a JSON object like `{ "paused": true/false }`, which works better to later extend the status with more info.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -314,7 +353,9 @@ class QueueProcessor extends EventEmitter {
             });
             this._consumer.on('ready', () => {
                 consumerReady = true;
-                this._consumer.subscribe();
+                const disableConsumer = (options && options.status === false);
```
jonathan-gramain: Boolean operations on boolean values shall normally only use boolean operators, so `options && !options.status`

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -282,6 +320,7 @@ class QueueProcessor extends EventEmitter {
      * @param {boolean} [options.disableConsumer] - true to disable
      *   startup of consumer (for testing: one has to call
      *   processQueueEntry() explicitly)
+     * @param {boolean} [options.status] - if false, kafka consumer is disabled
```
jonathan-gramain: Same idea than said above, if carrying a status, better have an object containing the elements of the status (e.g. `status.paused`), otherwise better rename the parameter to `paused` (+ invert its meaning) to help understanding.

extensions/replication/queueProcessor/task.js
```diff
@@ -20,17 +24,96 @@ const log = new werelogs.Logger('Backbeat:QueueProcessor:task');
...
+    zkClient.create(path, Buffer.from('true'), err => {
+        if (err && err.name !== 'NODE_EXISTS') {
+            log.fatal('could not setup zookeeper node', {
+                method: 'QueueProcessor:task',
+                zookeeperPath: path,
+                error: err,
+            });
+            return done(err);
+        }
+        if (err && err.name === 'NODE_EXISTS') {
```
jonathan-gramain: For readability and avoiding extra unnecessary condition checks, I would recommend to change the structure to:

```js
if (err && err.name === 'NODE_EXISTS') {
    ...
    return done();
}
if (err) {
    return done(err);
}
...
```

extensions/replication/queueProcessor/task.js
```diff
@@ -20,17 +24,96 @@ const log = new werelogs.Logger('Backbeat:QueueProcessor:task');
...
+            zkClient.getData(path, (err, data) => {
+                if (err) {
+                    log.fatal('could not check check site status in zookeeper',
```
jonathan-gramain: double `check check`

extensions/replication/queueProcessor/task.js
```diff
@@ -20,17 +24,96 @@ const log = new werelogs.Logger('Backbeat:QueueProcessor:task');
...
+        const { connectionString, autoCreateNamespace } = zkConfig;
+        log.info('opening zookeeper connection for replication processors');
+        zkClient = zookeeper.createClient(connectionString, {
+            autoCreateNamespace,
         });
-        return undefined;
+        zkClient.connect();
+        zkClient.once('error', err => {
+            log.fatal('error connecting to zookeeper', {
+                error: err,
```
jonathan-gramain: `error: err.message`

extensions/replication/queueProcessor/task.js
```diff
@@ -20,17 +24,96 @@ const log = new werelogs.Logger('Backbeat:QueueProcessor:task');
...
+        });
+        zkClient.once('ready', () => {
+            zkClient.removeAllListeners('error');
+            const path = getCRRStatusZkPath();
+            zkClient.mkdirp(path, err => {
+                if (err) {
+                    log.fatal('could not create path in zookeeper', {
+                        method: 'QueueProcessor:task',
+                        zookeeperPath: path,
+                        error: err,
```
jonathan-gramain: ditto

lib/BackbeatConsumer.js
```diff
@@ -196,10 +195,16 @@ class BackbeatConsumer extends EventEmitter {
...
+    *   consumer should not subscribe to its topic
     * @return {this} current instance
     */
-    subscribe() {
-        this._consumer.subscribe([this._topic]);
+    subscribe(disableConsumer) {
+        if (!disableConsumer) {
+            this._consumer.subscribe([this._topic]);
+        } else {
+            this._log.info(`consumer not subcribed to topic ${this._topic}`);
```
jonathan-gramain: At this level I would rather log in debug mode, as the message is likely obscure for a normal user. Such message would better fit at the service level (stating that it's paused, for example).

lib/api/BackbeatAPI.js
```diff
@@ -236,7 +239,9 @@ class BackbeatAPI {
     getHealthcheck(details, cb) {
         return this._healthcheck.getHealthcheck((err, data) => {
             if (err) {
-                this._logger.error('error getting healthcheck', err);
+                this._logger.error('error getting healthcheck', {
+                    error: err,
```
jonathan-gramain: `error: err.message`

jonathan-gramain (changes requested):

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: For scheduled resume, I realized it would probably be best to hold status/state for a queueprocessor at a common node, so I think using a JSON object would work best. I'll reflect the changes to a JSON object in this PR. Following this model:
`{ "paused": true/false, "scheduledResume": <DateString> }` or if you have a better format, please let me know and I can reflect them in the relevant commits

jonathan-gramain: I like the idea that the scheduled resume is set in zookeeper, so that we do not rely on orbit to re-enable replication at the requested time.

lib/api/BackbeatAPI.js
philipyoo: oops, this is my mistake, I want to pass `err` here as is because its a unique object returned from the Healthcheck class. I'll look into figuring out a better way of implementing this in a future refactor. 
Change is reverted

alexanderchan-scality (dismissed):

jonathan-gramain (dismissed):

philipyoo:
bert-e help

philipyoo:
@bert-e help

jonathan-gramain (approved):

alexanderchan-scality (approved):

rahulreddy:
@bert-e reset

PR #343, opened 2018-06-28
ft: ZENKO-239 scheduled resume crr

philipyoo:
Schedule CRR resumes by location

Example request through api:
`curl -X POST -d "{\"hours\":10}" localhost:8900/_/crr/resume/aws-location1/schedule`

If POST request is sent to the schedule route without specifying "hours" in the body of the request, a default 6 hours is applied as specified in JIRA.

If a scheduled resume already exists for a given site, and the user sends in a new schedule resume request, the old resume date should be overwritten.

### DIFF
--- a/docs/crr-pause-resume.md
+++ b/docs/crr-pause-resume.md
@@ -5,6 +5,11 @@
 This feature offers a way for users to manually pause and resume cross-region
 replication (CRR) operations by storage locations.
 
+Users may also choose to resume CRR operations for a given storage location by a
+specified number of hours from the current time. This is particularly useful
+for cases where the user knows a destination location will be down for a
+certain amount of time and would like to schedule a time to resume CRR.
+
 ## Design
 
 A RESTful API will expose methods for users to pause and resume cross-region
@@ -82,6 +87,23 @@ entries from its last offset.
     {}
     ```
 
+* GET `/_/crr/resume/<location-name>`
+
+    This GET request checks if the given location has a scheduled cross-region
+    replication resume job. If a resume job is scheduled, you will see the
+    expected date when the resume occurs.
+
+    You may specify "all" as the `<location-name>` to get all scheduled resume
+    jobs, if any.
+
+    Response:
+    ```json
+    {
+        "location1": "2018-06-28T05:40:20.600Z",
+        "location2": "none"
+    }
+    ```
+
 * POST `/_/crr/resume`
 
     This POST request is to manually resume the cross-region replication
@@ -94,9 +116,33 @@ entries from its last offset.
 
 * POST `/_/crr/resume/<location-name>`
 
-    This POST request is to manually resume the cross-region replication
+    This POST request is to manually resume the cross-region replication service
+    for a specified location configured as a destination replication endpoint.
+
+    Response:
+    ```json
+    {}
+    ```
+
+* POST `/_/crr/resume/<location-name>/schedule`
+
+    This POST request is to schedule resuming the cross-region replication
     service for a specified location configured as a destination replication
-    endpoint.
+    endpoint. You may specify "all" as a location name in order to schedule
+    a resume for all available destination locations.
+
+    Providing a POST request body object with an `hours` key and a valid
+    integer value will schedule a resume to occur by given hours.
+
+    If no request body is provided for this route, a default of 6 hours is
+    applied.
+
+    Request Body Example:
+    ```json
+    {
+        "hours": 6
+    }
+    ```
 
     Response:
     ```json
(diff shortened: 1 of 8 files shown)


docs/crr-pause-resume.md
```diff
@@ -94,9 +116,33 @@ entries from its last offset.
...
-    This POST request is to manually resume the cross-region replication
+    This POST request is to manually resume the cross-region replication service
+    for a specified location configured as a destination replication endpoint.
+
+    Response:
+    ```json
+    {}
+    ```
+
+* POST `/_/crr/resume/<location-name>/schedule`
```
jonathan-gramain, line 127: Instead of having a specific different route to schedule, we could have this as an optional request param on `/_/crr/resume/<location-name>` (either in POST body or as query string). It's also more consistent with the GET route. Let's keep this for now.

rahulreddy (approved):

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -229,10 +231,10 @@ class QueueProcessor extends EventEmitter {
                     resumeService: this._resumeService.bind(this),
                 };
                 try {
-                    const { action } = JSON.parse(message);
+                    const { action, date } = JSON.parse(message);
                     const cmd = validActions[action];
                     if (channel === channelName && typeof cmd === 'function') {
-                        cmd();
+                        cmd(date);
```
jonathan-gramain, line 237: ok for now, but ideally we should pass the whole parsed `message` object to the function.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -269,13 +271,18 @@ class QueueProcessor extends EventEmitter {
...
      * Resume replication consumers
+     * @param {Date} [date] - optional date object for scheduling resume
      * @return {undefined}
      */
-    _resumeService() {
+    _resumeService(date) {
         const enabled = this._consumer.getServiceStatus();
-        if (!enabled) {
+        const now = new Date();
+        if (date && now < new Date(date)) {
```
jonathan-gramain, line 280: We should check validity of the `new Date` object (keep for later).

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -417,8 +452,28 @@ class QueueProcessor extends EventEmitter {
...
+     * @return {undefined}
+     */
+    removeZkState(cb) {
+        const path = this._getZkSiteNode();
+        this.zkClient.remove(path, err => {
+            if (err && err.name !== 'NO_NODE') {
+                this.logger.error('failed removing zookeeper state node', {
+                    method: 'QueueProcessor.removeZkState',
+                    zookeeperPath: path,
+                    error: err,
```
jonathan-gramain, line 466: `error: err.message`

lib/api/BackbeatAPI.js
```diff
@@ -617,6 +620,38 @@ class BackbeatAPI {
         return cb(null, {});
     }
 
+    /**
+     * Validate that the POST request body has the necessary content
+     * @param {String} body - the POST request body string
+     * @return {Object} object containing an error and the request body
+     */
+    _parseScheduleResumeBody(body) {
+        const msg = 'The body of your POST request is not well-formed';
```
jonathan-gramain, line 629: minor: I think this message is superfluous since the error code already tells this.

lib/api/BackbeatAPI.js
```diff
@@ -617,6 +620,38 @@ class BackbeatAPI {
...
+        if (!body) {
+            return defaultRes;
+        }
+        try {
+            reqBody = JSON.parse(body);
+            if (!reqBody.hours) {
+                return defaultRes;
+            }
+            if (reqBody.hours === '' || isNaN(reqBody.hours) ||
+                parseInt(reqBody.hours, 10) <= 0) {
```
jonathan-gramain, line 641: Having sub-hour granularity could be useful in some cases. (either allowing a floating-point value, or allowing a "minutes" param as well could do). Not for this PR.

lib/api/BackbeatAPI.js
```diff
@@ -625,35 +660,55 @@ class BackbeatAPI {
...
         if (details.site === 'all') {
             sites = details.extensions.crr.filter(s => s !== 'all');
         } else {
             sites = [details.site];
         }
+
+        if (typeof details.schedule === 'boolean') {
+            if (!details.schedule) {
+                // escalate error
+                this._logger.error('error scheduling resume, wrong route path');
```
jonathan-gramain, line 674: That message looks odd, consider removing it

jonathan-gramain (approved):

alexanderchan-scality (approved):

docs/crr-pause-resume.md
philipyoo: I tried to have an optional body param passed, but for this feature, the user story states a default of 6 hours. In order to comply with that, it was necessary to create this other route

PR #344, opened 2018-06-28
improvement: ZENKO-353 Avoid storing all objectMD

bennettbuchanan:
Thanks for your review! 🚀

Make a request to cloud server to get the failed source object's metadata instead of storing it in Redis.

Previously, a Redis key storing all object metadata consumed ~2625 bytes of memory. With this change the key will consume ~115 bytes by only storing the IAM role ARN (which is used to authenticate the GET metadata request to cloud server). 

If a failure occurs each second of the 24 hours (the default key expiry time) this change saves ~216MB per day. Extrapolated to many days and multiple failures per second (in a 1-M scenario) the memory usage saved could be on the order of GBs.

For this purpose, I’ve extracted some common code from `UpdateReplicationStatus` for creating a backbeat client and getting metadata from the source.


extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -30,166 +24,33 @@ class UpdateReplicationStatus extends BackbeatTask {
...
-                return cbOnce(err);
-            }
-            return cbOnce(null, data);
-        });
-    }
-
     _refreshSourceEntry(sourceEntry, log, cb) {
-        this._getMetadata(sourceEntry, log, (err, blob) => {
+        const params = {
+            logInfo: sourceEntry.getLogInfo(),
```
jonathan-gramain: I think it's better not to pass `logInfo` and let the function log whatever it feels good to log. (same for other similar cases)

jonathan-gramain (changes requested):
`logInfo` param to be removed, otherwise LGTM

extensions/replication/tasks/UpdateReplicationStatus.js
bennettbuchanan: I'm not sure about this since that method `getLogInfo` was determining what was being logged before. Do you mean just keep calling `getLogInfo` in the helper method itself rather than passing it as a param?

jonathan-gramain: No, `getLogInfo` is used as a helper whenever an `entry` is passed to log its relevant info, but since we don't pass `entry` anymore there's no point in passing this as a `logInfo` param. I would suggest to just log relevant info directly (say, `bucket` and `objectKey`).

bennettbuchanan: Okay, I see what you mean. I'll update this.

jonathan-gramain:
@bert-e wait

jonathan-gramain (dismissed):

jonathan-gramain (approved):

philipyoo (approved):

PR #347, opened 2018-06-29
Feature/zenko 265 ingestion producer snapshotter

JianqinWang:


rahulreddy (approved):
I recorded the technical debt introduced by this PR (unused code not relevant to bucketd's ingestion). I will follow up with cleanup PRs later.

rahulreddy:
@bert-e help

rahulreddy:
@bert-e bypass_peer_approval

rahulreddy:
@bert-e reset

rahulreddy:
@bert-e force_reset

bin/ingestion.js
```diff
@@ -0,0 +1,100 @@
...
+});
+
+
+async.waterfall([
+    done => ingestionPopulator.open(done),
+    done => {
+        const taskState = {
+            batchInProgress: false,
+        };
+        schedule.scheduleJob(qpConfig.cronRule, () => {
```
jonathan-gramain, line 67: don't you want to use the `extensions.ingestion.cronRule`?

bin/ingestion.js
```diff
@@ -0,0 +1,100 @@
...
+            queueBatch(ingestionPopulator, taskState, qpConfig, log);
+        });
+        done();
+    },
+    done => {
+        healthServer.onReadyCheck(log => {
+            const state = ingestionPopulator.zkStatus();
+            if (state.code === zookeeper.State.SYNC_CONNECTED) {
```
jonathan-gramain: That has been fixed elsewhere, it has to be `SYNC_CONNECTED.code`

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -0,0 +1,364 @@
...
+        this._mProducer = new MetricsProducer(this.kafkaConfig, this.mConfig);
+        this._mProducer.setupProducer(cb);
+    }
+
+    /**
+     * Set up and start the consumer for retrying failed CRR operations.
+     * @param {Function} cb - The callback function
+     * @return {undefined}
+     */
+    _setupFailedCRRClients(cb) {
```
jonathan-gramain, line 132: I think we should not have a failed CRR consumer here (not critical)

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -0,0 +1,364 @@
...
+        this.zkClient.connect();
+        this.zkClient.once('error', done);
+        this.zkClient.once('ready', () => {
+            // just in case there would be more 'error' events emitted
+            this.zkClient.removeAllListeners('error');
+            done();
+        });
+    }
+
+    _loadExtensions() {
```
jonathan-gramain, line 271: We probably don't need to load the extensions for the ingestion daemon, keep for later refactor.

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -0,0 +1,364 @@
...
+                    (result, i) => Object.assign(result, {
+                        logSource: this.logReaders[i].getLogInfo(),
+                        logOffset: this.logReaders[i].getLogOffset(),
+                    }));
+                return done(null, annotatedResults);
+            });
+    }
+
+    processAllLogEntries(params, done) {
+        if (this.logReadersUpdate.length >= 1) {
```
jonathan-gramain, line 347: minor: this special case should not be necessary

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -0,0 +1,364 @@
...
+                    return done(err);
+                }
+                return this._processAllLogEntries(params, done);
+            });
+        }
+        return this._processAllLogEntries(params, done);
+    }
+
+    zkStatus() {
+        return this.zkClient.getState();
```
jonathan-gramain, line 359: since `zkClient` can be `null` before setup it would be useful to check it first.

lib/queuePopulator/IngestionReader.js
```diff
@@ -0,0 +1,233 @@
...
+            logger });
+        super({ zkClient, kafkaConfig, logConsumer, logId: `raft_${raftId}`,
+                logger, extensions, metricsProducer });
+        this.raftId = raftId;
+        this._iProducer = new IngestionProducer(bucketdConfig, qpConfig,
+            s3Config);
+        this.bucketdPrefix = bucketdConfig.prefix ? bucketdConfig.prefix :
+            bucketdConfig.name;
+        this.newIngestion = false;
+        this.removeLogOffset = null;
```
jonathan-gramain: You meant `remoteLogOffset` I guess

lib/queuePopulator/IngestionReader.js
```diff
@@ -0,0 +1,233 @@
...
+            this._extensions.forEach(ext => ext.filter({
+                type: entry.type,
+                bucket: entry.bucket,
+                key: entry.key,
+                value: entry.value,
+            }));
+        } else {
+            if (record.db === 'users..bucket') {
+                const keySplit = entry.key.split('..|..');
+                entry.key =
```
jonathan-gramain, line 95: There seems to be some black magic happening here that could benefit to be clarified (comments, helper functions, use of constants), though I know metadata does some black magic too. Not critical for this PR (but I give up trying to get it now :rofl:)

lib/queuePopulator/IngestionReader.js
```diff
@@ -0,0 +1,233 @@
...
+
+    _processPrepareEntries(batchState, done) {
+        const { entriesToPublish, logRes, logStats } = batchState;
+
+        this._setEntryBatch(entriesToPublish);
+
+        if (this.newIngestion) {
+            logRes.log.forEach(entry => {
+                logStats.nbLogRecordsRead += 1;
+                this._processLogEntry(batchState, entry, entry);
```
jonathan-gramain, line 123: There I don't get why we're passing `entry` for a `record`... if no record, why not `{}`? (as it seems it is used as an object, ideally it would be `null`).

lib/queuePopulator/IngestionReader.js
```diff
@@ -0,0 +1,233 @@
...
+        }, err => {
+            // TODO: On error, produce error metrics
+            if (err) {
+                return done(err);
+            }
+            // Find the CRR Class extension
+            const crrExtension = this._extensions.find(ext => (
+                ext instanceof ReplicationQueuePopulator
+            ));
+            const extMetrics = crrExtension.getAndResetMetrics();
```
philipyoo: I think we should remove metrics here, may duplicate crr metrics since publishing to same extension with same redis keys

lib/queuePopulator/IngestionReader.js
```diff
@@ -0,0 +1,233 @@
...
+
+        if (this.newIngestion) {
+            logRes.log.forEach(entry => {
+                logStats.nbLogRecordsRead += 1;
+                this._processLogEntry(batchState, entry, entry);
+            });
+            return done();
+        }
+        if (logRes.info.start === null || logRes.log === null) {
+            return done(null);
```
jonathan-gramain, line 128: There may be a hidden reason why we're passing `null` here instead of nothing, but eventually it should be rework to avoid having to pass `null` explicitly.

lib/queuePopulator/IngestionReader.js
```diff
@@ -0,0 +1,233 @@
...
+            return done();
+        });
+        return undefined;
+    }
+
+    _processPublishEntries(batchState, done) {
+        const { entriesToPublish, logRes, logStats } = batchState;
+
+        if (this.remoteLogOffset /* &&
+        config.queuePopulator.logSource.indexOf('ingestion') > -1 */) {
```
jonathan-gramain, line 158: comments to be cleaned up

lib/queuePopulator/IngestionReader.js
```diff
@@ -0,0 +1,233 @@
...
+        if (ingestionPath) {
+            this.pathToLogOffset = `${ingestionPath}` +
+                `${bucketdConfig.zookeeperSuffix}` +
+                `/logState/raft_${raftId}/logOffset`;
+        }
+    }
+
+    /* eslint-disable no-param-reassign */
+
+    _processReadRecords(params, batchState, done) {
```
jonathan-gramain, line 39: As a general comment, overriding private functions to get the desired behavior is usually a poor design choice, as the parent class should not care about side-effects of changing its private implementation that could break children classes. If inheritance is to be used (but usually it can be avoided with other means) it's better to define clear interfaces of what functions the child class is allowed to use and what is it allowed to override.

lib/queuePopulator/IngestionReader.js
```diff
@@ -0,0 +1,233 @@
...
+                batchState.publishedEntries[topic] = topicEntries;
+                return done();
+            });
+        }, err => {
+            // TODO: On error, produce error metrics
+            if (err) {
+                return done(err);
+            }
+            // Find the CRR Class extension
+            const crrExtension = this._extensions.find(ext => (
```
jonathan-gramain: Are there any CRR metrics to publish here?

jonathan-gramain (changes requested):
I will take care of the important issue(s) for now, we can keep comments as backlog for cleanups later.

jonathan-gramain (approved):

lib/queuePopulator/IngestionReader.js
jonathan-gramain: fixed

PR #356, opened 2018-07-02
improvement: ZENKO-483 Monitor CRR Upload doc

bennettbuchanan:


docs/metrics.md
```diff
@@ -27,16 +27,22 @@ HTTP code.
 ## Routes
 
 Routes are organized in the following fashion:
-`/_/metrics/<extension-type>/<location-name>/<metric-type>`
+`/_/metrics/<extension-type>/<site-name>/<metric-type>/<bucket>/<key>/<version-id>`
```
rahulreddy: I recommend keeping `location-name` as we are trying to be consistent with our wording.

rahulreddy (changes requested):
Please change the references of `site` with `location`. This is just to be consistent and avoid people mixing up locations/sites/storage_classes etc.

docs/metrics.md
jonathan-gramain: Also maybe enclose in `[` and `]` the set of optional bucket, key and version-id fields which depend on the route.

bennettbuchanan: Good call, this also needs to be updated to take `version-id` as a query param, reflecting recent changes.

rahulreddy (approved):

docs/metrics.md
bennettbuchanan: Updated accordingly.

jeremyds (approved):

PR #358, opened 2018-07-02
feature: ZENKO-483 Monitor CRR upload

bennettbuchanan:
See feature docs: https://github.com/scality/backbeat/pull/356.

### DIFF
--- a/extensions/replication/failedCRR/FailedCRRConsumer.js
+++ b/extensions/replication/failedCRR/FailedCRRConsumer.js
@@ -1,4 +1,5 @@
 'use strict'; // eslint-disable-line strict
+const async = require('async');
 
 const Logger = require('werelogs').Logger;
 const redisClient = require('../../replication/utils/getRedisClient')();
@@ -87,7 +88,53 @@ class FailedCRRConsumer {
             log.end();
             return cb();
         }
-        return this._setRedisKey(data, kafkaEntry, log, cb);
+        return async.series([
+            next => this._removePreexistingKeys(data, err => {
+                if (err) {
+                    this.logger.error('error deleting preexisting CRR object ' +
+                    'metrics redis keys', {
+                        method: 'FailedCRRConsumer.processKafkaEntry',
+                        error: err,
+                    });
+                }
+                return next();
+            }),
+            next => this._setRedisKey(data, kafkaEntry, log, next),
+        ], cb);
+    }
+
+    /**
+     * Delete any pre-existing object CRR metrics keys. This way the object
+     * metrics during the retry will not be affected by the prior attempt.
+     * @param {Object} data - The kafka entry value
+     * @param {Function} cb - callback to call
+     * @return {undefined}
+     */
+    _removePreexistingKeys(data, cb) {
+        // eslint-disable-next-line no-unused-vars
+        const [service, extension, status, bucketName, objectKey, versionId,
+            site] = data.key.split(':');
+        // CRR object-level metrics key schema.
+        const queryString = `${site}:${bucketName}:${objectKey}:${versionId}:` +
+            `${service}:${extension}:*`;
+        return redisClient.scan(queryString, undefined, (err, keys) => {
+            if (err) {
+                return cb(err);
+            }
+            if (keys === 0) {
+                return cb();
+            }
+            return redisClient.batch([['del', ...keys]], (err, res) => {
+                if (err) {
+                    return cb(err);
+                }
+                const [cmdErr] = res[0];
+                if (cmdErr) {
+                    return cb(cmdErr);
+                }
+                return cb();
+            });
+        });
     }
 
     /**
--- a/extensions/replication/queueProcessor/task.js
+++ b/extensions/replication/queueProcessor/task.js
@@ -17,6 +17,7 @@ const kafkaConfig = config.kafka;
 const repConfig = config.extensions.replication;
 const sourceConfig = repConfig.source;
 const redisConfig = config.redis;
+const mConfig = config.metrics;
 const { connectionString, autoCreateNamespace } = zkConfig;
 
 const log = new werelogs.Logger('Backbeat:QueueProcessor:task');
@@ -157,7 +158,7 @@ function initAndStart(zkClient) {
                     if (!activeSites.includes(site)) {
                         const qp = new QueueProcessor(
                             zkClient, kafkaConfig, sourceConfig, destConfig,
-                            repConfig, redisConfig, site);
+                            repConfig, redisConfig, mConfig, site);
                         activeQProcessors[site] = qp;
                         setupZkSiteNode(qp, zkClient, site, (err, data) => {
                             if (err) {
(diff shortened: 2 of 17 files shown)


tests/functional/api/BackbeatServer.js
```diff
@@ -272,6 +272,9 @@ describe('Backbeat Server', () => {
             '/_/metrics/crr/all/completions',
             '/_/metrics/crr/all/failures',
             '/_/metrics/crr/all/throughput',
+            `/_/metrics/crr/${site1}/progress/bucket/object/version`,
```
bennettbuchanan: Need to update routing design to allow for a key with a `/` character.

bennettbuchanan:
@bert-e help

bennettbuchanan:
@bert-e wait

tests/functional/api/BackbeatServer.js
bennettbuchanan: To be updated to use query params instead.

extensions/replication/failedCRR/FailedCRRConsumer.js
```diff
@@ -87,7 +88,57 @@ class FailedCRRConsumer {
             log.end();
             return cb();
         }
-        return this._setRedisKey(data, kafkaEntry, log, cb);
+        return async.series([
+            next => this._removePrexistingKeys(data, err => {
```
jonathan-gramain: minor typo: missing `e` (`_removePreExistingKeys`)

extensions/replication/failedCRR/FailedCRRConsumer.js
```diff
@@ -87,7 +88,57 @@ class FailedCRRConsumer {
...
+                if (err) {
+                    this.logger.error('error deleting prexisting CRR object ' +
```
jonathan-gramain: `preexisting`

Maybe a more accurate (and less "scary") formulation could be "error deleting preexisting CRR metrics key"

extensions/replication/failedCRR/FailedCRRConsumer.js
```diff
@@ -87,7 +88,57 @@ class FailedCRRConsumer {
             log.end();
             return cb();
         }
-        return this._setRedisKey(data, kafkaEntry, log, cb);
+        return async.series([
+            next => this._removePrexistingKeys(data, err => {
+                if (err) {
```
jonathan-gramain: Not sure, if the key does not exist does it return an error? In which case it should be filtered to avoid logging it (as I think this is a normal case).

extensions/replication/queueProcessor/task.js
```diff
@@ -157,14 +158,13 @@ function initAndStart(zkClient) {
...
-                            repConfig, redisConfig, site);
+                            repConfig, redisConfig, mConfig, site);
                         activeQProcessors[site] = qp;
                         setupZkSiteNode(qp, zkClient, site, (err, data) => {
                             if (err) {
                                 return next(err);
                             }
-                            qp.start({ paused: data.paused });
-                            return next();
+                            return qp.start({ paused: data.paused }, next);
```
jonathan-gramain: Why this change since `QueuePopulator.start()` does not appear to take a callback argument

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -359,7 +362,23 @@ class ReplicateObject extends BackbeatTask {
                 return doneOnce(err);
             }
             partObj.setDataLocation(data.Location[0]);
-            return doneOnce(null, partObj.getValue());
+            return async.series([
+                next => {
+                    if (partNumber === 1) {
+                        const extMetrics = getExtMetrics(this.site,
```
jonathan-gramain: Maybe you should consider doing this at the beginning of the process rather than after having put the first part (if possible), so you have the total available as soon as the object starts being put, instead of after the first part is complete. Note that here you can also have some completion metrics written for other parts even before the "queued" key is written, e.g. if part #2 and #3 are completed but not #1 yet. Not sure this has an important impact but just want to point it out, and pushing the "queued" key first would avoid this also.

lib/api/BackbeatAPI.js
```diff
@@ -157,13 +162,29 @@ class BackbeatAPI {
...
+        }
 
         // All routes at this point have extensions property in use
         // Match http verb and match extension name
-        let filteredRoutes = routes.filter(r => (
-            bbRequest.getHTTPMethod() === r.httpMethod &&
-            r.extensions[rDetails.extension]
-        ));
+        let filteredRoutes = routes.filter(r => {
+            const isCorrectMethod = bbRequest.getHTTPMethod() === r.httpMethod;
```
jonathan-gramain: minor naming issue: "correct" may not be the appropriate wording, since we are trying to find a route that "matches" what we provide, it's not really "incorrect" not to match a specific route. Maybe `isMatchingXxx` instead would work better.

jonathan-gramain (changes requested):

extensions/replication/failedCRR/FailedCRRConsumer.js
bennettbuchanan: If the key doesn't exist there is no error, just an empty response. The error would occur if the fetch from Redis failed for some reason (e.g. Redis is offline).

extensions/replication/queueProcessor/task.js
bennettbuchanan: Nice catch, this is left-over from a rebase. I've removed it.

extensions/replication/tasks/ReplicateObject.js
bennettbuchanan: Very good point. I've updated this.

tests/functional/api/BackbeatServer.js
```diff
@@ -542,6 +551,30 @@ describe('Backbeat Server', () => {
...
         });
 
+        it(`should return all metrics for route: /_/metrics/crr/${site1}` +
+            `/progress/${BUCKET_NAME}/${OBJECT_KEY}?versionId=${VERSION_ID}`,
+            done =>
+            getRequest(`/_/metrics/crr/${site1}/progress/${BUCKET_NAME}/` +
+                `${OBJECT_KEY}?versionId=${VERSION_ID}`, (err, res) => {
+                assert.ifError(err);
+                assert(res.description);
+                assert.equal(res.pending, 50);
```
jonathan-gramain: minor: we usually use `strictEqual` to also check the type

jonathan-gramain (approved):

extensions/replication/failedCRR/FailedCRRConsumer.js
```diff
@@ -87,7 +88,57 @@ class FailedCRRConsumer {
...
+     * @return {undefined}
+     */
+    _removePreexistingKeys(data, cb) {
+        // eslint-disable-next-line no-unused-vars
+        const [service, extension, status, bucketName, objectKey, versionId,
+            site] = data.key.split(':');
+        // CRR object-level metrics key schema.
+        const queryString = `${site}:${bucketName}:${objectKey}:${versionId}:` +
+            `${service}:${extension}:*`;
+        return redisClient.batch([['keys', queryString]], (err, res) => {
```
rahulreddy: Using `keys *` pattern is a bad idea to use on a production Redis. Please use `scan` instead.
Reference from redis.io/commands/keys

> Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases.

rahulreddy (changes requested):

extensions/replication/failedCRR/FailedCRRConsumer.js
bennettbuchanan: Good call, scan stream is handy.

rahulreddy (approved):

PR #361, opened 2018-07-03
bugfix: ZENKO-638 Use query params for API route

bennettbuchanan:
For the GET failed CRR API route `/_/crr/failed/<bucket>/<key>/<versionId>` use query params instead to properly handle key names with `/` character.

Route is thus updated to `/_/crr/failed/<bucket>/<key>?versionId=<version-id>`.

### DIFF
--- a/tests/functional/api/BackbeatServer.js
+++ b/tests/functional/api/BackbeatServer.js
@@ -557,7 +557,7 @@ describe('Backbeat Server', () => {
 
         const retryPaths = [
             '/_/crr/failed',
-            '/_/crr/failed/test-bucket/test-key/test-versionId',
+            '/_/crr/failed/test-bucket/test-key?versionId=test-versionId',
         ];
         retryPaths.forEach(path => {
             it(`should get a 200 response for route: ${path}`, done => {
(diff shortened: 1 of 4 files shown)


dora-korpar (approved):

nicolas2bert (approved):

tests/functional/api/BackbeatServer.js
```diff
@@ -794,8 +796,8 @@ describe('Backbeat Server', () => {
                 ];
                 setKey(redisClient, keys, err => {
                     assert.ifError(err);
-                    const route =
-                        '/_/crr/failed/test-bucket/test-key/test-versionId';
+                    const route = '/_/crr/failed?bucket=test-bucket&' +
```
jonathan-gramain: Is there a need to pass bucket & key as query params as well? If not necessary I would recommend to stick with AWS SDK way of doing GETs on objects to avoid adding new conventions when not necessary (which would give `/_/crr/failed/bucket/object?versionId=xyz`)

jonathan-gramain (changes requested):

jonathan-gramain:
@bert-e wait

tests/functional/api/BackbeatServer.js
bennettbuchanan: Good call, will update this.

bennettbuchanan: Updated.

jonathan-gramain (approved):

PR #368, opened 2018-07-09
Bugfix: ZENKO-621 Failed CRR metrics retry

bennettbuchanan:
Depends on https://github.com/scality/Arsenal/pull/531.

If a failed CRR object is being retried, find any keys tracking failed metrics for the site and decrement the bytes failed and ops failed of the latest key. Failed metrics thus accounts for any retried objects.

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -502,6 +502,69 @@ class BackbeatAPI {
     }
 
     /**
+     * Find the matching metric timestamp key and decrement it accordingly.
+     * @param {Array} keys - Array of keys that matched the query string
+     * @param {Number} timestamp - The normalized timestamp of the failed key
+     * @param {Number} decrement - The value to decrement the matching key by
+     * @param {Function} cb - Callback to call
+     * @return {undefined}
+     */
+    _decrementIntervalKey(keys, timestamp, decrement, cb) {
+        if (keys.length === 0) {
+            return process.nextTick(cb);
+        }
+        const intervalKey = keys.find(key => {
+            const schema = key.split(':');
+            return timestamp === schema[schema.length - 1];
+        });
+        // Edge case in which the key was created immediately before the
+        // normalized interval has elapsed (e.g. 11:59:59).
+        if (!intervalKey) {
+            return process.nextTick(cb);
+        }
+        const cmds = [['decrby', intervalKey, decrement]];
+        return this._redis.batch(cmds, (err, res) => {
+            if (err) {
+                return cb(err);
+            }
+            const [cmdErr] = res[0];
+            return cb(cmdErr);
+        });
+    }
+
+    /**
+     * Decrement any existing failed metrics to account for the retry.
+     * @param {Number} timestamp - The normalized timestamp of the failed key
+     * @param {String} site - The site of the object
+     * @param {Number} contentLength - The size of the object
+     * @param {Function} cb - The callback to call
+     * @return {undefined}
+     */
+    _decrementFailedMetrics(timestamp, site, contentLength, cb) {
+        const opsFailQuerystring = `${site}:${redisKeys.opsFail}:*`;
+        const bytesFailQuerystring = `${site}:${redisKeys.bytesFail}:*`;
+        async.parallel([
+            // Decrement ops failed count by one
+            next => this._redis.scan(opsFailQuerystring, undefined,
+                (err, keys) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    return this._decrementIntervalKey(keys, timestamp, 1, next);
+                }),
+            // Decrement bytes failed by the size of the object which had failed
+            next => this._redis.scan(bytesFailQuerystring, undefined,
+                (err, keys) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    return this._decrementIntervalKey(keys, timestamp,
+                        contentLength, next);
+                }),
+        ], cb);
+    }
+
+    /**
      * Process the stringified kafka entries that were stored in Redis, passing
      * each to `_pushToCRRRetryKafkaTopics`. For each kafka entry, construct the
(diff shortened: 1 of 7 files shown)


jonathan-gramain (approved):

JianqinWang (approved):

lib/api/BackbeatAPI.js
```diff
@@ -501,6 +501,50 @@ class BackbeatAPI {
...
+    _decrementFailedMetrics(site, contentLength, cb) {
+        const opsFailQS = `${site}:${redisKeys.opsFail}:*`;
+        const bytesFailQS = `${site}:${redisKeys.bytesFail}:*`;
+        async.series([
+            // Decrement ops failed count by one
+            next => this._redis.scan(opsFailQS, undefined, (err, keys) => {
+                if (keys.length === 0) {
+                    return next();
+                }
+                const key = keys[keys.length - 1];
```
rahulreddy: I don't think order is guaranteed by the SCAN command, it's better to sort on our end to pick the latest key.

rahulreddy (changes requested):

lib/api/BackbeatAPI.js
bennettbuchanan: Was thinking similarly. However I actually realized some additional complexity here: even if decrementing the latest key, the failure could have happened > 20 min ago (beyond the expiry time) so I will have to be sensitive to this fact or else the metrics could go into the negative.

bennettbuchanan: I'm planning to add a normalized timestamp to the failure key so that I can then match the key here, or else not decrement the failures metric.

rahulreddy: Gotcha. Also, duplicates. SCAN can return duplicates.

lib/api/BackbeatAPI.js
```diff
@@ -560,17 +615,21 @@ class BackbeatAPI {
...
-                // If the key did not exist, value is `null`.
-                if (sourceRole !== null) {
-                    entries.push(new ObjectFailureEntry(key, sourceRole));
-                }
-                return next(cmdErr);
+            return this._redis.scan(`${key}:*`, undefined, (err, keys) => {
+                // There can only be one failed key. We are only using the glob
+                // because we do not know the original timestamp.
+                const [key] = keys;
+                return this._redis.batch([['get', key]], (err, res) => {
```
rahulreddy: Why use batch when you are only accessing `get`? Shouldn't `this._redis.get` just work?

lib/api/BackbeatAPI.js
```diff
@@ -501,6 +501,56 @@ class BackbeatAPI {
...
+    _decrementIntervalKey(keys, timestamp, decrement, cb) {
+        if (keys.length === 0) {
+            return process.nextTick(cb);
+        }
+        const intervalKey = keys.find(key => {
+            const schema = key.split(':');
+            return timestamp === schema[schema.length - 1];
+        });
+        const cmds = [['decrby', intervalKey, decrement]];
+        return this._redis.batch(cmds, (err, res) => {
```
rahulreddy: `this._batch.decrby`, this should work.

lib/api/BackbeatAPI.js
```diff
@@ -501,6 +501,56 @@ class BackbeatAPI {
...
+     */
+    _decrementIntervalKey(keys, timestamp, decrement, cb) {
+        if (keys.length === 0) {
+            return process.nextTick(cb);
+        }
+        const intervalKey = keys.find(key => {
+            const schema = key.split(':');
+            return timestamp === schema[schema.length - 1];
+        });
+        const cmds = [['decrby', intervalKey, decrement]];
```
rahulreddy: What happens when no key is found here?

lib/api/BackbeatAPI.js
```diff
@@ -501,6 +501,56 @@ class BackbeatAPI {
...
+     * @param {Function} cb - The callback to call
+     * @return {undefined}
+     */
+    _decrementFailedMetrics(timestamp, site, contentLength, cb) {
+        const opsFailQuerystring = `${site}:${redisKeys.opsFail}:*`;
+        const bytesFailQuerystring = `${site}:${redisKeys.bytesFail}:*`;
+        async.parallel([
+            // Decrement ops failed count by one
+            next => this._redis.scan(opsFailQuerystring, undefined,
+                (err, keys) =>
```
rahulreddy: Not a good idea to ignore the `err` here

lib/api/BackbeatAPI.js
```diff
@@ -501,6 +501,56 @@ class BackbeatAPI {
...
+                    this._decrementIntervalKey(keys, timestamp, 1, next)),
+            // Decrement bytes failed by the size of the object which had failed
+            next => this._redis.scan(bytesFailQuerystring, undefined,
+                (err, keys) =>
```
rahulreddy: Same, please don't ignore the `err`

rahulreddy (changes requested):

lib/api/BackbeatAPI.js
bennettbuchanan: `this._redis` is an instance of the `RedisClient` wrapper imported from Arsenal and it only exposes certain methods. It would be cleaner to add those wrapper methods though that would work with `ioredis`. Perhaps this could be a follow-up refactor?

lib/api/BackbeatAPI.js
bennettbuchanan: Same as above: this is the `RedisClient` wrapper from Arsenal, not `ioredis`.

lib/api/BackbeatAPI.js
bennettbuchanan: Good point. Ideally this could never happen, but I think there could be a case where the key is set right before the interval time and this would end up as `undefined`. I've added a check for this, and a comment.

PR #373, opened 2018-07-10
improvement/ZENKO-647-backbeat_API_Routes_Doc

wabernat:
Changed routes to backbeat paths for metrics.md. Copy edited as well.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,6 +1,6 @@
 # Zenko Backbeat Metrics
 
-Zenko Backbeat exposes various metric routes which return a response with an
+Zenko Backbeat exposes various metric routes that return a response with an
 HTTP code.
 
 ## Response Codes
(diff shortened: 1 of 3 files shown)


docs/metrics.md
```diff
@@ -26,37 +26,36 @@ HTTP code.
 
 ## Routes
 
-Routes are organized in the following fashion:
-`/_/metrics/<extension-type>/<location-name/[<metric-type>]/[<bucket>]/[<key>]?[versionId=<version-id>]`
+Routes are organized as follows:
+`/_/backbeat/api/<extension-type>/<location-name/[<metric-type>]/[<bucket>]/[<key>]?[versionId=<version-id>]`
```
jonathan-gramain: We have to keep the `/metrics` after `api`

docs/metrics.md
```diff
@@ -26,37 +26,36 @@ HTTP code.
...
+   metrics for the given extension and location. 
+- `<bucket>` is an optional field. It carries the name of the bucket in which
+   the object is expected to exist.
+- `<key>` is an optional field. When getting CRR metrics for a particular object,
+   it contains the object's key. 
+- `<version-id>` is an optional field. When getting CRR metrics for a particular
+   object, it contains the object's version ID.
 
-### `/_/metrics/crr/<location-name>`
+### `/_/backbeat/api/<location-name>`
```
jonathan-gramain: Add `/metrics/crr` between `api` and `/<location-name>` (same for the following occurrences).

jonathan-gramain (changes requested):

jonathan-gramain:
@wabernatScality can you change the branch name to e.g. `improvement/ZENKO-647-backbeatAPIRoutesDoc`? It has to contain the ticket name to have an automatic link from Jira ticket, and the `improvement/` prefix is there to allow bert-e to merge it as it is one of its recognized prefixes (AFAIK `feature/`, `bugfix/` and `improvement/` can be used).

docs/metrics.md
wabernat: Thanks!

wabernat:
Complied (I hope!) with Jonathan's requests.

jonathan-gramain (approved):

docs/metrics.md
```diff
@@ -26,37 +26,36 @@ HTTP code.
 
 ## Routes
 
-Routes are organized in the following fashion:
-`/_/metrics/<extension-type>/<location-name/[<metric-type>]/[<bucket>]/[<key>]?[versionId=<version-id>]`
+Routes are organized as follows:
+`/_/backbeat/api/metrics/crr/<extension-type>/<location-name/[<metric-type>]/[<bucket>]/[<key>]?[versionId=<version-id>]`
```
bennettbuchanan, line 30: 'crr' is an `<extension-type>` so we don't want to add this here while also keeping the `<extension-type>` placeholder. Alternatively, we could remove `/<extension-type>/` since 'crr' is only supported in the API (but may be extended later). Either way is good for me 😸

bennettbuchanan (changes requested):
Nice edits! Just one comment about a change in the route.

docs/metrics.md
wabernat: Let's not close any doors! <extension-type> it is.

wabernat: By which I meant to say: "extension-type it is." or somesuch...

PR #375, opened 2018-07-11
S3C-1550 zk provisioning race condition

jonathan-gramain:
Notes:

- reviewing each commit individually may be easier, as tests commit introduces mostly indentation changes

- the fix applies to rel/7.2 as well, but it seems wiser to backport it if the need arises later, so that we can get the fix quickly in 7.4 branch (and not sure if any customer using 7.2 + backbeat and willing the fix exists).

### DIFF
--- a/lib/provisioning/ProvisionDispatcher.js
+++ b/lib/provisioning/ProvisionDispatcher.js
@@ -46,6 +46,8 @@ class ProvisionDispatcher {
         this._isLeader = false;
         this._owners = null;
         this._provisions = null;
+        this._redispatchInProgress = false;
+        this._redoRedispatch = false;
 
         this._log = new Logger('Backbeat:ProvisionDispatcher');
     }
@@ -151,6 +153,13 @@ class ProvisionDispatcher {
             this._owners.length === 0) {
             return undefined;
         }
+        if (this._redispatchInProgress) {
+            // queue up a redispatch after the current one completes
+            // to avoid race conditions with owner changes
+            this._redoRedispatch = true;
+            return undefined;
+        }
+        this._redispatchInProgress = true;
         // dispatch provisions to owners
         this._owners.forEach(owner => {
             provisionsByOwner[owner] = [];
@@ -187,8 +196,15 @@ class ProvisionDispatcher {
             );
         };
         return this._withRandomDelayIfTest(
-            () => async.eachLimit(Object.keys(provisionsByOwner), 20,
-                                  _provisionOwner));
+            () => async.eachLimit(
+                Object.keys(provisionsByOwner), 20, _provisionOwner, () => {
+                    this._redispatchInProgress = false;
+                    if (this._redoRedispatch) {
+                        this._log.debug('redoing provisions redispatch');
+                        this._redoRedispatch = false;
+                        this._redispatchProvisions();
+                    }
+                }));
     }
 
     _registerLeader(cb) {
--- a/lib/provisioning/ProvisionDispatcher.js
+++ b/lib/provisioning/ProvisionDispatcher.js
@@ -7,6 +7,8 @@ const zookeeper = require('node-zookeeper-client');
 
 const Logger = require('werelogs').Logger;
 
+const runningCI = process.env.CI === 'true';
+
 function _genRandomHexChars(len) {
     return crypto.randomBytes(Math.ceil(len / 2))
         .toString('hex') // convert to hexadecimal format
@@ -48,6 +50,14 @@ class ProvisionDispatcher {
         this._log = new Logger('Backbeat:ProvisionDispatcher');
     }
 
+    _withRandomDelayIfTest(cb) {
+        if (runningCI) {
+            setTimeout(cb, Math.random() * 100);
+        } else {
+            cb();
+        }
+    }
+
     /**
      * Add new provisions to the list of provisions to be dispatched
      *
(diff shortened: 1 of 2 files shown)


tmacro (approved):

rahulreddy:
@bert-e wait

eve/main.yml
```diff
@@ -37,4 +37,4 @@ stages:
           workdir: '%(prop:builddir)s/build'
       - ShellCommand:
           name: run feature tests
-          command: npm run ft_test
+          command: TEST_SWITCH=1 npm run ft_test
```
ssalaues: You can also add `TEST_SWITCH` to the environment.
Something like
```
       - ShellCommand:
            name: run feature tests
            command: npm run ft_test
            env:
              TEST_SWITCH: "1"
```

rahulreddy: You can introduce env `CI=true` globally. This will be consistent with CircleCI for opensource. See https://github.com/scality/S3/blob/development/8.0/eve/main.yml for reference.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -7,6 +7,8 @@ const zookeeper = require('node-zookeeper-client');
 
 const Logger = require('werelogs').Logger;
 
+const testIsOn = process.env.TEST_SWITCH === '1';
```
rahulreddy: Same thing, relying on `CI=true` is better as it's consistent with CI's that support OSS.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -242,103 +281,111 @@ class ProvisionDispatcher {
...
+                        return cb(err);
+                    }
+                    if (data !== undefined) {
+                        const provisionList = JSON.parse(data);
+                        this._log.info('provisioning update',
+                                       { zkPath: this._zkEndpoint + myPath,
+                                         provisionList });
+                        return cb(null, provisionList);
+                    }
+                    return undefined;
```
rahulreddy, line 361: Is there a case where `data` is defined and we will not be returning the callback?

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -242,103 +281,111 @@ class ProvisionDispatcher {
...
                 }
-                return undefined;
-            }
+            )
         );
     }
 
     _onProvisionChange() {
```
rahulreddy, line 367: Why not add callbacks to all methods which are `async` instead of adding timers to manage race? I feel like all the race is happening because async methods are being treated as sync.

rahulreddy (changes requested):

eve/main.yml
jonathan-gramain: I was actually relying on an existing variable, but I agree a global CI variable would make sense.

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: Do you mean if `data` is undefined? In such case, we don't call the callback and that's expected because there is no change in the provisioning.

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: Not clear for me what timer you're referring to (just in case the `_withRandomDelayIfTest` wrapper only introduces delays for testing purpose). In a nutshell, this is an event handler, not an async function, so there's no callback for this reason. This is called when zookeeper notifies us of a change in the nodes.

The race that I fixed was due to two events happening but the first event handler leading to call `_redispatchProvisions()` had to restart its processing due to a change coming from another event handler (`_onOwnersChange()`), but the first processing may be still writing stale state while the second one triggers, hence introducing a race condition. Making sure we retrigger `_redispatchProvisions` after the current one ends when new events occurred in the meantime solves this race.

eve/main.yml
jonathan-gramain: Actually the way I did it matches what has been introduced in development/8.0 branch. I can nevertheless do this change also on the integration branch.

lib/provisioning/ProvisionDispatcher.js
rahulreddy: Ah yeah undefined. If we expect to hit that case, we should return callback for that case, an empty one should be fine.

lib/provisioning/ProvisionDispatcher.js
rahulreddy: I meant methods like these `_onProvisionChange`, `_redispatchProvisions`, `_onLeaderChange`,`_onOwnerChange`, `_onProvisionChange`. All of these methods are asynchronous in nature and they are being called as if they are synchronous. This could lead to unmanageable race(?)

jonathan-gramain: All those functions are triggered from event handlers of zookeeper. (the `cb` argument of `_onMyselfChange()` is not a callback but an event handler function as well that can be called multiple times). So I don't see in which respect adding a callback argument to them would help managing race conditions, as any of those functions can be called at *any* time from zookeeper notifications. What we care about here should be taken care of, which is, making sure we redispatch all provisions *after* the latest notification that has an impact on the provisioning has been received, at which point what we redispatch is guaranteed to be the latest state.

Actually I started adding a callback to `_redispatchProvisions()`, so that I could embed it in to an `async.queue` to serialize all calls, but I realized this is extra work since we only want the leader to redispatch everything *once* after the latest event has been received, not once per every single event that has been received. So I removed the queue in favor of a set of flags to know if a redispatch is in progress or not.

lib/provisioning/ProvisionDispatcher.js
jonathan-gramain: No we don't want to call the callback because there's no change in provisioning (again this is not a regular callback, rather an event notification system).

lib/provisioning/ProvisionDispatcher.js
rahulreddy: Approving this after our offline discussion.

rahulreddy (approved):

jonathan-gramain:
@bert-e wait

jonathan-gramain:
@bert-e reset

PR #380, opened 2018-07-12
FIX: ZENKO-701 CRR stats error ":preferred_read" issue

nicolas2bert:
Remove ":preferred_read" suffix from Kafka metrics entry site

### DIFF
--- a/lib/MetricsConsumer.js
+++ b/lib/MetricsConsumer.js
@@ -61,6 +61,10 @@ class MetricsConsumer {
             this.logger.info('metrics processor is ready to consume entries');
         });
     }
+    _removePreferredRead(site) {
+        return site.endsWith(':preferred_read') ?
+            site.split(':')[0] : site;
+    }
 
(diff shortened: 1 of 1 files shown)


lib/MetricsConsumer.js
```diff
@@ -124,6 +128,9 @@ class MetricsConsumer {
             log.end();
             return done();
         }
+        if (data.site) {
+            data.site = this._removePreferredRead(data.site);
```
jonathan-gramain, line 132: I don't think this is the right place to do this processing, the metrics topic messages should not include preferred read info as it's not related. Let me figure where could be a better place.

nicolas2bert: ok, thanks @jonathan-gramain

nicolas2bert:
As discussed with @jonathan-gramain, this issue has been fixed by https://github.com/scality/S3/pull/1352/files

PR #381, opened 2018-07-12
bugfix: ZENKO-623 Use a ranged GET for MPU replication

bennettbuchanan:
Use ranged gets for the source object instead of using a part number. A similar change needs to be applied for CRR to a zenko destination in a separate, refactor PR to follow.

Rationale:

When the source object is stored in an external backend the location array does not include all the part information. Instead it stores the parts as a single element in the array (for example, [here](https://github.com/scality/S3/blob/development/8.0/lib/api/completeMultipartUpload.js#L211)):

```
[ { key: 'source-key',
    start: 0,
    size: 10485782,
    dataStoreName: 'source-location',
    dataStoreETag: 'dbff4c86d9f3daa594119662fb8e94b1-2' } ]
```

When this occurs, the multiple backend task attempts to retrieve the part number from the `dataStoreETag` value [here](https://github.com/scality/Arsenal/blob/development/8.0/lib/models/ObjectMDLocation.js#L46), but since the part numbers were never stored with the splitter `:`, we unintentionally send `NaN` as the part number resulting in the following logs:

```
{"name":"Backbeat:Replication:QueueProcessor","method":"MultipleBackendTask._putMPUPart","entry":{"bucket":"mehmeh","objectKey":"Transmit 5.0.5.zip","versionId":"98469702552383999999RG001 23.1","isDeleteMarker":""},"origin":"source","peer":{"host":"zenko-cloudserver-replicator","port":8000},"error":"Part number must be a number.","httpStatus":400,"time":1530297452207,"req_id":"ccdf12d7104ba92bdb98","level":"error","message":"an error occurred on getObject from S3","hostname":"backbeat-consumer-5f768fd94c-n9jv9","pid":27}
```

While the object is replicated properly, this change has the affect that the destination ETag will differ from the source because the MPU parts are composed of different ranges.

This change means that we create MPUs with part counts between 2 and 10,000. The sizes target a part count between 500-1000 parts for larger objects, starting with 16MB part sizes and increasing exponentially from there up to 1GB part sizes.

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -101,17 +101,6 @@ class MultipleBackendTask extends ReplicateObject {
         });
     }
 
-    _getAndPutMPUPart(sourceEntry, destEntry, part, uploadId, log, cb) {
-        this.retry({
-            actionDesc: 'stream part data',
-            logFields: { entry: sourceEntry.getLogInfo() },
-            actionFunc: done => this._getAndPutMPUPartOnce(sourceEntry,
-                destEntry, part, uploadId, log, done),
-            shouldRetryFunc: err => err.retryable,
-            log,
-        }, cb);
-    }
-
     /**
      * This is a retry wrapper for calling _getRangeAndPutMPUPartOnce.
      * @param {ObjectQueueEntry} sourceEntry - The source object entry
@@ -139,7 +128,7 @@ class MultipleBackendTask extends ReplicateObject {
      * Get the ranged object, calculate the data's size, then put the part.
      * @param {ObjectQueueEntry} sourceEntry - The source object entry
      * @param {ObjectQueueEntry} destEntry - The destination object entry
-     * @param {Object} range - The range to get an object with
+     * @param {Object} [range] - The range to get an object with
      * @param {Number} range.start - The start byte range
      * @param {Number} range.end - The end byte range
      * @param {Number} partNumber - The part number for the current part
@@ -159,11 +148,12 @@ class MultipleBackendTask extends ReplicateObject {
             Bucket: sourceEntry.getBucket(),
             Key: sourceEntry.getObjectKey(),
             VersionId: sourceEntry.getEncodedVersionId(),
-            Range: `bytes=${range.start}-${range.end}`,
+            // A 0-byte part has no range.
+            Range: range && `bytes=${range.start}-${range.end}`,
             LocationConstraint: sourceEntry.getDataStoreName(),
         });
-        // Range is inclusive, hence + 1.
-        const size = range.end - range.start + 1;
+        // A 0-byte object has no range, otherwise range is inclusive.
+        const size = range ? range.end - range.start + 1 : 0;
         return this._putMPUPart(sourceEntry, destEntry, sourceReq, size,
             uploadId, partNumber, log, doneOnce);
     }
@@ -201,7 +191,7 @@ class MultipleBackendTask extends ReplicateObject {
                 // eslint-disable-next-line no-param-reassign
                 err.origin = 'source';
                 log.error('an error occurred on completing MPU to S3', {
-                    method: 'MultipleBackendTask._getAndPutMultipartUploadOnce',
+                    method: 'MultipleBackendTask._completeMPU',
                     entry: destEntry.getLogInfo(),
                     origin: 'target',
                     peer: this.destBackbeatHost,
@@ -295,33 +285,6 @@ class MultipleBackendTask extends ReplicateObject {
         });
     }
 
-    /**
-     * Get an object by its part number, then put the part.
-     * @param {ObjectQueueEntry} sourceEntry - The source object entry
-     * @param {ObjectQueueEntry} destEntry - The destination object entry
-     * @param {Object} part - The single data location info
-     * @param {String} uploadId - The upload ID of the initiated MPU
-     * @param {Werelogs} log - The logger instance
-     * @param {Function} done - The callback to call
-     * @return {undefined}
-     */
-    _getAndPutMPUPartOnce(sourceEntry, destEntry, part, uploadId, log, done) {
-        log.debug('getting object part', { entry: sourceEntry.getLogInfo() });
-        const doneOnce = jsutil.once(done);
-        const partObj = new ObjectMDLocation(part);
-        const sourceReq = this.backbeatSource.getObject({
-            Bucket: sourceEntry.getBucket(),
-            Key: sourceEntry.getObjectKey(),
-            VersionId: sourceEntry.getEncodedVersionId(),
-            PartNumber: partObj.getPartNumber(),
-            LocationConstraint: sourceEntry.getDataStoreName(),
-        });
-        const size = partObj.getPartSize();
-        const partNumber = partObj.getPartNumber();
-        return this._putMPUPart(sourceEntry, destEntry, sourceReq, size,
-            uploadId, partNumber, log, doneOnce);
-    }
-
     _getAndPutMultipartUpload(sourceEntry, destEntry, part, uploadId, log, cb) {
         this.retry({
             actionDesc: 'stream part data',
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
@bert-e wait

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -372,14 +334,30 @@ class MultipleBackendTask extends ReplicateObject {
...
+     * two (aside from the last part).
      * @param {Number} contentLen - The content length of the whole object
+     * @param {Boolean} isGCP - Whether the object is being replicated to GCP
      * @return {Array} The array of byte ranges.
      */
-    _getRanges(contentLen) {
-        // Creates ranges that have sizes that are powers of 2.
+    _getRanges(contentLen, isGCP) {
+        // 5MB min part size rounded up to the nearest power of two is 8388608.
+        let partCount = Math.pow(2,
```
rahulreddy: This works for now but we need a more efficient way so that we can maximize the usage of 10K parts.

rahulreddy (approved):

extensions/replication/tasks/MultipleBackendTask.js
jonathan-gramain: Not sure what is the benefit of ensuring the number of parts is a power of two? Something that may benefit from being a power of two instead (or being divisible by a sufficiently large power-of-two number) could be the number of bytes per part, since a lot of layers in our systems tend to perform better when processing data aligned with disk or memory pages, generally speaking.

Usually using a fixed default range size that can be optionally tuned works well in practice, and we should increase the range size if needed so that we stay under 10K ranges.

bennettbuchanan: I like this idea @jonathan-gramain, I'll work on implementing that.

bennettbuchanan: I've updated so that it creates MPUs with part counts between 2 and 10,000. It targets a part count between 500-1000 parts for larger objects, starting with 16MB part sizes and increasing exponentially from there up to 512MB part sizes, at which point it uses some 1GB parts to get to the max object size of 5TB.

JianqinWang (approved):

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -332,6 +332,44 @@ class MultipleBackendTask extends ReplicateObject {
...
+     * that the range size sums to the content length when multiplied by a value
+     * between 2 and 1000, excepting a very large object. This has the effect
+     * of creating MPU parts of the given range size.
+     * @param {Number} contentLen - The content length of the whole object
+     * @return {Number} The range size to use
+     */
+    _getRangeSize(contentLen) {
+        let rangeSize = (1024 * 1024) * 16; // Default 16MB part size.
+        if (contentLen <= rangeSize) {
+            rangeSize = (1024 * 1024) * 5; // AWS minimum MPU part size.
```
jonathan-gramain: I don't think you need that special case, if one wants to create an MPU of 10MB he could get that MPU with just one part of 10MB? If you think it's important to have multiple parts, then you would also need to take care of object sizes between 16MB to 32 or 48MB as well to give them a "reasonable" number of parts, but I don't think it's necessary.

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -332,6 +332,44 @@ class MultipleBackendTask extends ReplicateObject {
...
+        }
+        // Target creation of an MPU that is between a 2 and 1000 parts.
+        while (contentLen / rangeSize > 1000) {
```
jonathan-gramain: I'm good with the logic of this loop

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -343,26 +381,21 @@ class MultipleBackendTask extends ReplicateObject {
...
         let start = 0;
         let end = 0;
         while (end < contentLen - 1) {
-            end = start + range - 1;
+            // If the object is large enough to reach 9760 parts of 512MB, then
+            // at this point we need to increase the part size to 1GB, such that
+            // the largest object size of 5TB can be accounted for by performing
+            // a 10,000 part multipart upload. That is, 9760 parts of size 512MB
+            // and 240 parts of size 1GB is 5TB.
+            if (ranges.length === 9760 && size === 512 * (1024 * 1024)) {
```
jonathan-gramain: I think this logic is brittle and not really needed, I would rather have the same part size for all to avoid surprises. Hence if you need 1GB parts because of the object being close to 5TB, you could add this in `_getRangeSize` before `return rangeSize`:

```js
// make sure we do not need more than 10K parts for the biggest allowable objects
while (contentLen / rangeSize > 10000) {
    rangeSize *= 2;
}
```

jonathan-gramain (changes requested):

extensions/replication/tasks/MultipleBackendTask.js
bennettbuchanan: I'll have to double check this, but I think you need at least two parts to complete an MPU.

extensions/replication/tasks/MultipleBackendTask.js
bennettbuchanan: Great, I like the simplification! Potentially using up to 10,000 parts of 512MB each and then increasing the part size from there seems good enough for me.

extensions/replication/tasks/MultipleBackendTask.js
jonathan-gramain: In that case, ok for me

bennettbuchanan: I didn't realize this: it is possible to have an MPU of 1 part, and, as the last part, it can be < 5MB. I've updated to just return the content length as the range size if <= 16MB.

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -381,6 +383,9 @@ class MultipleBackendTask extends ReplicateObject {
      * @return {Array} The array of byte ranges.
      */
     _getRanges(contentLen, isGCP) {
+        if (contentLen === 0) { // 0-byte object has no range.
+            return [undefined];
```
jonathan-gramain: why not `return []`? is it a required special case for 0-byte object?

bennettbuchanan: So that it still puts the part (i.e. the `async.timesLimit` is run once) but the range is not included in the GET request [here](https://github.com/scality/backbeat/pull/381/files#diff-bc4e752f356b967b5c1e8b8b77266bb2R152).

jonathan-gramain: OK, I think it's useful to have this explanation in the comment above, could you add it? Otherwise good for me.

jonathan-gramain (approved):

extensions/replication/tasks/MultipleBackendTask.js
bennettbuchanan: Added a comment for this, thanks!

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -370,21 +333,73 @@ class MultipleBackendTask extends ReplicateObject {
...
      */
-    _getRanges(contentLen) {
-        // Creates ranges that have sizes that are powers of 2.
-        const pow = Math.pow(2, Math.ceil(Math.log(contentLen) / Math.log(2)));
-        const range = Math.ceil(pow / MPU_GCP_MAX_PARTS);
+    _getRanges(contentLen, isGCP) {
+        if (contentLen === 0) {
+            // 0-byte object has no range. However we still want to put a single
+            // part so the range in the subsequent GET request is undefined.
+            return [undefined];
```
rahulreddy: One small change. Can we use `null` here instead?

rahulreddy (changes requested):

extensions/replication/tasks/MultipleBackendTask.js
bennettbuchanan: Updated.

rahulreddy (approved):

bennettbuchanan:
Build failed due to outdated node modules (i.e. I needed to force rebuild eve after rebasing).

PR #389, opened 2018-07-20
bugfix: ZENKO-793 GCP MPU range calculation

bennettbuchanan:
When uploading a small object to GCP, the `_getGCPRangeSize` was being used outright and could potentially calculate a high number of small parts. This is inefficient and potentially results in an invalid part size upon completion of the MPU. Instead, we want to first use `_getRangeSize` to ensure the minimum part size is met and then only use `_getGCPRangeSize` when the part count exceeds `MPU_GCP_MAX_PARTS` (i.e. 1024 parts).

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -341,8 +341,12 @@ class MultipleBackendTask extends ReplicateObject {
      * @return {Number} The range size to use
      */
     _getGCPRangeSize(contentLen) {
-        const pow = Math.pow(2, Math.ceil(Math.log(contentLen) / Math.log(2)));
-        const rangeSize = Math.ceil(pow / MPU_GCP_MAX_PARTS);
+        let rangeSize = this._getRangeSize(contentLen);
+        if (contentLen / rangeSize > MPU_GCP_MAX_PARTS) {
+            const pow =
+                Math.pow(2, Math.ceil(Math.log(contentLen) / Math.log(2)));
+            rangeSize = Math.ceil(pow / MPU_GCP_MAX_PARTS);
+        }
         return rangeSize;
     }
 
--- a/tests/unit/replication/MultipleBackendTask.js
+++ b/tests/unit/replication/MultipleBackendTask.js
@@ -180,6 +180,21 @@ describe('MultipleBackendTask', () => {
             });
         });
 
+        it('should get single part count for GCP', () => {
+            const contentLength = (1024 * 1024) * 5;
+            const ranges = task._getRanges(contentLength, true);
+            assert(ranges.length === 1);
+        });
+
+        it('should use GCP calculation for ranges exceeding 512MB * 1024',
+        () => {
+            const contentLength = ((1024 * 1024) * 512) * 1024;
+            let ranges = task._getRanges(contentLength, true);
+            assert(ranges.length === 1024);
+            ranges = task._getRanges(contentLength + 1, true);
+            assert(ranges.length === 513);
+        });
+
         it('should get <= 1024 ranges for part count 1025-10000', () => {
             const partSize = 1024 * 1024 * 1024 + 1;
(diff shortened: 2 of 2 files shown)


extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -392,8 +392,10 @@ class MultipleBackendTask extends ReplicateObject {
             // part so the range in the subsequent GET request is undefined.
             return [null];
         }
-        const size = isGCP ?
-            this._getGCPRangeSize(contentLen) : this._getRangeSize(contentLen);
+        let size = this._getRangeSize(contentLen);
+        if (isGCP && (contentLen / size > MPU_GCP_MAX_PARTS)) {
```
jonathan-gramain: Although I'm good with the logic, I think it would be cleaner if this logic goes in `_getGCPRangeSize()` (which then would call `_getRangeSize()`)

jonathan-gramain (changes requested):

alexanderchan-scality (approved):

extensions/replication/tasks/MultipleBackendTask.js
bennettbuchanan: Sounds good! I've updated accordingly and added a unit test for that case in `_getRangeSize`.

philipyoo (approved):

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -375,6 +376,9 @@ class MultipleBackendTask extends ReplicateObject {
         while (contentLen / rangeSize > 10000) {
             rangeSize *= 2;
         }
+        if (isGCP && (contentLen / rangeSize > MPU_GCP_MAX_PARTS)) {
+            rangeSize = this._getGCPRangeSize(contentLen);
```
jonathan-gramain: Sorry to insist, but wouldn't it be clearer if `_getGCPRangeSize` was computing the GCP range as a whole instead of being valid only for ranges bigger than GCP limit? Which brings to calling the generic `_getRangeSize` from inside, leaving the GCP logic to the GCP specialized function.

btw, "no" is a valid answer, just my personal though ;)

bennettbuchanan: I see your point. I think this inaccuracy is coming from the fact that we've refactored the design to use ranges always and have kept the `_getGCPRangeSize` function as-is. Perhaps, as a simple solution we could rename `_getGCPRangeSize` to something like `_largeGCPRangeSize`, or even just remove that method and move the code into `_getRangeSize`. What do you think?

jonathan-gramain: Helpers can be reworked if it helps :)

Here's the outline of how I think the code would gain in clarity:

```js

function _getGenericRangeSize(contentLen) {
    // nothing related to GCP here
    return someSize;
}

function _getGCPRangeSize(contentLen) {
    range = _getGenericRangeSize(contentLen);
    if (/* range too big for GCP */) {
        // put here the GCP logic for ranges too big for GCP with standard computation
    } else {
        return range;
    }
}

if (isGCP) {
    _getGCPRangeSize(contentLen);
} else {
    _getGenericRangeSize(contentLen);
}
```

The general idea is that we don't have to know that `_getGCPRangeSize` is not supposed to be called only with bigger ranges, as the logic to decide what is a big range is itself in the function. Now I may miss some logic which makes this not practical, or maybe it does not matter so much, so feel free to ignore!

bennettbuchanan: Okay, that makes sense. We can view the `_getRangeSize` as the "generic" one and `_getGCPRangeSize` as the special case. I prefer less boolean params being passed around so this works for me. 

Updated.

jonathan-gramain (approved):
code porn :rofl:

rahulreddy (approved):

PR #391, opened 2018-07-23
bf: ZENKO-812 fix LifecycleTask tags matching

jonathan-gramain:
Two important fixes in tags matching:

- do not crash when object has a no TagSet at all and a tag matching
  rule nesting tags with 'And'

- fix matching for objects with no TagSet at all (was selecting all
  such objects before when a tag-matching rule was present)

Side improvements:

- remove duplicate unit test

- rework code flow of LifecycleTask_filterRules() to avoid duplicating
  check logic

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -385,22 +385,22 @@ class LifecycleTask extends BackbeatTask {
             if (rule.Status === 'Disabled') {
                 return false;
             }
-            if (rule.Filter && rule.Filter.And) {
-                // multiple tags or prefix w/ tag(s)
-                const prefix = rule.Filter.And.Prefix;
-                if (prefix && !item.Key.startsWith(prefix)) {
-                    return false;
-                }
-                return deepCompare(rule.Filter.And.Tags, objTags.TagSet);
-            }
-            // only one prefix or one tag to compare in this rule
+            // check all locations where prefix could possibly be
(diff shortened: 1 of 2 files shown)


tmacro (approved):

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -385,22 +385,22 @@ class LifecycleTask extends BackbeatTask {
...
-                if (prefix && !item.Key.startsWith(prefix)) {
-                    return false;
-                }
-                return deepCompare(rule.Filter.And.Tags, objTags.TagSet);
-            }
-            // only one prefix or one tag to compare in this rule
+            // check all locations where prefix could possibly be
             const prefix = rule.Prefix ||
-                (rule.Filter && rule.Filter.Prefix);
+                  (rule.Filter && (rule.Filter.And ?
```
rahulreddy, line 390: ~~Isn't the second part of the conditional returning a boolean and not the actual prefix?~~ 
nvm, I see that logical operators are in play

jonathan-gramain: If you talk specifically about `rule.Filter.And ? .. : ..`, it's the ternary operator that returns either the 2nd or 3rd operand depending on the evaluation of the 1st, so not a boolean.

rahulreddy: Thanks the `&&` throw me off

rahulreddy (approved):

PR #398, opened 2018-07-31
bf: ZENKO-644 auto apply EODM for lifecycle service account

philipyoo:
When a valid Expiration rule exists and the current owner
is the lifecycle service account, automatically consider
the ExpiredObjectDeleteMarker rule if it hasn't already
been explicitly declared from another lifecycle rule.

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -13,6 +13,12 @@ const MAX_KEYS = process.env.CI === 'true' ? 3 : 1000;
 // concurrency mainly used in async calls
 const CONCURRENCY_DEFAULT = 10;
 
+function isLifecycleUser(canonicalID) {
+    const canonicalIDArray = canonicalID.split('/');
+    const serviceName = canonicalIDArray[canonicalIDArray.length - 1];
+    return serviceName === 'lifecycle';
+}
+
 class LifecycleTask extends BackbeatTask {
     /**
      * Processes Kafka Bucket entries and determines if specific Lifecycle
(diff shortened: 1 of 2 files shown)


philipyoo:
@bert-e wait

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -478,6 +488,13 @@ class LifecycleTask extends BackbeatTask {
             }
             return store;
         }, {});
+        // if EODM was not explicitly set and current account is LC account
+        // enable EODM rule by default
+        if (isLCUser && filteredRules.Expiration &&
+        filteredRules.Expiration.ExpiredObjectDeleteMarker !== false) {
```
jonathan-gramain, line 494: I would have preferred if we didn't have to tweak the applicable rules to support this, but instead do the check whenever we're ready to do the expiration of delete markers. (the reason is that this rule has a specific meaning in AWS API, we're extending that meaning to fit another special case for internal purpose which can bring some confusion - not a huge issue, if that requires some refactoring let's keep this for now as I prefer the amount of changes to be minimal).

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -410,11 +416,13 @@ class LifecycleTask extends BackbeatTask {
...
      * @param {array} rules - list of filtered rules that apply to a specific
      *   object, version, or upload
+     * @param {boolean} isLCUser - is current account a lifecycle service
+     *   account
      * @return {object} all applicable rules with earliest dates of action
      *  i.e. { Expiration: { Date: <DateObject>, Days: 10 },
      *         NoncurrentVersionExpiration: { NoncurrentDays: 5 } }
      */
-    _getApplicableRules(rules) {
+    _getApplicableRules(rules, isLCUser) {
```
jonathan-gramain, line 425: I would rather not shorten the variable name to win a few chars (so `isLifecycleUser`), but this may conflict with the helper function name (maybe?), so we can keep in doubt.

jonathan-gramain (approved):

extensions/lifecycle/tasks/LifecycleTask.js
philipyoo: maybe since its a boolean, i can call it something like `defaultEnable`? or something along those lines?

bennettbuchanan (approved):

extensions/lifecycle/tasks/LifecycleTask.js
philipyoo: I'll look into refactoring this. I think it'll be an easy refactor, but will not go into this release, but probably the next.
All relevant data is available in the function that sends the expiration object entries. But will need to edit the logic in the function that handles eodm versions

PR #399, opened 2018-07-31
bf: ZENKO-860, S3C-1820 maximize queue concurrency usage

jonathan-gramain:
Note for previous reviewers of this PR: the first commit is mostly the previous state of this PR that fixes the queue being blocked while one entry is slow to complete, except:

- removed a `setTimeout()` in case the queue is full, which was useless

- added a check with `isPaused()` to avoid calling `offsetsStore()` on a paused consumer, which causes an exception to be thrown because the consumer is not subscribed to a topic.

**bf: ZENKO-860 replace replication batching by continuous task flow**

Keep consuming kafka messages as soon as new workers become available
in the queue, rather than relying on the 'drain' callback from
async.queue.

This prevents being stuck from consuming new messages in case one of
the tasks takes a long time to finish (e.g. if retrying on a failing
cloud).

Change the way we manage Kafka consumer offsets: instead of disabling
auto-commit by default, always enable it and disable automatic offset
store of consumed messages. This allows to set the offset based on the
processed messages rather than the consumed messages, relying on a
helper class OffsetLedger to maintain knowledge of the highest offset
that can be committed, i.e. that is a higher-bound of already
processed message offsets.

**ft: ZENKO-860 improve replication task pool efficiency**

By reducing drastically the time it takes for a replication task to
complete, for the same concurrency number, a lot more tasks can be
executed, leading to more throughput and quicker completion of
replication queue tasks especially for small objects, while using
cloud resources more consistently and efficiently with the same bound
on the number of open connections.

To achieve this reduction in execution time, tasks do not wait anymore
for a delivery report to be returned from Kafka producers (for metrics
and backbeat-replication-status topics) before finishing. Most of the
wait time was spent waiting for the delivery reports, which are only
issued once Kafka commits changes in batch (every few seconds), hence
the tasks in progress were mostly waiting for those delivery reports
instead of transferring bytes of data between locations.

Instead, after the data replication is completed (success or failure),
and the task completes, we keep track of the delivery report pending
for the message produced to backbeat-replication-status topic. Only
when this status comes back (success or failure), we acknowledge that
this message has been consumed and processed, to allow eventually
committing the consumer group offset forward.

Delivery reports for messages produced to metrics topic are ignored,
because it does not seem to bring any consistency benefit to wait for
them (but we could instead implement a dedupe cache to get rid of
duplicates).


lib/BackbeatConsumer.js
```diff
@@ -224,18 +226,14 @@ class BackbeatConsumer extends EventEmitter {
...
-        // consumed partitions and try consuming new messages right away
+
+        // bookkeeping tasks when queue is empty
         this._processingQueue.drain = () => {
             // when auto-commit is set we don't have to commit
             // explicitly, so let's not do it for performance reasons
             if (!this._autoCommit) {
                 this._consumer.commit();
             }
-            this.emit('consumed', this._messagesConsumed);
```
rahulreddy, line 158: If autocommit is not enabled (which is the case for crr) and now that we are switching to flowing mode, when there are a lot of messages, the queue would never be empty (correct?) So the manual commit would probably happen after n messages and n could be 10/100/200?

lib/BackbeatConsumer.js
```diff
@@ -265,8 +263,20 @@ class BackbeatConsumer extends EventEmitter {
...
+              - this._processingQueue.running()
+              - this._processingQueue.length()
+              - this._nConsumePendingRequests;
+        if (nNewConsumeRequests <= 0) {
+            // processing pipeline is 100% busy, retry later
+            return setTimeout(this._tryConsume.bind(this), 1000);
+        }
+        this._nConsumePendingRequests += nNewConsumeRequests;
+        return this._consumer.consume(nNewConsumeRequests, (err, entries) => {
+            this._nConsumePendingRequests -= nNewConsumeRequests;
```
rahulreddy, line 218: `this._nConsumePendingRequests -= nNewConsumeRequests;` this can be decremented only if there is no error fetching the entries correct?

rahulreddy (approved):

rahulreddy (changes requested):
Approved accidentally

lib/BackbeatConsumer.js
jonathan-gramain: Actually I want to decrement it unconditionally, because this is the bookkeeper for pending *requests*, so even if the requests fail they are not pending anymore, and we don't want the counter to be off thereafter.

lib/BackbeatConsumer.js
jonathan-gramain: We are actually not switching to flowing mode, we are still running in non-flowing mode by consuming a specific amount of messages (https://github.com/scality/backbeat/pull/399/files#diff-e2120238e136b010d97978588107a8cbR278).

The `'consumed'` event is right now only used in tests AFAICS, and I moved it from this place to https://github.com/scality/backbeat/pull/399/files#diff-e2120238e136b010d97978588107a8cbR278, where each ending task has an opportunity to emit this event.

tests/functional/BackbeatConsumer.js
```diff
@@ -43,37 +39,43 @@ describe('BackbeatConsumer', () => {
...
-            _doneIfReady();
+        consumerConc = new BackbeatConsumer({
+            zookeeper: zookeeperConf,
+            kafka: kafkaConf, groupId: groupIdConc, topic,
+            queueProcessor,
+            concurrency: 10,
+            bootstrap: true,
         });
+        async.parallel([
+            done => producer.on('ready', done),
```
JianqinWang: I assume that having the outer function callback named `done` does not conflict with the `async` callbacks also named `done` - it was still confusing to read at a first glance xD.

JianqinWang (approved):

tests/functional/BackbeatConsumer.js
jonathan-gramain: it does not conflict, though may be confusing indeed

philipyoo (dismissed):
lgtm 👍

philipyoo:
are tests failing because requires more time than current set timeout to complete?

lib/BackbeatConsumer.js
jonathan-gramain: I realize this deserves a more specific answer.

Even though we're still in non-flowing mode, you're still right in saying the queue may never get empty, this until we stop adding new messages to the queue at least. This behavior has the same kind of validity that it used to, but it may have become less useful since there's no guarantee that it will be called timely.

To overcome this, the caller either has to use auto-commit, or we should add a `commit()` wrapper that the user is responsible for calling when he wishes to commit the current offset. For our use-case it seems reasonable to use auto-commit for replication. It may be also worth enabling it by default, like the kafka client does.

jonathan-gramain:
The latest changes should address remaining concerns (about the offset commit especially).

jonathan-gramain:
Postponed to Zenko 1.1

lib/BackbeatConsumer.js
jonathan-gramain: This has been solved by enabling auto-commit on the kafka consumer.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -132,33 +132,39 @@ class ReplicateObject extends BackbeatTask {
...
+            // Commit whether there was an error or not to allow
+            // progress of the consumer, as best effort measure when
+            // there are errors. We can count on the sweeper to retry
+            // entries that failed to be published to kafka (because
+            // they will keep their PENDING status).
+            if (this.consumer) {
+                this.consumer.onEntryCommittable(kafkaEntry);
+            }
+        });
+        log.end().info('replication status published', {
```
philipyoo, line 163: maybe only log this if an error didn't occur above https://github.com/scality/backbeat/pull/399/commits/63b260d0b2827d15501f984815fd0ace8fae6b76#diff-e5fe2d1a77d806ec6632afc36abf102fR145

philipyoo (approved):
lgtm 👍

extensions/replication/tasks/ReplicateObject.js
jonathan-gramain: This is logged in the synchronous context, so at this point we can only assert that we have published but did not get the feedback yet.

jonathan-gramain:
Manual run of backbeat E2E tests passes. (using custom branch test/backportCRRRetryZenkoE2ETests and my two lab deployments, I gave up for now trying to run them as part of eve automation).

It has also been tested with large workloads (100000+ objects) without issues.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -132,33 +132,39 @@ class ReplicateObject extends BackbeatTask {
...
-            log.end().info('replication status published', {
-                topic: this.repConfig.replicationStatusTopic,
-                entry: updatedSourceEntry.getLogInfo(),
-                replicationStatus,
-                reason,
-            });
-            return cb();
+            // Commit whether there was an error or not to allow
+            // progress of the consumer, as best effort measure when
+            // there are errors. We can count on the sweeper to retry
```
rahulreddy, line 156: When you say `sweeper`, do you mean the background task that publishes stalled entries? If yes, that would mean that we have to enable it by default in Zenko and install something comparable in S3C as well.

lib/BackbeatConsumer.js
```diff
@@ -191,29 +186,47 @@ class BackbeatConsumer extends EventEmitter {
...
-                        entry, err => this._onEntryProcessingDone(err, entry));
+                    this._processingQueue.push(entry, (err, completionArgs) => {
+                        this._onEntryProcessingDone(err, entry, completionArgs);
+                    });
                 });
             }
             if (err || entries.length === 0) {
-                // retry later to fetch new messages in case no one is
-                // available yet in the message queue
+                // no message is available yet, retry later
```
rahulreddy: Let's add a debug log here for visibility

lib/BackbeatConsumer.js
```diff
@@ -65,28 +68,31 @@ class BackbeatConsumer extends EventEmitter {
...
         this._queueProcessor = queueProcessor;
         this._concurrency = concurrency;
         this._fetchMaxBytes = fetchMaxBytes;
         this._bootstrap = bootstrap;
+        this._offsetLedger = new OffsetLedger();
+
         this._processingQueue = null;
         this._messagesConsumed = 0;
-
+        this._nConsumePendingRequests = 0;
```
rahulreddy, line 91: Can you please comment what this property represents?

lib/BackbeatConsumer.js
```diff
@@ -191,29 +186,47 @@ class BackbeatConsumer extends EventEmitter {
     _tryConsume() {
         // use non-flowing mode of consumption to add some flow
         // control: explicit consumption of messages is required,
-        // needs explicit polling until new messages become available.
-        this._consumer.consume(this._concurrency, (err, entries) => {
+        // needs explicit polling to get new messages
+
+        const nNewConsumeRequests =
```
rahulreddy: Can you make this a method that is more descriptive? This represents the queue status and we use it to decide if more items can be pushed to the queue.

lib/BackbeatConsumer.js
rahulreddy: It's worth logging debug log here.

lib/OffsetLedger.js
```diff
@@ -0,0 +1,143 @@
...
+        // since it means all lower offsets have already been
+        // processed. If nothing is being processed, the latest
+        // consumed message offset can be committed (so +1 for the
+        // commit offset).
+
+        if (partitionOffsets.processing.length !== 0) {
+            return partitionOffsets.processing[0];
+        }
+        if (partitionOffsets.latestConsumed !== null) {
+            return partitionOffsets.latestConsumed + 1;
```
rahulreddy, line 44: What if this is somehow a `NaN` because `latestConsumed` is `undefined` 😄? Does `noderdkafaka` throw an exception in this case?

lib/OffsetLedger.js
```diff
@@ -0,0 +1,143 @@
...
+    onOffsetConsumed(topic, partition, offset) {
+        const partitionOffsets = this._getPartitionOffsets(topic, partition);
+        if (partitionOffsets.latestConsumed !== null &&
+            offset !== partitionOffsets.latestConsumed + 1) {
+            // replay situation (at-least-one semantics does not
+            // prevent this from happening): get rid of all higher
+            // offsets currently being processed, as they will be
+            // reprocessed
+            partitionOffsets.processing =
+                partitionOffsets.processing.filter(pOff => pOff < offset);
```
rahulreddy, line 71: When you say higher offset, isn't the conditional `pOff > offset`?

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -0,0 +1,221 @@
...
+        for (let i = 0; i < 1000; ++i) {
+            boatloadOfMessages.push({
+                key: `message-${i}`,
+                message: `{"message_index":"${i}"}`,
+            });
+        }
+        async.series([
+            next => {
+                setTimeout(() => producer.send(boatloadOfMessages, err => {
+                    assert.ifError(err);
```
rahulreddy, line 63: This is asynchronous and the code following this i synchronous, I think it deserves its own step.

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -0,0 +1,221 @@
...
+                // test, so wait for 2 seconds
+                setTimeout(() => {
+                    assert.strictEqual(
+                        consumer.getOffsetLedger().getProcessingCount(),
+                        0);
+                    done();
+                }, 2000);
+            }
+        });
+    }).timeout(30000);
```
rahulreddy: For timeouts, let's set them globally at once. It can be done in package.json when invoked by command or mocha.opts.

rahulreddy (changes requested):

extensions/replication/tasks/ReplicateObject.js
jonathan-gramain: Indeed we should have a sweeper for those cases. Note that the behavior here has not changed with this PR: before we were simply committing when 'drain' even was issued, whether there were errors processing entries or not. So I think we should consider this separately from this PR, as it's not directly related.

lib/OffsetLedger.js
jonathan-gramain: Not sure it can happen, but I can make it bullet-proof by checking that it's an integer instead.

lib/OffsetLedger.js
jonathan-gramain: `filter` will keep the lower offsets only, so getting rid of the higher ones.

tests/functional/lib/BackbeatConsumer.js
jonathan-gramain: The synchronous code executes immediately without waiting for`producer.send` to complete, this is done on purpose so that we produce messages only after we are ready to consume them. If you have another preference for the code style let me know.

rahulreddy (approved):

jonathan-gramain:
@bert-e reset

philipyoo (approved):

jonathan-gramain:
Zenko tests ran and passed here, with a custom backbeat image: https://eve.devsca.com/github/scality/zenko/#/builders/2/builds/1046

jonathan-gramain:
@bert-e approve

PR #404, opened 2018-08-20
bf: ZENKO-1005 set earliest offset for new locations

philipyoo:
New storage locations will have no initial offset for each
partition. This causes problems if the user were to add a
new or fairly new storage location and were to pause the
location. Since pausing causes consumers to unsubscribe
from the crr topic, any partitions that did not receive any
prior data will not be able to determine the LAG for that
partition.

When the location is resumed, since potentially no LAG is
shown, the crr attempt for the new storage location objects
will stay stuck at PENDING.

While testing, the following command helped a lot to see what
was happening:

```
$ bin/kafka-consumer-groups.sh --describe \
--bootstrap-server <queue-service-ip>:9092 \
--group backbeat-replication-group-<location-name>
```

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -98,8 +98,7 @@
             "queueProcessor": {
                 "groupId": "backbeat-replication-group",
                 "retryTimeoutS": 300,
-                "concurrency": 10,
-                "fromOffset": "earliest"
+                "concurrency": 10
             },
             "replicationStatusProcessor": {
                 "groupId": "backbeat-replication-group",
--- a/conf/config.json
+++ b/conf/config.json
@@ -98,7 +98,8 @@
             "queueProcessor": {
                 "groupId": "backbeat-replication-group",
                 "retryTimeoutS": 300,
-                "concurrency": 10
+                "concurrency": 10,
+                "fromOffset": "earliest"
             },
             "replicationStatusProcessor": {
                 "groupId": "backbeat-replication-group",
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -129,13 +129,14 @@ class BackbeatConsumer extends EventEmitter {
             'enable.auto.commit': this._autoCommit,
             'offset_commit_cb': this._onOffsetCommit.bind(this),
         };
+        const topicParams = {};
         if (this._fromOffset !== undefined) {
-            consumerParams['auto.offset.reset'] = this._fromOffset;
+            topicParams['auto.offset.reset'] = this._fromOffset;
         }
         if (this._fetchMaxBytes !== undefined) {
             consumerParams['fetch.message.max.bytes'] = this._fetchMaxBytes;
         }
-        this._consumer = new kafka.KafkaConsumer(consumerParams);
+        this._consumer = new kafka.KafkaConsumer(consumerParams, topicParams);
         this._consumer.connect({ timeout: 10000 }, () => {
             const opts = {
                 topic: withTopicPrefix('backbeat-sanitycheck'),
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -61,7 +61,6 @@ const joiSchema = {
         groupId: joi.string().required(),
         retryTimeoutS: joi.number().default(300),
         concurrency: joi.number().greater(0).default(10),
-        fromOffset: joi.string().required(),
     },
     replicationStatusProcessor: {
         groupId: joi.string().required(),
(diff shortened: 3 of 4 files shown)


philipyoo:
@bert-e wait

bennettbuchanan (approved):

lib/BackbeatConsumer.js
```diff
@@ -129,13 +129,14 @@ class BackbeatConsumer extends EventEmitter {
             'enable.auto.commit': this._autoCommit,
             'offset_commit_cb': this._onOffsetCommit.bind(this),
         };
+        const topicParams = {};
         if (this._fromOffset !== undefined) {
-            consumerParams['auto.offset.reset'] = this._fromOffset;
+            topicParams['auto.offset.reset'] = this._fromOffset;
```
jonathan-gramain, line 134: good catch :+1:

conf/config.json
```diff
@@ -98,7 +98,8 @@
             "queueProcessor": {
                 "groupId": "backbeat-replication-group",
                 "retryTimeoutS": 300,
-                "concurrency": 10
+                "concurrency": 10,
+                "fromOffset": "earliest"
```
jonathan-gramain: This should not be configurable IMO (internal fix)

jonathan-gramain (changes requested):
I'm not fond of setting the "earliest" mode in all cases. I think it would fix this particular issue but introduce some unwanted overhead (and possibly side-effects?) when adding new replication policies. In the particular case of resume, I think it's correct to start from the beginning of a not-yet-consumed partition because normally the paused consumer that has not yet recorded its offset for a partition wants to consume from the beginning, but for a new location it seems better to start from the latest offset, as there may be a lot of backlog that we do not want to process for the new site.

Instead, could it be better to only set "earliest" when doing a resume, and "latest" when subscribing on a new location, i.e. during process startup?

philipyoo:
@jonathan-gramain I tried testing setting the `auto.offset.reset` to "earliest" when a resume request is made using 2 different methods. First, by setting directly on existing consumer, and second by recreating a new consumer instance with the topic config set.
In both cases, I wasn't able to recover prior offsets, and it seems like the config cannot be set once a prior consumer has already been set to "latest". Or at least that is what it looks like to me based on each partitions current-offset.

Ideally, for a new consumer, I want its CURRENT-OFFSET to be set to LOG-END-OFFSET on startup, similar to setting `auto.offset.reset` to "latest". But, the way the "latest" config seems to work is whenever that partition receives an entry for consumption, it looks like it will look directly at LOG-END-OFFSET + its current LAG and when consuming that current LAG, will then set CURRENT-OFFSET to the new offset. So it only updates a partitions CURRENT-OFFSET on new entry received.

Maybe there is an alternative of setting the CURRENT-OFFSET on startup by looking at the LOG-END-OFFSET?

rahulreddy:
I am deferring this fix to 1.1 as it needs a lot more testing. There is a good reason we have `latest` as the consumer offset behavior, this change fundamentally alters the behavior of consumers.

jonathan-gramain:
@philipyoo To clarify:

I'm not familiar with the LAG metric of kafka, for me kafka maintains (persists) an offset for each consumerGroup/topicPartition pair that has been already consumed in the past, and will use this offset for any further consumption from the same consumer group. I don't think there's an easy way to override this behavior, and I don't think we want it either (this would break the at-least-once semantic).

Considering a consumer group *has to* consume every message of a partition from a certain point, that point can be chosen to be the latest or earliest message available in the partition. In some cases it's desirable to have one or the other behavior. Once the consumer has started consuming from the desired (earliest/latest) point and committed its offset for a certain partition, there is no (easy) way back, even if recreating a new instance of KafkaConsumer, the offset is persisted by kafka so will be reused.

Sorry if this is already clear for you but I was not sure considering your comment so just in case :)

Now on the particular issue we're dealing with, I would think that for a new replication policy being set, it's desirable to start from the latest point in any kafka partition, because any prior message would not be eligible for replication, while for the particular case of resuming a paused location, if a partition has not been consumed yet it should mean that it had been paused while nothing had been consumed yet from this partition, so it can be desirable in such case to start consuming from the earliest message to process everything that has been put during the paused state.

I agree with Rahul that this needs a fair amount of testing, so 1.1 looks more reasonable.

philipyoo:
sgtm, I agree more testing should be done. I'll keep testing and seeing if I can figure out a fix where we can change the consumer config based on resume requests

jonathan-gramain:
One way to deal with this could be to enforce committing the latest offset for all partitions when creating a new location. Then a resume would start from this offset for all partitions.

philipyoo:
closing in favor of: https://github.com/scality/backbeat/pull/423

PR #405, opened 2018-08-22
ft: ZENKO-1019 delete scheduled resume

philipyoo:
**Looking to merge this for pre-GA build**

- Add a DELETE route to delete/cancel a scheduled resume for given site(s)
- Small refactor to reuse a test helper method for both POST and DELETE requests

### DIFF
--- a/docs/crr-pause-resume.md
+++ b/docs/crr-pause-resume.md
@@ -1,4 +1,4 @@
-# Cross-Region Replication (CRR) Manual Pause and Resume
+# Cross-Region Replication (CRR) Pause and Resume
 
 ## Description
 
@@ -140,6 +140,19 @@ is resumed, it again resumes consuming entries from its last offset.
         "hours": 6
     }
     ```
+
+    Response:
+    ```json
+    {}
+    ```
+
+* DELETE `/_/backbeat/api/crr/resume/<location-name>/schedule`
+
+    This is a DELETE request to remove a scheduled resume for cross-region
+    replication to a specified location configured as a destination replication
+    endpoint.
+    Specify "all" as a location name to make this request to all available
+    destinations.
 
     Response:
     ```json
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -736,6 +736,7 @@ class BackbeatAPI {
         }
         try {
             reqBody = JSON.parse(body);
+            // default 6 hours if no body value sent by user
             if (!reqBody.hours) {
                 return defaultRes;
             }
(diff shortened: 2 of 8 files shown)


philipyoo:
@bert-e wait

rahulreddy:
@RachedBenMustapha  This may be of interest to you. Did you already implement a cancel by sending an `{}` body from Orbit?

rachedbenmustapha:
@alexanderchan-scality can you confirm that the immediate resume is using an empty object as a body placeholder already?

alexanderchan-scality:
@RachedBenMustapha Yes, the immediate resume/pause requests have empty bodies. As for @rahulreddy 's question about `{}` for the schedule requests, an empty body will default to scheduling a resume in 6 hours, so the empty will not unset the scheduled times.

lib/api/BackbeatAPI.js
```diff
@@ -736,6 +736,9 @@ class BackbeatAPI {
         }
         try {
             reqBody = JSON.parse(body);
+            if (reqBody.cancel) {
```
JianqinWang: If a user set this to 'false', wouldn't this skip?
In the documentation it says:
"Providing a POST request body object with a `cancel` key with any value will remove any scheduled resumes for the given destination replication endpoint."
We probably shouldn't give users this much freedom of usage + interpretation.

JianqinWang (changes requested):

alexanderchan-scality (approved):

lib/api/BackbeatAPI.js
philipyoo: ah good catch! I think I'm going to change this so rather than any value, I expect user to send `true`. Seems sort of weird to just accept any value from a ux perspective

docs/crr-pause-resume.md
```diff
@@ -141,6 +142,20 @@ is resumed, it again resumes consuming entries from its last offset.
     }
     ```
 
+    Providing a POST request body object with a `cancel` key with any value
```
bennettbuchanan: Looks like it has to be a truthy value, rather than any value.

bennettbuchanan (approved):

alexanderchan-scality (approved):

lib/api/BackbeatAPI.js
```diff
@@ -736,6 +736,17 @@ class BackbeatAPI {
         }
         try {
             reqBody = JSON.parse(body);
+            // prioritize cancel request
+            if (Object.keys(reqBody).includes('cancel')) {
+                if (reqBody.cancel !== true) {
+                    return {
+                        error: errors.MalformedPOSTrequest.customizeDescription(
+                            `${msg}: cancel key found with incorrect value`),
```
bennettbuchanan: I think it would be worthwhile to include in the message that the `cancel` property key must be `true`.

philipyoo:
Sorry instead of cancel via POST, should be a DELETE request with no body

rahulreddy:
It's good to be inline with the REST API semantics 💯

JianqinWang (approved):

dora-korpar (approved):

rahulreddy (approved):

PR #410, opened 2018-08-22
bf: ZENKO-1052 send text/plain type for promclient

philipyoo:
Our use of promclient route in backbeat expects the response
to be content-type of text/plain as determined by the promclient
library itself using `client.register.contentType`

### DIFF
--- a/lib/api/BackbeatServer.js
+++ b/lib/api/BackbeatServer.js
@@ -182,15 +182,22 @@ class BackbeatServer {
         const log = backbeatRequest.getLog();
         const req = backbeatRequest.getRequest();
         const res = backbeatRequest.getResponse();
+        const contentType = backbeatRequest.getContentType();
+        const code = backbeatRequest.getStatusCode();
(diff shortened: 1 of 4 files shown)


giacope (approved):

alexanderchan-scality (approved):

ssalaues (approved):

lib/api/BackbeatServer.js
```diff
@@ -182,15 +182,22 @@ class BackbeatServer {
...
             method: 'BackbeatServer._response',
         });
-        const code = backbeatRequest.getStatusCode();
-        const payload = Buffer.from(JSON.stringify(data), 'utf8');
+
+        let payload;
+        if (contentType === 'application/json') {
+            payload = Buffer.from(JSON.stringify(data), 'utf8');
+        } else {
+            payload = data;
```
rahulreddy, line 195: Minor: Regardless of the content-type payload can be a buffer.

rahulreddy (approved):

PR #412, opened 2018-08-23
bugfix: ZENKO-1024 Add pending counters

bennettbuchanan:
Opening this PR for high-level reviews. Extensive testing for this new metric is still needed.

Depends on https://github.com/scality/Arsenal/pull/555.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -49,6 +49,25 @@ Where:
 This route gathers all the metrics below, returning one JSON object for the
 specified extension type and location name.
 
+### `/_/backbeat/api/metrics/crr/<location-name>/pending`
+
+This route returns the pending operations in number of objects and number
+of total bytes for the specified extension type and location. This metric does
+not expire.
+
+**Example Output**:
+
+```
+"pending":{
+    "description":"Number of pending replication operations (count) and bytes
+    (size)",
+    "results":{
+        "count":"1",
+        "size":"1024"
+    }
+}
+```
+
 ### `/_/backbeat/api/metrics/crr/<location-name>/backlog`
 
 This route returns the replication backlog in number of objects and number of
@@ -64,8 +83,8 @@ for an object fails, failed object metrics are considered backlog.
     "description":"Number of incomplete replication operations (count) and
     number of incomplete bytes transferred (size)",
     "results":{
-        "count":4,
-        "size":"6.12"
+        "count":"1",
+        "size":"1024"
     }
 }
 ```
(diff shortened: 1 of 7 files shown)


bennettbuchanan:
@bert-e wait

lib/MetricsConsumer.js
```diff
@@ -61,12 +61,24 @@ class MetricsConsumer {
...
         const { type, site, ops, bytes } = data;
         if (type === 'completed') {
+            this._statsClient
+                .decrementKey(`${site}:${redisKeys.opsPending}`, ops);
+            this._statsClient
+                .decrementKey(`${site}:${redisKeys.bytesPending}`, bytes);
             this._sendRequest(`${site}:${redisKeys.opsDone}`, ops);
             this._sendRequest(`${site}:${redisKeys.bytesDone}`, bytes);
         } else if (type === 'failed') {
+            this._statsClient
```
philipyoo, line 71: Greedy request, we will still have problems with backlog and I was hoping we could use the no-expire redis data to determine backlog.

Is it possible to have these global keys for backlog and separate keys for fails?

bennettbuchanan: I see your concern that backlog still does not factor in failures, and that it may be inaccurate (e.g. the case that the CRR operation completes in a later interval). I guess it's a matter of design philosophy—i.e. what is the reasoning behind using interval keys. Do we eventually want some kind of timeline history of events?

bennettbuchanan: Perhaps we can move the 'backlog' metric to instead be a 'queued' metric for that interval (i.e. do not subtract completions when calculating backlog). Then we end up with four unique metrics:

* pending (operations actively pending)
* queued (operations queued by interval)
* completed (operations completed by interval)
* failed (operations failed by interval)

This way we could use the intervals for building a timeline of system usage and still have a view of active pending operations.

docs/metrics.md
```diff
@@ -64,8 +83,8 @@ for an object fails, failed object metrics are considered backlog.
     "description":"Number of incomplete replication operations (count) and
     number of incomplete bytes transferred (size)",
     "results":{
-        "count":4,
-        "size":"6.12"
+        "count":"1",
+        "size":"1024"
```
philipyoo, line 87: 👍

package.json
```diff
@@ -32,7 +32,7 @@
   },
   "homepage": "https://github.com/scality/backbeat#readme",
   "dependencies": {
-    "arsenal": "scality/Arsenal#06dfdd9",
+    "arsenal": "scality/Arsenal#bugfix/ZENKO-1024/add-global-counters",
```
philipyoo: just a reminder to update this. I ran tests locally and the new pending tests were passing

philipyoo (approved):
lgtm 👍

rahulreddy (approved):

PR #415, opened 2018-09-01
bf: ZENKO-1024 use pending metrics for backlog

philipyoo:
Pending metrics don't expire which was a cause for problems
with current backlog. This quick fix is to use pending
metrics in place of backlog but keeping the same names
and routes in place to avoid regression.

Additional change includes failures metric validation in all metrics routes

To be merged with https://github.com/scality/Arsenal/pull/556

### DIFF
--- a/tests/functional/api/BackbeatServer.js
+++ b/tests/functional/api/BackbeatServer.js
@@ -400,10 +400,9 @@ describe('Backbeat Server', () => {
             getRequest(`/_/metrics/crr/${site1}/backlog`, (err, res) => {
                 assert.ifError(err);
                 const key = Object.keys(res)[0];
-                // Backlog count = OPS - OPS_DONE
-                assert.equal(res[key].results.count, 1275);
-                // Backlog size = BYTES - BYTES_DONE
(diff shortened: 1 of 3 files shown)


anurag4DSB (approved):

alexanderchan-scality (approved):

philipyoo:
@bert-e wait

bennettbuchanan (approved):

tests/functional/api/BackbeatServer.js
```diff
@@ -562,17 +560,11 @@ describe('Backbeat Server', () => {
             getRequest('/_/metrics/crr/all', (err, res) => {
                 assert.ifError(err);
                 const keys = Object.keys(res);
-                assert(keys.includes('backlog'));
```
rahulreddy, line 565: We no longer include backlog in the responses? Wouldn't this be breaking compatibility with Orbit?

philipyoo: sorry, I shouldn't have removed this. Both backlog and pending are in the response and data matches

PR #416, opened 2018-09-01
bf: ZENKO-1024 fail retries metrics

philipyoo:
Quick fix for backbeat failure/pending metrics and failure object metrics:
- Failures (bucket-level) metrics should represent list of failures object keys
- Retry Fails GET listings and POST requests should filter out duplicate version ids 
   as they should be unique.
- Any duplicates found should try to normalize pending metrics by first filtering from
   the list, attempt to remove the duplicate keys from redis, and then increment
   pending metrics

To be merged AFTER https://github.com/scality/backbeat/pull/415

To be merged with https://github.com/scality/Arsenal/pull/557

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -344,7 +344,16 @@ class BackbeatAPI {
         if (response.IsTruncated) {
             response.NextMarker = Number.parseInt(cursor, 10);
         }
-        return async.eachLimit(keys, 10, (key, next) =>
+
+        // TODO: Hotfix made here
+        let filteredKeys = this._filterFailObjectKeys(keys);
+
+        if (process.env.CI === 'true') {
+            // ignore filtering keys for tests
+            filteredKeys = keys;
+        }
+
+        return async.eachLimit(filteredKeys, 10, (key, next) =>
             this._redisClient.batch([['get', key]], (err, res) => {
                 if (err) {
                     return next(err);
@@ -616,6 +625,43 @@ class BackbeatAPI {
     }
 
     /**
+     * TODO: Tempfix to normalize metrics on concurrency issue
+     */
+    _normalizeFailMetrics(keys) {
+    }
+
+    _filterFailObjectKeys(keys) {
+        // TODO: Hotfix made here
+        const filteredKeys = [];
+        const uniqueObjs = {};
+        const duplicateKeys = [];
+        keys.forEach(key => {
+            // 8 parts: [bb, crr, failed, <bucket>, <key>,
+            //           <versionID>, <location>, <timestamp>]
+            const parts = process.env.CI === 'true' ?
+                key.split(':').slice(4) : key.split(':').slice(3);
+            const [bucket, keyName, versionID, location] = parts;
+            const hash = `${bucket}:${keyName}:${versionID}:${location}`;
+            // if the key already exists, debug log because an error
+            // occurred somewhere
+            if (uniqueObjs[hash]) {
+                this._logger.debug('duplicate version id was found in failed ' +
+                'object redis keys', {
+                    method: 'BackbeatAPI._filterFailObjectKeys',
+                });
+                duplicateKeys.push(key);
+            } else {
+                uniqueObjs[hash] = true;
+                filteredKeys.push(key);
+            }
+            uniqVersionIDs[versionID] = key;
+        });
+        // attempt to remove without blocking request
+        this._normalizeFailMetrics(duplicateKeys);
+        return filteredKeys;
+    }
+
+    /**
      * Retry all CRR operations that have failed.
      * @param {Object} details - The route details
      * @param {String} body - The POST request body string
@@ -635,6 +681,16 @@ class BackbeatAPI {
             (err, keys) => {
                 if (err) {
                     return next(err);
+                }
+                // TODO: if multiple keys, an error occurred
+                if (keys.length > 1) {
+                    this._logger.debug('duplicate keys were found in failed ' +
+                    'object redis keys', {
+                        method: 'BackbeatAPI.retryFailedCRR',
+                    });
+                    // TODO: attempt to remove without blocking request
+                    const leftoverKeys = keys.slice(1);
+                    this._normalizeFailMetrics(leftoverKeys);
                 }
                 // There can only be one failed key. We are only using the glob
                 // because we do not know the original timestamp.
@@ -705,6 +761,23 @@ class BackbeatAPI {
                 error: errors.MalformedPOSTRequest.customizeDescription(errMsg),
             };
         }
+        // Bucket, Key, VersionId, StorageClass
+        // filter duplicate version ids if any
+        const uniqueObjs = {};
+        if (process.env.CI === 'true') {
+            // ignore filtering for tests
+            return { reqBody };
+        }
+        reqBody = reqBody.filter(entry => {
+            const { Bucket, Key, VersionId, StorageClass } = entry;
+            const hash = `${Bucket}:${Key}:${VersionId}:${StorageClass}`;
+            if (!uniqueObjs[hash]) {
+                uniqueObjs[hash] = entry;
+                return true;
+            }
+            return false;
+        });
+
         return { reqBody };
     }
 
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -618,6 +618,16 @@ class BackbeatAPI {
                     }),
                     // Delete the Redis key from the prior failure.
                     done => this._deleteFailedCRRKey(entry.getRedisKey(), done),
+                    // Delete the Redis lock key
+                    done => {
+                        const bucket = entry.getBucket();
+                        const key = entry.getRedisKey();
+                        const versionId = entry.getEncodedVersionId();
+                        const storageClass = entry.getSite();
+                        const lockKey = `${bucket}:${key}:${versionId}:` +
+                            `${storageClass}:processingFailures`;
+                        return this._deleteFailedCRRKey(lockKey, done);
+                    },
                 ], next);
             }),
         err => cb(err, response));
@@ -813,7 +823,19 @@ class BackbeatAPI {
                     }
                     // If the key did not exist, value is `null`.
                     if (sourceRole) {
-                        entries.push(new ObjectFailureEntry(key, sourceRole));
+                        const lockKey = `${Bucket}:${Key}:${VersionId}:` +
+                            `${StorageClass}:processingFailures`;
+                        return this._applyRetryLockKey(lockKey,
+                        (err, lockApplied) => {
+                            if (err) {
+                                return next(err);
+                            }
+                            if (lockApplied) {
+                                entries.push(new ObjectFailureEntry(key,
+                                    sourceRole));
+                            }
+                            return next();
+                        });
                     }
                     return next();
                 });
@@ -823,6 +845,37 @@ class BackbeatAPI {
                 return cb(err);
             }
             return this._processFailedKafkaEntries(entries, cb);
+        });
+    }
+
+    /**
+     * Apply a retry lock for a given request if none already applied
+     * @param {String} key - lock key
+     * @param {Function} cb - callback(err, boolean)
+     *   Where if boolean is true, a lock has been applied
+     *   Where if boolean is false, a lock already exists
+     * @return {undefined}
+     */
+    _applyRetryLockKey(key, cb) {
+        this._redisClient.batch([['get', key]], (err, res) => {
+            if (err) {
+                return cb(err);
+            }
+            const [cmdErr, lock] = res[0];
+            if (cmdErr) {
+                return cb(cmdErr);
+            }
+            if (!lock) {
+                // add lock and continue processing
+                return this._redisClient.batch([['set', key, 'true']], err => {
+                    if (err) {
+                        return cb(err);
+                    }
+                    return cb(null, true);
+                });
+            }
+            // lock already exists
+            return cb(null, false);
         });
     }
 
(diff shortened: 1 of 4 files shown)


philipyoo:
@bert-e wait

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,156 @@ class BackbeatAPI {
...
+    getFailedMetrics(details, cb, data) {
+        const { failedCRR } = redisKeys;
+        const site = details.site === 'all' ? '*' : details.site;
+        const pattern = `${failedCRR}:*:${site}:*`;
+
+        this._redisClient.scan(pattern, undefined, (err, keys) => {
+            if (err) {
+                this._logger.error('error getting metric: failures', {
+                    error: err,
+                    method: 'BackbeatAPI.getFailMetrics',
```
bennettbuchanan: ~getFailMetrics~ getFailedMetrics

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,156 @@ class BackbeatAPI {
...
+    _filterFailObjectKeys(keys) {
+        const filteredKeys = [];
+        const uniqVersionIDs = {};
+        const removeKeys = [];
+        keys.forEach(key => {
+            // 8 parts: [bb, crr, failed, <bucket>, <key>,
+            //           <versionID>, <location>, <timestamp>]
+            const parts = key.split(':');
+            const versionID = process.env.CI === 'true' ? parts[6] : parts[5];
+            if (versionID === '') {
```
bennettbuchanan: I guess this is the case for non-versioned objects?

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,156 @@ class BackbeatAPI {
...
+                filteredKeys.push(key);
+                return;
+            }
+            // if the versionID already exists, debug log because an error
+            // occurred somewhere
+            if (uniqVersionIDs[versionID]) {
```
bennettbuchanan: I'm wondering if there is an off-chance that there could be multiple version IDs across buckets and objects. Perhaps to be safe, we should check that the bucket, object, and versionId is unique. In that case, we could just update the hash key to include all of those values.

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,156 @@ class BackbeatAPI {
...
+                this._logger.debug('duplicate version id was found in failed ' +
+                'object redis keys', {
+                    method: 'BackbeatAPI._filterFailObjectKeys',
+                    message: 'Known bug with unknown origin. Debug and ' +
```
bennettbuchanan: I'm thinking the debug log message is enough here and instead add the sentiment of this message as a comment. For example, I can imagine a user seeing this and wondering "who, exactly, should I report the behavior to?".

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,156 @@ class BackbeatAPI {
...
+                        'report this behavior please',
+                });
+                removeKeys.push(key);
+            } else {
+                uniqVersionIDs[versionID] = key;
+            }
+        });
+        filteredKeys.push(...Object.keys(uniqVersionIDs)
```
bennettbuchanan: I think we can avoid this loop by just pushing the key to `filteredKeys` in the `else` block above.

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,156 @@ class BackbeatAPI {
...
+            };
+            return cb(null, response);
+        });
+    }
+    /* eslint-enable */
+
+    _filterFailObjectKeys(keys) {
+        const filteredKeys = [];
+        const uniqVersionIDs = {};
+        const removeKeys = [];
```
bennettbuchanan: I think `duplicateKeys` might be more expressive.

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,156 @@ class BackbeatAPI {
...
+                    }
+                    if (!sourceRole) {
+                        // error occurred, skip for next time
+                        return next(errors.InternalError);
+                    }
+                    const entry = new ObjectFailureEntry(key, sourceRole);
+                    return done(null, entry);
+                }),
+                (entry, done) => {
+                    if (!entry) {
```
bennettbuchanan: I'm not sure I understand when this would this occur.

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,156 @@ class BackbeatAPI {
...
+                    });
+                },
+            ], (err, bytes) => {
+                if (err) {
+                    this._logger.debug('failed to delete duplicate object ' +
+                    'fail key', {
+                        method: 'BackbeatAPI._normalizeFailMetrics',
+                        key,
+                    });
+                    return next();
```
bennettbuchanan: I think we want to return the error here if there is one.

lib/api/BackbeatAPI.js
```diff
@@ -636,6 +793,19 @@ class BackbeatAPI {
                 if (err) {
                     return next(err);
                 }
+                // NOTE: if multiple keys, an error occurred
+                // Filter/Remove duplicate object keys with same version id
+                if (keys.length > 1) {
```
bennettbuchanan, line 804: 👍

lib/api/BackbeatAPI.js
```diff
@@ -705,6 +875,22 @@ class BackbeatAPI {
...
+        // filter duplicate version ids if any
+        const trackVersionIDs = {};
+        if (process.env.CI === 'true') {
+            // ignore filtering for tests
+            return { reqBody };
+        }
+        reqBody = reqBody.filter(entry => {
+            if (!entry.VersionId) {
+                return true;
+            } else if (!trackVersionIDs[entry.VersionId]) {
```
bennettbuchanan: As mentioned in another comment, I think it could be worthwhile to also check the bucket and object names here.

Minor: Let's make this another `if` block (rather than `else..if`) because we return in the first one.

bennettbuchanan (commented):
Thanks for getting this! It looks like a good way to address the issue we were seeing where there were duplicate failure entries of the same object-key-versionId. I think this will help avoid that case in the future.

lib/api/BackbeatAPI.js
philipyoo: Ah, I did this to fix a test case. Looking at the redis constants, maybe I can match part size instead

tests/functional/api/BackbeatServer.js
```diff
@@ -263,7 +263,20 @@ describe('Backbeat Server', () => {
             statsClient.reportNewRequest(`${site2}:${OPS_FAIL}`, 55);
             statsClient.reportNewRequest(`${site2}:${BYTES_DONE}`, 1874);
             statsClient.reportNewRequest(`${site2}:${BYTES_FAIL}`, 575);
+
+            // TODO: quick fix made for failures/pending
```
JianqinWang: ? is a quick fix needed or is this an old TODO that has been completed in this commit

tests/functional/api/BackbeatServer.js
```diff
@@ -543,9 +556,9 @@ describe('Backbeat Server', () => {
 
                 assert(res.failures.description);
                 // Failures count = OPS_FAIL
-                assert.equal(res.failures.results.count, 150);
+                assert.equal(res.failures.results.count, 1);
```
JianqinWang: the values for these comparison have been changed (no longer matching the values set OPS_FAIL and BYTES_FAIL in `before`), perhaps comment needs updating?

tests/functional/api/BackbeatServer.js
philipyoo: Yup, I want to keep this todo. I want to improve these tests, but currently not sure how atm

JianqinWang: Can you update the TODO to be more descriptive of future work? maybe something like 'refactor/improve tests for failures/pending, current tests are modified as a quick fix'. Sorry for being so picky, the current `TODO` sounds more like 'here's what happened' than a 'this is what should happen'

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,150 @@ class BackbeatAPI {
...
+     * @return {undefined}
+     */
+    _normalizeFailMetrics(keys) {
+        // ignore deleting duplicate keys for tests
+        if (process.env.CI === 'true') {
+            return;
+        }
+        async.eachLimit(keys, 10, (key, next) => {
+            async.waterfall([
+                done => this._redisClient.batch([['get', key]], (err, res) => {
```
jonathan-gramain, line 651: Since we are not targeting efficiency I think it's fine as it is, but we could also have used a single batch to get all keys.

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,150 @@ class BackbeatAPI {
...
+                    if (err) {
+                        // error occurred, skip for next time
+                        return done(err);
+                    }
+                    const [cmdErr, sourceRole] = res[0];
+                    if (cmdErr) {
+                        // error occurred, skip for next time
+                        return done(cmdErr);
+                    }
+                    if (!sourceRole) {
```
jonathan-gramain, line 661: I don't get the link between `sourceRole` missing and the key being deleted. If this is correct, can you comment to explain why?

lib/api/BackbeatAPI.js
```diff
@@ -615,6 +622,150 @@ class BackbeatAPI {
...
+    /* eslint-enable */
+
+    _filterFailObjectKeys(keys) {
+        const filteredKeys = [];
+        const uniqueObjs = {};
+        const duplicateKeys = [];
+        keys.forEach(key => {
+            // 8 parts: [bb, crr, failed, <bucket>, <key>,
+            //           <versionID>, <location>, <timestamp>]
+            const parts = process.env.CI === 'true' ?
```
jonathan-gramain, line 759: I'm not fond of having to do this specific case for the CI as it brings discrepancy between testing and production, is there an easy alternative to avoid it? E.g. are tests using an extra field that cannot be put into the production keys?

lib/api/BackbeatAPI.js
```diff
@@ -705,6 +867,23 @@ class BackbeatAPI {
                 error: errors.MalformedPOSTRequest.customizeDescription(errMsg),
             };
         }
+        // Bucket, Key, VersionId, StorageClass
+        // filter duplicate version ids if any
+        const uniqueObjs = {};
+        if (process.env.CI === 'true') {
+            // ignore filtering for tests
```
jonathan-gramain: Why do tests need to ignore filtering?

jonathan-gramain (commented):
The method looks good to me, I have a few random comments though

lib/api/BackbeatAPI.js
philipyoo: When making the `get` to redis, if no key was found in redis, `sourceRole` will be null

lib/api/BackbeatAPI.js
philipyoo: there is an additional section for the failed test keys. I tried to match the length in the key and it broke many of the tests. Given lack of time, I decided to keep things as is. But definitely, I'd like to avoid doing this

lib/api/BackbeatAPI.js
philipyoo: failure keys rely on valid bucket, key, version ids. I wasn't fully sure how the test mock metadata worked, but the version id's in the failure keys for tests use the same mock metadata. I'd like to eventually adjust the tests, and left comment here https://github.com/scality/backbeat/pull/416/files/df1c6700ef3c549deebdfe3a6b0b6a7a2dd8cbdf#diff-8c901706f0bfac9a0fd4f1d1205a4924R267

I've mainly relied on manual testing for these fixes since the bug reported mainly has to do with user usage cases or timing

lib/api/BackbeatAPI.js
jonathan-gramain: Yes, but why `sourceRole`? Is it the value stored in the key?

jonathan-gramain: Or could be the argument needed to create the `ObjectFailureEntry`

lib/api/BackbeatAPI.js
jonathan-gramain: OK, maybe worth commenting with a TODO in the code as it does not come obvious.

lib/api/BackbeatAPI.js
philipyoo: yes, `sourceRole` is the arn saved in the key

jonathan-gramain (approved):

bennettbuchanan (approved):

lib/api/BackbeatAPI.js
```diff
@@ -662,6 +848,37 @@ class BackbeatAPI {
...
+            const [cmdErr, lock] = res[0];
+            if (cmdErr) {
+                return cb(cmdErr);
+            }
+            if (!lock) {
+                // add lock and continue processing
+                return this._redisClient.batch([['set', key, 'true']], err => {
+                    if (err) {
+                        return cb(err);
+                    }
```
bennettbuchanan, line 873: Looks like we also need to check for the command error here.

lib/api/BackbeatAPI.js
```diff
@@ -662,6 +848,37 @@ class BackbeatAPI {
...
+    /**
+     * Apply a retry lock for a given request if none already applied
+     * @param {String} key - lock key
+     * @param {Function} cb - callback(err, boolean)
+     *   Where if boolean is true, a lock has been applied
+     *   Where if boolean is false, a lock already exists
+     * @return {undefined}
+     */
+    _applyRetryLockKey(key, cb) {
+        this._redisClient.batch([['get', key]], (err, res) => {
```
bennettbuchanan, line 860: Minor: using the redis command 'exists' might be a bit cleaner.

PR #417, opened 2018-09-02
bf: ZENKO-1105 fix CI functional tests

philipyoo:
Fixes:
- use correct cloudserver module name
- update package-lock
- add check in bash test file to make sure the module was installed. If not, error right away

Other changes:
- update cloudserver to latest
- separate out the functional tests (feature and api) into their own sections for more visibility

I'm not sure why the package lock file always generates huge diffs. It looks like each persons environment affects the lock file :/

### DIFF
--- a/eve/workers/unit_and_feature_tests/run_ft_tests.bash
+++ b/eve/workers/unit_and_feature_tests/run_ft_tests.bash
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -x
+set -e
+
+if [ ! -d "node_modules/@zenko/cloudserver" ]; then
+    echo "cloudserver module was not found!"
+    exit 1
+fi
+
+
+killandsleep () {
+  kill -9 $(lsof -t -i:$1) || true
+  sleep 10
+}
+
+cd node_modules/@zenko/cloudserver && npm run mem_backend & bash tests/utils/wait_for_local_port.bash 8000 40 && npm run ft_test
+
+killandsleep 8000
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -57,7 +57,13 @@ stages:
           name: run unit tests
           command: npm test
       - ShellCommand:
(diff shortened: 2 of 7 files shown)


eve/workers/unit_and_feature_tests/run_ft_tests.bash
```diff
@@ -0,0 +1,21 @@
...
+killandsleep () {
+  kill -9 $(lsof -t -i:$1) || true
+  sleep 10
+}
+
+cd node_modules/@zenko/cloudserver && npm run mem_backend & bash tests/utils/wait_for_local_port.bash 8000 40 && npm run ft_test
+
+killandsleep 8000
+
+exit $?
```
jonathan-gramain: This line should not be necessary as the script should exit with this status anyway.

jonathan-gramain (approved):
This seems to fix, so I'm good to merge it as is.

Although, inserting artificially a failed functional test to observe the behavior, I see that in this case it waits for a 1200 seconds timeout. (see e.g. https://eve.devsca.com/github/scality/backbeat/#/builders/6/builds/3115). I'm pretty sure this was also the case before, but it does not seem ideal to me so we should consider reworking the scripts to have a more immediate and clean failure report. The fact that we run commands in a subshell is also not ideal in the sense that it may mask test failures if some other step fails in the process.

philipyoo:
@jonathan-gramain I totally agree, I also saw the 1200s timeout as well. 
I'm thinking for backbeat server, export the server class, and run the server locally on a test file and closing the server after tests complete.
And, for cloudserver, run it as a process in the background in eve setup. Not sure how to do this yet, but just my thoughts on trying to fix this

jonathan-gramain:
@philipyoo we could probably run the tests in a eve container with all standard components already running as well, which would avoid to have to manage their lifetime during the tests.

giacope (approved):

PR #422, opened 2018-09-06
bugfix: ZENKO-1115 Error status for healthcheck

bennettbuchanan:
Return a 500 HTTP status in the following case where an in-sync replica array is empty during a health check:

```
{"topics":{"backbeat-replication":{"name":"backbeat-replication","partitions":[{"id":2,"leader":2,"replicas":[2],"isrs":[2]},{"id":1,"leader":-1,"replicas":[1],"isrs":[]},{"id":0,"leader":0,"replicas":[0],"isrs":[0]}]}},"internalConnections":{"isrHealth":"error","zookeeper":{"status":"ok","details":{"name":"SYNC_CONNECTED","code":3}},"kafkaProducer":{"status":"ok"}}}
```

### DIFF
--- a/lib/api/Healthcheck.js
+++ b/lib/api/Healthcheck.js
@@ -60,6 +60,16 @@ class Healthcheck {
     }
 
     /**
+     * Ensure that there is an in-sync replica for each partition.
+     * @param {object} topicMd - topic metadata object
+     * @return {boolean} true if each partition has a replica, false otherwise
+     */
+    _isMissingISR(topicMd) {
+        return topicMd.partitions.some(partition =>
+            (partition.isrs && partition.isrs.length === 0));
+    }
+
+    /**
      * Builds the healthcheck response
      * @param {function} cb - callback(error, data)
      * @return {undefined}
@@ -83,6 +93,14 @@ class Healthcheck {
             if (topicMd) {
                 topics[this._repConfig.topic] = topicMd;
                 connections.isrHealth = this._checkISRHealth(topicMd);
+            }
+            if (topicMd && this._isMissingISR(topicMd)) {
+                const error = {
+                    method: 'Healthcheck.getHealthcheck',
+                    error: 'no in-sync replica for partition',
+                    topicMd,
+                };
+                return cb(error);
             }
             Object.assign(connections, this._getConnectionDetails());
             response.topics = topics;


bennettbuchanan:
@bert-e wait

lib/api/Healthcheck.js
```diff
@@ -84,6 +94,14 @@ class Healthcheck {
                 topics[this._repConfig.topic] = topicMd;
                 connections.isrHealth = this._checkISRHealth(topicMd);
             }
+            if (topicMd && this._isMissingISR(topicMd)) {
+                const error = {
+                    method: 'Healthcheck.getHealthcheck',
+                    error: 'missing in-sync replicas',
```
jonathan-gramain: The error message could benefit to be rephrased, as it can be confused as actually missing *some* in-sync replicas, while it shall mean at least one partition has *no* in-sync replica available.

bennettbuchanan: Agree. I've updated the message to 'partition missing in-sync replica'.

lib/api/Healthcheck.js
```diff
@@ -84,6 +94,14 @@ class Healthcheck {
                 topics[this._repConfig.topic] = topicMd;
                 connections.isrHealth = this._checkISRHealth(topicMd);
             }
+            if (topicMd && this._isMissingISR(topicMd)) {
+                const error = {
+                    method: 'Healthcheck.getHealthcheck',
+                    error: 'partition missing in-sync replica',
```
jonathan-gramain: Sorry to come back again but I still see this message as ambiguous, my first understanding of this message is "ther e is a missing in-sync replica to a partition" while I should read "there is no in-sync replica for a partition", correct?

bennettbuchanan: Okay, I think the confusion is around whether the wording communicates a count of in-sync replicas. Using 'no in-sync replica ...' will avoid that problem.

lib/api/Healthcheck.js
```diff
@@ -84,6 +94,14 @@ class Healthcheck {
                 topics[this._repConfig.topic] = topicMd;
                 connections.isrHealth = this._checkISRHealth(topicMd);
             }
+            if (topicMd && this._isMissingISR(topicMd)) {
+                const error = {
+                    method: 'Healthcheck.getHealthcheck',
+                    error: 'no in-sync replica for partition',
+                    topicMd,
```
philipyoo, line 101: 👍

philipyoo (approved):

lib/api/Healthcheck.js
```diff
@@ -84,6 +94,14 @@ class Healthcheck {
                 topics[this._repConfig.topic] = topicMd;
                 connections.isrHealth = this._checkISRHealth(topicMd);
             }
+            if (topicMd && this._isMissingISR(topicMd)) {
+                const error = {
+                    method: 'Healthcheck.getHealthcheck',
```
rahulreddy, line 99: Always return `Arsenal` style errors as they carry http code. I have fixed a bugs recently as the error was not Arsenal type. It's good to be consistent.

rahulreddy (changes requested):

tmacro (approved):

lib/api/Healthcheck.js
bennettbuchanan: I agree with that practice, although I am following the pattern of this method. There is a wrapper for this method that returns the `Arsenal` error [here](https://github.com/scality/backbeat/blob/development/8.0/lib/api/BackbeatAPI.js#L275). The way logging is done in that wrapper makes it so this method returns an error object like so.

rahulreddy: Ok..it's like we are accruing unnecessary tech debt. We need to clean that up sometime.

rahulreddy (approved):

lib/api/Healthcheck.js
bennettbuchanan: I can refactor to make it consistent, but was hesitant to introduce changes beyond the scope of this healthcheck change.

bennettbuchanan: Opened a PR to come back and clean this up: https://github.com/scality/backbeat/pull/425.

PR #423, opened 2018-09-07
bf: ZENKO-1005 set offset via canary for new locations

philipyoo:
Consumer group on startup, send noop/canary entries to set current offsets

Changes in this PR:
- Update node-rdkafka version. See: https://github.com/Blizzard/node-rdkafka/issues/375
- Add specifying partition in kafka entries only for "canary" entries
- On consumer init, check consumer assignment to partition and then send off canary entries.
- On consumption, if entry is canary, skip further processing

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -480,7 +480,8 @@ class QueueProcessor extends EventEmitter {
                     consumerReady = true;
                     const paused = options && options.paused;
                     this._consumer.subscribe(paused);
-
+                });
+                this._consumer.on('canary', () => {
                     this.logger.info('queue processor is ready to consume ' +
                                      'replication entries');
                     this.emit('ready');
@@ -546,6 +547,10 @@ class QueueProcessor extends EventEmitter {
             this.logger.error('error processing source entry',
                               { error: sourceEntry.error });
             return process.nextTick(() => done(errors.InternalError));
+        }
+        if (sourceEntry.skip) {
+            // skip message, noop
+            return process.nextTick(done);
         }
         let task;
         if (sourceEntry instanceof BucketQueueEntry) {
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -3,12 +3,13 @@ const kafka = require('node-rdkafka');
 const assert = require('assert');
 const async = require('async');
 const joi = require('joi');
-
-const zookeeperHelper = require('./clients/zookeeper');
-const BackbeatProducer = require('./BackbeatProducer');
 const Logger = require('werelogs').Logger;
 const zookeeper = require('node-zookeeper-client');
 
+const zookeeperHelper = require('./clients/zookeeper');
+const BackbeatProducer = require('./BackbeatProducer');
+
+const replicationTopic = require('../conf/Config').extensions.replication.topic;
 // controls the number of messages to process in parallel
 const CONCURRENCY_DEFAULT = 1;
 const CLIENT_ID = 'BackbeatConsumer';
@@ -219,6 +220,10 @@ class BackbeatConsumer extends EventEmitter {
         } else {
             this._log.debug(`consumer is paused for topic ${this._topic}`);
         }
+
+        this._sendCanary(() => {
+            this.emit('canary');
+        });
 
         this._processingQueue = async.queue(
             this._queueProcessor, this._concurrency);
@@ -540,6 +545,69 @@ class BackbeatConsumer extends EventEmitter {
     getServiceStatus() {
         const subscriptions = this._consumer.subscription();
         return subscriptions.length > 0;
+    }
+
+    /**
+     * Helper method to wait for consumers to be assigned to partitions before
+     * proceeding to send canary messages.
+     * @param {number} wait - current accumulated wait time
+     * @param {callback} cb - callback to invoke
+     * @return {undefined}
+     */
+    _waitForAssignment(wait, cb) {
+        setTimeout(() => {
+            const assignments = this._consumer.assignments();
+            if (assignments.length === 0) {
+                if (wait > 50000) {
+                    return cb(true);
+                }
+                return this._waitForAssignment(wait + 2000, cb);
+            }
+            return cb();
+        }, 2000);
+    }
+
+    /**
+     * On start, send canary (noop) messages to set partition offsets for this
+     * consumer group. Necessary to avoid pause/resume bug where new locations
+     * on pause will not be able to queue entries as no current offset to
+     * determine number of entries (lag) to consume on resume.
+     * @param {callback} cb - callback to invoke
+     * @return {undefined}
+     */
+    _sendCanary(cb) {
+        const crrTopic = withTopicPrefix(replicationTopic);
+        if (this._topic !== crrTopic) {
+            return process.nextTick(cb);
+        }
+        return this._waitForAssignment(0, err => {
+            if (err) {
+                this._log.debug('could not send canary on consumer init, ' +
+                'waiting for consumer assignment to partition took too long', {
+                    method: 'BackbeatConsumer._sendCanary',
+                });
+                return cb();
+            }
+            const topicDetails = this._consumer._metadata.topics
+                .find(topic => topic.name === crrTopic);
+            const entries = topicDetails.partitions.map(p => {
+                const contents = '{"canary":true}';
+                return {
+                    key: 'canary',
+                    message: contents,
+                    partition: p.id,
+                };
+            });
+            const producer = new BackbeatProducer({
+                kafka: { hosts: this._kafkaHosts },
+                topic: this._topic,
+            });
+            return producer.on('ready', () => {
+                producer.send(entries, () => {
+                    producer.close(cb);
+                });
+            });
+        });
     }
 
     /**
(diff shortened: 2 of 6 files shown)


philipyoo:
@bert-e wait

lib/BackbeatConsumer.js
```diff
@@ -3,12 +3,13 @@ const kafka = require('node-rdkafka');
...
 const joi = require('joi');
+const Logger = require('werelogs').Logger;
+const zookeeper = require('node-zookeeper-client');
 
 const zookeeperHelper = require('./clients/zookeeper');
 const BackbeatProducer = require('./BackbeatProducer');
-const Logger = require('werelogs').Logger;
-const zookeeper = require('node-zookeeper-client');
 
+const replicationTopic = require('../conf/Config').extensions.replication.topic;
```
rahulreddy, line 12: Isn't this already available from the constructor as `this._topic`? This class has to be generic, enough to handle all topics.

lib/BackbeatConsumer.js
```diff
@@ -542,6 +547,68 @@ class BackbeatConsumer extends EventEmitter {
...
+                .find(topic => topic.name === crrTopic);
+            const entries = topicDetails.partitions.map(p => {
+                const contents = '{"canary":true}';
+                return {
+                    key: 'canary',
+                    message: contents,
+                    partition: p.id,
+                };
+            });
+            const producer = new BackbeatProducer({
```
rahulreddy, line 601: Can this be part of init step, so that you don't have to create a new producer and close it every time someone hits pause/resume?

rahulreddy (changes requested):

lib/BackbeatConsumer.js
philipyoo: This conditional is needed I think, otherwise the canary will be sent for all topics. I'd have to check for canary messages on all consumers that extend BackbeatConsumer. Maybe I can change it so that the QueueProcessor triggers the canary being sent? And the QueueProcessor will check if its a replication group?

lib/BackbeatConsumer.js
philipyoo: This is within the QueueProcessor `start` process. Pause/resume requests won't go through here unless we have a paused site on startup. If queueprocessor goes down and is being brought up again, it'll send the canary again even though its not a new location.
Pause/resume requests will call the kafka consumer subscribe/unsubscribe directly

Also, we need to subscribe initially and then wait for consumers to be assigned to partitions, otherwise the messages won't reach the new consumers in time

philipyoo:
ci passing here: https://eve.devsca.com/github/scality/backbeat/#/builders/9/builds/1836

lib/BackbeatConsumer.js
rahulreddy: May be not in this PR, this needs to be done cleanly.

rahulreddy (approved):

lib/BackbeatConsumer.js
```diff
@@ -542,6 +547,68 @@ class BackbeatConsumer extends EventEmitter {
...
+            return process.nextTick(cb);
+        }
+        return this._waitForAssignment(0, err => {
+            if (err) {
+                this._log.debug('could not send canary on consumer init', {
+                    method: 'BackbeatConsumer._sendCanary',
+                });
+                return cb();
+            }
+            const topicDetails = this._consumer._metadata.topics
```
philipyoo, line 591: I had trouble finding a means of getting partitions for a consumer in a synchronous way. I'm trying to make this request as quick as possible, so rather than wait on consumers md to set, i'm getting the partition info directly from the topic

lib/BackbeatProducer.js
```diff
@@ -177,14 +177,21 @@ class BackbeatProducer extends EventEmitter {
...
-                this._topic,
-                null, // partition
-                new Buffer(item.message), // value
-                item.key, // key (for keyed partitioning)
-                Date.now(), // timestamp
-                sendCtx // opaque
-            ));
+            entries.forEach(item => {
+                let partition = null;
+                if (Object.keys(item).includes('partition') &&
```
jonathan-gramain: why not simply `if (item.partition !== undefined && ...)`

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -480,7 +480,8 @@ class QueueProcessor extends EventEmitter {
                     consumerReady = true;
                     const paused = options && options.paused;
                     this._consumer.subscribe(paused);
```
jonathan-gramain, line 482: What if you subscribe when already in paused state, will you ever receive the canary?

lib/BackbeatConsumer.js
jonathan-gramain: Instead I would rather have it configured via an option to the constructor. OK to do it later if time is too short.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -547,6 +548,10 @@ class QueueProcessor extends EventEmitter {
                               { error: sourceEntry.error });
             return process.nextTick(() => done(errors.InternalError));
         }
+        if (sourceEntry.skip) {
+            // skip message, noop
```
jonathan-gramain, line 552: I would add a debug log here as it can be useful if there is any concern with canary processing.

lib/BackbeatConsumer.js
```diff
@@ -542,6 +547,68 @@ class BackbeatConsumer extends EventEmitter {
...
+     * @param {number} wait - current accumulated wait time
+     * @param {callback} cb - callback to invoke
+     * @return {undefined}
+     */
+    _waitForAssignment(wait, cb) {
+        setTimeout(() => {
+            const assignments = this._consumer.assignments();
+            if (assignments.length === 0) {
+                if (wait > 50000) {
+                    return cb(true);
```
jonathan-gramain, line 562: Can you log a warning here stating that it has timed out waiting to be assigned a partition?

lib/BackbeatConsumer.js
```diff
@@ -542,6 +547,68 @@ class BackbeatConsumer extends EventEmitter {
...
+     * @return {undefined}
+     */
+    _sendCanary(cb) {
+        const crrTopic = withTopicPrefix(replicationTopic);
+        if (this._topic !== crrTopic) {
+            return process.nextTick(cb);
+        }
+        return this._waitForAssignment(0, err => {
+            if (err) {
+                this._log.debug('could not send canary on consumer init', {
```
jonathan-gramain: Is this a normal situation? If not, can you make it an error log instead?

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: The canary wont get sent. I'm thinking that a new storage location cannot already be paused because a pause request will be rejected for a non-existing storage location. So whenever the new storage location is added to the list of boostraplist, the user can then pause.

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: there will be a lot of canary messages and I wanted to keep them silent because all consumer groups for replication topic will receive these messages. When a new storage location is added, it will send off more canaries, and all these consumers will receive again. Theres definitely room for improvement

lib/BackbeatConsumer.js
philipyoo: I added a debug log on L#585. I'll improve the message there

lib/BackbeatConsumer.js
philipyoo: I'm not sure how to best handle these situations. I added this debug log in case assignment takes too long. The 50 sec timeout is not a calculated estimate or anything, but it seems like consumers will attach to partitions fairly quickly normally. Maybe around 10-20 secs is average i'd say. 

If it doesn't assign in 50 sec, honestly probably something wrong internally with kafka and wouldn't reach this phase or fail here probably? : https://github.com/scality/backbeat/blob/4781125a0c89db2280c1ae32971686bbcdaabba4/lib/BackbeatConsumer.js#L244

JianqinWang (approved):

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: As long as messages do not go at a crazy rate like 10+ messages per second, it's good to be able to track what is going on, that is what debug logs are for. Imagine a race condition prevents the message from being received on time, or some issue with timing, it will be hard to diagnose without logs showing exactly when messages have been consumed. On kafka we can fetch the message timestamps to know when messages have been published.

I think there is value in having a debug log here, but if you really think this will be too much logging, you can leave as is.

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: Or actually, the canary should be sent but these consumers won't receive it, but should be added as `LAG`.
An improvement I can do is to only send canaries on initial subscribe because I don't think it will be possible to have a new storage location that is already paused

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: I have an issue here where we will have multiple pods, so each new queueprocessor on each pod will send the canary, so there will be duplicate entries sent. Since I get the kafka topic md and send for all existing partitions, each consumer will send a msg to each partition. I haven't been able to figure out a way to do this per consumer on a partition because a consumer may subscribe to more than one partition if one or more consumers never connect, or if we have more partitions than consumers. So again, room for improvements here

PR #424, opened 2018-09-07
bugfix: ZENKO-1102 fix code issues + new test

jonathan-gramain:
MultipleBackendTask fixes:

- use proper async callback next() when raising error after COMPLETED
  status is detected

- change error to InvalidObjectState so that it can be handle by
  skipping the entry instead of setting it to FAILED

- new specific "functional with mock" test for ZENKO-1102

### DIFF
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -285,11 +285,13 @@ class ReplicateObject extends BackbeatTask {
             const partObj = new ObjectMDLocation(part);
             return partObj.getDataStoreETag() === undefined;
         })) {
-            log.error('cannot replicate object without dataStoreETag ' +
-                      'property',
-                      { method: 'ReplicateObject._getAndPutData',
-                        entry: sourceEntry.getLogInfo() });
-            return cb(errors.InvalidObjectState);
+            const errMessage =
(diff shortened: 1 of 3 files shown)


extensions/replication/tasks/ReplicateObject.js
```diff
@@ -285,11 +285,13 @@ class ReplicateObject extends BackbeatTask {
...
-                      { method: 'ReplicateObject._getAndPutData',
-                        entry: sourceEntry.getLogInfo() });
-            return cb(errors.InvalidObjectState);
+            const errMessage =
+                  'cannot replicate object without dataStoreETag property';
+            log.error(errMessage, {
+                method: 'ReplicateObject._getAndPutData',
+                entry: sourceEntry.getLogInfo(),
+            });
+            return cb(errors.InternalError.customizeDescription(errMessage));
```
rahulreddy, line 294: Wouldn't a 500/InternalError be retried though?

jonathan-gramain: No, as it is above the retry layer here (retries happen in the `_getAndPutPart` function).

rahulreddy (approved):

philipyoo (approved):

PR #431, opened 2018-09-15
bugfix: ZENKO-1144 Use sorted sets for failures

bennettbuchanan:
Depends on https://github.com/scality/Arsenal/pull/570

Use Redis sorted sets for failed CRR API routes. 

This allows for O(1) metrics retrieval, and avoids linear scaling for API listings and retries.


bennettbuchanan:
@bert-e wait

extensions/replication/failedCRR/FailedCRRConsumer.js
```diff
@@ -160,28 +116,30 @@ class FailedCRRConsumer {
                 const entry = Buffer.from(kafkaEntry.value).toString();
                 return this._failedCRRProducer.publishFailedCRREntry(entry, cb);
             }
-            log.info('successfully set redis key');
+            if (err) {
+                log.error('could not add redis sorted set member', {
+                    error: err,
+                    data,
+                });
```
philipyoo, line 123: maybe invoke cb here before info log

tests/functional/api/BackbeatServer.js
```diff
@@ -35,16 +35,21 @@ function getUrl(options, path) {
...
-function setKey(redisClient, keys, cb) {
-    const cmds = keys.map(key => {
-        const value = 'arn:aws:iam::604563867484:test-role';
-        return ['set', `${TEST_REDIS_KEY_FAILED_CRR}:${key}`, value];
-    });
+function addMembers(redisClient, site, members, cb) {
+    const statsClient = new StatsModel(redisClient);
+    const epoch = Date.now();
+    const twelveHouseAgo = epoch - (60 * 60 * 1000) * 12;
+    let score = statsClient.normalizeTimestampByHour(new Date(twelveHouseAgo));
```
philipyoo: `_normalizeTimestampByHour`

tests/functional/api/BackbeatServer.js
```diff
@@ -35,16 +35,21 @@ function getUrl(options, path) {
...
 
-function setKey(redisClient, keys, cb) {
-    const cmds = keys.map(key => {
-        const value = 'arn:aws:iam::604563867484:test-role';
-        return ['set', `${TEST_REDIS_KEY_FAILED_CRR}:${key}`, value];
-    });
+function addMembers(redisClient, site, members, cb) {
+    const statsClient = new StatsModel(redisClient);
+    const epoch = Date.now();
+    const twelveHouseAgo = epoch - (60 * 60 * 1000) * 12;
```
philipyoo: small mispell: Hours

extensions/replication/failedCRR/FailedCRRConsumer.js
bennettbuchanan: Good catch! Updated.

tests/functional/api/BackbeatServer.js
```diff
@@ -267,7 +277,15 @@ describe('Backbeat Server', () => {
...
 
+            const testVersionId =
+                '3938353030303836313334343731393939393939524730303120203';
+            const members = [
+                `test-bucket:test-key:${testVersionId}0:${site1}`,
+                `test-bucket:test-key:${testVersionId}1:${site2}`,
+            ];
+
             return async.parallel([
+                next => addMembers(redisClient, site1, members, next),
```
philipyoo, line 288: nice 👍 my fake data added below on L#297 and L#308 may create conflicts with tests possibly?

I basically put in 10 fake entries per hour for each `site1` and `site2`. I also did not form the score and member values as I did it solely for metrics. We can probably drop my changes and use your added members instead? Or I can form my score and members to be more realistic

tests/functional/api/BackbeatServer.js
bennettbuchanan: Oh, yes. I'm going to open a PR that changes this to a public method. Will follow with one.

tests/functional/api/BackbeatServer.js
bennettbuchanan: I added +2 to those value checks. Does that work?

lib/api/BackbeatAPI.js
```diff
@@ -106,11 +107,15 @@ class BackbeatAPI {
...
     validateQuery(bbRequest) {
-        const { marker } = bbRequest.getRouteDetails();
+        const { marker, sitename } = bbRequest.getRouteDetails();
         if (marker !== undefined && (marker === '' || isNaN(marker))) {
             return errors.InvalidQueryParameter
                 .customizeDescription('marker must be a number');
         }
+        if (sitename !== undefined && sitename === '') {
+            return errors.InvalidQueryParameter
+                .customizeDescription('marker must be a string');
```
philipyoo: minor: ~~marker~~ sitename

lib/api/BackbeatAPI.js
```diff
@@ -428,23 +398,142 @@ class BackbeatAPI {
...
+        let i = 0;
+        return async.doWhilst(next => {
+            const key = keys[i++];
+            async.waterfall([
+                done => {
+                    if (score) {
+                        // If the request included a next marker we want to list
+                        // the sorted set from the member with that score.
+                        return this._redisClient
+                            .zrangebyscore(key, score, '+inf', (err, res) => {
```
philipyoo: wouldn't we want the older entries in the sorted sets starting from the marker? so '-inf' up to score

lib/api/BackbeatAPI.js
```diff
@@ -428,23 +398,142 @@ class BackbeatAPI {
...
+                                score = undefined;
+                                return done(null, res);
+                            });
+                    }
+                    return this._redisClient.zrange(key, 0, -1, done);
+                },
+                (results, done) => {
+                    if (results.length + entries.length >= listingLimit) {
+                        const index = listingLimit - entries.length;
+                        const member = results[index];
```
philipyoo: i think next marker member is the start of the sorted set, which is an older member

lib/api/BackbeatAPI.js
```diff
@@ -428,23 +398,142 @@ class BackbeatAPI {
...
+                                    return done(err);
+                                }
+                                nextMarker = score;
+                                return done(null, results, index, true);
+                            });
+                    }
+                    return done(null, results, undefined, true);
+                },
+                (results, index, hasLastMember, done) => {
+                    const slice = results.slice(0, index);
```
philipyoo: same here as above, older vs newer entries?

philipyoo (commented):
i guess comes down to semantics, but i'm seeing listing from older entries in an hourly set, but listing from newer hourly entries

tests/functional/api/BackbeatServer.js
philipyoo: sounds good

lib/api/BackbeatAPI.js
philipyoo: Sorry, it won't be the start of the sorted set, but rather the index at something like 
results.length - (listingLimit - entries.length)

lib/api/BackbeatAPI.js
```diff
@@ -482,18 +571,37 @@ class BackbeatAPI {
         });
     }
 
+    _findAndDeleteMember(member, site, cb) {
+        const keys = this._getSortedSetKeys(site);
+        let i = 0;
+        let shouldContinue = true;
+        return async.doWhilst(next => {
+            const key = keys[i++];
+            this._redisClient.zscore(key, member, (err, res) => {
```
philipyoo, line 581: 👍 I was just thinking using `zrem` on each set, but this should be a faster overall, nice!

bennettbuchanan: Nice to consider the best-case scenario, too! 💃

lib/api/BackbeatAPI.js
bennettbuchanan: You're right, I've overlooked the listing within the hour. I've updated this and subsequent values to keep it ordered.

lib/api/BackbeatAPI.js
bennettbuchanan: Updated.

lib/api/BackbeatAPI.js
bennettbuchanan: Updated.

lib/api/BackbeatAPI.js
bennettbuchanan: Nice catch!

extensions/replication/failedCRR/FailedCRRConsumer.js
```diff
@@ -24,6 +24,7 @@ class FailedCRRConsumer {
         this.logger = new Logger('Backbeat:FailedCRRConsumer');
         this._failedCRRProducer = new FailedCRRProducer(this.kafkaConfig);
         this._backbeatTask = new BackbeatTask();
+        this._statsClient = new StatsModel(redisClient);
```
rahulreddy, line 27: Is this ok to set without passing interval and expiry as params?

bennettbuchanan: Should be okay. It's not asserted in the constructor, and the only method used from that class  here is `addToSortedSet` which does not need the interval or expiry values.

extensions/replication/utils/ObjectFailureEntry.js
```diff
@@ -4,48 +4,18 @@
...
-        const schema = this.redisKey.split(':');
-        if (process.env.CI === 'true') {
-            schema.shift(); // Test keys are prefixed with 'test:'.
-        }
-        const [service, extension, status, bucket, objectKey, encodedVersionId,
-            site, timestamp] = schema;
-        this.service = service;
-        this.extension = extension;
-        this.status = status;
+    constructor(member, sitename) {
```
jonathan-gramain, line 11: FYI the role will likely be needed for the backport in progress to 7.4 and then the future 9.0, but ok to keep as it is since we focus exclusively on 8.0 for this PR, and I think it makes more sense not to rely on checking the `role` value where there is no such thing in 8.0, as it should only be a side artefact to access source metadata when a role is required.

lib/api/BackbeatAPI.js
```diff
@@ -106,11 +107,15 @@ class BackbeatAPI {
...
     validateQuery(bbRequest) {
-        const { marker } = bbRequest.getRouteDetails();
+        const { marker, sitename } = bbRequest.getRouteDetails();
         if (marker !== undefined && (marker === '' || isNaN(marker))) {
             return errors.InvalidQueryParameter
                 .customizeDescription('marker must be a number');
         }
+        if (sitename !== undefined && sitename === '') {
+            return errors.InvalidQueryParameter
+                .customizeDescription('sitename must be a string');
```
jonathan-gramain, line 117: Could rephrase as `must be a non-empty string`

lib/util/sortedSetHelper.js
```diff
@@ -1,19 +1,31 @@
...
     // NFS bucket), maintain the Redis key schema by using and empty string.
     const schemaVersionId = versionId === undefined ? '' : versionId;
-    return `${failedCRR}:${bucket}:${key}:${schemaVersionId}:${storageClass}`;
+    return `${bucket}:${key}:${schemaVersionId}`;
+}
+
+/**
+ * Returns the sorted set key.
+ * @param {String} storageClass - The storage class of the object
+ * @param {Number} epoch - The normalized epoch
```
rahulreddy: You mean normalized timestamp? epoch would mean `January 1, 1970 00:00:00 UTC`

philipyoo (approved):
lgtm :)

lib/api/BackbeatAPI.js
```diff
@@ -330,89 +335,54 @@ class BackbeatAPI {
 
     /**
      * Builds the failed CRR response.
-     * @param {String} cursor - The Redis HSCAN cursor
-     * @param {Array} keys - The collection of Redis keys for the iteration
+     * @param {Number} score - The score used for the NextMarker in the response
```
jonathan-gramain, line 338: Why not calling this `nextMarkerScore` instead

lib/util/sortedSetHelper.js
bennettbuchanan: Agree, that's misleading. I've updated to just call it `timestamp`.

lib/api/BackbeatAPI.js
```diff
@@ -428,23 +398,143 @@ class BackbeatAPI {
...
+                this._getFailedCRRResponse(nextMarker, entries, cb));
+    }
+
+    /**
+     * Get all entries across all site sorted sets.
+     * @param {String} member - The member in the sorted set to find
+     * @param {Function} cb - The callback to call
+     * @return {undefined}
+     */
+    _getEntriesAcrossSite(member, cb) {
```
jonathan-gramain, line 427: do you mean "across site**s**"? (i.e. for all sites)

lib/api/BackbeatAPI.js
```diff
@@ -428,23 +398,143 @@ class BackbeatAPI {
...
+    /**
+     * Use the current score to find all sorted sets that are >= to the score.
+     * @param {Number} score - The score to use for reduction
+     * @param {Array} keys - All the possible site keys
+     * @return {Array} - The reduced ke
+     */
+    _reduceSortedSetKeys(score, keys) {
+        const date = new Date(score);
+        const startHour = this._statsClient.normalizeTimestampByHour(date);
+        const index = keys.findIndex(key => key.includes(startHour));
```
jonathan-gramain, line 460: I'm wondering if you do not need to distinguish here, in case the `startHour` is not found in the list of keys (if that may happen at all) whether you are behind or above, in which case you shall return all keys or an empty list, respectively, but here all you will get is `-1` to pass to `keys.slice()` which I think returns all.

alexanderchan-scality (approved):

jonathan-gramain (commented):
partial review only, FYI since it's merged

jonathan-gramain (commented):
Remainder of review: LGTM (did not review tests)

extensions/replication/utils/ObjectFailureEntry.js
bennettbuchanan: Yes, we considered this and were thinking along those lines. It's possible to alter the API to send the role to the user in the response and then require it when making a retry.

lib/api/BackbeatAPI.js
bennettbuchanan: I agree, this could be cleaned up. It could happen if the marker the user gives is a timestamp they've created. 

For now, it looks harmless, as `-1` will get the last sorted set key. The response will be an empty array because that sorted set wouldn't contain a member with the given marker as its score.

PR #436, opened 2018-09-18
improvement: ZENKO-1170 move async out of trycatch

philipyoo:
Changes in this PR:
- Move async call outside of try/catch block because errors
  may be propagated back to the catch block


rahulreddy (approved):

bennettbuchanan (approved):
LGTM, thanks!

PR #438, opened 2018-09-20
improvement: change producer acks to min insync replicas

rahulreddy:
This commit sets the request.required.acks to all, which means Producer gets
an ack when the entry is committed to all of the in sync replicas (which is 3
in case of 5 node cluster and 2 in case of 3 node cluster)


jonathan-gramain (approved):
The comment could be improved, but not this PR's focus so it looks good to me.

philipyoo (approved):

alexanderchan-scality (approved):

PR #439, opened 2018-09-20
bf: ZENKO-1183 fix scheduled crr resume

philipyoo:
Changes in this PR:
- Fix wrong property name used on init 😨 
- Re-schedule resume when a triggered scheduled resume
  fails. Only error that can occur is if update to
  zookeeper fails

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -26,10 +26,10 @@ const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
 const BucketQueueEntry = require('../../../lib/models/BucketQueueEntry');
 const MetricsProducer = require('../../../lib/MetricsProducer');
 
-const { zookeeperReplicationNamespace } = require('../constants');
-const ZK_CRR_STATE_PATH = '/state';
-
 const {
+    zookeeperReplicationNamespace,
+    zkCRRStatePath,
+    zkCRRStateProperties,
     proxyVaultPath,
     proxyIAMPath,
     replicationBackends,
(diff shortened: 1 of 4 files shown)


JianqinWang (approved):

tmacro (approved):

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -383,26 +383,35 @@ class QueueProcessor extends EventEmitter {
     }
 
     scheduleResume(date) {
+        function triggerResume() {
+            this._updateZkStateNode('scheduledResume', null, err => {
+                if (err) {
+                    this.logger.error('error occurred saving state ' +
```
bennettbuchanan, line 396: Maybe add a log indicating it is being rescheduled for a minute later.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -383,26 +383,35 @@ class QueueProcessor extends EventEmitter {
...
+                    'to zookeeper', {
+                        method: 'QueueProcessor.scheduleResume',
```
bennettbuchanan: Let's add the error here as well?

bennettbuchanan (approved):

extensions/replication/queueProcessor/QueueProcessor.js
philipyoo: I'll update the message to:
`error occurred saving state to zookeeper for resuming a scheduled resume. Retry again in 1 minute`

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -383,26 +383,37 @@ class QueueProcessor extends EventEmitter {
     }
 
     scheduleResume(date) {
+        function triggerResume() {
+            this._updateZkStateNode('scheduledResume', null, err => {
```
rahulreddy, line 394: This node name `scheduledResume` is the same one which was mis-spelt, correct? Why not use a constant 
`const RESUME_NODE = 'scheduledResume';`

rahulreddy (approved):

PR #440, opened 2018-09-20
bugfix: ZENKO-1165 No duplicate entries in listing

bennettbuchanan:
Depends on https://github.com/scality/Arsenal/pull/575.

The design for storing failed entries in sorted sets using the millisecond in which the failed entry was added did not account for entries being added in the same millisecond. In such a scenario, if the listing happens to provide a marker from a score that occurs more than once in the sorted set, then some failed objects will be listed in subsequent GET requests when passed the value of `NextMarker`.

This scenario, however rare it sounds, actually did occur. Any duplicate entries were filtered out and prevented by the retry locking mechanism, but we just shouldn't have duplicate listings in the first place. 

The proposed solution is to check that the score does not occur more than once, and if it does continue searching the sorted set until the next unique score is found. This means that a listing can potentially be > 100 entries.

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -482,68 +482,59 @@ class BackbeatAPI {
             // If a given marker did not match a sorted set key.
             return process.nextTick(() => cb(null, undefined, []));
         }
-        const listingLimit = 100; // Don't exceed 100 entries in the response.
+        // Don't exceed 100 entries in the response unless there are entries
+        // in the sorted set with the same score (i.e. they were added in the
+        // same millisecond).
+        const listingLimit = 100;
         const entries = [];
         let nextMarker;
-        let shouldContinue;
+        let lastMemberScore;
         let i = 0;
         return async.doWhilst(next => {
             const key = keys[i++];
-            async.waterfall([
-                done => {
-                    if (score) {
-                        // If the request included a next marker we want to list
-                        // the sorted set from the member with that score.
-                        return this._redisClient
-                            .zrangebyscore(key, '-inf', score, (err, res) => {
-                                if (err) {
-                                    return done(err);
-                                }
-                                // Don't use the score in subsequent sorted set
-                                // range requests.
-                                score = undefined;
-                                return done(null, res);
-                            });
-                    }
-                    return this._redisClient.zrange(key, 0, -1, done);
-                },
-                (results, done) => {
-                    const totalMembers = results.length + entries.length;
-                    if (totalMembers >= listingLimit) {
-                        const index = totalMembers - listingLimit;
-                        const member = results[index - 1];
-                        if (member === undefined) {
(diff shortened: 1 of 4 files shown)


bennettbuchanan:
@bert-e wait

lib/api/BackbeatAPI.js
```diff
@@ -482,68 +482,59 @@ class BackbeatAPI {
...
+            const totalMembers = results.length / 2 + entries.length;
+            if (totalMembers >= listingLimit) {
+                resultsIndex = totalMembers * 2 - listingLimit * 2;
+                lastMemberScore = results[resultsIndex + 1];
+                nextMarker = results[resultsIndex - 1];
+                // We keep searching backward until we find a new score. In this
+                // case, we'll include additional members in the listing. If we
+                // reach the end of this sorted set, nextMarker is undefined and
+                // we go on to the next sorted set to find nextMarker.
+                while (nextMarker === lastMemberScore) {
```
philipyoo, line 522: I'm thinking this will also need to make sure `resultsIndex` >= 2

lib/api/BackbeatAPI.js
```diff
@@ -482,68 +482,59 @@ class BackbeatAPI {
...
+            }
+            const resultsSlice = results.slice(resultsIndex);
+            for (let i = 0; i < resultsSlice.length; i += 2) {
+                entries.push(new ObjectFailureEntry(resultsSlice[i], sitename));
+            }
+            // If the listing result didn't include the next marker, we should
+            // continue checking for one in subsequent sorted sets. Otherwise we
+            // only continue if there are more sorted sets.
+            return nextMarker ?
+                i < keys.length && entries.length < listingLimit :
```
philipyoo, line 535: if we have the same score continuing into the next sorted set, i think we will still need to check for that or we can run into the same problem

philipyoo (commented):
nice refactors in here :) i've added some comments for possible changes

lib/api/BackbeatAPI.js
bennettbuchanan: In this case `nextMarker` will be set to `undefined` here: https://github.com/scality/backbeat/pull/440/files#diff-7012bb0401146916f9385c1c205fe3dcR524.

That logic is a bit esoteric though, so I can make it more obvious.

lib/api/BackbeatAPI.js
bennettbuchanan: I'm relying `nextMarker` being `undefined` in that case. Since `lastMemberScore` will be defined, it ends the loop.

philipyoo (approved):
nice! lgtm :)

bennettbuchanan:
Seeing as this is an edge case and we're already far along in the RC process for Zenko 1.1., I’m putting this on hold so that https://scality.atlassian.net/browse/ZENKOIO-11  can take precedence.

PR #442, opened 2018-09-22
bugfix: ZENKO-1195 Object monitoring during retry

bennettbuchanan:
Depends on https://github.com/scality/Arsenal/pull/576.

During a retry any object-level metrics keys should be deleted so the new operation's progress is tracked and is not affected by the previous operation.


bennettbuchanan:
@bert-e wait

bennettbuchanan:
@bert-e reset

lib/api/BackbeatAPI.js
```diff
@@ -601,33 +605,29 @@ class BackbeatAPI {
...
+        const keys = [
+            getObjectBytesKey(storageClass, bucket, key, versionId),
+            getObjectBytesDoneKey(storageClass, bucket, key, versionId),
+        ];
+        return async.each(keys, (key, next) =>
+            this._redisClient.keys(`${key}*`, (err, res) => {
+                if (err) {
+                    return next(err);
+                }
+                return async.eachLimit(res, 10, (key, done) =>
```
alexanderchan-scality, line 626: the implementation looks good. was wondering if it is possible to use `batch` for deleting the multiple keys

alexanderchan-scality (approved):

lib/api/BackbeatAPI.js
```diff
@@ -601,33 +605,29 @@ class BackbeatAPI {
...
-            }
-            return cb();
-        });
+    _deleteProgressKeys(bucket, key, versionId, storageClass, cb) {
+        const keys = [
+            getObjectBytesKey(storageClass, bucket, key, versionId),
+            getObjectBytesDoneKey(storageClass, bucket, key, versionId),
+        ];
+        return async.each(keys, (key, next) =>
+            this._redisClient.keys(`${key}*`, (err, res) => {
```
jonathan-gramain, line 622: Shouldn't the filter be `${key}:*` instead?

jonathan-gramain (commented):
I wonder if we should not apply this fix to 7.4 as well

lib/api/BackbeatAPI.js
bennettbuchanan: I think either will work for this case, but I agree that keeping the `:` is a bit more readable. I will double check that the timestamp is always appended to confirm.

lib/api/BackbeatAPI.js
bennettbuchanan: 👍 I agree, that would be more efficient. My one concern is the case of a big upload where there could be hundreds of interval keys to delete. I'll just confirm some performance assumptions around `batch`.

bennettbuchanan:
@jonathan-gramain No need for this in 7.4 unless we backport this CRR monitor feature as well.

lib/api/BackbeatAPI.js
jonathan-gramain: My concern would be if we included other keys that begin with the target key as prefix (e.g. you want to process "foo" but you also process "foo/bar", I'm not sure this is a possibility but I wanted to raise the point though.

bennettbuchanan:
Seeing as this is an edge case and we're already far along in the RC process for Zenko 1.1., I’m putting this on hold so that https://scality.atlassian.net/browse/ZENKOIO-11 can take precedence.

PR #443, opened 2018-09-26
improvement: allow non-failed entries to be queued

rahulreddy:
This piggybacks on failed entries processing logic to queue stalled entries

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -650,7 +650,11 @@ class BackbeatAPI {
                     // If we cannot retrieve the source object, then we no
                     // longer can consider it a failed operation. Delete the
                     // member used to monitor the failure.
-                    return this._redisClient.zrem(key, member, next);
+                    if (key) {
+                        return this._redisClient.zrem(key, member, next);
+                    }
+                    // If it is a forced retry we do not remove the member.
+                    return next();
                 }
                 const site = failureEntry.getSite();
                 const contentLength = queueEntry.getContentLength();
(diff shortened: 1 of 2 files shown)


lib/api/BackbeatAPI.js
```diff
@@ -864,6 +864,7 @@ class BackbeatAPI {
                                 return done();
                             });
                     }
+                    entries.push({ key, member, StorageClass });
```
jonathan-gramain: I don't know what the "applyRetryLockKey` protects against, if just to avoid duplicates in the retry list, should be ok, just want to make sure you have considered what are the implications of not locking for this particular case.

jonathan-gramain (approved):

lib/api/BackbeatAPI.js
rahulreddy: I am re-using this method to queue stalled entries which don't have any redis keys set.

alexanderchan-scality (approved):

lib/api/BackbeatAPI.js
rahulreddy: I actually went ahead and made the change, it's better to use the lock to avoid race.

bennettbuchanan (approved):
I guess this allows queueing a retry of any keys passed to the API route regardless if they failed or not. Seems fine given the context.

JianqinWang (approved):

bennettbuchanan:
@bert-e wait

philipyoo (approved):

JianqinWang (approved):

bennettbuchanan (changes requested):
This has the impact that failures may not be removed from sorted sets (as it does not first check that the member exists in the sorted set) and consequently failure metrics / listings will be inaccurate.

alexanderchan-scality (approved):

bennettbuchanan (approved):

PR #445, opened 2018-09-28
bugfix: ZENKO-697 Check MD5 content for NFS CRR

bennettbuchanan:
Depends on https://github.com/scality/cloudserver/pull/1571.

When an object is non-versioned (i.e. in an NFS source bucket) we need to first check that the object state has not changes since the CRR entry was queued. These changes ensure that the latest source object is the same as that which is queued for replication by comparing the MD5 hash from the kafka entry with latest object metadata MD5 hash.

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -3,7 +3,9 @@ const uuid = require('uuid/v4');
 
 const errors = require('arsenal').errors;
 const jsutil = require('arsenal').jsutil;
+const ObjectMD = require('arsenal').models.ObjectMD;
 const ObjectMDLocation = require('arsenal').models.ObjectMDLocation;
+const BackbeatMetadataProxy = require('../utils/BackbeatMetadataProxy');
 const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
 
 const ReplicateObject = require('./ReplicateObject');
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
@bert-e wait

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -333,6 +333,47 @@
...
+                    "StorageClass": {
+                        "location": "header",
+                        "locationName": "X-Scal-Storage-Class"
+                    },
+                    "UploadId": {
+                        "location": "header",
+                        "locationName": "X-Scal-Upload-Id"
+                    }
+                },
+                "payload": "Body"
```
alexanderchan-scality: Minor: `Body` isn't defined as a member of the request.
Since the request doesn't require a body, this can be removed.

alexanderchan-scality (approved):

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -578,6 +702,93 @@ class MultipleBackendTask extends ReplicateObject {
...
+     * @param {ObjectQueueEntry} sourceEntry - The source object entry
+     * @param {ObjectQueueEntry} destEntry - The destination object entry
+     * @param {Werelogs} log - The logger instance
+     * @param {Function} cb - The callback to call
+     * @return {undefined}
+     */
+    _checkObjectState(sourceEntry, destEntry, log, cb) {
+        return this._getSourceMD(sourceEntry, destEntry, log, (err, res) => {
+            if (err && err.code === 'ObjNotFound' &&
+            !sourceEntry.getIsDeleteMarker()) {
```
philipyoo, line 730: just a question, if the err code is 'ObjNotFound' and the source entry is a delete marker, why wouldn't this be considered InvalidObjectState?

bennettbuchanan: In the case that the source object is deleted we want to replicate the delete marker while skipping any current CRR operations for the previous object (i.e. the one that was just deleted). If it's CRR of a delete marker we are checking the state of, then this is a valid state of the source.

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -789,7 +1023,16 @@ class MultipleBackendTask extends ReplicateObject {
...
             (sourceRole, next) =>
-                this._refreshSourceEntry(sourceEntry, log, next),
+                this._refreshSourceEntry(sourceEntry, log, (err, res) => {
+                    if (err && err.code === 'ObjNotFound' &&
+                        sourceEntry.getReplicationIsNFS() &&
+                        !sourceEntry.getIsDeleteMarker()) {
+                        // The object was deleted before entry is processed, we
+                        // can safely skip this entry.
+                        return next(errors.InvalidObjectState);
+                    }
```
philipyoo, line 1033: i think we might need to catch a general err here as well

bennettbuchanan: Oh, nice catch! Will update this. Thanks 😄

philipyoo (approved):
lgtm 👍

rahulreddy:
@bert-e reset

rahulreddy:
@bert-e reset

bennettbuchanan:
@bert-e help

PR #446, opened 2018-10-01
bf: fix kafka consumer topic params field

philipyoo:
Changes in this PR:
- Kafka consumer param 'auto.offset.reset' should be applied
  as a topic level param


bennettbuchanan (approved):
Nice catch 🚀In the kafka docs it does look like a global consumer config, while in `node-rdkafka` it is listed under topic configs, so this LGTM.

jonathan-gramain (approved):

PR #447, opened 2018-10-03
bf: ZENKO-1020 eodm rule change for delete markers

philipyoo:
Changes in this PR:
- ExpiredObjectDeleteMarker applies the same for delete
  markers in a versioning suspended/enabled bucket. It
  does not matter if the delete marker has a null
  version id or not.
- Expiration rule for IsLatest applies normally to
  Versions
- Expiration for expired object delete markers still
  apply. If we have an IsLatest delete marker that has 1+
  non-current versions, expiration rule will not apply

### DIFF
--- a/tests/functional/lifecycle/LifecycleTask.js
+++ b/tests/functional/lifecycle/LifecycleTask.js
@@ -818,6 +818,172 @@ describe('lifecycle task functional tests', function dF() {
             });
         });
 
+        // 1 Day Expiration rule - handling the IsLatest versions
+        [
+            {
+                message: 'should expire a version in a versioning enabled ' +
+                    'bucket with 0 non-current versions using basic ' +
+                    'expiration rule',
+                isDeleteMarker: false,
+                hasNonCurrentVersions: false,
+                versionStatus: 'Enabled',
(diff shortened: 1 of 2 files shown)


philipyoo:
@bert-e wait

tests/functional/lifecycle/LifecycleTask.js
```diff
@@ -915,59 +915,6 @@ describe('lifecycle task functional tests', function dF() {
             });
         });
 
-        it('should expire a version or delete marker in a versioning ' +
```
dora-korpar, line 918: Can you test that versions/delete markers are NOT being expired in a versioning suspended bucket? Or is that covered in like e2e tests?

philipyoo: yup for sure, will add these tests soon!

philipyoo:
I think this implementation is still not correct :/ I'll update the jira according to what each expected behavior is per scenario and update these PR's accordingly

bennettbuchanan (approved):

alexanderchan-scality (approved):

PR #449, opened 2018-10-03
improvement: ZENKO-1217 tunables for retries

rahulreddy:
Expose tunable for retries for replication queue processor and status
processor


docker-entrypoint.sh
```diff
@@ -113,6 +113,14 @@ if [[ "$EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST" ]]; then
     fi
 fi
 
+if [[ "$EXTENSIONS_REPLICATION_REPLICATION_STATUS_PROCESSOR_RETRY_TIMEOUT_S" ]]; then
```
philipyoo, line 116: should we just reduce the name to just use "replication" once?

bennettbuchanan (approved):

docker-entrypoint.sh
rahulreddy: They have separate configs at the moment, so I introduced both. And honestly, there is no need to change replication status processor retries. It's the QueueProcessor that counts as it affects the backoff time for retries to cloud locations.

anurag4DSB (approved):

PR #450, opened 2018-10-05
feature: configurable retry params for clouds

rahulreddy:
This introduces configurable retry parameters for all classes that use
BackbeatTask for handling retry mechanism. For now, for Zenko, retry
tunables are available per-cloud which affect how many times an entry
is retried during temporary failures.


rahulreddy:
@bert-e wait

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -34,10 +34,14 @@ class ReplicateObject extends BackbeatTask {
     constructor(qp) {
         const qpState = qp.getStateVars();
         super({
-            retryTimeoutS: qpState.repConfig.queueProcessor.retryTimeoutS,
+            retry: qpState.repConfig.queueProcessor.retry,
         });
         Object.assign(this, qpState);
-
+        this.destType = this.destConfig.bootstrapList
+        .find(endpoint => endpoint.site === this.site);
```
bennettbuchanan: After finding the site object: e.g.

```
{ "site": "aws-location", "type": "aws_s3" }
```

It looks like we need to also get the `type` from that object.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -34,10 +34,14 @@ class ReplicateObject extends BackbeatTask {
...
+        const retryParams = this.repConfig.queueProcessor.retry[this.destType];
```
bennettbuchanan: We'll have to have a way of accounting for a bootstrap list type of `aws_s3` and the newly added retry config property `aws`.

conf/config.json
```diff
@@ -97,7 +97,48 @@
...
+                        "maxRetries": 5,
+                        "timeoutS": 300,
+                        "backoff": {
+                            "min": 1000,
+                            "max": 30000,
+                            "jitter": 0.1,
+                            "factor": 1.5
+                        }
+                    },
+                    "default": {
```
bennettbuchanan: Looks like this will also apply to a Scality RING destination because there is no `site` property. Thinking aloud here: maybe we could move the default values to the Joi config validator (i.e. using the `default()` method) and then rename this something like "scality".

lib/tasks/BackbeatTask.js
```diff
@@ -26,10 +24,9 @@ class BackbeatTask {
     retry(args, done) {
         const { actionDesc, logFields,
                 actionFunc, shouldRetryFunc, onRetryFunc, log } = args;
-        const backoffCtx = new BackOff(BACKOFF_PARAMS);
+        const backoffCtx = new BackOff(this.retryParams.backoff);
```
bennettbuchanan, line 23: 👍

extensions/replication/tasks/ReplicateObject.js
rahulreddy: Ah you are right. I should be pulling `type` instead of site.

extensions/replication/tasks/ReplicateObject.js
rahulreddy: Yeah it should be called `aws_s3`. I need to find a way to have one source of truth for types (refactor for later).

conf/config.json
rahulreddy: Yeah, that's a good idea. I will make the change.

conf/config.json
```diff
@@ -97,12 +105,61 @@
             "monitorReplicationFailureExpiryTimeS": 86400,
             "queueProcessor": {
                 "groupId": "backbeat-replication-group",
-                "retryTimeoutS": 300,
+                "retry": {
+                    "aws": {
+                        "maxRetries": 5,
+                        "timeoutS": 300,
+                        "backoff": {
+                            "min": 1000,
```
jonathan-gramain: Didn't you want to take the opportunity to adjust the default values for all clouds, e.g. setting a larger min backoff and larger timeout by default? Not perfect for all environments of course but hopefully closer to reasonable defaults in production env.

lib/tasks/BackbeatTask.js
```diff
@@ -1,35 +1,16 @@
...
-    constructor(config) {
-        this.config = joi.attempt(config || {}, backbeatTaskConfigJoi,
-                                  'invalid backbeat task config');
-    }
 
     retry(args, done) {
         const { actionDesc, logFields,
                 actionFunc, shouldRetryFunc, onRetryFunc, log } = args;
-        const backoffCtx = new BackOff(BACKOFF_PARAMS);
+        const backoffCtx = new BackOff(this.retryParams.backoff);
```
jonathan-gramain, line 23: Since we don't provide the retry params in the constructor anymore, I would rather populate them with default values in the constructor, and provide a setter to change them. That would avoid taking the risk of forgetting to set the retry params and raising an exception.

lib/tasks/BackbeatTask.js
```diff
@@ -42,30 +23,32 @@ class BackbeatTask {
...
 
                 const now = Date.now();
-                if (now > (startTime + self.config.retryTimeoutS * 1000)) {
-                    log.error(
-                        `retried for too long to ${actionDesc}, giving up`,
+                const retriesExceeded = nbRetries > this.retryParams.maxRetries;
+                const timeoutReached = now >
+                    (startTime + this.retryParams.timeoutS * 1000);
+                if (retriesExceeded || timeoutReached) {
+                    log.error('giving up replication as retries exceeded',
```
jonathan-gramain: `actionDesc` should be used for the log instead of "replication", also I would rephrase "retries exceeded" to "giving up retries" to be agnostic as to why we stopped retrying (the reason can be found as `retriesExceeded` or `timeoutReached` in the log params).

lib/tasks/BackbeatTask.js
```diff
@@ -42,30 +23,32 @@ class BackbeatTask {
...
             }
             if (nbRetries > 0) {
-                const now = Date.now();
-                log.info(`succeeded to ${actionDesc} after retries`,
-                         Object.assign({
-                             nbRetries,
-                             retryTotalMs: `${now - startTime}`,
-                         }, logFields || {}));
+                const retryTotalMs = Date.now() - startTime;
+                log.info('replication succeeded after retries',
```
jonathan-gramain: Same thing, use `actionDesc`.

jonathan-gramain (commented):
Some early comments

jonathan-gramain:
I was thinking also as a general idea, could it be better to consider stopping retries after *both* timeout *and* maximum number of retries are exceeded, instead of *either* one? I'm considering the case of putting a very large object, where it can be useful to retry, say at least once, but the timeout will surely be exceeded after the first failure if not very large. As an alternative we could also provide a minimum number of retries.

conf/config.json
rahulreddy: Yeah this PR is a bit premature at the moment. I will be apply a default of 15 min for all clouds. Since we don't have any real world data at the moment we can live with that and the tunable should help as it can be changed from Zenko.

lib/tasks/BackbeatTask.js
rahulreddy: I will make it more generic as it will be easier to filter and search for that specific `actionDesc` in Kibana.

rahulreddy:
I think having both timeout and minimum number of retries would have negative effect on small objects esp. 0-byte ones or MD updates only entries. I like the idea of doing min. retries. but it's confusing as to which one trumps the other (#of retries or timeout). It's easier to think of the logic the other way around i.e `either or`, it would mean we would retry the object upto the maxRetries param in the case of small objects and we would try at least timeoutS (this needs to be adjusted to factor in large objects).

jonathan-gramain:
Actually, for 0 byte object, or any small object, it seems to me desirable to satisfy at least the maximum retry time specified, to guarantee we can replicate everything if a cloud is down for some time less than this maximum, absorbing transient unavailability (may be due to the cloud but also to the local client network as well).

On the other hand, large objects may still be retried for longer, because they may have been hit by a random connectivity issue in the middle of a large transfer. In either case, especially for small objects, the total number of retries will be kept low by the exponential backoff (so roughly `nretries = O(log(retryTimeout/minRetryDelay))` which should be relatively low, ignoring the `maxRetryDelay` param).

rahulreddy:
> Actually, for 0 byte object, or any small object, it seems to me desirable to satisfy at least the maximum retry time specified, to guarantee we can replicate everything if a cloud is down for some time less than this maximum, absorbing transient unavailability (may be due to the cloud but also to the local client network as well).

I don't think factoring issues like cloud downtime is feasible, that would be handled by the manual retrying of failed objects feature. This is to absorb interim nuances like AWS giving 404s even when bucket exists, socket hangups etc. At the moment we are either retrying too quickly or failing too quickly, that's what we are trying to address in this PR.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -33,11 +33,19 @@ class ReplicateObject extends BackbeatTask {
...
-            retryTimeoutS: qpState.repConfig.queueProcessor.retryTimeoutS,
-        });
+        super();
         Object.assign(this, qpState);
-
+        this.retryParams = this.repConfig.queueProcessor.retry.scality;
+        if (this.destConfig && this.destConfig.bootstrapList) {
+            const destination = this.destConfig.bootstrapList
+                .find(endpoint => endpoint.site === this.site) || {};
+            this.destType = destination.type;
```
philipyoo, line 42: I think we can remove this?

rahulreddy: I wanted to change this https://github.com/scality/backbeat/blob/development/8.1/extensions/replication/tasks/MultipleBackendTask.js#L22 as this is being evaluated on every request.
I will follow up with another PR for that.

philipyoo (approved):
lgtm 👍

conf/config.json
```diff
@@ -97,12 +106,62 @@
             "monitorReplicationFailureExpiryTimeS": 86400,
             "queueProcessor": {
                 "groupId": "backbeat-replication-group",
-                "retryTimeoutS": 300,
+                "retry": {
+                    "aws_s3": {
+                        "maxRetries": 5,
+                        "timeoutS": 900,
```
bennettbuchanan, line 107: This increase looks good to me. Just want to confirm that it's intended to increase the default timeout for public cloud locations.

lib/tasks/BackbeatTask.js
```diff
@@ -1,37 +1,30 @@
...
-    constructor(config) {
-        this.config = joi.attempt(config || {}, backbeatTaskConfigJoi,
-                                  'invalid backbeat task config');
+
+    constructor() {
+        this.retryParams = {
+            timeoutS: 300,
+            backoff: {
+                min: 1000,
+                max: 300000,
```
bennettbuchanan, line 13: ~300000~ 30000

lib/config/configItems.joi.js
```diff
@@ -51,10 +51,22 @@ const zenkoAuthJoi = joi.object({
     account: joi.string().required(),
 });
 
+const retryParamsJoi = joi.object({
```
bennettbuchanan, line 54: Seems like these default values are a combination of scality and public cloud destinations. Also, wondering why `0.2` jitter, whereas elsewhere it is consistently `0.1`.

lib/tasks/BackbeatTask.js
```diff
@@ -42,33 +35,40 @@ class BackbeatTask {
...
-                    log.error(
-                        `retried for too long to ${actionDesc}, giving up`,
+                const retriesMaxedOut = nbRetries >=
+                    this.retryParams.maxRetries;
+                const timeoutReached = now >=
+                    (startTime + this.retryParams.timeoutS * 1000);
+                // Give up if max retries reached or if timeout reached and
+                // the entry has been retried at least once
+                if (retriesMaxedOut || (timeoutReached && nbRetries > 0)) {
+                    log.error('giving up processing as retries are maxed out',
```
bennettbuchanan: Minor: could be useful to make the message conditional whether the timeout or number of retries was reached.

bennettbuchanan (approved):
I've left some minor comments around config value consistency and logging. Otherwise, LGTM! 🚢

conf/config.json
rahulreddy: That's correct, 15 min. or 5 retries should be a good start. We will refine this as we know better.

lib/tasks/BackbeatTask.js
rahulreddy: It's 300K ms. as in 300S which is the timeout. Isn't that correct?

lib/tasks/BackbeatTask.js
rahulreddy: I added that as log properties so that the message can be used for searching consistently and filtering the log properties `retriesMaxedOut` or `timeoutReached` would give the info where it timed out or had max retries reached.

lib/tasks/BackbeatTask.js
bennettbuchanan: I see, it looks like the other values set at `30000` need to be updated in that case.

lib/config/configItems.joi.js
rahulreddy: It's a remnant of changes that did not propagate. I will amend this, `0.2` or `0.1` has medium impact, so just sticking to one for consistency.

lib/tasks/BackbeatTask.js
bennettbuchanan: Good point re: searching logs.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -58,7 +58,7 @@ class QueueProcessor extends EventEmitter {
      *   specific to queue processor
      * @param {String} repConfig.queueProcessor.groupId - kafka
      *   consumer group ID
-     * @param {String} repConfig.queueProcessor.retryTimeoutS -
+     * @param {Object} repConfig.queueProcessor.retry -
      *   number of seconds before giving up retries of an entry
```
jonathan-gramain: The comment is off now

lib/tasks/BackbeatTask.js
rahulreddy: I just saw that, thanks for pointing out. I updated the values correctly to be 300K

lib/tasks/BackbeatTask.js
jonathan-gramain: "retries are maxed out" may be understood as "number of retries maxed out", while it may also have timed out

jonathan-gramain (approved):

rahulreddy:
@bert-e reset

PR #454, opened 2018-10-10
bugfix: ZENKO-697 Do not skip completed NFS CRR

bennettbuchanan:
If an object is overwritten during the time window that the initial object is being streamed to the destination, the refreshed source entry for the overwritten object will be COMPLETED. In this scenario we don't want to skip the entry. Instead, we want to replicate this object so that it becomes the latest version on the destination.


bennettbuchanan:
@bert-e wait

alexanderchan-scality (approved):

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -1048,7 +1048,9 @@ class MultipleBackendTask extends ReplicateObject {
                     site: this.site, siteStatus: status,
                     content,
                 });
-                if (status === 'COMPLETED' && content.includes('DATA')) {
+                if ((status === 'COMPLETED' &&
+                    !sourceEntry.getReplicationIsNFS()) &&
```
jonathan-gramain: I am not fond of this solution, as this code has been introduced to fix a critical bug with transient source (https://scality.atlassian.net/browse/ZENKO-1102).

I don't recall if we decided that NFS was not compatible with transient source, if so it should not be an issue, if not we have to take this into account and see if it's acceptable.

Let's discuss this tomorrow.

philipyoo (approved):
lgtm 👍

jonathan-gramain (approved):

PR #490, opened 2018-10-23
feature: S3C-1640 Add functional tests

bennettbuchanan:
Add functional tests that use role to authenticate requests when retrying CRR operations.

* Refactor functional tests into individual test suites based on feature.
* Run the retry tests two times: once with account auth, and another time with role auth


index.js
```diff
@@ -7,6 +7,11 @@ const runServer = require('./lib/api/BackbeatServer');
 const Config = process.env.CI !== 'true' ? require('./conf/Config') :
     require('./tests/config.json');
```
jonathan-gramain: There can be a good reason to specifically load a test file here, but if at all possible we should set BACKBEAT_CONFIG_FILE env var and let Config.js parse it (would avoid duplicating special cases and allow config validation to take place).

jonathan-gramain (approved):

index.js
bennettbuchanan: I agree that would be cleaner. 👍I'll aim to add multiple config files each auth case.

philipyoo (approved):
nice! lgtm 👍

bennettbuchanan:
@bert-e wait

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@ironman-machine reset

bennettbuchanan:
@bert-e reset

tests/functional/api/retry.js
```diff
@@ -0,0 +1,678 @@
...
+const config = require('../../config.json');
+const getUrl = require('../utils/getUrl');
+const { addMembers, addManyMembers } = require('../utils/sortedSetHelpers');
+const { makeRetryPOSTRequest, getRequest, getResponseBody } =
+    require('../utils/httpHelpers');
+const fakeLogger = require('../utils/fakeLogger');
+
+const redisConfig = { host: '127.0.0.1', port: 6379 };
+const redisClient = new RedisClient(redisConfig, fakeLogger);
+const useRoleAuth = process.env.CI_AUTH_TYPE === 'role';
```
jonathan-gramain: Now maybe it could make more sense to check the "authType" specified in the config rather than rely on an env var.

bennettbuchanan: Good call, updated.

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

jonathan-gramain (approved):

PR #498, opened 2018-10-26
ft: ZENKO-1175 tailable cursor to consume mongo oplog

jonathan-gramain:
Use a tailable custor to keep ordering guarantees for the records we
read. This also means we have to read from the beginning when we
reconnect (at startup), and start processing when we encountered the
unique ID previously stored in zookeeper.

Also removed dispatcher mode with MongoLogReader (was only used for
the short-lived Federation deployment of Zenko).

Depends on PR https://github.com/scality/Arsenal/pull/594

### DIFF
--- a/bin/ingestion.js
+++ b/bin/ingestion.js
@@ -29,20 +29,14 @@ function queueBatch(queuePopulator, taskState, qConfig, log) {
     log.debug('start queueing replication batch');
     taskState.batchInProgress = true;
     const maxRead = qpConfig.batchMaxRead;
-    queuePopulator.processAllLogEntries({ maxRead }, (err, counters) => {
+    queuePopulator.processLogEntries({ maxRead }, err => {
         taskState.batchInProgress = false;
         if (err) {
             log.error('an error occurred during replication', {
                 method: 'QueuePopulator::task.queueBatch',
                 error: err,
             });
-            return undefined;
         }
-        const logFunc = (counters.some(counter =>
-            Object.keys(counter.queuedEntries).length > 0) ?
-            log.info : log.debug).bind(log);
-        logFunc('replication batch finished', { counters });
-        return undefined;
     });
     return undefined;
 }
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -326,33 +326,23 @@ class IngestionPopulator {
         return process.nextTick(done);
     }
 
-    _processAllLogEntries(params, done) {
+    _processLogEntries(params, done) {
         return async.map(
             this.logReaders,
-            (logReader, done) => logReader.processAllLogEntries(params, done),
-            (err, results) => {
-                if (err) {
-                    return done(err);
-                }
-                const annotatedResults = results.map(
-                    (result, i) => Object.assign(result, {
-                        logSource: this.logReaders[i].getLogInfo(),
-                        logOffset: this.logReaders[i].getLogOffset(),
-                    }));
-                return done(null, annotatedResults);
-            });
+            (logReader, cb) => logReader.processLogEntries(params, cb),
+            done);
     }
 
-    processAllLogEntries(params, done) {
+    processLogEntries(params, done) {
         if (this.logReadersUpdate.length >= 1) {
             return this._setupUpdatedReaders(err => {
                 if (err) {
                     return done(err);
                 }
-                return this._processAllLogEntries(params, done);
+                return this._processLogEntries(params, done);
             });
         }
-        return this._processAllLogEntries(params, done);
+        return this._processLogEntries(params, done);
     }
 
     zkStatus() {
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -1,5 +1,7 @@
 const async = require('async');
 
+const jsutil = require('arsenal').jsutil;
+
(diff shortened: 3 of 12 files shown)


jonathan-gramain:
@bert-e wait

JianqinWang (approved):

lib/queuePopulator/LogReader.js
```diff
@@ -213,74 +217,69 @@ class LogReader {
...
+                };
+                // Use heuristics to log when:
+                // - at least one entry is pushed to any topic
+                // - many records have been skipped (e.g. when starting up
+                //   consumption from mongo oplog)
+                // - the batch took a significant time to complete.
+                const useInfoLevel =
+                      Object.keys(stats.queuedEntries).length > 0 ||
+                      stats.skippedRecords > 1000 ||
+                      stats.elapsedMs > 10000;
```
philipyoo: I think this is undefined. Is it supposed to be based off of `batchState.startTime`?

lib/queuePopulator/LogReader.js
```diff
@@ -316,21 +318,87 @@ class LogReader {
...
+        };
+        const endEventHandler = () => {
+            logger.debug('ending record stream', {
+                method: 'LogReader._processPrepareEntries',
+            });
+            return shipBatchCb();
+        };
+        logRes.log.on('data', dataEventHandler);
+        logRes.log.on('error', errorEventHandler);
+        logRes.log.on('end', endEventHandler);
```
philipyoo, line 349: maybe an event might emit before `shipBatchCb` is set

lib/queuePopulator/LogReader.js
jonathan-gramain: Good catch, was a left-over from a previous way to check batch timeout, I will simply remove this check.

lib/queuePopulator/LogReader.js
jonathan-gramain: I don't think it's a possibility, since new event handlers should not get events already queued for delivery, I think, only the newer events will be dispatched, which can occur only after we return from the function `_processPrepareEntries()`

jonathan-gramain:
Update:

Testing has shown that when doing pause/resume on the cursor or streams linked to mongo oplog, if an error occurs during the time the cursor is paused, no event (especially `'close'` or `'error'`) is forwarded to the consumer, even after a resume. This means the queue populator can be stuck for an indefinite time if mongodb connection has an issue or if mongodb pods die, until manually restarted.

When errors occur while the cursor (or its attached transform stream) is not paused, even when the tailable cursor is idle waiting for new data to come, those errors are correctly received as events.

One way to deal with this would be to avoid doing pause/resume, but this requires some design changes to ensure flow control, by using a down-to-top instead of top-down approach. It would mean stepping away from using cron-style jobs triggered from the highest level, instead have a long-running log reader task, and trigger `producer.send()` to kafka when the cursor has retrieved enough records or when a timeout is reached (because we still need to regularly commit the current offset to zookeeper, we still need a timeout). It will be also a more efficient approach for a tailable cursor as we will not be idle for some time between batches even when we're actively reading records, as it is the case in the current state of this PR.

I'm aware we need to keep compatibility with non-tailable logs (raft) at least for the ingestion feature. I can think about either making them both compatible with the new design, or separating into two different queue populator flavors, deployed with two separate deployments and pods, with the current design kept for ingestion, and the new design for mongo log consumption.

Comments welcome!

JianqinWang (approved):

bin/ingestion.js
```diff
@@ -29,20 +29,14 @@ function queueBatch(queuePopulator, taskState, qConfig, log) {
     log.debug('start queueing replication batch');
     taskState.batchInProgress = true;
     const maxRead = qpConfig.batchMaxRead;
-    queuePopulator.processAllLogEntries({ maxRead }, (err, counters) => {
+    queuePopulator.processLogEntries({ maxRead }, err => {
```
ploki, line 32: Isn't processLogEntries supposed to take a second parameter for a callback function?

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -326,33 +326,23 @@ class IngestionPopulator {
...
-                    return done(err);
-                }
-                const annotatedResults = results.map(
-                    (result, i) => Object.assign(result, {
-                        logSource: this.logReaders[i].getLogInfo(),
-                        logOffset: this.logReaders[i].getLogOffset(),
-                    }));
-                return done(null, annotatedResults);
-            });
+            (logReader, done) => logReader.processLogEntries(params, done),
```
rahulreddy: `done` works because of variable scoping but I think using `next` or `cb` or something is more readable.

bin/ingestion.js
jonathan-gramain: Not anymore, as the stats are now logged inside LogReader class, I deemed useless to expose them in the callback.

rahulreddy (approved):

jonathan-gramain:
@bert-e reset

philipyoo:
On your comment about non-tailable logs, I think our plan for now is to implement our own populator (for ingestion) that will create/add readers on new ingestion buckets. Maybe we can consider a new design once things are more fleshed out for ingestion

jonathan-gramain:
@philipyoo sounds like a good idea, at least in the meantime. It also avoids the risk of mongo populator impacting the ingestion from raft.

PR #507, opened 2018-11-05
bf: fix ci tests

philipyoo:
CI functional tests were not being run because modules were missing. https://eve.devsca.com/github/scality/backbeat/#/builders/6/builds/3678/steps/8/logs/stdio

Thomas helped in order to delete images on registry, but another problem:
https://eve.devsca.com/github/scality/backbeat/#/builders/6/builds/3726

Removing node modules before installing seems to have helped.
Other changes address the error not being escalated. We were getting green checks even though the install step failed and didn't run any tests. Now the error should be shown and escalated. For example, same error, but with the bash script changes:
https://github.com/scality/backbeat/compare/bugfix/testing-eve-failures?expand=1

Changes in this PR:
- Zenko cloudserver package name changed to match cloudserver
  (named based on a fresh git install of the module).
- Ensure previous node modules are removed and reinstalled.
- In bash files, do not chain server run with npm script. We can capture error on either step this way
- The pause/resume tests were failing because it relied on hard-coded zk path (that recently changed, oops..). Rely on paths defined in constants file instead

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -7,7 +7,6 @@ branches:
   development/*:
     stage: "post-merge"
 
-
 models:
   - SetProperty: &docker_image_name
       name: Set docker image name property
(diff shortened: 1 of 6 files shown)


philipyoo:
sorry, closing and will tag with a jira + re-target for 7.4

PR #513, opened 2018-11-05
bf: ZENKO-1306 eve test failing due to cloudserver

philipyoo:
CI functional tests were not being run because modules were missing. https://eve.devsca.com/github/scality/backbeat/#/builders/6/builds/3678/steps/8/logs/stdio

Thomas helped in order to delete images on registry, but another problem:
https://eve.devsca.com/github/scality/backbeat/#/builders/6/builds/3726

Removing node modules before installing seems to have helped. Also, cloudserver repo name updated to match what a fresh npm install of the module name would be.

Other changes address the error not being escalated. We were getting green checks even though the install step failed and didn't run any tests. Now the error should be shown and escalated. For example, same error, but with the bash script changes:
https://github.com/scality/backbeat/compare/bugfix/testing-eve-failures?expand=1

Changes in this PR:
- Zenko cloudserver package name changed to match cloudserver (named based on a fresh git install of the module). Update package-lock as well
- In bash files, do not chain server run with npm script. We can capture error on either step this way
- The pause/resume tests were failing because it relied on hard-coded zk path (that recently changed, oops..). Rely on paths defined in constants file instead
- Any functional test files not within a subdir has been moved to a directory called `models`. This is so we can easily run individual tests by feature/category
- Re-enable functional tests: lifecycle, models


philipyoo:
@bert-e reset

jonathan-gramain (approved):

philipyoo:
@bert-e wait

philipyoo:
@bert-e reset

package.json
```diff
@@ -14,6 +14,8 @@
     "mongo_queue_processor": "node extensions/mongoProcessor/mongoProcessorTask.js",
     "garbage_collector": "node extensions/gc/service.js",
     "test": "mocha --recursive tests/unit",
+    "ft_test:models": "mocha --recursieve $(find tests/functional/models -name '*.js')",
+    "ft_test:lifecycle": "mocha --recursive $(find tests/functional/lifecycle -name '*.js')",
```
bennettbuchanan, line 19: Thanks for catching this 👍

bennettbuchanan (approved):

philipyoo:
@bert-e wait

package.json
```diff
@@ -14,6 +14,8 @@
     "mongo_queue_processor": "node extensions/mongoProcessor/mongoProcessorTask.js",
     "garbage_collector": "node extensions/gc/service.js",
     "test": "mocha --recursive tests/unit",
+    "ft_test:models": "mocha --recursieve $(find tests/functional/models -name '*.js')",
```
jonathan-gramain: I would rather call suffix with `:lib`, as "model" usually applies specifically to object mapping to storage underneath, where we're actually testing some general library code (including models)

philipyoo:
@bert-e reset

package.json
```diff
@@ -14,6 +14,8 @@
     "mongo_queue_processor": "node extensions/mongoProcessor/mongoProcessorTask.js",
     "garbage_collector": "node extensions/gc/service.js",
     "test": "mocha --recursive tests/unit",
```
bennettbuchanan, line 16: I think it could also be worthwhile for development to reintroduce the global `ft_test` command to run all functional tests. i.e.

`"ft_test": "mocha --recursieve $(find tests/functional -name '*.js')",`

philipyoo:
@bert-e wait

package.json
philipyoo: what do you think about the directory name? I wasn't sure honestly, so named it "models". I think I'll rename to "lib" for now

jonathan-gramain: Yes, lib is good as it matches the suffix it's less confusing.

jonathan-gramain: It also matches the directory where the code being tested lies, which can help to link the code and tests together.

philipyoo: Great, thank you jonathan!

philipyoo:
@bert-e reset

philipyoo:
@bert-e help

bennettbuchanan (approved):

philipyoo:
@reset

philipyoo:
@bert-e reset

philipyoo:
@bert-e force-reset

philipyoo:
@bert-e force_reset

bennettbuchanan (approved):

jonathan-gramain (approved):

PR #522, opened 2018-11-13
ft: ZENKO-1240 ingestion init management populator

philipyoo:
Sits on top of PR https://github.com/scality/backbeat/pull/521

**Purpose of this PR**:
Add the management layer, check for new ingestion sources (currently querying mongo directly), and add/remove ingestion readers accordingly.
Each config for ingestion sources can use a refactor, but to keep compliance with existing code, I kept the ingestion source config as is with minor changes.

**Changes in this PR**:
- Add management layer to get config location constraint
  updates
- Each queueBatch call, update ingestion sources list first

PR #525, opened 2018-11-15
ZENKO-1235: logreader to filter by bucket

JianqinWang:


JianqinWang:
@bert-e wait

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -241,10 +243,15 @@ class IngestionPopulator {
...
                         qpConfig: this.qpConfig,
                         s3Config: this.s3Config,
+                        bucket: this.bucket,
                     });
+                    return ingestionReader.setup(() => (ingestionReader));
                 });
-            newRaftReaders.forEach(reader =>
-                this.logReadersUpdate.push(reader));
+            newRaftReaders.forEach(reader => {
+                if (reader.logConsumer && reader.logConsumer !== {}) {
```
jonathan-gramain: I think `reader.logConsumer !== {}` does not do what you want, as the comparison operates on the references to the objects instead of the object contents (so would always return `false` here because the `logConsumer` object **is** not the empty object you just created).

If you want to check that the object is empty, you could do `Object.keys(reader.logConsumer).length === 0`.

lib/queuePopulator/IngestionProducer.js
```diff
@@ -61,10 +62,51 @@ class IngestionProducer {
...
+     * Find the raft session that the bucket exists on
+     * @param {string} bucketName - name of sourcebucket that needs logs
+     * @param {number} raftRange -
+     * @param {function} done -
+     * @return {number} the raftId that has logs for the bucket
+     */
+    getRaftId(bucketName, raftRange, done) {
+        assert.equal(typeof raftRange, 'number');
+        return async.waterfall([
+            next => async.timesSeries(raftRange, (raftId, cb) =>
```
jonathan-gramain: If `raftId + 1` is returned, then the loop variable is not really a raftId, so maybe you could rather call it `i` to avoid confusion.

lib/queuePopulator/IngestionProducer.js
```diff
@@ -61,10 +62,51 @@ class IngestionProducer {
...
+                    if (id) {
+                        return id;
+                    }
+                    return false;
+                });
+                if (raftId && raftId[0]) {
+                    return next(null, raftId[0]);
+                }
+                this.log.error(`could not find bucket ${bucketName} in any` +
+                    'raft sessions', { method: 'getRaftId', bucketName });
```
jonathan-gramain: `in any raft session`

lib/queuePopulator/IngestionProducer.js
```diff
@@ -61,10 +62,51 @@ class IngestionProducer {
...
+        return async.waterfall([
+            next => async.timesSeries(raftRange, (raftId, cb) =>
+                this._getBuckets((raftId + 1).toString(), (err, res) => {
+                    if (res && res.includes(bucketName)) {
+                        return cb(null, raftId + 1);
+                    }
+                    return cb();
+                }), next),
+            (val, next) => {
+                const raftId = val.filter(id => {
```
jonathan-gramain: I think you can use `Array.find` more easily for the same purpose: `const raftId = val.find(id => id); if (raftId) ...`

lib/queuePopulator/IngestionReader.js
```diff
@@ -9,31 +9,52 @@ const LogReader = require('./LogReader');
...
+            // new per-bucket schema
             this.pathToLogOffset = `${ingestionPath}` +
                 `${bucketdConfig.zookeeperSuffix}` +
-                `/logState/raft_${raftId}/logOffset`;
+                `/logState/${this.bucket}/logOffset`;
         }
     }
 
+    setup(done) {
+        return async.waterfall([
```
jonathan-gramain: You may remove the waterfall, since the 2nd step is fully synchronous.

tests/unit/ingestion/IngestionProducer.js
```diff
@@ -86,10 +86,10 @@ describe('ingestion producer unit tests with mock', () => {
     });
 
     it('can generate a valid snapshot', done => {
-        this.iProducer.snapshot('1', (err, res) => {
-            // we expect 6 logs from the MockMetadataServer: 2 buckets with 2
-            // logs per bucket, and 1 object in each bucket with 1 log entry
-            assert.strictEqual(res.length, 6);
+        this.iProducer.snapshot('bucket1', (err, res) => {
+            // we expect 3 logs from the MockMetadataServer: 1 bucket with 2 log
+            // entriesper bucket, and 1 object in each bucket with 1 log entry
```
jonathan-gramain: `entries per`

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -241,10 +243,15 @@ class IngestionPopulator {
                         ingestionPath,
                         qpConfig: this.qpConfig,
                         s3Config: this.s3Config,
+                        bucket: this.bucket,
                     });
+                    return ingestionReader.setup(() => (ingestionReader));
```
philipyoo: the `setup` here is also called in `_setupUpdatedReaders`

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -241,10 +243,15 @@ class IngestionPopulator {
...
                 });
-            newRaftReaders.forEach(reader =>
-                this.logReadersUpdate.push(reader));
+            newRaftReaders.forEach(reader => {
+                if (Object.keys(reader.logConsumer).length > 0) {
```
philipyoo: I'll need to add this change into maybe my commit, or maybe we can move this out to another PR.
I have some conflicting changes in https://github.com/scality/backbeat/pull/522

Just to make sure, are you trying to find if an existing & same raft ID/source (but different bucket name) exists. If so, don't instantiate a new ingestion reader?

Isn't `reader.logConsumer` and instance of `LogConsumer`?

lib/queuePopulator/IngestionReader.js
```diff
@@ -9,31 +9,48 @@ const LogReader = require('./LogReader');
...
     }
 
+    setup(done) {
+        return this._iProducer.getRaftId(this.bucket, (err, data) => {
+            if (err || !this.raftId) {
+                return done();
+            }
+            this.raftId = data;
+            this.logConsumer = new LogConsumer({
+                bucketClieng: this.bucketClient,
```
philipyoo: small typo `bucketClient`

lib/queuePopulator/IngestionProducer.js
```diff
@@ -61,10 +62,41 @@ class IngestionProducer {
...
 
-    snapshot(raftId, done) {
+    /**
+     * Find the raft session that the bucket exists on
+     * @param {string} bucketName - name of sourcebucket that needs logs
+     * @param {number} raftRange -
+     * @param {function} done -
+     * @return {number} the raftId that has logs for the bucket
+     */
+    getRaftId(bucketName, raftRange, done) {
```
philipyoo: I guess `raftRange` no longer needed here?

lib/queuePopulator/IngestionReader.js
```diff
@@ -9,31 +9,48 @@ const LogReader = require('./LogReader');
...
+        this.bucketClient = new BucketClient(`${host}:${port}`);
         this._iProducer = new IngestionProducer(bucketdConfig, qpConfig,
-            s3Config);
+            s3Config, bucket);
+        this.bucket = bucket;
+        this.raftRange = raftRange;
+        this.logId = `raft_${this.raftId}`;
+        this.logger = logger;
+        this.logger.info('initializing raft log reader',
+            { method: 'RaftLogReader.constructor',
```
philipyoo: `IngestionReader`

lib/queuePopulator/IngestionPopulator.js
philipyoo: I was thinking, what if we get raft id in the IngestionPopulator and its a required parameter when instantiating a new IngestionReader. This way, we can keep same format for calling `setup`
Just thinking out loud, I guess depends on below comment on L#251

lib/queuePopulator/IngestionReader.js
```diff
@@ -9,31 +9,48 @@ const LogReader = require('./LogReader');
...
+            // TODO: rename zookeeper paths for ingestion to properly reflect
+            // new per-bucket schema
             this.pathToLogOffset = `${ingestionPath}` +
                 `${bucketdConfig.zookeeperSuffix}` +
-                `/logState/raft_${raftId}/logOffset`;
+                `/logState/${this.bucket}/logOffset`;
         }
     }
 
+    setup(done) {
```
philipyoo, line 34: I think we will need to check log offset as well in here, similar to how LogReader does it, in case this isn't a new ingestion source and state was saved in zookeeper from previous ingestion.

lib/queuePopulator/IngestionPopulator.js
JianqinWang: For IngestionReader, in order to use the cloudserver routes, we had to remove the logConsumer + bucketclient dependencies. Originally I was thinking I had to wrap the getRaftLogs to act like logConsumer (which I would initialize to {}), but now IngestionReader no longer has a LogConsumer dependency so this should not be an issue anymore.

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -256,20 +258,22 @@ class IngestionPopulator {
...
+                    kafkaConfig: this.kafkaConfig,
+                    bucketdConfig: bucketd,
+                    raftId: token,
+                    logger: this.log,
+                    extensions: [this._extension],
+                    metricsProducer: this._mProducer,
+                    ingestionPath,
+                    qpConfig: this.qpConfig,
+                    s3Config: this.s3Config,
+                    bucket: this.bucket,
```
philipyoo: `bucket` should be provided in `bucketd`. If relying on the unique zenko bucket name, should be saved as `bucketd.name`

we won't be able to keep `this.bucket` on `IngestionPopulator` because we will need to add/remove IngestionReader's based on different sources

JianqinWang: Aren't we making a new `IngestionPopulator` workflow for each separate bucket?

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -256,20 +256,22 @@ class IngestionPopulator {
...
+                    raftId: token,
+                    logger: this.log,
+                    extensions: [this._extension],
+                    metricsProducer: this._mProducer,
+                    ingestionPath,
+                    qpConfig: this.qpConfig,
+                    s3Config: this.s3Config,
+                    bucket: bucketd.name,
+                });
+                return ingestionReader.setup(() => (ingestionReader));
```
rahulreddy: Unless you return `ingestionReader` this would be an empty array as setup does not return anything.

lib/queuePopulator/IngestionProducer.js
```diff
@@ -77,10 +77,39 @@ class IngestionProducer {
...
+        });
+
+        attachReqUids(req, this.requestLogger);
+        req.send((err, data) => {
+            if (err) {
+                this.log.error(`could not find bucket ${bucketName} in any` +
+                'raft session', { method: 'getRaftId', bucketName });
+                return done(err);
+            }
+            return done(null, data['0']);
```
rahulreddy: I recommend validating this `data && data[0]`. Wait, why is the key stringified `0`?

rahulreddy (changes requested):

lib/queuePopulator/IngestionProducer.js
JianqinWang: When we run `getRaftId` against S3C the data response is `{ '0': '5' }`.
It looks like just `0` works as well so I can remove the string `''`.

lib/queuePopulator/IngestionReader.js
```diff
@@ -1,37 +1,42 @@
...
         this.remoteLogOffset = null;
-        // i.e.: '/ingestion/${bucket}/logState/raft_${raftId}/logOffset'
-        this.pathToLogOffset = `${ingestionPath}${zookeeperSuffix}` +
-            `/logState/raft_${raftId}/logOffset`;
-
-        // identifier - zenko bucket name
         this._targetZenkoBucket = name;
+        if (ingestionPath) {
+            // TODO: rename zookeeper paths for ingestion to properly reflect
+            // new per-bucket schema
```
philipyoo, line 27: 👍 ya i think zk paths need a revision, i'll go through the code to check this

JianqinWang:
@bert-e approve

lib/queuePopulator/IngestionReader.js
```diff
@@ -1,37 +1,42 @@
...
-        this.raftId = raftId;
         this._iProducer = new IngestionProducer(bucketdConfig, qpConfig,
-            s3Config);
+            s3Config, bucket);
+        this.bucket = bucket;
+        this.logger = logger;
+        this.logger.info('initializing raft log reader',
+            { method: 'IngestionReader.constructor',
+                bucketdConfig, raftId: this.raftId });
+        this.bucketdPrefix = bucketdConfig.prefix ? bucketdConfig.prefix :
```
philipyoo, line 20: minor: `this.bucketdPrefix` has been removed in favor of `this._targetZenkoBucket` 
which is the zenko bucket name and should be unique

philipyoo (approved):
left a minor comment, shouldn't impact anything though, lgtm

rahulreddy (approved):

lib/queuePopulator/IngestionReader.js
rahulreddy: Ok. @JianqinWang Please follow up with a PR for this.

rahulreddy:
@bert-e status

rahulreddy (approved):

PR #531, opened 2018-11-28
ZENKO-1347: A. update BackbeatClient to support GetRaftLog

JianqinWang:


lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -949,10 +949,64 @@
...
+                }
+            },
+            "output": {
+                "type": "integer"
+            }
+        },
+        "GetRaftLog": {
+            "http": {
+                "method": "GET",
+                "requestUri": "/_/metadata/admin/raft_sessions/{LogId}/log?begin={Begin}&end={End}&targetLeader={TargetLeader}"
```
alexanderchan-scality: ~should the params `Begin/End/TargetLeader` be required? (since these seem to be hardcoded here)
If they aren't required, I would suggest `"location": "querytstring"` in the member definition instead of "uri".~

**update**
```suggestion
                "requestUri": "/_/metadata/admin/raft_sessions/{LogId}/log"
```

make `Begin/End/TargetLeader` required params
and change "location" from "uri" to "querystring" and "locationName" to lower-cased values

JianqinWang: I see, it should be in `querystring` in that case.

Do you have a reference doc for editing this file? The past few times I've had to work with this file have been guess and check (based on previous examples here and in the aws json files), but I haven't been able to dig up any documentation.

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -949,10 +949,64 @@
                 }
             }
         },
+        "GetRaftID": {
+            "http": {
+                "method": "GET",
+                "requestUri": "/_/metadata/admin/buckets/{Bucket}/id"
```
bennettbuchanan, line 955: The methods that involve raft appear to be using the namespace `raft_sessions` (e.g. `/_/metadata/admin/raft_sessions/[...]`). Do we need to add this for API consistency here?

lib/clients/backbeat-2017-07-01.api.json
alexanderchan-scality: Same. I don't know of any formal documentation on this. I have been pulling examples from files in https://github.com/aws/aws-sdk-js/tree/master/apis

JianqinWang: I've been using those examples as well, was hoping for something more clear and definitive.

If it's in the querystring, why does it have to be a required param?

tmacro (approved):

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -994,7 +1048,7 @@
         "GetObjectMetadata": {
             "http": {
                 "method": "GET",
-                "requestUri": "/_/metadata/getobject/{Bucket}/{Key+}"
+                "requestUri": "/_/metadata/default/bucket/{Bucket}/{Key+}"
```
bennettbuchanan, line 1054: I'm not sure I follow the namespace of the API. Is this right: 

`/_/metadata/<user-type>/<resource-type>/<bucket-name>/<key-name>`

I guess I'm unsure about the use of `attributes` vs. `bucket`.

bennettbuchanan (commented):
👍

lib/clients/backbeat-2017-07-01.api.json
JianqinWang: the `/_/metadata/admin` and `/_/metadata/default` are prefixes that we are adding to the paths in scality/Metadata.

For this case, the original path does not reference `raft_sessions` https://github.com/scality/MetaData/blob/development/8.0/documentation/PROTOCOL.ADMIN.md#get-the-id-of-a-raft-session-handling-a-bucket

lib/clients/backbeat-2017-07-01.api.json
JianqinWang: Rahul did a great refactor here: https://github.com/scality/cloudserver/blob/development/8.1/lib/routes/routeMetadata.js
Since I haven't fully tested it yet, my current api changes are based off: https://github.com/scality/cloudserver/commit/d09b6da7beb0c9f0fa5d8b8b8fbbc9235c525de3

However, based on my understanding,  `attributes` would be specifying we want the information of the bucket itself, whereas specifying `bucket` would just mean we're accessing a bucket for information on the contents of the bucket.

bennettbuchanan: As it's already implemented in cloudserver, it's probably a minor thing at this point. I just would have expected:

Route to retrieve bucket metadata:
`/_/metadata/default/attributes/{Bucket}`

Route to retrieve object metadata:
`/_/metadata/default/attributes/{Bucket}/{Key+}`

lib/clients/backbeat-2017-07-01.api.json
bennettbuchanan: Cool, thanks that helps. 👍

lib/clients/backbeat-2017-07-01.api.json
alexanderchan-scality: Looking at the `aws-sdk` generator code, if the querystrings are defined in `requestUri` and the substitution values aren't provided, the `uri` will be malformed. (e.g. literal string value `{Begin}` if the `Begin` member is missing)

with the change from `location: uri` to `querystring`, this will no longer be an issue

By making these required, the client will throw an error if any these members are missing.
In https://github.com/scality/MetaData/blob/development/8.0/documentation/PROTOCOL.ADMIN.md#get-the-id-of-a-raft-session-handling-a-bucket, these querystring parameters seem to be mandatory

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -949,10 +949,67 @@
                 }
             }
         },
+        "GetRaftID": {
+            "http": {
+                "method": "GET",
+                "requestUri": "/_/metadata/admin/buckets/{Bucket}/id"
+            },
+            "input": {
+                "type": "string",
```
jonathan-gramain: Should it be "structure"?

jonathan-gramain (approved):

lib/clients/backbeat-2017-07-01.api.json
JianqinWang: you're right, thanks for catching it!

lib/clients/backbeat-2017-07-01.api.json
JianqinWang: My bad, I phrased the question wrong, I was asking if the `location` was `querystring`, why would it be required. Good to confirm that in that case, it isn't. The edits in the initial comment made it seem that the two changes both needed to be made.

Considering other `querystring` params in this file are not specified as required, I will go forward with just changing the `location` param.

On the metadata api side, I am not convinced that all of the parameters are required for the querystring. In the documentation, it specifies a default for `targetLeader` and a default (limited) max return of 10k logs.

Even if the values are required, that would (imo) should still be added during the usage of `BackbeatClient` and not added as a limitation here in the api definitions.

bennettbuchanan (approved):

alexanderchan-scality (approved):

lib/clients/backbeat-2017-07-01.api.json
jonathan-gramain: You should find the spec and examples in the OpenAPI repo: https://github.com/OAI/OpenAPI-Specification

JianqinWang: :pray: Thank you!

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -949,10 +949,67 @@
...
+                ],
+                "members": {
+                    "Bucket": {
+                        "location": "uri",
+                        "locationName": "Bucket"
+                    }
+                }
+            },
+            "output": {
+                "type": "integer"
```
philipyoo: may need to change this to `"string"`. Not getting any output with "integer"

philipyoo (approved):

tests/unit/clients/BackbeatClient.js
```diff
@@ -0,0 +1,86 @@
+const http = require('http');
```
rahulreddy: I think this is more of a functional test than unit

rahulreddy (approved):

JianqinWang:
@bert-e approve

PR #537, opened 2018-12-04
Feature/zenko 1347: B. Add getRaftLog to IngestionProducer

JianqinWang:


lib/queuePopulator/IngestionProducer.js
```diff
@@ -69,6 +70,25 @@ class IngestionProducer {
         ], err => done(err, this.resLog));
     }
 
+    getRaftLog(raftId, begin, end, targetLeader, done) {
+        const req = this.ringReader.getRaftLog({
+            LogId: raftId,
```
philipyoo: When testing this out, `raftId` must be a string. To be safe, can we convert `raftId.toString()`.

lib/queuePopulator/IngestionProducer.js
```diff
@@ -69,6 +70,25 @@ class IngestionProducer {
...
+            Begin: begin,
+            End: end,
+            TargetLeader: targetLeader,
+        });
+        attachReqUids(req, this.requestLogger);
+        const readStream = req.createReadStream();
+        const jsonResponse = readStream.pipe(jsonStream.parse('log.*'));
```
philipyoo, line 105: getting an epiphany looking at this pr lol.. the transform stream makes a lot more sense now

I think we should utilize the transform as well to apply `entry.type` on any entries that do not have it. This is used on processor side. Haven't looked more into it, but may also be used in the logreader elsewhere?

lib/queuePopulator/IngestionProducer.js
JianqinWang: ty for the catch

lib/queuePopulator/IngestionProducer.js
philipyoo: actually, even delete object entries don't have a "type" field specified.. hmm

JianqinWang: You're right, I should have the transform here to ensure that `put` is applied if there is no specified `entry.type`.

I don't think the the `logReader` side needs the entry type specification

philipyoo: sry, false alarm. I'm not sure the reason, but when using a stream, the type 'del' will be defined. But when just sending a request and logging the response data, the entry doesn't have the type 'del' field defined.

lib/queuePopulator/IngestionReader.js
```diff
@@ -58,7 +59,21 @@ class IngestionReader extends LogReader {
...
+                            this.log.error('error retrieving logs',
+                            { err, raftId: this.raftId, method:
+                            'IngestionReader._processReadRecords' });
+                            return done(err);
+                        }
+                        this.log.debug('readRecords got raft logs', {
+                            method: 'IngestionReader._processReadRecords',
+                            params });
+                        batchState.logRes = data;
+                        return done();
```
bennettbuchanan: I think this is meant to be `next`, and the same callback above.

package.json
```diff
@@ -15,7 +15,7 @@
     "garbage_collector": "node extensions/gc/service.js",
     "test": "mocha --recursive tests/unit",
     "ft_test": "mocha --recursive $(find tests/functional -name '*.js') --timeout 30000",
-    "ft_test:lib": "mocha --recursieve $(find tests/functional/lib -name '*.js') --timeout 30000",
+    "ft_test:lib": "mocha --recursive $(find tests/functional/lib -name '*.js') --timeout 30000",
```
bennettbuchanan, line 18: 🙌

tests/functional/lib/IngestionProducer.js
```diff
@@ -7,7 +7,7 @@ const IngestionProducer =
 const testConfig = require('../../config.json');
 const MetadataMock = require('../../utils/MockMetadataServer');
 
-describe('ingestion producer unit tests with mock', () => {
```
bennettbuchanan, line 10: 👍

lib/queuePopulator/IngestionProducer.js
```diff
@@ -69,6 +87,27 @@ class IngestionProducer {
         ], err => done(err, this.resLog));
     }
 
+    getRaftLog(raftId, begin, end, targetLeader, done) {
+        const recordResponse = new ListRecordStream(this.log);
```
philipyoo: not sure but maybe to be safe, should we add the on 'error' catch for this?
Also, can we rename this variable to use "stream" instead of "response"?

lib/queuePopulator/IngestionProducer.js
jonathan-gramain: Isn't this because of using `console.log` to log a javascript response object? `console.log` prunes inner objects reaching some depth so will not log everything, but it can be prior `JSON.stringify` for this purpose.

JianqinWang: The `type` field is at the same level (or should be) as `key` and `value` if it exists, as per the metadata documentation https://github.com/scality/MetaData/blob/development/8.0/documentation/PROTOCOL.ADMIN.md#get-raft-session-logs-2
so it shouldn't be the inner object issue?

philipyoo: even with `JSON.stringify`, i'm seeing the following from a `send` response
`{"db":"something","entries":[{"key":"bay1"}],"method":8}`
and the following from stream:
`{"db":"something","entries":[{"key":"bay1","type":"del"}],"method":8}`

I'm most likely doing something wrong in my script x.x If I figure it out, I'll mention it here

philipyoo (approved):

lib/queuePopulator/IngestionProducer.js
rahulreddy: Ok, that's very odd. @JianqinWang Can you add a TODO here mentioning the possible missing type attribute for delete ops? We can come back to after this PR.

rahulreddy (approved):
I think we can revisit the delete ops weirdness without holding this PR as hostage. LGTM to make progress.

JianqinWang:
@bert-e approve

PR #538, opened 2018-12-05
feature: ZENKO-733 Lifecycle transitions policies

bennettbuchanan:
Introduces the lifecycle transition policy support:
* Checks lifecycle configurations for transition policies.
* Finds the most recent transition rule in the lifecycle configuration—i.e. the one to apply.
* Pushes lifecycle object data to the replication topic and replication status topic.
* Begins decoupling the replication topic by introducing a new kafka entry field `operationType`.
* Refactors methods for simpler code paths that are used by both lifecycle and replication (in particular, for classes `ReplicateObject`, `UpdateReplicationStatus`).

Remaining TODO items (as commented in the code):
* Introduce garbage collection for transitions.
* Add `NonCurrentVersionTransition` support.

Remaining items to consider:
* Add functional and Zenko integration tests.
* Possible optimization: the transition feature does not need to list and parse object versions as it is reusing the same code path that expiration, even when a bucket is versioning enabled or suspended.
* For reliability, we need a way to ensure the same rule isn't applied in two subsequent lifecycle tasks. For example, if the first transition is still in the progress when, 5 minutes later, a new lifecycle task is started, we risk transitioning the object twice.

Depends on https://github.com/scality/Arsenal/pull/612.


extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -37,12 +37,18 @@ class LifecycleBucketProcessor {
      * @param {Object} kafkaConfig - kafka configuration object
      * @param {string} kafkaConfig.hosts - list of kafka brokers
      *   as "host:port[,host:port...]"
-     * @param {Object} lcConfig - lifecycle config
-     * @param {Object} lcConfig.auth - authentication info
-     * @param {String} lcConfig.bucketTasksTopic - lifecycle bucket topic name
-     * @param {Object} lcConfig.bucketProcessor - kafka consumer object
-     * @param {String} lcConfig.bucketProcessor.groupId - kafka
+     * @param {Object} extensions - backbeat config extenstions object
```
jonathan-gramain: typo on "extensions" in comment

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -482,11 +488,124 @@ class LifecycleTask extends BackbeatTask {
...
+     * @return {object} The most applicable transition rule
+     */
+    _getTransitionRule(params) {
+        const { transitions, objectLastModified, currentDate } = params;
+        let result;
+        transitions.forEach(transition => {
+            const shouldTransition =
+                this._getTimestamp(currentDate) >=
+                this._getTransitionTimestamp(transition, objectLastModified);
+            if (shouldTransition === false) {
```
jonathan-gramain: Usually written `!shouldTransition` when the value is guaranteed to be boolean.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -482,11 +488,124 @@ class LifecycleTask extends BackbeatTask {
...
+            return rule2;
+        }
+        if (rule2 === undefined) {
+            return rule1;
+        }
+        const { objectLastModified, currentDate } = params;
+        const time1 = this._getTransitionTimestamp(rule1, objectLastModified);
+        const time2 = this._getTransitionTimestamp(rule2, objectLastModified);
+        const now = this._getTimestamp(currentDate);
+        return (now - time1) < (now - time2) ? rule1 : rule2;
```
jonathan-gramain: You could rewrite this as `time1 > time2` for simplicity.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -671,6 +790,94 @@ class LifecycleTask extends BackbeatTask {
         return false;
     }
 
+    /**
+     * Build the transition entry from metadata retrieved from CloudServer.
```
jonathan-gramain: This description seems related to what the function is used for (from `_getTransitionEntry`) rather than what is does. It's probably useful to keep it as an extra note after the main description though.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -147,13 +147,28 @@ class ReplicateObject extends BackbeatTask {
         }, cb);
     }
 
+    _getUpdatedSourceEntry(params) {
+        const { sourceEntry, replicationStatus } = params;
+        if (sourceEntry.isReplicationOperation()) {
+            const entry = replicationStatus === 'COMPLETED' ?
+                sourceEntry.toCompletedEntry(this.site) :
+                sourceEntry.toFailedEntry(this.site);
+            const id = sourceEntry
```
jonathan-gramain: minor: you could call this variable with a slightly more descriptive name like `versionId`

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -79,17 +79,22 @@ class UpdateReplicationStatus extends BackbeatTask {
...
-     * @param {String} status - The replication status
-     * @param {ObjectQueueEntry} entry - updated object entry
-     * @param {String} site - site recently updated
+     * @param {ObjectQueueEntry} sourceEntry - The original entry
+     * @param {ObjectQueueEntry} updatedSourceEntry - updated object entry
      * @return {undefined}
      */
-    _reportMetrics(status, entry, site) {
+    _reportMetrics(sourceEntry, updatedSourceEntry) {
+        if (sourceEntry.isReplicationOperation() === false) {
```
jonathan-gramain: better style if written `if (!sourceEntry.isReplicationOperation())`

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -134,64 +140,127 @@ class UpdateReplicationStatus extends BackbeatTask {
         return refreshedEntry.toCompletedEntry(sourceEntry.getSite());
     }
 
+    _checkStatus(sourceEntry) {
+        // This check only applies to replication operations.
+        if (sourceEntry.isReplicationOperation() === false) {
```
jonathan-gramain: ditto

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -134,64 +140,127 @@ class UpdateReplicationStatus extends BackbeatTask {
...
+            return undefined;
+        }
+        const site = sourceEntry.getSite();
+        const status = sourceEntry.getReplicationSiteStatus(site);
+        const statuses = ['COMPLETED', 'FAILED', 'PENDING'];
+        if (statuses.includes(status) === false) {
```
jonathan-gramain: `if (!statuses.includes(status))`

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -134,64 +140,127 @@ class UpdateReplicationStatus extends BackbeatTask {
...
+            const msg = `unknown status in replication info: ${status}`;
+            return errors.InternalError.customizeDescription(msg);
+        }
+        return undefined;
+    }
+
+    _getUpdatedReplicationEntry(params) {
```
jonathan-gramain, line 158: Nice refactor of the old monolithic code into specialized functions :+1:

lib/models/ObjectQueueEntry.js
```diff
@@ -37,6 +37,24 @@ class ObjectQueueEntry extends ObjectMD {
...
+    getOperationType() {
+        return this.operationType;
+    }
+
+    isLifecycleOperation() {
+        return this.operationType === 'lifecycle';
+    }
+
+    isReplicationOperation() {
+        // For backward compatibility.
```
jonathan-gramain, line 54: :+1:

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -414,20 +417,23 @@ class LifecycleTask extends BackbeatTask {
...
         // Assumes if for example a rule defines expiration and transition
         // and if backbeat disables expiration and enables transition,
         // we still consider this rules transition to apply
-
         // these are rules enabled in config.json
         const enabledRules = Object.keys(this.enabledRules).filter(rule =>
             this.enabledRules[rule].enabled);
 
+        // Declare the current date before the reducing function so that all
+        // rule comparisons use the same date.
```
jonathan-gramain, line 435: The following comment `//NOTE: only care about expiration...` at line 440 should be removed now.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -482,11 +488,124 @@ class LifecycleTask extends BackbeatTask {
...
+            if (hasTransitions && enabledRules.includes('transitions')) {
+                // Find the most applicable transition rule.
+                const transitionRule = this._getTransitionRule({
+                    currentDate,
+                    transitions: rule.Transitions,
+                    objectLastModified: metadata.LastModified,
+                    objectStorageClass: metadata.StorageClass,
+                });
+                // Compare across rules to determine which is most applicable.
+                const ruleToApply = this._compareTransitionRules({
```
jonathan-gramain: For clarity, I would advise the following changes, that as pseudo-code would give:

```
if transitions are enabled and rule.Transitions is an array:
    for each transition in rule.Transitions:
        if applicable(transition, rule.criteria, currentDate) and
           (!store.Transition or compareTransitionRules(transition, store.Transition) > 0):
                store.Transition = transition
```

- `applicable()` returns true if single transition satisfies current criteria of date etc.
- `compareTransitionRules()` now only compares two rules that have been checked as applicable, and as a generally adopted convention returns a number > 0 when first argument is "higher" than second (i.e. more recent here)
- that avoids having to use `undefined` as a special value to compare for non-applicable rules, and I think makes the code a bit easier to follow and understand as well

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -26,14 +26,30 @@ const lp = {
...
+function getDate(params) {
+    const numberOfDaysFromNow = params.numberOfDaysFromNow || 0;
+    const oneDay = 24 * 60 * 60 * 1000; // Milliseconds in a day.
+    const milliseconds = numberOfDaysFromNow * oneDay;
+    const timestamp = Date.now() + milliseconds;
+    return new Date(timestamp);
+}
+
+// Get the metadata object.
+function getMetadatObject(lastModified) {
```
jonathan-gramain: typo: `getMetadataObject`

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -273,6 +289,388 @@ describe('lifecycle task helper methods', () => {
...
+        it('should not return transition when Transitions is undefined', () => {
+            const applicableRules = [
+                new Rule()
+                    .addExpiration('Days', 1)
+                    .build(),
+            ];
+            const rules = lct._getApplicableRules(applicableRules, {});
+            assert.strictEqual(rules.Transitions, undefined);
+        });
+    });
```
jonathan-gramain: You could add here one or more tests with no applicable rule

jonathan-gramain (commented):
Looks great overall, a few comments mostly minor, good to see transitions materializing :gear:

extensions/lifecycle/tasks/LifecycleTask.js
jonathan-gramain: If the only check done on applicable rules is the recency, you could as well have a `isMoreRecentRule()` helper returning true/false.

extensions/lifecycle/tasks/LifecycleTask.js
jonathan-gramain: Nevermind, you fixed this :)

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -110,6 +128,9 @@ class LifecycleBucketProcessor {
         return {
             sendBucketEntry: this.sendBucketEntry.bind(this),
             sendObjectEntry: this.sendObjectEntry.bind(this),
+            sendTransitionEntry: this.sendTransitionEntry.bind(this),
+            replicationSource: this._repConfig.source,
+            bootstrapList: this._repConfig.destination.bootstrapList,
```
philipyoo, line 133: in the `bucketProcess/task.js`, we will also need to update bootstrapList, similar to https://github.com/scality/backbeat/blob/642e7baf7694282ae3be91edc29ea42f2f10ca44/index.js#L31-L36

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -482,11 +488,124 @@ class LifecycleTask extends BackbeatTask {
...
+        const { transitions, objectLastModified, currentDate } = params;
+        let result;
+        transitions.forEach(transition => {
+            const shouldTransition =
+                this._getTimestamp(currentDate) >=
+                this._getTransitionTimestamp(transition, objectLastModified);
+            if (shouldTransition === false) {
+                return undefined;
+            }
+            // Is this transition rule the most recent one in the transitions?
```
philipyoo: wouldn't we want the transition rule that is valid and oldest? Why most recent one?

philipyoo (commented):
nice! :D

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: That's right, and a more precise way of wording this because "recent" can be interpreted in different ways. I'll update accordingly.

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
bennettbuchanan: Thanks for pointing this out.

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: Thanks! I had the delta comparison stuck in my head. 😆

extensions/lifecycle/tasks/LifecycleTask.js
philipyoo: ah sorry, my misunderstanding, thanks for clarifying 👍

bennettbuchanan: No misunderstanding—the comment should be written such that it cannot be misinterpreted. Thanks for bringing it up! 😸

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: I've taken these concepts and refactored accordingly, though I've elected to maintain the use of `undefined` in some cases. For example, the comparison of a new transition with the `store.Transition` involved multiple conditionals. Using a method that may return `undefined` felt like a cleaner option.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -482,11 +485,110 @@ class LifecycleTask extends BackbeatTask {
...
+                    transition1: transition,
+                    transition2: store.Transition,
+                    lastModified: metadata.LastModified,
+                });
+            }
+            // TODO: Add support for NoncurrentVersionTransitions.
             return store;
         }, {});
+        // Do not transition to a location where the object is already stored.
+        if (applicableRules.Transition &&
```
jonathan-gramain, line 502: I think this check would be better placed inside `getApplicableTransition`, it's generally better if all related criteria to select a rule can be done in a single place to ease understanding why a rule is applicable or not.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -482,11 +485,110 @@ class LifecycleTask extends BackbeatTask {
...
+            const isApplicable = // Is the transition time in the past?
+                this._getTimestamp(currentDate) >=
+                this._getTransitionTimestamp(transition, lastModified);
+            if (!isApplicable) {
+                return result;
+            }
+            if (!result) {
+                return transition;
+            }
+            const delta =
```
jonathan-gramain: Couldn't we rely on `_getStoreTransition` to do the recency check? It would avoid comparing dates at two different places for the same purpose.

Also instead of using `delta` variable you can simply compare timestamps directly.

extensions/lifecycle/tasks/LifecycleTask.js
jonathan-gramain: The main idea I was suggesting was to "unroll" the transitions into a flat list, to simplify the checks logic. Although I'm fine with the way it's done, a few things may hinder understandability. For example calling `getApplicableRules` on a single Transition rule may hide the less recent rules if overriden by a more recent one, though the returned rule considered applicable may be skipped by `getStoreRules` itself if also overriden by another rule, hence the concept of applicability becomes a bit fuzzy. By unrolling the transitions and applying the same check to all in a flat way, we can define applicability in a stricter way, then compare applicable rules directly based solely on their recency.

Maybe the mix with expiration checks would require some refactor which makes things more complex, and I'm ok to keep as is, just wanted to explain more precisely what I meant in the original comment.

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: No, because I first want to get the most applicable transition across all the rules, which is found once the reduce method is finished. If it is a part of `getApplicableTransition` then it's possible that another, older rule would end up as "most applicable".

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: Good call, I'll make that a more generic comparison function.

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: I agree that would be a lot easier to understand. I think, as you mention, it may have to do with the fact that the method is being called within the reducer function. Though it doesn't entirely "unroll" the transitions, I think I have made an improvement to this end by moving all the checks within `_getApplicableTransition`.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -575,18 +573,18 @@ class LifecycleTask extends BackbeatTask {
...
         }
         if (transition2 === undefined) {
             return transition1;
         }
-        const delta =
-            this._getTransitionTimestamp(transition1, lastModified) -
-            this._getTransitionTimestamp(transition2, lastModified);
-        return delta > 0 ? transition1 : transition2;
+        // Which transition is the most recent?
+        return this._getTransitionTimestamp(transition1, lastModified) -
```
jonathan-gramain: it could be simply `this._getTransitionTimestamp(...) > this._getTransitionTimestamp(...) ? ...`

jonathan-gramain (approved):
Thanks for doing the changes, it looks neat!

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: Thanks! Updated this. 😄

jonathan-gramain (approved):

bennettbuchanan:
@bert-e approve

philipyoo (approved):
noice 👍

PR #541, opened 2018-12-06
ft: ZENKO-1240 ingestion init management processor

philipyoo:
**Purpose of this PR**:
Add management layer on ingestion processor side.
Primarily necessary for pause and resume service for ingestion. Creates separate consumers by each ingestion location

**Changes in this PR**:
- Add management layer to get config location constraint
  updates
- Add consumers/MongoQueueProcessors by site in order to
  support pause/resume later on

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -48,14 +48,17 @@ class MongoQueueProcessor {
      * @param {number} [mongoProcessorConfig.retry.backoff.factor] -
      *  backoff factor
      * @param {Object} mongoClientConfig - config for connecting to mongo
+     * @param {String} site - site name
      */
-    constructor(kafkaConfig, mongoProcessorConfig, mongoClientConfig) {
+    constructor(kafkaConfig, mongoProcessorConfig, mongoClientConfig, site) {
         this.kafkaConfig = kafkaConfig;
         this.mongoProcessorConfig = mongoProcessorConfig;
+        this.mongoClientConfig = mongoClientConfig;
+        this.site = site;
+
         this._consumer = null;
         this.logger =
-            new Logger('Backbeat:MongoProcessor');
-        this.mongoClientConfig = mongoClientConfig;
+            new Logger(`Backbeat:Ingestion:MongoProcessor:${this.site}`);
         this.mongoClientConfig.logger = this.logger;
         this._mongoClient = new MongoClient(this.mongoClientConfig);
     }
@@ -75,7 +78,7 @@ class MongoQueueProcessor {
             let consumerReady = false;
             this._consumer = new BackbeatConsumer({
                 topic: this.mongoProcessorConfig.topic,
-                groupId: `${this.mongoProcessorConfig.groupId}`,
+                groupId: `${this.mongoProcessorConfig.groupId}-${this.site}`,
                 kafka: { hosts: this.kafkaConfig.hosts },
                 queueProcessor: this.processKafkaEntry.bind(this),
             });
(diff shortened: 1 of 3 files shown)


extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -122,6 +125,13 @@ class MongoQueueProcessor {
                               { error: sourceEntry.error });
             return process.nextTick(() => done(errors.InternalError));
         }
+        // TODO-FIX:
+        // Depends on the filter data. Need a way of determining the
+        // zenko bucket.
+        // if entry is for another site, simply skip/ignore
+        if (this.site !== kafkaEntry.bucket) {
```
philipyoo, line 132: to support site-level consumers, need to ignore entries for another site. This may need some revision based on kafka entries produced

extensions/mongoProcessor/mongoProcessorTask.js
```diff
@@ -16,18 +19,82 @@ const healthServer = new HealthProbeServer({
...
+
+        let bootstrapList = config.getBootstrapList();
+        config.on('bootstrap-list-update', () => {
+            bootstrapList = config.getBootstrapList();
+
+            const active = Object.keys(activeProcessors);
+            const update = bootstrapList.map(i => i.site);
+            const allSites = [...new Set(active.concat(update))];
+
+            async.each(allSites, (site, next) => {
```
bennettbuchanan: Is this, in fact, an async operation?

extensions/mongoProcessor/mongoProcessorTask.js
```diff
@@ -16,18 +19,82 @@ const healthServer = new HealthProbeServer({
...
+                if (update.includes(site)) {
```
bennettbuchanan: Minor: for readability, you can avoid the `if...else` and `if` nesting by first checking if the site is not in the updated list.

extensions/mongoProcessor/mongoProcessorTask.js
```diff
@@ -16,18 +19,82 @@ const healthServer = new HealthProbeServer({
...
+                        mqp.start();
+                        activeProcessors[site] = mqp;
+                    }
+                } else {
+                    // site no longer in bootstrapList
+                    activeProcessors[site].stop(() => {});
+                    delete activeProcessors[site];
+                }
+                return next();
+            }, err => {
```
bennettbuchanan: I don't see errors being passed to `next`, but I think this can be a synchronous operation in any case.

extensions/mongoProcessor/mongoProcessorTask.js
```diff
@@ -16,18 +19,82 @@ const healthServer = new HealthProbeServer({
...
+                if (err) {
+                    process.exit(1);
+                }
+            });
+        });
+
+        // Start Processors for each site
+        const siteNames = bootstrapList.map(i => i.site);
+        async.each(siteNames, (site, next) => {
```
bennettbuchanan: Same comment about async and error passing to `next`.

extensions/mongoProcessor/mongoProcessorTask.js
```diff
@@ -16,18 +19,82 @@ const healthServer = new HealthProbeServer({
...
+                }
+            });
+            return passed;
+        });
+        log.info('Starting HealthProbe server');
+        healthServer.start();
+    });
+}
+
+initAndStart();
```
bennettbuchanan: Minor: `initAndStart` seems to encompass a lot of different ideas (i.e., initiate the management, update the bootstrap list, start processors, and check the health of the system with the health probe). I think we could improve maintainability and understanding by breaking down into multiple, simpler functions.

extensions/mongoProcessor/mongoProcessorTask.js
philipyoo: ah yes you're right, good catch!

bennettbuchanan (approved):
🚢 Nice!

JianqinWang (approved):

extensions/mongoProcessor/MongoQueueProcessor.js
rahulreddy: Ok, this is inline with what I mentioned yesterday. We need to move the logic of processing of an entry to be ingested from producer to consumer.

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #548, opened 2018-12-12
bf: ZENKO-1374 do not escalate error provisioning

philipyoo:
Provisioning log source may take a while. Do not escalate
error, but instead allow for retries.
Escalating an error will crash the pod


JianqinWang (approved):

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #551, opened 2018-12-13
bf: ZENKO-1374 ingestion provisioner cb fix

philipyoo:
ProvisionDispatcher may take a while to setup, and might
eventually try to call the callback a second time.
This commit is to address this issue

### DIFF
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -239,10 +239,19 @@ class IngestionPopulator {
     }
 
     _subscribeToLogSourceDispatcher(zkEndpoint, bucketd, cb) {
+        const key = bucketd.name;
+        if (this._provisionedIngestionSources[key]) {
+            this.log.info('waiting to be provisioned a log source',
+                          { zkEndpoint });
+            return cb();
+        }
+
         const ingestionPath = this.ingestionConfig.zookeeperPath;
         // TODO: Remove/Replace ProvisionDispatcher
         const raftIdDispatcher =
             new ProvisionDispatcher({ connectionString: zkEndpoint });
+        this._provisionedIngestionSources[key] = raftIdDispatcher;
+
         raftIdDispatcher.subscribe((err, items) => {
             if (err) {
                 this.log.error(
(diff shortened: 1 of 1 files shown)


lib/queuePopulator/IngestionPopulator.js
```diff
@@ -275,10 +282,11 @@ class IngestionPopulator {
                 return ingestionReader;
             });
             this.logReadersUpdate.push(...newRaftReaders);
-            return cb();
+            return undefined;
         });
         this.log.info('waiting to be provisioned a log source',
                       { zkEndpoint });
+        return cb();
```
rahulreddy: This is odd, you are calling the callback outside of the `subscribe` call which is async...

rahulreddy (changes requested):

lib/queuePopulator/IngestionPopulator.js
philipyoo: `subscribe` is async but I'm relying on `queueBatch` (similar to how QueuePopulator does it). Once the ProvisionDispatcher completes its async call, the `logReadersUpdate` list gets updated with the new Ingestion readers

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -239,21 +239,28 @@ class IngestionPopulator {
     }
 
     _subscribeToLogSourceDispatcher(zkEndpoint, bucketd, cb) {
```
jonathan-gramain: What is `cb` used for at the end, by the caller, and isn't it possible to remove `cb` if not used? I think it could help reduce confusion as it's not a regular callback, the `subscribeToLogSourceDispatcher` logic is essentially synchronous as it only kicks off the provisioning, it does not wait for something afterwards.

philipyoo: ah ok, sorry I was so fixated on keeping it async somehow because of the previous methods, but I see what you mean. No need to pass `cb` here and just make it a sync call that will update the readers when provisioned

jonathan-gramain (approved):

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #552, opened 2018-12-14
ft: ZENKO-1380 source credentials for ingestion

philipyoo:
Use the source credentials we save from mongo.
Add ability to decrypt secret key.
The decrypted secret key is only available within
the IngestionProducer. Otherwise, the encrypted key is
the one passed around between classes.

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -49,7 +49,11 @@
                     "port": 8000,
                     "https": false,
                     "type": "scality_s3",
-                    "raftCount": 8
+                    "raftCount": 8,
+                    "auth": {
+                        "accessKey": "myAccessKey",
+                        "secretKey": "myEncryptedSecretKey"
+                    }
                 }
             ]
         },
--- a/lib/management/index.js
+++ b/lib/management/index.js
@@ -44,6 +44,27 @@ function decryptSecret(instanceCredentials, secret) {
     });
 }
 
+/**
+ * Decrypt encrypted secret key. Needed for ingestion source credentials
+ * @param {string} encryptedSecret - encrypted secret key
+ * @param {werelogs.Logger} log - Logger object
+ * @param {function} cb - callback(error, decryptedKey)
+ * @return {undefined}
+ */
+function decryptLocationSecret(encryptedSecret, log, cb) {
+    return metadata.getObjectMD(serviceBucket, tokenKey, {}, log,
+    (err, instanceAuth) => {
+        if (err) {
+            log.error('failed to fetch decryption key from metadata for ' +
+            'deciphering authentication secret', {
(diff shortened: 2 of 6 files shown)


conf/config.json
```diff
@@ -49,7 +49,11 @@
                     "port": 8000,
                     "https": false,
                     "type": "scality_s3",
-                    "raftCount": 8
+                    "raftCount": 8,
```
JianqinWang, line 52: just a note, I think we don't need this anymore since we ingest by bucket
probably need one pr to do cleanup to remove unused vals/variables

lib/queuePopulator/IngestionProducer.js
```diff
@@ -49,24 +47,18 @@ class IngestionProducer {
...
-        this.auth = sourceConfig.auth || {
-            type: 'account',
-            account: 'bart',
-            vault: {
-                host: '127.0.0.1',
-                port: 8500,
-                adminPort: 8600,
-            },
-        };
+        this.s3sourceCredentials = new AWS.Credentials({
```
JianqinWang, line 50: :+1:

JianqinWang (approved):

lib/management/index.js
```diff
@@ -44,6 +44,24 @@ function decryptSecret(instanceCredentials, secret) {
...
+ * @param {string} encryptedSecret - encrypted secret key
+ * @param {werelogs.Logger} log - Logger object
+ * @param {function} cb - callback(error, decryptedKey)
+ * @return {undefined}
+ */
+function decryptLocationSecret(encryptedSecret, log, cb) {
+    return metadata.getObjectMD(serviceBucket, tokenKey, {}, log,
+    (err, instanceAuth) => {
+        if (err) {
+            log.error('failed to decrypt given encrypted key');
```
jonathan-gramain: A more accurate error message could be "failed to fetch decryption key from metadata for deciphering authentication secret", possibly logging `tokenKey` as well.

conf/config.json
philipyoo: Ya I agree 👍 
`raftCount` is still used though w/ ProvisionDispatcher. I'd like to remove use of the ProvisionDispatcher though, but I want things working first before making this change

jonathan-gramain (approved):

conf/config.json
rahulreddy: Agreed, let's focus on having the feature working and then we can do cleanups.

rahulreddy (approved):

philipyoo:
force pushed a rebase to update with 8.1, no conflicts

philipyoo:
@bert-e approve

PR #553, opened 2018-12-14
bf: ZENKO-1378 fix zk path in Reader, rf configs

philipyoo:
Changes in this commit:
- Fix IngestionReader pathToLogOffset
- Update BackbeatClient endpoint
- Remove unused configs

### DIFF
--- a/lib/queuePopulator/IngestionProducer.js
+++ b/lib/queuePopulator/IngestionProducer.js
@@ -32,25 +32,22 @@ class IngestionProducer {
      * pre-existing RING backend
      *
      * @constructor
-     * @param {object} sourceConfig - source config with value for
-     *                                bucketdBootstrap
+     * @param {object} sourceConfig - source config (also called bucketdConfig)
      * @param {object} qpConfig - queuePopulator config object with value for
      *                            sslEnabled
      * @param {object} s3Config - S3 config object with value for host and port
      *                            of BackbeatClient endpoint
-     * @param {string} bucket - name of source bucket to ingest from
      */
     constructor(sourceConfig, qpConfig, s3Config) {
         this.log = new Logger('Backbeat:IngestionProducer');
-        this.ringParams = {
-            bucketdBootstrap: [`${sourceConfig.host}:${sourceConfig.port}`],
-            bucketdLog: undefined,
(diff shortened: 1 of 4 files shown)


JianqinWang (approved):

lib/queuePopulator/IngestionProducer.js
```diff
@@ -60,10 +57,9 @@ class IngestionProducer {
         this.sourceHTTPAgent = new http.Agent({ keepAlive: true });
 
         this.ringReader = new BackbeatClient({
-            endpoint: 'http://' +
-                `${this.s3source.host}:${this.s3source.port}`,
+            endpoint: `${protocol}://${host}:${port}`,
             credentials: this.s3sourceCredentials,
-            sslEnabled: this.qpConfig.transport === 'https',
+            sslEnabled: sourceConfig.auth.https,
```
rahulreddy: `protocol === 'https'`, wouldn't this be sufficient so that we can remove `sourceConfig.auth.https`?

rahulreddy (changes requested):

lib/queuePopulator/IngestionProducer.js
philipyoo: good call 👍 updated

rahulreddy (approved):

lib/queuePopulator/IngestionReader.js
```diff
@@ -61,8 +57,10 @@ class IngestionReader extends LogReader {
                 }
                 this.raftId = data;
                 this.logId = `raft_${this.raftId}`;
+                this.pathToLogOffset = `/${this._targetZenkoBucket}/logState` +
```
jonathan-gramain, line 60: I would advise to add a base path like `ingestionPath` to avoid populating the root (or chroot) directly with bucket names, it may clash with other unrelated nodes (and it's better to separate ingestion matters from the rest).

philipyoo: the base path is added to the ingestion zookeeper client instance so all usage of ingestion related things will automatically have this base path 👍

philipyoo:
@bert-e approve

PR #555, opened 2018-12-15
bf: S3C-1886 fix "processing" metrics retention

jonathan-gramain:
The fix consists of a refactor of how metrics are published, in a way
that maps what has been done in 8.0 branch but has not been backported
to 7.4, however the differences between the two branches called for a
custom implementation on 7.4, a simple cherry-pick was not possible.

A simpler fix removing a conditional check on publishing 'metrics'
event would have been possible, but it would not have been easily
testable and would have kept the technical debt that has been fixed in
8.0 branch (i.e. that BackbeatConsumer should not deal with
CRR-specific semantics).

Previously the metrics were published as a "metrics" event in the
BackbeatConsumer class, possibly gathering multiple stats together
before publishing. The condition to publish them was on
"messagesConsumed > 0", which was incorrect as new metrics could have
been published after "messagesConsumed" was reset to 0. By moving the
metrics publishing outside of BackbeatConsumer and in the CRR
extension code, this gets fixed on the way, and it becomes possible to
test published metrics without relying on a specific topic name being
published to.

Note: This PR is only touching 7.4 branch, 8.0 and 8.1 are left untouched.

### DIFF
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -12,6 +12,7 @@ const AccountCredentials =
           require('../../../lib/credentials/AccountCredentials');
 const RoleCredentials =
           require('../../../lib/credentials/RoleCredentials');
+const { metricsExtension, metricsTypeProcessed } = require('../constants');
 
 const MPU_CONC_LIMIT = 10;
 
@@ -373,6 +374,13 @@ class ReplicateObject extends BackbeatTask {
                 return doneOnce(err);
             }
             partObj.setDataLocation(data.Location[0]);
+            const extMetrics = {};
+            extMetrics[this.site] = {
+                ops: 1,
+                bytes: partObj.getPartSize(),
+            };
+            this.mProducer.publishMetrics(
+                extMetrics, metricsTypeProcessed, metricsExtension, () => {});
             return doneOnce(null, partObj.getValue());
         });
     }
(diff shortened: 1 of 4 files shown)


extensions/replication/tasks/ReplicateObject.js
```diff
@@ -373,6 +374,13 @@ class ReplicateObject extends BackbeatTask {
                 return doneOnce(err);
             }
             partObj.setDataLocation(data.Location[0]);
+            const extMetrics = {};
+            extMetrics[this.site] = {
+                ops: 1,
+                bytes: partObj.getPartSize(),
+            };
+            this.mProducer.publishMetrics(
+                extMetrics, metricsTypeProcessed, metricsExtension, () => {});
```
philipyoo, line 383: Maybe needed also in `MultipleBackendTask`

jonathan-gramain: Correct me if I'm wrong, normally MultipleBackendTask is dead code for 7.4, this class was included when we started developing zenko on top of S3C code but should normally not be used by running processes. Since this PR is a no-op for 8.0 and 8.1 (as the multiple backend metrics support is already there and I preferred a minimal functional PR than a complete backport), I think we can leave as is.

bennettbuchanan: I believe the MultipleBackendTask was added before zenko, or at least, it wasn't added specifically for zenko. If I recall it was the release in which we introduced CRR to public clouds, starting with AWS.

bennettbuchanan (approved):

philipyoo (approved):

jonathan-gramain:
@bert-e approve

PR #562, opened 2018-12-18
ZENKO-1390: fix ingestionProducer.getRaftLog bugs

JianqinWang:
a. 'this.logger' is undefined
b. 'info' in response from getRaftLog was formatted incorrectly

### DIFF
--- a/lib/queuePopulator/IngestionProducer.js
+++ b/lib/queuePopulator/IngestionProducer.js
@@ -123,7 +123,7 @@ class IngestionProducer {
     getRaftLog(raftId, begin, end, targetLeader, done) {
         const recordStream = new ListRecordStream(this.log);
         recordStream.on('error', err => {
-            this.logger.error('error receiving raft log',
+            this.log.error('error receiving raft log',
             { error: err.message });
             return done(errors.InternalError);
         });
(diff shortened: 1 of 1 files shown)


philipyoo (approved):

lib/queuePopulator/IngestionProducer.js
```diff
@@ -123,7 +123,7 @@ class IngestionProducer {
     getRaftLog(raftId, begin, end, targetLeader, done) {
         const recordStream = new ListRecordStream(this.log);
         recordStream.on('error', err => {
-            this.logger.error('error receiving raft log',
+            this.log.error('error receiving raft log',
```
bennettbuchanan, line 126: Can we also update to follow the same naming scheme to help avoid this in the future?

https://github.com/scality/backbeat/blob/6ec8bd753cc8843305f53879f768e4d28092ebe0/lib/queuePopulator/IngestionProducer.js#L18 

There may be some magic with setting that for the inherited `stream.Transform`, but maybe we could just remove that in the `ListRecordStream` constructor altogether.

alexanderchan-scality (approved):

jonathan-gramain (approved):

lib/queuePopulator/IngestionProducer.js
JianqinWang: I think keeping the 2 different can help clarify that they are in different classes.

This is code copied from Arsenal and will be rf'ed in the future, at the moment if it's not breaking anything I think it's better to leave as is.

JianqinWang:
@bert-e approve

PR #577, opened 2018-12-19
feature: ZENKO-1387 Filter transition entries

bennettbuchanan:
Because the lifecycle conductor may process buckets at relatively small intervals (i.e. every five minutes, by default, based on the cron rule), we need to ensure that multiple transitions do not occur for the same object if the replication is taking longer than the interval.


tests/unit/TaskScheduler.spec.js
```diff
@@ -0,0 +1,176 @@
...
+                            contentMD5: hasUniqueContent && getRandomHash(),
+                        }),
+                    };
+                    objects.push(object);
+                    taskScheduler.push(ctx, customDoneFunc || doneFunc);
+                }
+            }
+        }
+
+        it('should each task when keys are unique', done => {
```
jonathan-gramain: `should process ...`

extensions/utils/TaskScheduler.js
```diff
@@ -0,0 +1,84 @@
...
+*/
+class TaskScheduler {
+    constructor(processingFunc) {
+        this._processingFunc = processingFunc;
+        this._replicationTasks = {};
+        this._lifecycleTasks = {};
+    }
+
+    /**
+     * Set the unqiue key in the relevant task or push to that queue.
```
jonathan-gramain: typo: `unique`

extensions/utils/TaskScheduler.js
```diff
@@ -0,0 +1,84 @@
+const async = require('async');
```
jonathan-gramain, line 1: I liked this helper being scoped to replication module, because it's doing replication-specific guards (serializing to avoid overlapping copy). Even in the context of using it for lifecycle transitions, essentially it's still a helper for the data replication service (if we consider the extensions/replication module to be a generic replication internal service, not just used for CRR).

Now the fact that lifecycle requires specificity to avoid duplicate work can be implemented in the replication module itself, since the problem lies on replication shoulders. On the other hand, it's better to try to keep loose binding between modules, i.e. not add to much module-specific logic into one another (I mean lifecycle-specific checks into replication module), but it's not always easy to abstract correctly hence I'm good with it if that helps development effort.

extensions/utils/TaskScheduler.js
```diff
@@ -0,0 +1,84 @@
...
+        const { entry } = ctx;
+        const canonicalKey = entry.getCanonicalKey();
+        const versionId = entry.getVersionId() || '';
+        const contentMD5 = entry.getContentMd5();
+        const key = `${canonicalKey}:${versionId}:${contentMD5}`;
+        // If this object is already queued, skip it.
+        if (this._lifecycleTasks[key]) {
+            return cb();
+        }
+        return this._pushToQueue(this._lifecycleTasks, ctx, key, cb);
```
jonathan-gramain, line 63: I think the `key` is not appropriate for the purpose of serializing operations correctly: for transition purpose, the problem is basically the same than with replication: cloud backends need serialized PUTs to the same object in order to provide consistent versioning, so the key must be the original object key without its version ID (and without the content MD5). That will ensure all PUT requests to the same master key will be queued in the same queue object bound to this master key.

Also, let's use the same set of queues than for replication if possible, so `_replicationTasks`, because lifecycle-triggered and CRR-triggered tasks may step on each other otherwise (even if in such case the configuration of the bucket is probably weird to trigger both replication and lifecycle at the same time, it's better to stay safe and consistent).

Note that the full key containing the version ID and MD5 looks appropriate to deduplicate identical requests (that we can consider only when they come from lifecycle, e.g. to not care about metadata-only update cases, eventually we could make it more generic with a few extra checks).

jonathan-gramain (changes requested):

extensions/utils/TaskScheduler.js
bennettbuchanan: I had debated this, and felt it could go either way. But I can agree with this, considering it is, in fact, still performing CRR in the end regardless of what feature (i.e. replication or lifecycle) it is doing it for.

bennettbuchanan:
Closing this as it needs significant refactors due to forthcoming data-mover additions. I will reopen with the necessary changes once that has been merged such that the diff makes sense and can be applied.

PR #578, opened 2018-12-20
ZENKO-1377: tests for IngestionReader

JianqinWang:
Depends on https://github.com/scality/Arsenal/pull/691


tests/utils/MockZkClient.js
```diff
@@ -0,0 +1,11 @@
+class MockZkClient {
```
philipyoo: I feel like we should keep this mock solely in the ingestion reader test file because I don't know if this could be re-used by other test components

tests/utils/TestIngestionQP.js
```diff
@@ -0,0 +1,25 @@
+const assert = require('assert');
+
+/**
+ * The QueuePopulatorExtension class sends entries to kafka, but for testing
+ * purposes we may not have kafka setup for quick testing. This class will
+ * mock some of the functions that are used by the logReader classes, and will
+ * assert that the input from the logReaders is as expected.
+ */
+class TestingQueuePopulator {
```
philipyoo: Same for this mock

tests/utils/DummyRequestLogger.js
```diff
@@ -0,0 +1,59 @@
+class DummyRequestLogger {
```
philipyoo: nice! 👍

philipyoo (approved):

tests/config.json
```diff
@@ -34,20 +34,25 @@
...
-            "topic": "backbeat-test-ingestion",
+            "topic": "backbeat-ingestion",
             "zookeeperPath": "/ingestion",
+            "cronRule": "*/5 * * * * *",
             "sources": [
                 {
-                    "name": "ingest-site-1",
+                    "name": "zenkobucket",
+                    "bucket": "bucket1",
+                    "prefix": "location-name",
```
philipyoo: minor: prefix, cronRule, zookeeperSuffix, and raftCount can be removed

philipyoo (approved):
the package-lock seems weird.. the diff is really big for an arsenal package update

tests/config.json
JianqinWang: `-I would prefer to open a separate pr if this ends up being the only request for changes-`

package-lock.json is bloated, going to fix that and this, thanks

philipyoo: sounds good

tests/config.json
```diff
@@ -34,20 +34,21 @@
...
                 "account": "md-ingestion"
             },
-            "topic": "backbeat-test-ingestion",
+            "topic": "backbeat-ingestion",
             "zookeeperPath": "/ingestion",
+            "cronRule": "*/5 * * * * *",
             "sources": [
                 {
-                    "name": "ingest-site-1",
+                    "name": "zenkobucket",
```
rahulreddy, line 42: We no longer get this config from `config.json`. Does it make sense to still rely on this?

JianqinWang: https://github.com/scality/backbeat/blob/485f64744a0e17b5d1dbde95227ebb4c80440162/lib/queuePopulator/IngestionReader.js#L31

I did it so I could create a simpler producer used in testing

tests/utils/DummyRequestLogger.js
bennettbuchanan: Can you use the mock logger in `tests/utils/fakeLogger.js`?  I see this tracks the logging with a counter, but not sure how that's used, or if we can combine these two logger mocks into a single utility.

JianqinWang: done :) thanks

tests/config.json
philipyoo: For pause/resume, I created tests that mocks out IngestionPopulator's MongoClient (for fetching all ingestion enabled buckets), and the Config event emitter for getting updates on storage locations.

I think it would be very hard to build tests w/out mocking this out. Based off how pause/resume went, I can do the IngestionPopulator functional tests similarly where I mock similar functions, and maybe refactor so we have mock helper classes for this stuff?

Also, in IngestionReader, with decrypted location secret, I feel we should mock this out (like how Jay did it in this PR) because it relies on an ssh key generated in Orbit world. Ultimately, all we care about is passing the "verySecretKey1" to the IngestionProducer class as the secret key so we can continue using the CI in-memory cloudserver client.

tests/functional/ingestion/IngestionReader.js
```diff
@@ -0,0 +1,251 @@
...
+                    return next();
+                }),
+            next => this.ingestionReader.processLogEntries({}, err => {
+                assert.ifError(err);
+                return next();
+            }),
+        ], done);
+    });
+
+    // it('should successfully ', done => {
```
philipyoo: i guess this can be removed

philipyoo (commented):
Can you try rebuilding the package-lock file? Remove your current lock file changes and use current 8.1 lock file, remove node_modules dir, and then try installing again?
I know some modules and versions were changed in this PR, but I still think the lock file is behaving strangely

JianqinWang:
@philipyoo updated package-lock

tests/functional/ingestion/IngestionReader.js
```diff
@@ -0,0 +1,247 @@
...
+                }, {});
+                return next();
+            },
+            next => {
+                this.m.insertOne(dummySSHKey, {});
+                return next();
+            },
+            next => {
+                zkClient.connect();
+                zkClient.once('error', next);
```
rahulreddy: This event can be caused multiple times, try protecting it with `cbOnce` utility in Backbeat.

tests/functional/ingestion/IngestionReader.js
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * The QueuePopulatorExtension class sends entries to kafka, but for testing
+ * purposes we may not have kafka setup for quick testing. This class will
+ * mock some of the functions that are used by the logReader classes, and will
+ * assert that the input from the logReaders is as expected.
+ */
+class TestIngestionQP {
+    constructor(params) {
+        this.config = params.config;
+        this.expectedEntry = params.expectedEntry;
```
rahulreddy, line 27: You should initialize this._batch in the constructor.

tests/functional/ingestion/IngestionReader.js
```diff
@@ -0,0 +1,247 @@
...
+        const logOffset = this.ingestionReader.getLogOffset();
+        // value initialized when creating MockZkClient
+        assert.equal(logOffset, mockLogOffset);
+        done();
+    });
+
+    it('should succesfully ingest new bucket with existing objects', done => {
+        async.waterfall([
+            next => zkClient.setData(this.ingestionReader.pathToLogOffset,
+                Buffer.from('1'), -1, err => {
```
rahulreddy, line 241: As discussed, you can add a comment here that reading from logOffset 1 would mean ingesting existing objects.

rahulreddy (changes requested):
Some minor comments

tests/functional/ingestion/IngestionReader.js
JianqinWang: https://github.com/scality/backbeat/commit/0c876539c83ef11190a41e140cab2a2756b2dc2e

In the follow-up I actually end up deleting `TestIngestionQP` since we're already checking through kafka, should I add that change to the pr here

rahulreddy: Ok you can just leave this as-is then and follow up with the changes in another PR.

rahulreddy (approved):

JianqinWang:
@bert-e approve

PR #582, opened 2018-12-31
ZENKO-1414 avoid replication service account in lifecycle

jonathan-gramain:
Modify BackbeatMetadataProxy helper:

- move it from replication extension to backbeat lib so all modules
  can use it

- pass an endpoint and authentication params to the constructor
  (instead of relying on replication config directly) so each module
  can tailor to its needs


bennettbuchanan (approved):

philipyoo (approved):

jonathan-gramain:
@bert-e approve

PR #583, opened 2019-01-02
refactor: ZENKO-734 remove destEntry from MultipleBackendTask

jonathan-gramain:
The "destEntry" param is a relic from ReplicateObject base class, that
does not play a role in zenko replication code: "destEntry" in CRR
code represents the metadata pushed to the target site, which has no
materialization in zenko replication as every operation is carried out
by the multiple backend layer.

Also, when it comes to changes related to using the queue processor
for lifecycle transitions, there are more reasons why it can become an
important cleanup.

This commit is a pure refactor, the few extra functions directly come
from ReplicateObject class without their "destEntry" and CRR-specific
behavior, that I deemed appropriate to "duplicate" to avoid
possibilities of code path incompatibility between both classes and
separate the matters better between zenko replication and CRR.

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -138,14 +138,13 @@ class MultipleBackendTask extends ReplicateObject {
      * Used when an operation performed in the process of replicating a
      * multipart upload fails.
      * @param {ObjectQueueEntry} sourceEntry - The source object entry
-     * @param {ObjectQueueEntry} destEntry - The destination object entry
      * @param {String} uploadId - The upload ID of the initiated MPU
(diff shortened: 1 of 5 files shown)


bennettbuchanan (approved):
Thanks for this cleanup!

philipyoo (approved):

jonathan-gramain:
I'm going to create a new task for lifecycle transitions separate from MultipleBackendTask, hence this PR is not required anymore in the context of lifecycle transitions. It can be a nice cleanup still, so keeping it around.

jonathan-gramain:
And since it's potentially impactful I wish to have a run of zenko tests before merging.

jonathan-gramain:
@bert-e approve

PR #584, opened 2019-01-02
bugfix: ZENKO-1418 update Dockerfile

rahulreddy:
This commit addresses the quirks needed to build a clean docker
image. It also changes the base image to alpine-based reducing the
image size by half.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,33 @@
-FROM node:8
+FROM node:8-alpine
 
 WORKDIR /usr/src/app
 
-RUN apt-get update \
-    && apt-get install -y jq wget --no-install-recommends
+RUN apk --no-cache add \
+    bash \
+    g++ \
+    ca-certificates \
+    lz4-dev \
+    musl-dev \
+    cyrus-sasl-dev \
+    openssl-dev \
+    make \
+    python
+
+RUN apk add --no-cache --virtual .build-deps gcc zlib-dev libc-dev bsd-compat-headers py-setuptools bash git
 
 ENV DOCKERIZE_VERSION v0.6.1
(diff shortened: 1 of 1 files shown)


jonathan-gramain (approved):

giacope (approved):

ssalaues (approved):

rahulreddy:
@bert-e approve

Dockerfile
```diff
@@ -1,22 +1,33 @@
...
+    g++ \
+    ca-certificates \
+    lz4-dev \
+    musl-dev \
+    cyrus-sasl-dev \
+    openssl-dev \
+    make \
+    python
+
+RUN apk add --no-cache --virtual .build-deps gcc zlib-dev libc-dev bsd-compat-headers py-setuptools bash git
```
NicolasT, line 16: Aren't `g++`, `lz4-dev`, `musl-dev`, `cyrus-sasl-dev`, `openssl-dev`, `make` and `python` also build-only deps?

Dockerfile
```diff
@@ -1,22 +1,33 @@
...
 
 COPY package.json /usr/src/app
 RUN npm install --production \
     && rm -rf /var/lib/apt/lists/* \
     && npm cache clear --force \
     && rm -rf ~/.node-gyp \
-    && rm -rf /tmp/npm-*
-
+    && rm -rf /tmp/npm-* \
+    && apk del .build-deps
```
NicolasT, line 30: Because there are various steps/layers between this and the install phase, this doesn't really do much to reduce the image size.

Dockerfile
rahulreddy: They are but it was taking longer to figure out which ones of them are needed for runtime as `noderdkafka` was bailing out on startup. I figured that I can follow up later.

Dockerfile
rahulreddy: I agree. The initial size reduction is because of the switch to alpined-based image, I am gonna clean up build time dependencies after this.

PR #593, opened 2019-01-08
bugfix: ZENKO-1430 Pass correct arguments

bennettbuchanan:
The function parameters do not match those passed here: https://github.com/scality/backbeat/blob/development/7.4/extensions/replication/tasks/MultipleBackendTask.js#L589. 

The effect of this is that a failed initiate-multipart-upload operation, which would otherwise be retried, does not use the exponential back-off.

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -181,12 +181,12 @@ class MultipleBackendTask extends ReplicateObject {
         });
     }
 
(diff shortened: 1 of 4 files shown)


alexanderchan-scality (approved):

rahulreddy (approved):

rahulreddy:
This is scary though. Can we add a test (unit may be) verifying the params?

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #599, opened 2019-01-29
rf: S3C-1944 backport route parsing

philipyoo:
This isn't a perfect backport as 7.4 differs quite a bit
with 8.0/8.1. The general parsing logic is being
backported (with additions of additional newer changes
in 8.1) so parsing logic matches between versions.

This is in preparation to adding an additional route to 7.4
and easier forward ports later on. This can also help with
easier backporting in the future from 8.1 back to 7.4. For
example, we still need to backport site-level metrics

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -96,77 +96,123 @@ class BackbeatAPI {
     }
 
     /**
-     * Check if incoming request is valid
+     * Find valid route from list of routes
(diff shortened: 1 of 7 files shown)


philipyoo:
deep healthcheck feature is being pushed out, so this is not required immediately

bennettbuchanan (approved):

jonathan-gramain (approved):

PR #601, opened 2019-01-30
bf: ZENKO-1424 bucket cseq snapshot

philipyoo:
Only used during snapshot phase

Changes in this PR:
- Introduce metadata route to fetch bucket cseq with
  BackbeatClient
- Fetch cseq and set as `logStats.nbLogRecordsRead`
  during snapshot phase only

### DIFF
--- a/lib/clients/backbeat-2017-07-01.api.json
+++ b/lib/clients/backbeat-2017-07-01.api.json
@@ -1076,6 +1076,35 @@
             "output": {
                 "shape": "ObjectMetadataObject"
             }
+        },
+        "GetBucketCseq": {
+            "http": {
+                "method": "GET",
+                "requestUri": "/_/metadata/default/informations/{Bucket}"
+            },
+            "input": {
+                "type": "structure",
+                "required": [
+                    "Bucket"
+                ],
+                "members": {
+                    "Bucket": {
+                        "location": "uri",
+                        "locationName": "Bucket"
+                    }
+                }
+            },
+            "output": {
+                "type": "list",
+                "member": {
+                    "type": "structure",
+                    "members": {
+                        "cseq": {
+                            "type": "integer"
+                        }
+                    }
+                }
+            }
         }
     },
     "shapes": {
--- a/lib/queuePopulator/IngestionProducer.js
+++ b/lib/queuePopulator/IngestionProducer.js
@@ -149,6 +149,7 @@ class IngestionProducer {
      * @param {string} [state.keyMarker] - NextKeyMarker
      * @param {function} done - callback(error, response) where response has:
      *   logRes {object} - metadata logs formed as RaftLogEntry put entries
+     *   cseq {integer} - cseq at start of snapshot phase
      *   initState {object} - returns status for snapshot process
      *   initState.isStatusComplete {boolean} - true/false
      *   [initState.versionMarker] {string} - NextVersionIdMarker, if any
@@ -156,7 +157,21 @@ class IngestionProducer {
      * @return {undefined}
      */
     snapshot(bucketName, state, done) {
+        // get cseq ONLY on first snapshot request, indicated by markers
+        let initialCseq;
         async.waterfall([
+            next => {
+                if (!state.versionMarker && !state.keyMarker) {
+                    return this._getBucketCseq(bucketName, (err, cseq) => {
+                        if (err) {
+                            return next(err);
+                        }
+                        initialCseq = cseq;
+                        return next();
+                    });
+                }
+                return process.nextTick(next);
+            },
             next => this._getObjectVersionsList(bucketName, state, next),
             (data, next) => {
                 const {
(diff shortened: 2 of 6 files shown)


philipyoo:
putting on hold as this will most likely need to change based on jira ticket 1452

jonathan-gramain (approved):

lib/queuePopulator/IngestionProducer.js
```diff
@@ -171,6 +186,9 @@ class IngestionProducer {
                         logRes,
                         initState: { status, versionMarker, keyMarker },
                     };
+                    if (initialCseq) {
```
bennettbuchanan, line 193: Is there a chance the `cseq` can be zero?

lib/queuePopulator/IngestionProducer.js
```diff
@@ -381,6 +399,28 @@ class IngestionProducer {
...
+    /**
+     * Get bucket cseq
+     * @param {string} bucket - bucket name
+     * @param {function} done - callback(err, cseq) where `cseq` is an integer
+     * @return {undefined}
+     */
+    _getBucketCseq(bucket, done) {
+        return this._ringReader.getBucketCseq({ Bucket: bucket },
+        (err, data) => {
+            if (err || !data[0] || !data[0].cseq) {
```
bennettbuchanan: In what case would an error be returned? Could it ever by worthwhile to retry this request before declaring it an internal error?

lib/queuePopulator/IngestionProducer.js
philipyoo: not that I'm aware of. I've observed cseq to be greater than 1 even when there are no objects in a given bucket. The first few log entries are the bucket entries.
I don't think (at least for oob ring) cseq ever being less than 2 - one for the bucket creation, and another for versioning on bucket

lib/queuePopulator/IngestionProducer.js
philipyoo: Definitely, I agree, we need to make the service more durable adding retries where applicable (especially on processor side).

I'm not sure a retry should be added here. Since this is on populator side, it'll trigger another try on the next scheduled timer. Instead, I think adding validations of the source should be added in the IngestionPopulator class. I have another ticket (1449) for this specific case

I think it's a little strange to check for data response. That should be the expected. I also agree, I should escalate the original error rather than declaring this an internal error. Thank you, updating!

JianqinWang (approved):

lib/queuePopulator/IngestionProducer.js
```diff
@@ -383,6 +401,28 @@ class IngestionProducer {
...
+            if (err) {
+                this.log.error('error getting bucket cseq', {
+                    method: 'IngestionProducer._getBucketCseq',
+                    error: err,
+                    bucket,
+                });
+                return done(err);
+            }
+            // cseq returned by all nodes. Just return the first node response
+            return done(null, data[0].cseq);
```
nicolas2bert, line 431: It might be safe to check that `data.length > 0` and `data[0]` is not undefined.

alexanderchan-scality (approved):

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -1076,6 +1076,35 @@
             "output": {
                 "shape": "ObjectMetadataObject"
             }
+        },
+        "GetBucketCseq": {
```
jonathan-gramain, line 1080: We can later extend this route to return more (or all) available info eventually, so could be renamed (something like GetBucketMetadataInfo?)

lib/queuePopulator/IngestionProducer.js
jonathan-gramain: I recently learnt that cseq/aseq start from 1.

lib/queuePopulator/IngestionProducer.js
philipyoo: ty, updated :)
oops, I made a mistake, the BackbeatClient likes to return as an object with key of '0' :/ fixing now

nicolas2bert (approved):

lib/queuePopulator/IngestionProducer.js
```diff
@@ -383,6 +401,36 @@ class IngestionProducer {
...
+        (err, data) => {
+            if (err) {
+                this.log.error('error getting bucket cseq', {
+                    method: 'IngestionProducer._getBucketCseq',
+                    error: err,
+                    bucket,
+                });
+                return done(err);
+            }
+            if (!data || !data.length > 0 || !data[0]) {
```
jonathan-gramain: `!data.length > 0` is redundant since if the case, `!data[0]` will also return true.

lib/clients/backbeat-2017-07-01.api.json
philipyoo: I think this route only gets the raft session information for a given bucket. The response contains aseq/cseq/vseq, etc. I'm also selecting only cseq on response, but if there is a need for this information elsewhere, I can expand it here for future use case 👍

philipyoo:
@bert-e approve

PR #603, opened 2019-02-02
ft: ZENKO-1452 update snapshot to use s3 api

philipyoo:
Goes with https://github.com/scality/Arsenal/pull/713

Changes in this PR:
- Add S3 client
- Object list versions to use S3 api
- Snapshot methods to reference a single bucket instead of list of buckets
- Use NextKeyMarker and NextVersionIdMarker for splitting the snapshot phase.
   This allows snapshot to complete processing smaller batches. This also helps with
   pause/resume of ingestion.
- Save state in zookeeper as paths:
      BASE: `/ingestion/<bucket>/logState/init`
      NODES: `status`, `versionMarker`, `keyMarker`
   where status is a string of either `complete` || `incomplete`

Suggestions on naming, especially zookeeper path names, is appreciated!

Edit:
Will implement Rahul's suggestion for following change to zookeeper path:
- Replace `<bucket>` in paths with the unique id of the bucket

This will be done on a later ticket, possibly along with ticket (1441) as I think they relate.
Other suggestions on zookeeper path is welcome!

### DIFF
--- a/lib/queuePopulator/IngestionProducer.js
+++ b/lib/queuePopulator/IngestionProducer.js
@@ -144,14 +144,40 @@ class IngestionProducer {
      * generate a listing of all current objects that exists on the source
      * bucket, including the bucket MD to create the correct entries
      * @param {string} bucketName - name of source bucket
-     * @param {function} done - callback(error, resLog)
+     * @param {object} state - previous state used to paginate version listing
+     * @param {string} [state.versionMarker] - NextVersionIdMarker
+     * @param {string} [state.keyMarker] - NextKeyMarker
+     * @param {function} done - callback(error, response) where response has:
+     *   logRes {object} - metadata logs formed as RaftLogEntry put entries
+     *   initState {object} - returns status for snapshot process
+     *   initState.isStatusComplete {boolean} - true/false
+     *   [initState.versionMarker] {string} - NextVersionIdMarker, if any
+     *   [initState.keyMarker] {string} - KeyMarker, if any
      * @return {undefined}
      */
-    snapshot(bucketName, done) {
+    snapshot(bucketName, state, done) {
         async.waterfall([
-            next => this._getObjectVersionsList(bucketName, next),
-            (versionList, next) =>
-                this._getBucketObjectsMetadata(bucketName, versionList, next),
+            next => this._getObjectVersionsList(bucketName, state, next),
+            (data, next) => {
+                const {
+                    IsTruncated, versionList, versionMarker, keyMarker,
+                } = data;
+                this._getBucketObjectsMetadata(bucketName, versionList,
+                (err, logRes) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    const response = {
+                        logRes,
+                        initState: {
+                            isStatusComplete: !IsTruncated,
+                            versionMarker,
+                            keyMarker
+                        },
+                    };
+                    return next(null, response);
+                });
+            },
         ], done);
     }
 
@@ -232,19 +258,27 @@ class IngestionProducer {
      * Get the list of object versions for a bucket
      *
      * @param {string} bucket - bucket name
+     * @param {object} state - previous state used to paginate version listing
+     * @param {string} [state.versionMarker] - NextVersionIdMarker
+     * @param {string} [state.keyMarker] - NextKeyMarker
      * @param {function} done - callback function
      * @return {object} list of objects for each bucket, including a duplicate
      *   entry for IsLatest versions
      */
-    _getObjectVersionsList(bucket, done) {
+    _getObjectVersionsList(bucket, state, done) {
         if (bucket === constants.usersBucket ||
             bucket === constants.metastoreBucket) {
             return done();
         }
-        // TODO: VersionIdMarker && KeyMarker
+        const { versionMarker, keyMarker } = state;
         const params = {
             Bucket: bucket,
         };
+        // if previous state, should paginate here
+        if (versionMarker && keyMarker) {
+            params.VersionIdMarker = versionMarker;
+            params.KeyMarker = keyMarker;
+        }
         // TODO: For testing, I can set MaxKeys
         const req = this._s3Client.listObjectVersions(params);
         attachReqUids(req, this.requestLogger);
@@ -259,24 +293,19 @@ class IngestionProducer {
             }
             const {
                 IsTruncated,
-                // NextKeyMarker,
-                // NextVersionIdMarker,
+                NextKeyMarker,
+                NextVersionIdMarker,
                 Versions,
                 DeleteMarkers
             } = data;
 
-            if (IsTruncated) {
-                // need NextKeyMarker
-                // need NextVersionIdMarker
-            }
-
-            const versionList = [...Versions, ...DeleteMarkers];
-            // find all IsLatest versions since we need these as separate
-            // metadata entries
-            const latestVersions = versionList.filter(v => v.IsLatest)
-                .map(v => Object.assign({}, v, { isLatest: true }));
-
-            return done(null, versionList.concat(latestVersions));
+            const response = {
+                versionList: [...Versions, ...DeleteMarkers],
+                IsTruncated,
+                versionMarker: NextVersionIdMarker,
+                keyMarker: NextKeyMarker,
+            };
+            return done(null, response);
         });
     }
 
@@ -294,15 +323,10 @@ class IngestionProducer {
             return done();
         }
 
-        return async.mapLimit(versionList, 10, (version, cb) => {
-            const { Key, VersionId, isLatest } = version;
-
-            let objectKey;
-            if (isLatest) {
-                objectKey = Key;
-            } else {
-                objectKey = `${Key}${VID_SEP}${decode(VersionId)}`;
-            }
+        const objectMDList = [];
+        return async.eachLimit(versionList, 10, (version, cb) => {
+            const { Key, VersionId, IsLatest } = version;
+            const objectKey = `${Key}${VID_SEP}${decode(VersionId)}`;
 
             const req = this._ringReader.getObjectMetadata({
                 Bucket: bucket,
@@ -317,17 +341,27 @@ class IngestionProducer {
                     });
                     return cb(err);
                 }
-                return cb(null, {
+                const objectEntry = {
                     res: data,
                     objectKey,
                     bucketName: bucket,
-                });
+                };
+                objectMDList.push(objectEntry);
+                if (IsLatest) {
+                    // duplicate the entry w/out the version id in the object
+                    // key to represent the master key
+                    objectMDList.push(Object.assign({}, objectEntry, {
+                        // key name w/out version id
+                        objectKey: Key,
+                    }));
+                }
+                return cb();
             });
-        }, (err, objectMDs) => {
+        }, err => {
             if (err) {
                 return done(err);
             }
-            return this._createAndPushEntry(objectMDs, done);
+            return this._createAndPushEntry(objectMDList, done);
         });
     }
 
--- a/lib/queuePopulator/IngestionProducer.js
+++ b/lib/queuePopulator/IngestionProducer.js
@@ -1,10 +1,15 @@
 const async = require('async');
 const AWS = require('aws-sdk');
 const http = require('http');
+const https = require('https');
 const jsonStream = require('JSONStream');
 const stream = require('stream');
 const Logger = require('werelogs').Logger;
 const { constants, errors } = require('arsenal');
+
+const { decode } = require('arsenal').versioning.VersionID;
+const VID_SEP = require('arsenal').versioning.VersioningConstants
+          .VersionId.Separator;
 
 const BackbeatClient = require('../clients/BackbeatClient');
 const { attachReqUids } = require('../clients/utils');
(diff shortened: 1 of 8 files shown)


lib/queuePopulator/IngestionReader.js
```diff
@@ -1,9 +1,15 @@
 const async = require('async');
+const VID_SEP = require('arsenal').versioning.VersioningConstants
+          .VersionId.Separator;
 
 const IngestionProducer = require('./IngestionProducer');
 const LogReader = require('./LogReader');
 const { decryptLocationSecret } = require('../management/index');
 
+function _isVersionedLogKey(key) {
+    return key.split(VID_SEP)[1];
```
jonathan-gramain: I would advise to convert the value to an actual boolean before returning it.

lib/queuePopulator/IngestionReader.js
```diff
@@ -202,13 +306,23 @@ class IngestionReader extends LogReader {
...
-        if (this.remoteLogOffset /* &&
-        config.queuePopulator.logSource.indexOf('ingestion') > -1 */) {
-            batchState.nextLogOffset = this.remoteLogOffset;
-            // return done();
-        } else {
+        // initState.cseq is only fetched at very start of snapshot phase.
+        // We want to save cseq right before we started snapshot
+        // phase to guarantee we don't miss any new entries while snapshot
+        // is in process
+        if (initState.cseq) {
```
jonathan-gramain: Is `initState` guaranteed to be an object here? Otherwise, it should be checked first.

lib/queuePopulator/IngestionProducer.js
```diff
@@ -41,26 +45,53 @@ class IngestionProducer {
...
+            secretAccessKey: sourceConfig.auth.secretKey,
+        });
+
+        this._ringReader = new BackbeatClient({
+            endpoint,
+            credentials: s3sourceCredentials,
             sslEnabled: protocol === 'https',
-            httpOptions: { agent: this.sourceHTTPAgent, timeout: 0 },
+            httpOptions: {
+                agent: new http.Agent({ keepAlive: true }),
```
rahulreddy: You have to setup https.Agent if protocol === 'https'

lib/queuePopulator/IngestionProducer.js
```diff
@@ -41,26 +45,53 @@ class IngestionProducer {
...
+                          `${protocol}://${host}:8000` :
+                          endpoint;
+        this._s3Client = new AWS.S3({
+            endpoint: s3endpoint,
+            credentials: s3sourceCredentials,
+            sslEnabled: false,
+            s3ForcePathStyle: true,
+            signatureVersion: 'v4',
+            httpOptions: {
+                agent: new http.Agent({ keepAlive: true }),
```
rahulreddy: Same here, setup https.Agent if protocol === 'https'

lib/queuePopulator/IngestionProducer.js
philipyoo: I think I need your input on implementing this. For https, do we require the remote source certificates?

I'll reach out tomorrow

tests/functional/ingestion/S3Mock.js
```diff
@@ -0,0 +1,96 @@
+const async = require('async');
+const AWS = require('aws-sdk');
+// const http = require('http');
+
+// const { MetadataMock } = require('arsenal').testing.MetadataMock;
+const testConfig = require('../../config.json');
+const BackbeatClient = require('../../../lib/clients/BackbeatClient');
+
+const sourceConfig = testConfig.extensions.ingestion.sources[0];
```
JianqinWang: instead of declaring sourceConfig here, can you pass it as a param for setupS3Mock

tests/functional/ingestion/S3Mock.js
```diff
@@ -0,0 +1,108 @@
...
+ */
+function setupS3Mock(sourceInfo, cb) {
+    const { bucket } = sourceInfo;
+    async.series([
+        next => awsClient.createBucket({ Bucket: bucket }, next),
+        next => awsClient.putBucketVersioning({
+            Bucket: bucket,
+            VersioningConfiguration: { Status: 'Enabled' },
+        }, next),
+        next => bbClient.getObjectList({ Bucket: sourceConfig.bucket },
```
JianqinWang: shouldn't this use the `bucket` from `sourceInfo`

lib/queuePopulator/IngestionProducer.js
```diff
@@ -185,95 +251,135 @@ class IngestionProducer {
...
         }
-        return async.mapSeries(bucketObjectList, (bucket, cb) => {
-            if (!bucket) {
-                return cb();
+
+        return async.mapLimit(versionList, 10, (version, cb) => {
+            const { Key, VersionId, isLatest } = version;
+
+            let objectKey;
+            if (isLatest || process.env.CI === 'true') {
```
philipyoo: the `CI` condition here was added because the object key entries will have their version id's appended to them otherwise. The IsLatest version will have 2 entries (one with a version id appended to object key name, and one without).
For test cases, the version id is assigned by the s3 client and there isn't a reasonable way of applying this version id to the md requests. Instead, for tests, I will leave version id stripped from the object keys. So for the IsLatest version, there will be 2 identical entries.

JianqinWang (approved):

lib/queuePopulator/IngestionReader.js
```diff
@@ -57,14 +60,114 @@ class IngestionReader extends LogReader {
...
+        async.mapSeries(initPathNodes, (pathNode, cb) => {
+            // if complete, we do not care to get other zookeeper data
+            if (statusResponse === 'complete') {
+                return process.nextTick(cb);
+            }
+            const path = `${this.bucketInitPath}/${pathNode}`;
+            return this.zkClient.getData(path, (err, data) => {
+                if (err) {
+                    if (err.name !== 'NO_NODE') {
+                        this.logger.error(
```
philipyoo: these should use the request logger from batchState

lib/queuePopulator/IngestionProducer.js
```diff
@@ -95,16 +144,37 @@ class IngestionProducer {
...
-        ], err => done(err, this.resLog));
+            next => this._getObjectVersionsList(bucketName, state, next),
+            (data, next) => {
+                const { versionList, versionMarker, keyMarker } = data;
+                this._getBucketObjectsMetadata(bucketName, versionList,
+                (err, logRes) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    const status = (!versionMarker && !keyMarker) ?
```
rahulreddy: `IsTruncated` from the response should be enough to detect the end of listing.

lib/queuePopulator/IngestionProducer.js
```diff
@@ -185,95 +251,135 @@ class IngestionProducer {
...
+            const {
+                IsTruncated,
+                NextKeyMarker,
+                NextVersionIdMarker,
+                Versions,
+                DeleteMarkers
+            } = data;
+
+            const versionList = [...Versions, ...DeleteMarkers];
+            // find all IsLatest versions since we need these as separate
```
rahulreddy: I don't understand the reasoning behind this. You are taking all the latest versions and appending them to the end of the list. Is is because you want to process them last?

lib/queuePopulator/IngestionProducer.js
rahulreddy: This is because in the Raft log, you have a master and the latest versioned key pointing to the master. For tests, you can generate VersionId using the Versioning utilities if you want and apply them. We should avoid using `CI` env as much as possible as it creates a plot hole.

lib/queuePopulator/IngestionReader.js
```diff
@@ -47,6 +47,9 @@ class IngestionReader extends LogReader {
     }
 
     setup(done) {
+        const bucketZkRootPath = `/${this._targetZenkoBucket}/logState`;
```
rahulreddy: This would be nice in the constructor as `this.zkBasePath`. It's more explicit and testable.

lib/queuePopulator/IngestionReader.js
```diff
@@ -57,14 +60,115 @@ class IngestionReader extends LogReader {
                 }
                 this.raftId = data;
                 this.logId = `raft_${this.raftId}`;
-                this.pathToLogOffset = `/${this._targetZenkoBucket}/logState` +
-                    `/${this.logId}/logOffset`;
+                this.pathToLogOffset =
+                    `${bucketZkRootPath}/${this.logId}/logOffset`;
```
rahulreddy: I recommend constructing these paths in a static method like `getLogOffsetPath`. These can be leveraged in unit tests and sanity checks.

lib/queuePopulator/IngestionReader.js
```diff
@@ -57,14 +60,115 @@ class IngestionReader extends LogReader {
...
+     * @param {Logger.newRequestLogger} logger - request logger object
+     * @param {function} done - callback(error, object)
+     *   where object.versionMarker is the NextVersionIdMarker
+     *   where object.keyMarker is the NextKeyMarker
+     * @return {undefined}
+     */
+    _readInitState(logger, done) {
+        const initPathNodes = ['status', 'versionMarker', 'keyMarker'];
+        let statusResponse;
+        async.mapSeries(initPathNodes, (pathNode, cb) => {
```
rahulreddy: Can't this just be `async.map`? What's the reason that it needs be fetched one at a time?

lib/queuePopulator/IngestionReader.js
```diff
@@ -57,14 +60,115 @@ class IngestionReader extends LogReader {
...
+     * @param {function} done - callback(error)
+     * @return {undefined}
+     */
+    _writeInitState(initState, logger, done) {
+        // initState is set by each request of processLogEntries. If undefined,
+        // we did not go through snapshot phase
+        if (!initState) {
+            return process.nextTick(done);
+        }
+        const initPathNodes = ['status', 'versionMarker', 'keyMarker'];
```
rahulreddy: This is being repeated. I would say just make a static method that gives out these paths.

lib/queuePopulator/IngestionReader.js
```diff
@@ -77,50 +181,42 @@ class IngestionReader extends LogReader {
...
-                return this._iProducer.snapshot(this.bucket, (err, res) => {
+                return this._iProducer.snapshot(this.bucket, initState,
+                (err, res) => {
                     if (err) {
                         logger.error('error generating snapshot for ' +
                         'ingestion', { err });
                         return next(err);
                     }
-                    batchState.logRes = { info: { start: 1 }, log: res };
+                    batchState.logRes = { info: { start: 1 }, log: res.logRes };
```
rahulreddy, line 227: Check if `res` is defined before accessing its properties.

lib/queuePopulator/IngestionReader.js
```diff
@@ -77,50 +181,42 @@ class IngestionReader extends LogReader {
...
-                if (offset === 1) {
-                    this.newIngestion = true;
-                }
-                return next();
-            }),
-            next => {
-                if (!this.newIngestion) {
+            next => this._readInitState(logger, next),
+            (initState, next) => {
+                if (initState.status === 'complete') {
```
rahulreddy: For the status strings, use constants declared either in the constructor. Something like this is more readable
```js
initState.status === this.STATUS_COMPLETE

```

rahulreddy (changes requested):

lib/queuePopulator/IngestionProducer.js
philipyoo: I should clean this up, I agree it's pretty weird logic.
Here, I wanted to accumulate all entries into a single list, including the master version, which I indicate by duplicating the `IsLatest` versions and putting a flag on them for the next method to process.

In addition, I realize it's probably best to not duplicate entry here, but after fetching object metadata (on next method) for the `IsLatest` version, I should create the duplicate entry there. It is possible that the `IsLatest` version and master version object metadata fetch might differ otherwise.

I'll look into removing the logic here, and creating the duplicate entry after fetching object metadata for `IsLatest` versions on the next method (`_getBucketObjectsMetadata`)

lib/queuePopulator/IngestionReader.js
philipyoo: I tried to reduce i/o operations here by first trying to check `status` value. If the `status` is "complete", we don't need the other values.
Reasoning for doing this is that this check (`_readInitState`) is applied every `processLogEntries`, and the snapshot phase is a temporary phase. The trade-off is the async series calls to fetch all nodes, but this is only applicable during the snapshot.

lib/queuePopulator/IngestionReader.js
philipyoo: Is it not a good idea to use a constant because it's an array?

rahulreddy: It's good but if it's being repeated, it's better to declare them once and share them across. This makes it easy to keep track of them instead of having to hunt for them in the code.

lib/queuePopulator/IngestionReader.js
philipyoo: little hesitant to do this change because this instance variable is also used in LogReader

lib/queuePopulator/IngestionProducer.js
```diff
@@ -95,16 +147,39 @@ class IngestionProducer {
...
+            (data, next) => {
+                const {
+                    IsTruncated, versionList, versionMarker, keyMarker,
+                } = data;
+                this._getBucketObjectsMetadata(bucketName, versionList,
+                (err, logRes) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    const status = IsTruncated ?
```
bennettbuchanan: Seems like the global vars are unneeded. Could it be simplified to just be a boolean? For example:

```
initState: {
    isStatusComplete: <bool>,
    [...]
}
```

That way you can just set it to whatever the value of `IsTruncated` is.

lib/queuePopulator/IngestionProducer.js
```diff
@@ -185,95 +256,142 @@ class IngestionProducer {
...
+            // metadata entries
+            // const latestVersions = versionList.filter(v => v.IsLatest)
+            //     .map(v => Object.assign({}, v, { isLatest: true }));
+
+            const response = {
+                versionList,
+                IsTruncated,
+            };
+            if (IsTruncated) {
+                response.versionMarker = NextVersionIdMarker;
```
bennettbuchanan: Minor: these can just be added to the `response` object since they'll be `undefined` anyway if it isn't truncated, and you check for that later.

lib/queuePopulator/IngestionProducer.js
```diff
@@ -185,95 +256,142 @@ class IngestionProducer {
...
-            if (bucketInfo === constants.usersBucket ||
-            bucketInfo === constants.metastoreBucket) {
-                return cb();
+        const { versionMarker, keyMarker } = state;
+        const params = {
+            Bucket: bucket,
+        };
+        // if previous state, should paginate here
+        if (versionMarker && keyMarker) {
+            params.VersionIdMarker = versionMarker;
```
bennettbuchanan, line 279: What happens if the version had been deleted or the objects in the bucket were modified. Is it okay to not have the listing continue from where it had left off before?

philipyoo: For modifications, should be fine. This initial listing/snapshot phase is a best attempt at capturing objects quickly. Then the md logs should pick up the rest of the objects accurately. Duplicating entries is also fine since we plug right into mongo and the md will be the same. Basically, write over the existing data in mongo, regardless of if the data is identical.

Good point on deleted version. I just checked and it seems if the marker is deleted (not a delete marker), the marker will still work and list from the expected spot. I'm guessing the version id's are in order. If the version id marker is provided but doesn't exist, it will know to just list by whichever is next in the listing

lib/queuePopulator/IngestionProducer.js
philipyoo: I like this idea 👍 I'll look into a rf

lib/queuePopulator/IngestionProducer.js
bennettbuchanan: For duplicates, if you overwrite the metadata of an object that had been replicated by Zenko, would you end up putting the object twice? Does backbeat log processor handle that scenario by filtering it?

philipyoo: ah, good call, the log processor won't handle this scenario. There is a plan to add a `headObject` check on processor side (for the PR applying replication info on each object md before saving to mongo).

I'm thinking, we will need to get object md from mongo instead of doing the `headObject` call, and seeing if replication info has already been set. If so, keep the status. But i'll need to update other md only fields in-case this is an md update from s3c.

bennettbuchanan: I agree there would need to be some logic around overwriting any existing metadata. 

I've been working intermittently on a metadata locking mechanism that might be applicable for this purpose. It allows you to update metadata only on certain conditions and handles race conditions when multiple clients are attempting concurrent writes to the same metadata.

philipyoo: nice! that sounds useful for sure

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #604, opened 2019-02-05
ft: ZENKO-1420 ActionQueueEntry model for background jobs

jonathan-gramain:
Provide a helper class to represent an action message that can be
passed along to backbeat services through Kafka queues to execute
specific actions. It can also represent the outcome of the actions
(status and potential results) returned by the service in another
queue.

It's meant to be generic to any kind of background job executed
through a service, but provides a structure that should eventually
help with instrumentation (tracing notably) and logging, while
improving consistency of our internal message passing.

The first user will be the "data mover" as a service called for
transition policies, where we'll be able to request a data copy to a
new location, and get back a result message with the access details to
the new data location, that the transition policies code can use to
update object metadata afterwards.

### DIFF
--- a/lib/models/ActionQueueEntry.js
+++ b/lib/models/ActionQueueEntry.js
@@ -1,0 +1,393 @@
+const uuid = require('uuid/v4');
+
+/**
+ * @class
+ * @classdesc Generic convenience class used to ask a remote service
+ * to execute an action through message-passing, with the possibility
+ * to retrieve the outcome in a separate message from a given topic.
+ *
+ * Here is an outline of the different workflows:
+ *
+ * When creating a new action object to query a service:
+ *
+ * - call create() to create the action object
+ * - then use the various setters to set the required info
+ * - optionally call setResultsTopic() in order to be notified of the
+ *   outcome on this Kafka topic (if the service supports it)
+ * - then serialize with toKafkaMessage()
+ * - then send this serialized message to the service Kafka topic
+ *
+ * When receiving an action message from a Kafka queue to execute an
+ * action:
+ *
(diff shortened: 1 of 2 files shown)


lib/models/ActionQueueEntry.js
```diff
@@ -0,0 +1,393 @@
...
+
+    /**
+     * Add one or more context attributes from a JS object, that will
+     * be stored in the "contextInfo" field of the action message and
+     * logged as part of getLogInfo().
+     *
+     * @param {object} contextObj - attributes to add to the action context
+     * @return {ActionQueueEntry} - this
+     */
+    addContext(contextObj) {
```
philipyoo, line 195: just curious, is this being added for better logging? Will it serve purposes other than logging?

philipyoo (approved):
nice!

lib/models/ActionQueueEntry.js
jonathan-gramain: Yes it will not only help for logging, but also to feed back some context to the action requester in the results topic messages. E.g., if transition policies request a data copy, in the response, the context can carry the necessary info about the original action, e.g. that it was a transition, or the original storage class for the purpose of conditional checks when updating the metadata. The idea is carrying state over to the next actions so that it's readily available when taking steps after the action is done.

bennettbuchanan (approved):

jonathan-gramain:
@bert-e approve

bennettbuchanan (approved):

rahulreddy (approved):

jonathan-gramain:
@bert-e approve

PR #605, opened 2019-02-05
bf: ZENKO-1464 ingestion log stream sync loop fix

philipyoo:
Avoid manipulating original record entries so our other entries in sync loop are not impacted

log stream has a sync loop that looks at each entry within a log record. These entries are (for a versioned bucket) one entry for the entry key without a version suffix (to represent IsLatest) and one entry for the entry key with version suffix.

### DIFF
--- a/lib/queuePopulator/IngestionReader.js
+++ b/lib/queuePopulator/IngestionReader.js
@@ -140,22 +140,30 @@ class IngestionReader extends LogReader {
                 value: entry.value,
             }));
         } else {
+            let key;
+            let db;
             if (record.db === 'users..bucket') {
                 const keySplit = entry.key.split('..|..');
-                entry.key = `${keySplit[0]}..|..${this._targetZenkoBucket}`;
+                key = `${keySplit[0]}..|..${this._targetZenkoBucket}`;
             } else if (record.db === 'metastore') {
                 const keySplit = entry.key.split('/');
-                entry.key = `${keySplit[0]}/${this._targetZenkoBucket}`;
+                key = `${keySplit[0]}/${this._targetZenkoBucket}`;
             } else {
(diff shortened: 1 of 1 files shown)


JianqinWang (approved):

lib/queuePopulator/IngestionReader.js
```diff
@@ -140,22 +140,24 @@ class IngestionReader extends LogReader {
...
                 }
-                record.db = this._targetZenkoBucket;
+                db = this._targetZenkoBucket;
             }
             this._extensions.forEach(ext => ext.filter({
                 type: entry.type,
-                bucket: record.db,
-                key: entry.key,
+                bucket: db || record.db,
+                key: key || entry.key,
```
jonathan-gramain: I would rather check `key !== undefined` to avoid issues with potential defined but falsy keys.

jonathan-gramain (commented):
I'm not familiar with the logic of this code, just keep in mind that to my experience, a record would not necessarily have two entries containing the (updated) master and the versioned key. I think it depends on how write gathering decides to group updates together in a single write, so potentially may have more than two entries, for different keys, or maybe even just one entry. I may be wrong but if it has implications on the design, just double check this.

philipyoo:
@jonathan-gramain Ah, ty for this info. It should be fine in this PR, but I'll need to make sure with other PR's that I'm not making this assumption.
I mentioned the observation in this PR because (at least with recent manual tests), I was observing behavior where log entries contained the new master and the latest versioned together, and we were overwriting one or the other, so only one of the entries was ingested

alexanderchan-scality (approved):

jonathan-gramain (approved):

philipyoo:
@bert-e approve

PR #607, opened 2019-02-06
ZENKO-1420 select topic per message in BackbeatProducer

jonathan-gramain:
Add the ability to select the topic to produce to for each call to
BackbeatProducer.send(), allowing the use of multiple or dynamic
target topics for one producer instance.

The intent is to ask a service to send the outcome of an action
execution to the topic specified by the user in the action message
itself, requiring a dynamic selection of the topic. This is not an
issue as Kafka producer supports this naturally in its API so we just
adapt the BackbeatProducer wrapper.

### DIFF
--- a/lib/BackbeatProducer.js
+++ b/lib/BackbeatProducer.js
@@ -31,7 +31,8 @@ class BackbeatProducer extends EventEmitter {
     /**
     * constructor
     * @param {Object} config - config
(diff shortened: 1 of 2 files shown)


bennettbuchanan (approved):

philipyoo (approved):

jonathan-gramain:
@bert-e approve

PR #611, opened 2019-02-08
cleanup: ZENKO-1420 more generic BackbeatMetadataProxy.putMetadata()

jonathan-gramain:
Make the putMetadata() call more generic by taking separate params
like bucket name etc. and a metadata blob, instead of using an
ObjectQueueEntry directly. This will be useful to update metadata
based on action message fields that will not be converted into an
ObjectQueueEntry.

For consistency, also renamed "encodedVersionId" to "versionId" in
getMetadata(), as the encoded form is the common form used in AWS API
the name can be kept short.

### DIFF
--- a/lib/BackbeatMetadataProxy.js
+++ b/lib/BackbeatMetadataProxy.js
@@ -40,27 +40,43 @@ class BackbeatMetadataProxy extends BackbeatTask {
         return new RoleCredentials(vaultclient, extension, role, log);
     }
 
-    putMetadata(entry, log, cb) {
+    /**
+     * Write raw object metadata blob in JSON to MongoDB
+     *
+     * @param {object} params - params object
+     * @param {string} params.bucket - bucket name
+     * @param {string} params.objectKey - object key
+     * @param {string} [params.versionId] - encoded version ID
+     * @param {Buffer} params.mdBlob - raw metadata blob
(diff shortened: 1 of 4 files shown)


philipyoo (approved):

lib/BackbeatMetadataProxy.js
```diff
@@ -40,27 +40,30 @@ class BackbeatMetadataProxy extends BackbeatTask {
         return new RoleCredentials(vaultclient, extension, role, log);
     }
 
-    putMetadata(entry, log, cb) {
+    putMetadata(params, mdBlob, log, cb) {
```
bennettbuchanan: Seems we could make the interface a bit simpler by making `mdBlob` one of the properties of `params`. But that could end up making the logging a bit more complex. Just thinking out loud about trade-offs.

bennettbuchanan (approved):

lib/BackbeatMetadataProxy.js
jonathan-gramain: I agree, it could make the API slightly simpler, for logging it's a minor concern and passing the fields to log explicitly is fine as there are only three of them.

bennettbuchanan (approved):
:shipit:

jonathan-gramain:
@bert-e approve

PR #612, opened 2019-02-12
bugfix: ZENKO-1470 Attempt to abort MPU on failure

bennettbuchanan:


bennettbuchanan:
@bert-e reset

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -132,6 +132,35 @@ class MultipleBackendTask extends ReplicateObject {
...
+                log.error('an error occurred when aborting a MPU', {
+                    method: 'MultipleBackendTask._handleMPUFailure',
+                    entry: sourceEntry.getLogInfo(),
+                    origin: 'target',
+                    peer: this.destBackbeatHost,
+                    error: err.message,
+                });
+                return cb(err);
+            }
+            log.info('successfully aborted a MPU', {
```
jonathan-gramain: very minor but easy change: when an operator sees "successfully" he may think the operation was a success, I would rather simply state "aborted MPU" so it's still clear something was wrong at the beginning.

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -132,6 +132,35 @@ class MultipleBackendTask extends ReplicateObject {
...
+        this._multipleBackendAbortMPU(sourceEntry, uploadId, log, err => {
+            if (err) {
+                log.error('an error occurred when aborting a MPU', {
+                    method: 'MultipleBackendTask._handleMPUFailure',
+                    entry: sourceEntry.getLogInfo(),
+                    origin: 'target',
+                    peer: this.destBackbeatHost,
+                    error: err.message,
+                });
+                return cb(err);
```
jonathan-gramain: You could do `return cb()` since the error is not meant to be reused.

jonathan-gramain (approved):

extensions/replication/tasks/MultipleBackendTask.js
bennettbuchanan: Good point—"success" is probably an overstatement of what had happened here.

extensions/replication/tasks/MultipleBackendTask.js
bennettbuchanan: I decided to pass it here even though I'm not using it because maybe someday, someone, somewhere, will expect an error from this method (as is the common callback pattern). I figured it best to allow the caller of the method to decide whether or not to use the error.

philipyoo (approved):

rahulreddy (approved):

bennettbuchanan:
@bert-e approve

PR #614, opened 2019-02-12
refactor: ZENKO-1420 only one producer in bucket processor

jonathan-gramain:
Since we're going to publish to a third topic and we now have a way to
use a single producer to publish to multiple topics, convert the code
to only use one producer instance. Rework a bit some helpers to send
messages so that they are defined in the LifecycleTask class directly.

### DIFF
--- a/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
+++ b/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
@@ -67,9 +67,7 @@ class LifecycleBucketProcessor {
         this._repConfig = extensions.replication;
         this._s3Endpoint = `${transport}://${s3Config.host}:${s3Config.port}`;
         this._transport = transport;
-        this._bucketProducer = null;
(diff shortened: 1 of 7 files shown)


philipyoo (approved):

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -432,61 +396,17 @@ class LifecycleBucketProcessor {
...
-                            error: err.message,
-                            method: 'LifecycleBucketProcessor.start',
-                        });
-                        return next(err);
-                    }
-                    this._transitionProducer = producer;
-                    return next();
-                }),
-        ], err => {
+        this._setupProducer((err, producer) => {
```
bennettbuchanan, line 399: ❤️ this diff!

bennettbuchanan (approved):

jonathan-gramain:
@bert-e approve

PR #622, opened 2019-02-22
bugfix: ZENKO-1522 fix CRR on copied MPU object

jonathan-gramain:
Change the logic to ignore the 'MPU' flag set in replicationInfo.content array (it may not always be set), instead rely on the object size and if a MPU or not based on its content-md5 metadata attribute containing a dash.

Note: the copy code lacks integrity checking now, this would need a
rework but it's out of scope of this fix.

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -3,7 +3,6 @@ const uuid = require('uuid/v4');
 
 const errors = require('arsenal').errors;
 const jsutil = require('arsenal').jsutil;
-const ObjectMDLocation = require('arsenal').models.ObjectMDLocation;
 const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
 
 const ReplicateObject = require('./ReplicateObject');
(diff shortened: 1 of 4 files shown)


jonathan-gramain:
@bert-e reset

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -825,7 +799,26 @@ class MultipleBackendTask extends ReplicateObject {
                     return this._deleteObjectTagging(sourceEntry, destEntry,
                         log, next);
                 }
-                return this._getAndPutData(sourceEntry, destEntry, log, next);
+                // Do a multipart upload when either the size is above
+                // a threshold or the source object is itself a MPU.
```
bennettbuchanan, line 803: What is the rationale for the 20MB threshold?

bennettbuchanan (approved):

extensions/replication/tasks/MultipleBackendTask.js
jonathan-gramain: It's a heuristic, if the object is "big enough" we want to MPU it to benefit from more parallelism (and later maybe better recovery on error by keeping the valid parts around), if the object is too small it's overkill to do a MPU (more requests and metadata overhead). So we need some boundary to decide between the two, 20MB makes a MPU of 4x5MB parts for example, which sounds like a reasonable size to start thinking about MPU.

Of course the size threshold can be debated, it just looks a reasonable tradeoff to me.

jonathan-gramain:
There is a new test in E2E as well: https://github.com/scality/Zenko/compare/bugfix/ZENKO-1522-copyObjectMPUCRRTest

rahulreddy (approved):

rahulreddy (commented):
Make sure you commit the package.lock changes as well.

ploki (approved):

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@bert-e approve

jonathan-gramain:
@bert-e reset

PR #625, opened 2019-02-26
ft: ZENKO-1531 add replication info from Zenko

philipyoo:
Zenko bucket replication info should be applied to newly ingested object metadata entries so that replication works on Zenko for an out-of-band RING bucket.

This is part 1 of the updates. This will allow for crr to work on new object versions only. MD updates will come in a later PR.

Changes in this PR:
- Get the Zenko bucket replication configuration info and apply to object metadata replication info
- Match each storage class to their relevant storage type. These are mapped in the Config bootstrapList object
- Replication content type field based strictly on data || no data. This will be revised for md-only crr updates in a later pr

To be handled in a follow-up PR:
- updates for md-only PUT_TAGGING, DELETE_TAGGING
- updates for mpu entries MPU

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -6,9 +6,13 @@ const http = require('http');
 
 const Logger = require('werelogs').Logger;
 const errors = require('arsenal').errors;
+const { replicationBackends } = require('arsenal').constants;
 const MongoClient = require('arsenal').storage
     .metadata.mongoclient.MongoClientInterface;
+const ObjectMD = require('arsenal').models.ObjectMD;
+const { isMasterKey } = require('arsenal/lib/versioning/Version');
 
+const Config = require('../../conf/Config');
 const BackbeatConsumer = require('../../lib/BackbeatConsumer');
 const QueueEntry = require('../../lib/models/QueueEntry');
 const DeleteOpQueueEntry = require('../../lib/models/DeleteOpQueueEntry');
@@ -71,6 +75,7 @@ class MongoQueueProcessor {
         this._s3Endpoint = `http://${s3Config.host}:${s3Config.port}`;
         this._s3Client = null;
         this._consumer = null;
+        this._bootstrapList = null;
         this.logger =
             new Logger(`Backbeat:Ingestion:MongoProcessor:${this.site}`);
         this.mongoClientConfig.logger = this.logger;
@@ -117,6 +122,12 @@ class MongoQueueProcessor {
                 this.logger.error('could not connect to MongoDB', { err });
                 process.exit(1);
             }
+
+            this._bootstrapList = Config.getBootstrapList();
+            Config.on('bootstrap-list-update', () => {
+                this._bootstrapList = Config.getBootstrapList();
+            });
+
             let consumerReady = false;
             this._consumer = new BackbeatConsumer({
                 topic: this.mongoProcessorConfig.topic,
@@ -149,6 +160,37 @@ class MongoQueueProcessor {
             return setImmediate(done);
         }
         return this._consumer.close(done);
+    }
+
+    _getDataContent(entry) {
+        const contentLength = entry.getContentLength();
+        if (contentLength > 0) {
+            return ['DATA', 'METADATA'];
+        } else {
+            return ['METADATA'];
+        }
+    }
+
+    _getZenkoObjectMetadata(entry, done) {
+        const bucket = entry.getBucket();
+        const key = entry.getObjectKey();
+        const params = { versionId: entry.getVersionId() };
+
+        this._mongoClient.getObject(bucket, key, params, this.logger,
+        (err, data) => {
+            if (err && err.NoSuchKey) {
+                return done();
+            }
+            if (err) {
+                this.logger.error('error getting zenko object metadata', {
+                    method: 'MongoQueueProcessor._getZenkoObjectMetadata',
+                    error: err,
+                    entry: entry.getLogInfo(),
+                });
+                return done(err);
+            }
+            return done(null, data);
+        });
     }
 
     /**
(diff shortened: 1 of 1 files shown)


JianqinWang (approved):

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -196,7 +204,20 @@ class MongoQueueProcessor {
...
-    _updateReplicationInfo(entry) {
+    /**
+     * Update replication info on ingested object MD to match Zenko defined
+     * replication info.
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @param {BucketInfo} bucketInfo - bucket info object
+     * @return {undefined}
+     */
+    _updateReplicationInfo(entry, bucketInfo) {
+        const bucketRepInfo = bucketInfo.getReplicationConfiguration();
```
bennettbuchanan: Could be worthwhile to abstract and reuse the logic in https://github.com/scality/cloudserver/blob/development/8.1/lib/api/apiUtils/object/getReplicationInfo.js#L95.

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -210,9 +231,39 @@ class MongoQueueProcessor {
             isNFS: null,
         };
 
+        if (role && destination && rules && rules.length > 0) {
```
bennettbuchanan: Could be worthwhile to abstract and reuse the logic [here](https://github.com/scality/cloudserver/blob/development/8.1/lib/api/apiUtils/object/getReplicationInfo.js#L95). I think that logic is a bit tangled, so it could a great opportunity for a nice refactor should we feel inspired 😄.

For example, first we'd need to check that the object is applicable to the configuration based on the key prefix.

philipyoo: In the interest of time, I'll be duplicating some code here rather than reusing the cloudserver code. I'll follow up to refactor the code

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -211,21 +288,47 @@ class MongoQueueProcessor {
             isNFS: null,
         };
 
-        entry.setReplicationInfo(replicationInfo);
-    }
+        // TODO: refactor based off cloudserver getReplicationInfo
+        if (bucketRepInfo && !_isServiceAccount(entry.getOwnerDisplayName())) {
```
rahulreddy: What's the relevance of the `_isServiceAccount` check?

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -269,30 +372,51 @@ class MongoQueueProcessor {
...
-            });
+            // update necessary metadata fields before saving to Zenko MongoDB
+            this._updateOwnerMD(sourceEntry, bucketInfo);
+            this._updateObjectDataStoreName(sourceEntry, location);
+            this._updateLocations(sourceEntry, location);
+            this._updateAcl(sourceEntry);
+            this._updateReplicationInfo(sourceEntry, bucketInfo, content);
+
+            const objVal = sourceEntry.getValue();
+            // Always call putObject with version params undefined so
```
rahulreddy, line 382: 💯

extensions/mongoProcessor/MongoQueueProcessor.js
philipyoo: None I think haha, I'll remove this. I wasn't sure if service accounts were used in s3c, but based off the branches, doesn't look like it

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -151,6 +160,41 @@ class MongoQueueProcessor {
         return this._consumer.close(done);
     }
 
+    _isMasterKey(entry) {
+        return entry.getObjectVersionedKey() === entry.getObjectKey();
```
bennettbuchanan: Looks like this comparison is already exposed by `ObjectQueueEntry` as the method `isVersion`. We could avoid some duplicate code by using that. https://github.com/scality/backbeat/blob/93980e8130a0a7c25d9f0ec52234e6b5d46e49ab/lib/models/ObjectQueueEntry.js#L77

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -151,6 +160,41 @@ class MongoQueueProcessor {
...
+        (err, data) => {
+            if (err && !err.NoSuchKey) {
+                this.logger.error('error getting zenko object metadata', {
+                    method: 'MongoQueueProcessor._getZenkoObjectMetadata',
+                    error: err,
+                    entry: entry.getLogInfo(),
+                });
+                return done(err);
+            }
+            if (err && err.NoSuchKey) {
```
bennettbuchanan: Minor suggestion: maybe reads a bit cleaner if this is the first error condition—that way you can simplify the above condition as `if (err)`.

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -196,7 +240,36 @@ class MongoQueueProcessor {
...
     }
 
-    _updateReplicationInfo(entry) {
+    /**
+     * Update acl info on ingested object MD to match Zenko defined bucket acl.
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @return {undefined}
+     */
+    _updateAcl(entry) {
+        // TODO: Rely on ObjectMD to set this
```
bennettbuchanan: Thinking out loud here: seems like it might be quicker to instantiate `ObjectMD` and then call `getAcl`. That way we wouldn't have to revisit this.

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -211,21 +284,47 @@ class MongoQueueProcessor {
             isNFS: null,
         };
 
-        entry.setReplicationInfo(replicationInfo);
-    }
+        // TODO: refactor based off cloudserver getReplicationInfo
+        if (bucketRepInfo) {
+            const { role, destination, rules } = bucketRepInfo;
+            const rule = rules.find(r =>
+                (entry.getObjectKey().startsWith(r.prefix) && r.enabled));
```
bennettbuchanan, line 271: 👍

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -211,21 +284,47 @@ class MongoQueueProcessor {
             isNFS: null,
         };
 
-        entry.setReplicationInfo(replicationInfo);
-    }
+        // TODO: refactor based off cloudserver getReplicationInfo
```
bennettbuchanan, line 267: Would it be possible to extract this into an Arsenal util so that a single place can manage this logic? Seeing this closely resemble [_getReplicationInfo](https://github.com/scality/cloudserver/blob/development/8.1/lib/api/apiUtils/object/getReplicationInfo.js#L33). Is that what's meant by the comment?

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -269,30 +368,52 @@ class MongoQueueProcessor {
...
+            // is sufficient to replay the version logic in the consumer.
+            return this._mongoClient.putObject(bucket, key, objVal, undefined,
+                this.logger, err => {
+                    if (err) {
+                        this.logger.error('error putting object metadata ' +
+                        'to mongo', { error: err });
+                        return done(err);
+                    }
+                    this.logger.info('object metadata put to mongo',
+                    { key });
```
bennettbuchanan: Maybe valuable also to log the bucket name because there could be multiple of the same key names in MD across buckets.

bennettbuchanan (approved):
Overall LGTM! :ship: Left some considerations if there's time.

extensions/mongoProcessor/MongoQueueProcessor.js
rahulreddy: Nope, service accounts are only used in Zenko.

extensions/mongoProcessor/MongoQueueProcessor.js
philipyoo: Yup, the todo was added to revisit this. I wasn't sure how to go about the next steps for setting content, so I decided to just borrow the logic from cloudserver.
But I think it'll be ok, will work on refactoring

Edit: Sorry, I'll add the refactor in separate PRs. For now, I'd like to get this feature in

extensions/mongoProcessor/MongoQueueProcessor.js
philipyoo: For this, I found a method in arsenal for checking master key. I'll use that because I think it more accurately performs the check I need and also works for readability

extensions/mongoProcessor/MongoQueueProcessor.js
philipyoo: great idea, i've went ahead and made this change

philipyoo:
@bert-e approve

PR #626, opened 2019-02-26
feature: ZENKO-1529 Add transition to manager

bennettbuchanan:


lib/management/applyWorkflowUpdates.js
```diff
@@ -1,5 +1,23 @@
 const async = require('async');
 
+function getNewBucketWorkflows(serviceName, workflows, bucketName) {
+    const newBucketWorkflows =
+        (workflows &&
+        workflows[serviceName] &&
+        workflows[serviceName][bucketName]) || [];
```
ploki, line 7: I find this construct confusing but very handy. Is it a widely used js idiom?

extensions/lifecycle/management.js
```diff
@@ -37,9 +37,17 @@ function putLifecycleConfiguration(bucketName, workflows, cb) {
...
-                    };
+                    if (wf.type.includes('expiration')) {
+                        workflow.Expiration = {
+                            Days: wf.currentVersionTriggerDelayDays,
+                        };
+                    }
+                    if (wf.type.includes('transition')) {
+                        workflow.Transitions = [{
+                            Days: wf.currentVersionTriggerDelayDays,
+                            StorageClass: wf.currentVersionLocations[0].name,
```
jonathan-gramain, line 48: Can we add a sanity check to ensure `wf.currentVersionLocations` array has exactly one element and return an error in the callback if not the case? Unless Orbit already restricts it to one item.

lib/management/applyWorkflowUpdates.js
jonathan-gramain: I think it's widely used indeed, as an object or array are always truish in JS, and the `&&` operator is lazy it will stop at the first falsish value (or the last one) and return it. The `||` is the opposite, stop at the first truish value (or the last one) and return it.

lib/management/index.js
```diff
@@ -164,28 +164,42 @@ function _saveServiceState(newState, cb) {
                                 JSON.stringify(newState), {}, logger, cb);
 }
 
+function _applyWorkflowUpdate(conf, currentState, params, workflow, cb) {
```
jonathan-gramain: Sorry if I'm being nitpicky here, I think this helper is confusing, since it has almost the same name than another function that it calls but does more things. Why not simply call `getWorkflowUpdates` from the main `async.each` loop before calling `applyWorkflowUpdates`?

bennettbuchanan: I agree with that—will update.

lib/management/applyWorkflowUpdates.js
bennettbuchanan: ❤️JS

extensions/lifecycle/management.js
bennettbuchanan: In Orbit there doesn't seem to be a way to add multiple transitions, so had assumed there's a limit. That's a bit suspect as an assumption though, considering it's an array. This was actually an open question for me.

lib/management/applyWorkflowUpdates.js
ploki: Got it. Not only lazy, what was surprising to my eyes was the train of conjunctions and its variability in the returned type

lib/management/index.js
bennettbuchanan: Ok, I've renamed the method called from `_refreshServiceState` to `_applyServiceState`. `_applyServiceState` will abstract the process of getting the workflow and applying them. I felt it helps with readability since the nesting was getting deep, having added the `async.each`.

extensions/lifecycle/management.js
bennettbuchanan: Maybe we want to use a `map` so that Orbit could eventually send multiple transitions if we decide to support it from the front-end.

bennettbuchanan: There's a validation that it must be 1 element in the API schema. See: https://github.com/scality/pensieve-api/blob/deploy/swagger.yml#L1963 (thanks @alexanderchan-scality for pointing to that schema definition).

bennettbuchanan: For transitioning to multiple destinations with the same prefix, a user would have to use a lifecycle config like:

```
{
    "Rules": [
        {
            "Filter": {
                "Prefix": "foo"
            },
            "Status": "Enabled",
            "Transitions": [
                {
                    "Days": 30,
                    "StorageClass": "aws-backend"
                }
            ],
            "ID": "59af0a70-3ad4-11e9-a6af-3e11a5f127c7"
        },
        {
            "Filter": {
                "Prefix": "foo"
            },
            "Status": "Enabled",
            "Transitions": [
                {
                    "Days": 60,
                    "StorageClass": "azure-backend"
                }
            ],
            "ID": "5f058a8d-3ad4-11e9-a6b9-aeb9ed4edffc"
        }
    ]
}
```

Instead of:

```
{
    "Rules": [
        {
            "Filter": {
                "Prefix": "foo"
            },
            "Status": "Enabled",
            "Transitions": [
                {
                    "Days": 30,
                    "StorageClass": "aws-backend"
                },
                {
                    "Days": 60,
                    "StorageClass": "azure-backend"
                }
            ],
            "ID": "5f058a8d-3ad4-11e9-a6b9-aeb9ed4edffc"
        }
    ]
}
```

jonathan-gramain (approved):

lib/management/applyWorkflowUpdates.js
```diff
@@ -1,5 +1,23 @@
...
+    // The lifecycle service manager handles both the expiration workflow and
+    // the transition workflow.
+    if (serviceName === 'lifecycle') {
+        const transitionWorkflows =
+            (workflows &&
+            workflows.transition &&
+            workflows.transition[bucketName]) || [];
+        transitionWorkflows.forEach(workflow =>
```
nicolas2bert: using `concat()` might be cleaner ?

lib/management/applyWorkflowUpdates.js
```diff
@@ -1,5 +1,23 @@
 const async = require('async');
 
+function getNewBucketWorkflows(serviceName, workflows, bucketName) {
+    const newBucketWorkflows =
+        (workflows &&
+        workflows[serviceName] &&
+        workflows[serviceName][bucketName]) || [];
+    // The lifecycle service manager handles both the expiration workflow and
```
nicolas2bert, line 8: Yes, it can be confusing but we use `lifecycle` for expiration only in Orbit. We might want to add a note about it :)

nicolas2bert (approved):

lib/management/applyWorkflowUpdates.js
bennettbuchanan: Nice tip. Thanks!

bennettbuchanan: Updated.

bennettbuchanan:
@bert-e approve

PR #635, opened 2019-03-08
feature: ZENKO-1420 data mover service

jonathan-gramain:
Create a new task type inside the queue processor to copy data between
locations.

It relies on ActionQueueEntry class to marshall/unmarshall requests
and responses to/from the backbeat-data-mover topic.

The current version does not allow choosing the key of the new object,
as it is chosen by the multiple backend code that executes the copy,
but it fits the need for transition policies.

**Note for reviews**: it may look daunting, but a large part of the logic of CopyLocationTask is derived from MultipleBackendTask, with argument passing changed to use action objects and return action results. I though it was hard and risky enough to factorize the code that I preferred to split it into different tasks instead, in a future version we could get rid of MultipleBackendTask in favor of CopyLocationTask (copy as a service approach).

Functionally depends on https://github.com/scality/cloudserver/pull/1719.

### DIFF
--- a/docs/data-mover.md
+++ b/docs/data-mover.md
@@ -1,0 +1,78 @@
+# Data Mover Service
+
+## OVERVIEW
+
+The data mover service can copy object data as a service.
+
+It consumes action messages from a kafka topic ("backbeat-data-mover"
+by default), and optionally returns the action outcome in another
+kafka topic, which name can be set in the action message.
+
+Each action message type defines the type of copy we want, described
+below. The recommended way to pass action messages and receive results
+is to use the lib/models/ActionQueueEntry class to create instances of
+actions that can be manipulated, but it's not mandatory as messages
+can be created and parsed directly to/from JSON format.
+
+Attributes in parenthesis are optional.
+
+## COPY LOCATION ACTION
+
+The "copy location" action copies a zenko object to a new location.
+
+### Action message format
+
+This is the format of "copy location" actions that the data mover
+expects:
+
+```
+{
+    "action": "copyLocation",
+    ("actionId": "uuid-xyz",)
+    "target": {
+        "bucket": "bucketName",
+        "key": "objectKey",
+        "contentMd5": "0123456789abcdef0123456789abcdef",
+        ("version": "versionNumber"),
+    },
+    "toLocation": "zenkoLocationName",
+    ("resultsTopic": "results-kafka-topic-name"),
+    ("context": { /* user-defined context fields */ },)
+}
+```
+
+### Action result format
+
+This is the format of "copy location" responses that the data mover
+will send to the results topic if set, with all original action
+message fields preserved:
+
+```
+{
+    "action": "copyLocation",
+    ("actionId": "uuid-xyz",)
+    "target": {
+        "bucket": "bucketName",
+        "key": "objectKey",
+        ("version": "versionNumber")
+    },
+    "toLocation": "zenkoLocationName",
+    ("resultsTopic": "results-kafka-topic-name",)
+    ("context": { /* user-defined context fields */ },)
+
+    "status": "success|error",
+    // if status is "success":
+    ("results": {
+        "location": [{
+            // location-specific info (key, dataStoreVersionId etc.)
+        }]
+    }),
+    // if status is "error":
+    ("error": {
+        "code": xxx // original error.code, usually HTTP error code
+        "message": "ErrorType" // original error.message, usually
+                               // arsenal error type
+        "description": "lengthy description", // original error.description
+    })
+}
+```
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -21,10 +21,12 @@ const getLocationsFromStorageClass =
     require('../utils/getLocationsFromStorageClass');
 const ReplicateObject = require('../tasks/ReplicateObject');
 const MultipleBackendTask = require('../tasks/MultipleBackendTask');
+const CopyLocationTask = require('../tasks/CopyLocationTask');
 const EchoBucket = require('../tasks/EchoBucket');
 
 const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
 const BucketQueueEntry = require('../../../lib/models/BucketQueueEntry');
+const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
 const MetricsProducer = require('../../../lib/MetricsProducer');
 
 const {
@@ -106,6 +108,7 @@ class QueueProcessor extends EventEmitter {
         this.destAdminVaultConfigured = false;
         this.replicationStatusProducer = null;
         this._consumer = null;
+        this._dataMoverConsumer = null;
         this._mProducer = null;
         this.site = site;
         this.mConfig = mConfig;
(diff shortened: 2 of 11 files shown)


docs/data-mover.md
```diff
@@ -0,0 +1,78 @@
...
+    ("actionId": "uuid-xyz",)
+    "target": {
+        "bucket": "bucketName",
+        "key": "objectKey",
+        "contentMd5": "0123456789abcdef0123456789abcdef",
+        ("version": "versionNumber"),
+    },
+    "toLocation": "zenkoLocationName",
+    ("resultsTopic": "results-kafka-topic-name"),
+    ("context": { /* user-defined context fields */ },)
```
bennettbuchanan, line 40: I like the versatility of this design 🥇

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -676,18 +714,67 @@ class QueueProcessor extends EventEmitter {
...
+     */
+    processDataMoverEntry(kafkaEntry, done) {
+        const actionEntry = ActionQueueEntry.createFromKafkaEntry(kafkaEntry);
+        if (actionEntry.error) {
+            this.logger.error('error processing source entry',
+                              { error: actionEntry.error,
+                                entry: kafkaEntry.value });
+            return process.nextTick(() => done(errors.InternalError));
+        }
+        if (actionEntry.skip) {
```
bennettbuchanan, line 745: Could be worthwhile to add this to the doc as is done for the `error` field. e.g. https://github.com/scality/backbeat/pull/635/files#diff-2a67cabc92f6a2562cc11d59e1ad64f0R70

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -0,0 +1,794 @@
...
+    };
+    return extMetrics;
+}
+
+class CopyLocationTask extends BackbeatTask {
+
+    _getReplicationEndpointType() {
+        const replicationEndpoint = this.destConfig.bootstrapList
+            .find(endpoint => endpoint.site === this.site);
+        return replicationEndpoint.type;
```
bennettbuchanan, line 37: Is there any scenario where this could be called but the config does not have `this.site` as a site? For example, a race condition when an update to the config is made.

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -0,0 +1,794 @@
...
+        actionEntry.addLoggedAttributes({
+            bucketName: 'target.bucket',
+            objectKey: 'target.key',
+            versionId: 'target.version',
+            copyToLocation: 'toLocation',
+        });
+        log.debug('action execution starts', actionEntry.getLogInfo());
+        return async.waterfall([
+            next => {
+                if (!this._setupClients(actionEntry, log)) {
```
bennettbuchanan: Minor suggestion: move the `_validateActionCredentials` conditional check here (instead of in `_setupClients`) and return the error. Then `_setupClients` can be responsible only for a single are of concern, i.e. setting up the clients.

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -0,0 +1,794 @@
...
+        ], err => this._publishCopyLocationStatus(
+            err, actionEntry, kafkaEntry, log, done));
+    }
+
+    _getSourceMD(actionEntry, log, cb) {
+        const params = {
+            bucket: actionEntry.getAttribute('target.bucket'),
+            objectKey: actionEntry.getAttribute('target.key'),
+            versionId: actionEntry.getAttribute('target.version'),
+        };
```
bennettbuchanan, line 159: Minor: it's arguably insignificant, but it would save some run-time by doing something like:

```
const { bucket, key, version } = actionEntry.getAttribute('target');
```

That way we don't have to loop through the message object more than once.

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -0,0 +1,794 @@
...
+    }
+
+    _getAndPutObject(actionEntry, objMD, log, cb) {
+        const partLogger = this.logger.newRequestLogger(log.getUids());
+        const extMetrics = _getExtMetrics(this.site,
+            objMD.getContentLength(), actionEntry);
+        this.mProducer.publishMetrics(extMetrics,
+            metricsTypeQueued, metricsExtension, () => {});
+        this.retry({
+            actionDesc: 'stream part data',
```
bennettbuchanan: ~stream part data~ stream object data

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -0,0 +1,794 @@
...
+                        method: 'CopyLocationTask._getAndPutObjectOnce',
+                        peer: this.sourceConfig.s3,
+                        error: err.message,
+                        httpStatus: err.statusCode,
+                    }, actionEntry.getLogInfo()));
+                return doneOnce(err);
+            });
+            incomingMsg = sourceReq.createReadStream();
+            incomingMsg.on('error', err => {
+                if (err.statusCode === 404) {
```
bennettbuchanan, line 234: Just curious here: does this error occur if the object is deleted mid-stream?

tests/unit/lib/models/ActionQueueEntry.spec.js
```diff
@@ -43,6 +43,25 @@ describe('ActionQueueEntry', () => {
             entry.getAttribute('details'), { roundTrip: true });
     });
 
+    it('should set an error when retrieving an unset attribute and ' +
+    '"required" option is set', () => {
+        const entry = ActionQueueEntry.create('sendPigeon')
+              .setAttribute('toCountry', 'Finland');
```
bennettbuchanan, line 49: 🤔 Why is this pigeon being sent to Finland?

tests/unit/lib/models/ActionQueueEntry.spec.js
```diff
@@ -43,6 +43,25 @@ describe('ActionQueueEntry', () => {
...
+        assert.strictEqual(
+            entry.getAttribute('cost', { required: true }), undefined);
+        assert(entry.getError());
+        assert.strictEqual(entry.getError().code, 400);
+        assert.strictEqual(entry.getError().message, 'MissingParameter');
+
+        // nested attributes should work too
```
bennettbuchanan: Seems like this should be a separate test since it is not checking for an error when a required attribute is unset.

bennettbuchanan (approved):
Great improvement 🔥. Overall, LGTM. Left some considerations.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -544,52 +579,45 @@ class QueueProcessor extends EventEmitter {
...
-                    this.logger.info('queue processor is ready to consume ' +
-                                     'replication entries');
-                    this.emit('ready');
-                });
-                return undefined;
-            });
+                this._consumer = this._createConsumer(
+                    this.repConfig.topic,
+                    this.processReplicationEntry.bind(this), options);
+                return this._consumer.on('canary', done);
```
philipyoo: I think the canary event triggers many times per location. Maybe use `.once` here?

Same for here: https://github.com/scality/backbeat/pull/635/files#diff-edb6155bab9639f454e2f3d93ffa45f4R615

extensions/replication/queueProcessor/QueueProcessor.js
jonathan-gramain: Actually the `error` field above is not the one mentioned in this doc: it's the conventional result from `createFromKafkaEntry()` when there is an error to *parse the kafka entry*. If there's no error, then we get an `ActionQueueEntry` object that may contain an error result retrieved with `getError()`, but not accessible directly with an `actionEntry.error` attribute.

For the "skip" it's used in CRR code to skip canary messages, since I kept the same mechanism (useful for pause/resume) I kept using the same convention here than in the other CRR queue.

extensions/replication/tasks/CopyLocationTask.js
jonathan-gramain: No, this is an invariant from the constructor of QueueProcessor class, and passed via `getStateVars()`, it does not change after it's been initialized. Actually the same is done in `MultipleBackendTask`, and I don't recall we've seen issues with this.

extensions/replication/tasks/CopyLocationTask.js
jonathan-gramain: I don't think so, I would expect some 50x, it likely depends on the backend type too, but it would be a good test to do at least once to know the behavior in such case (although hard to produce and even harder to automate...). But in any case it should lead to a failure that does not trigger a change on the source metadata.

tests/unit/lib/models/ActionQueueEntry.spec.js
jonathan-gramain: Because it is pigeons' favorite country! It's also a good pigeon hub for long haul flights.

tests/unit/lib/models/ActionQueueEntry.spec.js
jonathan-gramain: Ah, interesting, it looks this is just an erroneous cut&paste from the above test, will remove!

philipyoo (approved):

jonathan-gramain:
@bert-e approve

PR #638, opened 2019-03-09
bf: ZENKO-1560 get ingestion locations in initMgmt

philipyoo:
Change where/when we fetch ingestion buckets/locations.
Rather than making the fetch in IngestionPopulator, instead
get the information during the initManagement
patchConfiguration step and update the Config object.

Changed the Config get/set method name since the Config
field is now only specific to ingestion buckets/locations
now.

Changes in this PR:
- Change method name: Config.getLocations -> Config.getIngestionBuckets
- Change method name: Config.setLocations -> Config.setIngestionBuckets
- Move `patchConfiguration` in `lib/management/index.js` to its own file.
- Change `patchConfiguration` to also get ingestion buckets + information and update the Config object (`setIngestionBuckets`)
- `IngestionPopulator.applyUpdates` no longer done by the cron task from populator, but instead by the updates made by `initManagement` to the Config object. Event based.
- Remove unused components from IngestionPopulator


lib/queuePopulator/IngestionPopulator.js
```diff
@@ -140,92 +116,69 @@ class IngestionPopulator {
...
+        // Example buckets param object:
+        // [
+        //      {
+        //          accessKey: ...,
+        //          secretKey: ...,
+        //          endpoint: 'http://10.100.1.128:8000',
+        //          locationType: 'scality_s3',
+        //          bucketName: 'source-bucket-1',
+        //          zenkoBucket: 'my-zenko-bucket',
+        //          ingestion: { Status: 'enabled' },
```
philipyoo: this should be lowercase `status`

conf/Config.js
```diff
@@ -166,25 +166,34 @@ class Config extends EventEmitter {
...
-        this.locations = {};
-        Object.keys(locationConstraints).forEach(key => {
-            const locItem = locationConstraints[key];
-            this.locations[key] = Object.assign({}, locItem.details, {
-                locationType: ingestionTypeMatch[locItem.locationType],
-            });
+        this.ingestionBuckets = [];
+        buckets.forEach(bucket => {
+            const { name, ingestion, locationConstraint } = bucket;
+            const locationInfo = locationConstraints[locationConstraint];
```
nicolas2bert, line 178: Could `locationInfo` be undefined? If yes, `locationInfo.details` 
 here: https://github.com/scality/backbeat/pull/638/files#diff-f31fbe58af17434ce0c78133b65017d2R182 would throw an error.

conf/Config.js
```diff
@@ -166,25 +166,34 @@ class Config extends EventEmitter {
...
+                    locationType: ingestionTypeMatch[locationInfo.locationType],
+                    zenkoBucket: name,
+                    ingestion,
+                    locationConstraint,
+                }
+            );
+            this.ingestionBuckets.push(ingestionBucketDetails);
         });
-        this.emit('locations-update');
+        this.emit('ingestion-update');
```
nicolas2bert: It looks like ingestion-update is never used. :(

conf/Config.js
```diff
@@ -166,25 +166,34 @@ class Config extends EventEmitter {
...
-            });
+        this.ingestionBuckets = [];
+        buckets.forEach(bucket => {
+            const { name, ingestion, locationConstraint } = bucket;
+            const locationInfo = locationConstraints[locationConstraint];
+            const ingestionBucketDetails = Object.assign(
+                {},
+                locationInfo.details,
+                {
+                    locationType: ingestionTypeMatch[locationInfo.locationType],
```
nicolas2bert, line 192: Same for `locationInfo.locationType`, it could throw an error

lib/management/patchConfiguration.js
```diff
@@ -0,0 +1,89 @@
...
+            $type: 'object',
+        },
+    }).project({
+        'value.name': 1,
+        'value.ingestion': 1,
+        'value.locationConstraint': 1,
+    }).toArray((err, doc) => {
+        if (err) {
+            logger.error('error getting ingestion buckets',
+                { error: err.message });
```
nicolas2bert: You might want to add an error method to ease debugging ?

lib/management/patchConfiguration.js
```diff
@@ -0,0 +1,89 @@
+const arsenal = require('arsenal');
+const config = require('../../conf/Config');
+
+const USERSBUCKET = arsenal.constants.usersBucket;
+const PENSIEVE = 'PENSIEVE';
+const METASTORE = '__metastore';
+
+function getIngestionBuckets(metadata, logger, cb) {
```
philipyoo: moving this to arsenal MongoClientInterface and exposing through MDWrapper

lib/management/patchConfiguration.js
```diff
@@ -0,0 +1,89 @@
...
+            return cb(err);
+        }
+        return cb(null, doc.map(i => i.value));
+    });
+}
+
+function updateIngestionBuckets(locations, metadata, logger, cb) {
+    getIngestionBuckets(metadata, logger, (err, buckets) => {
+        if (err) {
+            return cb(err);
```
nicolas2bert, line 14: You might want to add a log here to ease debugging ?

lib/management/patchConfiguration.js
```diff
@@ -0,0 +1,89 @@
...
+    return cb(null, conf.version);
+}
+
+function patchConfiguration(overlayVersion, conf, metadata, logger, cb) {
+    if (conf.version === undefined) {
+        return process.nextTick(cb, null);
+    }
+    if (conf && conf.locations) {
+        // Update ingestion buckets regardless of config overlay version
+        return updateIngestionBuckets(conf.locations, metadata, logger, err => {
```
philipyoo: Pass an extra param to `initManagement`. Based on this param, we perform this query.
We should not impact the performance of other services because of this

conf/Config.js
philipyoo: The buckets should be buckets created from a given zenko storage location so I don't think this will be a problem, but i'm not 100% sure on if these values will be available, so will add a check with a debug log and will skip the bucket. Thanks!

rahulreddy (approved):

JianqinWang (approved):

philipyoo:
@bert-e approve

PR #641, opened 2019-03-12
rf: ZENKO-1566 api route parser map datapoint keys

philipyoo:
Goes with https://github.com/scality/Arsenal/pull/757

Rather than passing redis keys to route function defined in
Arsenal, we fetch the data point identifiers for a given
route from Arsenal and then map those to the redis keys
we want for a given service.

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -21,11 +21,12 @@ const { getSortedSetKey, getSortedSetMember } =
 const monitoringClient = require('../clients/monitoringHandler').client;
 
 const {
-    redisKeys,
+    redisKeys: crrRedisKeys,
     zookeeperNamespace: zookeeperReplicationNamespace,
     zkStatePath: zkReplicationStatePath,
 } = require('../../extensions/replication/constants');
 const {
+    redisKeys: ingestionRedisKeys,
     zookeeperNamespace: zookeeperIngestionNamespace,
     zkStatePath: zkIngestionStatePath,
 } = require('../../extensions/ingestion/constants');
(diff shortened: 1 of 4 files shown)


lib/api/BackbeatAPI.js
```diff
@@ -21,11 +21,12 @@ const { getSortedSetKey, getSortedSetMember } =
 const monitoringClient = require('../clients/monitoringHandler').client;
 
 const {
-    redisKeys,
+    redisKeys: crrRedisKeys,
```
JianqinWang, line 24: Is the  name `crrRedisKeys` updated in the `/../../extensisons/replication/constants` file?

lib/api/BackbeatAPI.js
```diff
@@ -21,11 +21,12 @@ const { getSortedSetKey, getSortedSetMember } =
...
     zookeeperNamespace: zookeeperReplicationNamespace,
     zkStatePath: zkReplicationStatePath,
 } = require('../../extensions/replication/constants');
 const {
+    redisKeys: ingestionRedisKeys,
```
JianqinWang, line 29: same thing in `../../extensions/ingestion/constants`

rahulreddy (approved):

lib/api/BackbeatAPI.js
rahulreddy: @JianqinWang  It's the opposite. Because of object destructuring he is assigning the value of redisKeys to a constant crrRedisKeys.

JianqinWang: my bad, thanks for clarifyinh

JianqinWang (approved):

philipyoo:
@bert-e approve

PR #643, opened 2019-03-13
feature: ZENKO-1520 transition policies metrics

bennettbuchanan:
Introduce a metric handler specifically for action entries. This behaves the same as the current metrics producer but does not process metrics during a lifecycle task, thereby preventing replication metrics pollution.

### DIFF
--- a/extensions/replication/ReplicationMetric.js
+++ b/extensions/replication/ReplicationMetric.js
@@ -1,0 +1,80 @@
+const { Logger } = require('werelogs');
+
+const MetricsModel = require('../../lib/models/MetricsModel');
+
+class ReplicationMetric {
+    constructor() {
+        this._log = new Logger('ReplicationMetric');
+    }
+
+    withProducer(producer) {
+        this._producer = producer;
+        return this;
+    }
+
+    withSite(site) {
+        this._site = site;
+        return this;
+    }
+
+    withObjectSize(objectSize) {
+        this._objectSize = objectSize;
+        return this;
+    }
+
+    /**
+     * Setter for an instance of ActionQueueEntry.
+     * @param {ActionQueueEntry} entry - The entry to publish a metric for
+     * @return {ReplicationMetric} - The current instance
+     */
+    withEntry(entry) {
+        this._entry = entry;
+        return this;
+    }
+
+    withMetricType(type) {
+        this._metricsType = type;
+        return this;
+    }
+
+    withExtension(extension) {
+        this._extension = extension;
+        return this;
+    }
+
+    _isLifecycleAction() {
+        const { origin } = this._entry.getContext();
+        return origin !== undefined && origin === 'lifecycle';
+    }
+
+    _createProducerMessage() {
+        const { bucket, key, version } = this._entry.getAttribute('target');
+        const metricsModel = new MetricsModel()
+            .withBytes(this._objectSize)
+            .withExtension(this._extension)
+            .withMetricType(this._metricsType)
+            .withSite(this._site)
+            .withBucketName(bucket)
+            .withObjectKey(key)
+            .withVersionId(version);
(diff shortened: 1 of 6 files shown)


lib/ActionMetric.js
```diff
@@ -0,0 +1,75 @@
+const { Logger } = require('werelogs');
+
+const MetricsModel = require('./models/MetricsModel');
+
+class ActionMetric {
```
jonathan-gramain: Since this class is pretty linked to replication semantics, it should not be generic, let's move it to extensions/replication/... and call it something like ReplicationMetric (or even CopyLocationMetric if specialized to the new task).

lib/ActionMetric.js
```diff
@@ -0,0 +1,75 @@
...
+        this._site = site;
+        return this;
+    }
+
+    withObjectSize(objectSize) {
+        this._objectSize = objectSize;
+        return this;
+    }
+
+    withEntry(entry) {
```
jonathan-gramain: Could maybe add JSDoc explaining that we expect an `ActionQueueEntry` here, since we are calling some of its member functions.

lib/ActionMetric.js
```diff
@@ -0,0 +1,75 @@
...
+        this._metricsType = type;
+        return this;
+    }
+
+    withExtension(extension) {
+        this._extension = extension;
+        return this;
+    }
+
+    _isLifecycleAction() {
```
jonathan-gramain: OK for now, I think eventually we should publish all replication metrics while keeping the context info, and do the demultiplexing based on the originator (lifecycle transition etc.) in the higher presentation layers.

lib/MetricsProducer.js
```diff
@@ -16,7 +16,7 @@ class MetricsProducer {
         this._kafkaConfig = kafkaConfig;
         this._topic = mConfig.topic;
 
-        this._producer = null;
```
jonathan-gramain: Could be cleaner to add an accessor `getProducer()` instead, to keep encapsulation.

lib/models/MetricsModel.js
```diff
@@ -26,6 +26,46 @@ class MetricsModel {
...
+        this._bytes = bytes;
+        return this;
+    }
+
+    withExtension(extension) {
+        this._extension = extension;
+        return this;
+    }
+
+    withType(type) {
```
jonathan-gramain: What is `type` here?

lib/ActionMetric.js
bennettbuchanan: Ok, I agree with that. This will serve as a kind of placeholder for following-up with a lifecycle metrics feature, and perhaps a service which can present a nice metrics API.

extensions/replication/ReplicationMetric.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    withExtension(extension) {
+        this._extension = extension;
+        return this;
+    }
+
+    _isLifecycleAction() {
+        const { origin } = this._entry.getContext();
+        return origin !== undefined && origin === 'lifecycle';
```
philipyoo, line 47: nice!

philipyoo (approved):

tmacro (approved):

extensions/replication/ReplicationMetric.js
jonathan-gramain: I'm not a fan of this to be honest :wink: but for now it's fine for me, eventually I would prefer that the replication metrics can "dispatch" based on some explicit originator value, but without having any semantic about a particular originator in the metrics code itself, as it should be agnostic of who is using the replication service.

jonathan-gramain (approved):

bennettbuchanan:
@bert-e approve

PR #644, opened 2019-03-13
feature: ZENKO-1387 filter transition entries

bennettbuchanan:
Prevent processing of duplicate lifecycle tasks. E.g. in the scenario where an object from a previous cron task has not finished the transition, the same object should not be queued again (unless the source data has been overwritten).

### DIFF
--- a/extensions/replication/utils/ReplicationTaskScheduler.js
+++ b/extensions/replication/utils/ReplicationTaskScheduler.js
@@ -1,24 +1,58 @@
 const async = require('async');
 
+const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
+const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
+
 /**
- * Schedule replication tasks for replicated buckets, by ensuring that
- * all updates to the same object are serialized
-*/
+ * Schedule replication tasks by ensuring that all updates to the same object
+ * are serialized. Any etnries that are duplicates (i.e. they have the same
+ * key, version, and content MD5) are skipped.
+ */
 class ReplicationTaskScheduler {
     constructor(processingFunc) {
         this._processingFunc = processingFunc;
         this._runningTasksByMasterKey = {};
+        this._runningTasks = {};
+    }
+
+    _getTaskKey(ctx) {
+        const { entry } = ctx;
+        if (entry instanceof ObjectQueueEntry) {
+            const key = entry.getCanonicalKey();
+            const version = entry.getVersionId();
+            const contentMd5 = entry.getContentMd5();
+            return `${key}:${version || ''}:${contentMd5}`;
+        }
+        if (entry instanceof ActionQueueEntry) {
+            const { key, version, contentMd5 } = entry.getAttribute('target');
+            return `${key}:${version || ''}:${contentMd5}`;
+        }
+        return undefined;
+    }
+
+    _getNewMasterKeyQueue(ctx, masterKey) {
+        const queue = async.queue(this._processingFunc);
+        queue.drain = () => delete this._runningTasksByMasterKey[masterKey];
+        this._runningTasksByMasterKey[masterKey] = queue;
+        return queue;
+    }
+
+    _getMasterKeyQueue(ctx, masterKey) {
+        const queue = this._runningTasksByMasterKey[masterKey];
+        return queue || this._getNewMasterKeyQueue(ctx, masterKey);
     }
 
     push(ctx, masterKey, done) {
-        let queue = this._runningTasksByMasterKey[masterKey];
-        if (queue === undefined) {
-            queue = async.queue(this._processingFunc);
-            queue.drain =
-                () => delete this._runningTasksByMasterKey[masterKey];
-            this._runningTasksByMasterKey[masterKey] = queue;
+        const taskKey = this._getTaskKey(ctx);
+        if (taskKey === undefined || this._runningTasks[taskKey]) {
+            return done();
         }
-        queue.push(ctx, done);
+        this._runningTasks[taskKey] = true;
+        const queue = this._getMasterKeyQueue(ctx, masterKey);
+        return queue.push(ctx, () => {
+            delete this._runningTasks[taskKey];
+            done();
+        });
     }
 }
 
--- a/tests/unit/replication/ReplicationTaskScheduler.spec.js
+++ b/tests/unit/replication/ReplicationTaskScheduler.spec.js
@@ -1,39 +1,309 @@
 const assert = require('assert');
+const uuid = require('uuid/v4');
 
-const ReplicationTaskScheduler = require(
-    '../../../extensions/replication/utils/ReplicationTaskScheduler');
+const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
+const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
 
-describe('replication task scheduler', () => {
-    it('should ensure serialization of updates to the same versioned object',
-    done => {
-        const taskScheduler = new ReplicationTaskScheduler(
-            (entry, done) => {
-                setTimeout(() => {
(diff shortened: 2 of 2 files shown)


extensions/replication/utils/ReplicationTaskScheduler.js
```diff
@@ -1,24 +1,63 @@
 const async = require('async');
 
+const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
+const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
+
 /**
- * Schedule replication tasks for replicated buckets, by ensuring that
- * all updates to the same object are serialized
-*/
+ * Schedule replication tasks for replication tasks, by ensuring that all
```
jonathan-gramain: "replication tasks for replication tasks" :egg: :repeat: :chicken:

extensions/replication/utils/ReplicationTaskScheduler.js
```diff
@@ -1,24 +1,63 @@
...
+
+    _isRunningTask(ctx) {
+        return this._runningTasks[this._getUniqueKey(ctx)];
+    }
+
+    _getNewQueue(ctx, masterKey) {
+        const queue = async.queue(this._processingFunc);
+        queue.drain = () => delete this._runningTasksByMasterKey[masterKey];
+        this._runningTasksByMasterKey[masterKey] = queue;
+        this._runningTasks[this._getUniqueKey(ctx)] = true;
```
jonathan-gramain: I think this update is misplaced. It should be done on *every* new incoming task that is not skipped, including those that get queued to an existing queue.

To test this you would have to add 3 entries:

- two with the same master key but different version and/or md5 (second entry should be queued after the first)
- one with the same master/version/md5 than the second entry

Then expect the latest entry to be skipped (I think in the current situation it would not be)

extensions/replication/utils/ReplicationTaskScheduler.js
```diff
@@ -1,24 +1,63 @@
...
-        if (queue === undefined) {
-            queue = async.queue(this._processingFunc);
-            queue.drain =
-                () => delete this._runningTasksByMasterKey[masterKey];
-            this._runningTasksByMasterKey[masterKey] = queue;
+        if (this._isRunningTask(ctx)) {
+            return done();
         }
-        queue.push(ctx, done);
+        const queue = this._getQueue(ctx, masterKey);
```
jonathan-gramain: I think this is the right place to update `_runningTasks`.

tests/unit/replication/ReplicationTaskScheduler.spec.js
```diff
@@ -1,39 +1,181 @@
...
+    function scheduleDuplicateVersions(params, cb) {
+        const { uniqueKeyCount, duplicateKeyCount } = params;
+
+        for (let i = 0; i < uniqueKeyCount; ++i) {
+            objects.push({ objectKey: `key_${i}`, value: -1 });
+            // the following inner operations shall be executed in
+            // order because they have the same objectKey (passed
+            // to replicationTaskScheduler.push())
+            const versionId = uuid();
+            for (let j = 0; j < duplicateKeyCount; ++j) {
```
jonathan-gramain: minor: looks a little confusing to have `duplicateKeyCount` include the original key (so duplicate == 1 means no duplicate, right), maybe simply shift its meaning by one and update loops etc. accordingly.

Also to add more fuzziness in the test (in order to increase coverage of cases) you could pre-create an array of entries, shuffle it, then insert entries in the shuffled order (or pick entries at random and remove them from the array as it goes).

jonathan-gramain (changes requested):

tests/unit/replication/ReplicationTaskScheduler.spec.js
bennettbuchanan: Sounds good. Updated the `duplicateKeyCount` value. Also added a randomized entries test.

extensions/replication/utils/ReplicationTaskScheduler.js
bennettbuchanan: 👍 Updated and added this test case.

extensions/replication/utils/ReplicationTaskScheduler.js
```diff
@@ -1,24 +1,57 @@
...
-            queue = async.queue(this._processingFunc);
-            queue.drain =
-                () => delete this._runningTasksByMasterKey[masterKey];
-            this._runningTasksByMasterKey[masterKey] = queue;
+        const taskKey = this._getTaskKey(ctx);
+        if (this._runningTasks[taskKey]) {
+            return done();
         }
-        queue.push(ctx, done);
+        this._runningTasks[taskKey] = true;
```
philipyoo, line 50: maybe won't happen because I think entries will be either ObjectQueueEntry or ActionQueueEntry, but `taskKey` can be undefined based on return from `_getTaskKey`

alexanderchan-scality (approved):

extensions/replication/utils/ReplicationTaskScheduler.js
```diff
@@ -1,24 +1,57 @@
 const async = require('async');
 
+const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
+const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
+
 /**
- * Schedule replication tasks for replicated buckets, by ensuring that
- * all updates to the same object are serialized
-*/
+ * Schedule replication tasks by ensuring that all updates to the same object
```
jonathan-gramain, line 7: Maybe you can also explain a bit about the just-introduced mechanism to skip entries :)

tests/unit/replication/ReplicationTaskScheduler.spec.js
```diff
@@ -1,39 +1,294 @@
...
+    }
+
+    function scheduleEntries(params, cb) {
+        const { uniqueKeyCount, hasUniqueVersions, hasUniqueContent } = params;
+        const duplicateKeyCount = params.duplicateKeyCount + 1;
+        for (let i = 0; i < uniqueKeyCount; ++i) {
+            const objectKey = `key_${i}`;
+            let versionId = uuid();
+            let contentMD5 = uuid();
+            for (let j = 0; j < duplicateKeyCount; ++j) {
```
jonathan-gramain: change to `j <= duplicateKeyCount`?

bennettbuchanan: I've added 1 to `duplicateKeyCount` above: https://github.com/scality/backbeat/pull/644/files/592eb2d9047720c614b25198e51b159ed0fe9e1c#diff-016c440d4159f39e05ea96d09aaaf274R68

So this should be good as is.

jonathan-gramain: Ah, ok, I would recommend though to avoid mixing params and local variables with the same name but a different meaning, as it brings confusion, you sure can use a local variable that has the same value as the param but a shorter form for convenience, and set `duplicateKeyCount + 1` in the loop or use `<=` instead.

extensions/replication/utils/ReplicationTaskScheduler.js
bennettbuchanan: Yeah, this shouldn't happen. But I agree it's better to be safe. I'll add a check for that.

tests/unit/replication/ReplicationTaskScheduler.spec.js
bennettbuchanan: Yeah, I agree that would improve readability. Updated!

extensions/replication/utils/ReplicationTaskScheduler.js
bennettbuchanan: Added a test for that case as well.

jonathan-gramain (approved):

bennettbuchanan:
@bert-e approve

PR #645, opened 2019-03-14
rf: ZENKO-1566 adjust metrics consumer

philipyoo:
The reason for a few of the changes here is to adapt to adding ingestion populator metrics. This means that metric entries need to be filtered by their given populator. The filter is done on each MetricsConsumer, and each MetricsConsumer is created by each QueuePopulator.

Helper method for reporting site-level metrics to StatsClient was adjusted to first check if the Redis key exists for the given populator and then check if the value is a non zero to avoid reporting empty metrics.

Plan is to add another MetricsConsumer on IngestionPopulator that reports pending metrics.

### DIFF
--- a/lib/MetricsConsumer.js
+++ b/lib/MetricsConsumer.js
@@ -2,9 +2,17 @@
 
 const Logger = require('werelogs').Logger;
 const { RedisClient, StatsModel } = require('arsenal').metrics;
+const errors = require('arsenal').errors;
 
 const BackbeatConsumer = require('./BackbeatConsumer');
-const redisKeys = require('../extensions/replication/constants').redisKeys;
+const {
+    redisKeys: crrRedisKeys,
+    metricsExtension: crrExtension,
+} = require('../extensions/replication/constants');
+const {
+    redisKeys: ingestionRedisKeys,
+    metricsExtension: ingestionExtension,
+} = require('../extensions/ingestion/constants');
 
 // StatsClient constant defaults for site metrics
 const INTERVAL = 300; // 5 minutes;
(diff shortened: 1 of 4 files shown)


lib/MetricsConsumer.js
```diff
@@ -58,35 +68,68 @@ class MetricsConsumer {
         });
     }
 
+    _getRedisKeys(extension) {
```
jonathan-gramain, line 71: Not sure how relevant it is or how intrusive relative to existing deployments (backward compat), but we may be able to just use the extension name to format the redis keys, instead of using a static lookup. That would make the consumer more generic and agnostic of which extension is using it. But I understand it might not be practically doable. Another option though can be to publish the redis key format in the extension manifest (e.g. extensions/replication/index.js).

lib/MetricsConsumer.js
```diff
@@ -141,8 +188,10 @@ class MetricsConsumer {
         return done();
     }
 
-    _sendRequest(key, value) {
-        this._statsClient.reportNewRequest(key, value);
+    _sendRequest(action, site, redisKeys, keyType, value) {
+        if (redisKeys[keyType] && value && value !== 0) {
```
jonathan-gramain: last test `value !== 0` looks useless - actually could we want to publish even if `value === 0`?

lib/MetricsConsumer.js
philipyoo: ah ya I really like the idea of formatting the redis keys by the entries' `extension` received. 

There were no options for users to change the format of redis keys, so I think if we keep the same format, there should be no backwards compat issues. The replication metrics entry extension `crr` matches with the redis key as well. Some tweaking for test keys would be needed, but that should be fine

Edit: Hm, actually this requires quite a few changes. I like the idea, but I think I'll wait on making the change, maybe after 1.1?

rahulreddy (approved):

lib/MetricsConsumer.js
philipyoo: This sends a request to redis to save the given value. Since it won't change the value in redis, I thought it'd be nice to avoid sending altogether

Edit: On second thought, I will keep functionality as is without ignoring 0 values. I want to minimize changes in this PR. Thanks for the input!

JianqinWang (approved):

philipyoo:
@bert-e approve

PR #646, opened 2019-03-14
bugfix: ZENKO-1582 check ETag before updating md on transition

jonathan-gramain:
If ETag does not match between transition action and object metadata,
skip update and garbage-collect the new location. This is to cover the
race condition that exists on non-versioned source bucket if a user
modifies the object while it is being transitioned, that would lead to
losing the latest write if updating the metadata anyway.

Add basic unit test coverage for object processor.


tmacro (approved):

bennettbuchanan (approved):

jonathan-gramain:
@bert-e approve

PR #648, opened 2019-03-15
bugfix: ZENKO-1582 skip update if new location == old location

jonathan-gramain:
It's possible to get duplicate updates, not always because of kafka
delivering duplicate messages, but also if a duplicate copy occurred
(possible even if we reduced the risk by deduplicating in the queue
processor).

In such case we need to skip the GC of the location otherwise we lose
the valid data.

Note that it's normally more likely to happen if the target is non-versioned (e.g. Azure bucket). But it also covers duplicate cases that may happen with versioned targets as well.


anurag4DSB (approved):

bennettbuchanan (approved):

jonathan-gramain:
@bert-e approve

PR #617, opened 2019-02-15
ft: ZENKO-1283 ingestion pause resume

philipyoo:
Changes in this PR:
- Add constants file for ingestion
- Add pause/resume for ingestion
- Update pause/resume readme to include RING-OOB
- Add functional tests for pause/resume ingestion

Follows same pattern for zookeeper as replication pause/resume:
`/backbeat/ingestion/state/<location-name>`

PR #650, opened 2019-03-18
ft: ZENKO-1441 allow updating ingestion locations

philipyoo:
Sits on top of https://github.com/scality/backbeat/pull/617

If changes to a storage location are made on an s3c ingestion location, we should detect these changes and apply them.

Changes in this PR:
- Add a method to check existing IngestionReaders'
  bucketdConfig information for any changes. If an
  update occurred, we want to update relevant, impacted
  instance variables and re-create IngestionProducer clients
- Update tests to check if updates are fetched properly

### DIFF
--- a/bin/ingestion.js
+++ b/bin/ingestion.js
@@ -53,17 +53,26 @@ function queueBatch(ingestionPopulator, taskState, log) {
 
     const maxRead = qpConfig.batchMaxRead;
     // apply updates to Ingestion Readers
-    ingestionPopulator.applyUpdates();
-    ingestionPopulator.processLogEntries({ maxRead }, err => {
-        taskState.batchInProgress = false;
+    ingestionPopulator.applyUpdates(err => {
         if (err) {
-            log.fatal('an error occurred during ingestion', {
-                method: 'bin::ingestion.queueBatch',
+            log.fatal('error occurred applying updates to ingestion readers', {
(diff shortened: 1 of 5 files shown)


bin/ingestion.js
```diff
@@ -53,17 +53,26 @@ function queueBatch(ingestionPopulator, taskState, log) {
...
                 error: err,
+                method: 'bin::ingestion.queueBatch',
             });
             scheduler.cancel();
             process.exit(1);
         }
+        ingestionPopulator.processLogEntries({ maxRead }, err => {
+            taskState.batchInProgress = false;
+            if (err) {
+                log.fatal('an error occurred during ingestion', {
```
rahulreddy, line 68: Did you make sure that `processLogEntries` does not return an error for non-critical errors?

JianqinWang (approved):

rahulreddy (approved):

bin/ingestion.js
philipyoo: Errors will only propagate here during setup of ingestion populator, or setup of new ingestion readers.
Errors produced from requests made to remote s3c are consumed within Ingestion Populator and logged. These errors are hard to discern because the error message is re-interpreted (I think) by backbeat client

philipyoo:
@bert-e approve

PR #654, opened 2019-03-20
bugfix: ZENKO-1601 fix replication status processor skipping entries

jonathan-gramain:
Modify the ReplicationTaskScheduler class with custom "queue key" and
"dedupe key" user functions: that allows the flexibility for queueing
and/or skipping tasks based on the provided keys. Also take the
opportunity to move it to global backbeat lib and rename the class to
TaskScheduler, because it's not specific to replication anymore and
might be potentially useful for other purposes.

The queue processor now passes explicitly queue key and dedupe key
functions, while the replication status processor only passes a queue
key (because it does not want deduplication).

I reworked the unit tests by splitting them conceptually to test the
scheduler and the key-generating functions separately. This makes them
simpler and more focused on testing each component.

On the scheduler tests, I think by reading the code that the removed
tests (recently introduced) were not effectively testing the queueing
behavior anymore, so I though cleaner to revert back to the previous
tests and add an extra test for deduplication.


lib/tasks/TaskScheduler.js
```diff
@@ -0,0 +1,91 @@
...
+ * Schedule replication tasks according to:
+ *
+ * - An optional queue key that guarantees all tasks sharing the same
+ *   queue key will be processed in order and serially
+ *
+ * - An optional dedupe key that will skip new tasks if they share the
+ *   same dedupe key than another task either in progress or already
+ *   queued
+ */
+class ReplicationTaskScheduler {
```
bennettbuchanan: ~ReplicationTaskScheduler~ TaskScheduler

Looks like we still need to update the name and path across files.

lib/tasks/TaskScheduler.js
```diff
@@ -0,0 +1,91 @@
...
+        let queueKey;
+        const onTaskEnd = () => {
+            if (dedupeKey !== undefined) {
+                delete this._dedupeCache[dedupeKey];
+            }
+            done();
+        };
+        if (this._getDedupeKeyFunc) {
+            dedupeKey = this._getDedupeKeyFunc(ctx);
+            if (typeof dedupeKey !== 'string') {
```
bennettbuchanan, line 67: Looks like it will either be a string or undefined in any case. I guess the extra check is just for extra safety?

jonathan-gramain: Kind of, say it's equivalent to checking if undefined, but also supports `null` as if no key is returned (so to skip either the dedupe or queueing behavior for a particular object). We could as well assert or error if it's something we don't understand (like an object or array).

lib/tasks/TaskScheduler.js
jonathan-gramain: Yes, I just forgot to commit that bit originally :rofl:

bennettbuchanan (approved):

lib/tasks/TaskScheduler.js
```diff
@@ -0,0 +1,91 @@
+const async = require('async');
+
+/**
+ * Schedule replication tasks according to:
```
philipyoo: minor: can remove replication

lib/tasks/TaskScheduler.js
```diff
@@ -0,0 +1,91 @@
...
+            queueKey = this._getQueueKeyFunc(ctx);
+            if (typeof queueKey !== 'string') {
+                queueKey = undefined;
+            }
+            if (queueKey !== undefined) {
+                const queue = this._getTaskQueue(ctx, queueKey);
+                return queue.push(ctx, onTaskEnd);
+            }
+        }
+        return process.nextTick(() => this._processingFunc(ctx, onTaskEnd));
```
philipyoo, line 87: not 100% sure, but `onTaskEnd`, would we need to bind `this`? Will it have reference to `this._dedupeCache`?

jonathan-gramain: I think so, because the function is a closure in the TaskScheduler object. But the particular case where we don't provide a queue func is not properly tested now, I can add a test case to make sure it works.

jonathan-gramain: Added a test for this line

philipyoo (approved):

jonathan-gramain:
@bert-e approve

PR #655, opened 2019-03-20
bf: ZENKO-1605 null version object md fetch

philipyoo:
When fetching object metadata for oob-ring, do not specify
the version id field. Instead fetch master which should have
`nullVersionId` defined with the null version id. Make
another request to get the null version specific object md.

In IngestionProducer._getBucketObjectsMetadata, master keys were
duplicated to emulate the additional metadata log entry to update
master. This should not occur for IsLatest null versions.


JianqinWang (approved):

philipyoo:
closing for now, I believe I found a better solution

PR #657, opened 2019-03-20
ft: ZENKO-1448 ingestion location cleanup

philipyoo:
Sits on top of https://github.com/scality/backbeat/pull/650

When a user removes an ingestion bucket, we should cleanup the bucket path in zookeeper.
In zookeeper, the path is prefixed like: `/ingestion/<bucket>`
This path has nodes that hold information about the snapshot phase and log offset

From a user perspective, in order for a cleanup to occur, they would need to remove all object versions from their Zenko bucket. This will remove object versions from their s3c bucket as well. Then, they can remove the Zenko bucket. The cleanup will occur after deletion of Zenko bucket and before we try and call `processLogEntries` for that given ingestion bucket.

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -186,10 +186,36 @@ class MongoQueueProcessor {
      * @return {undefined}
      */
     stop(done) {
-        if (!this._consumer) {
-            return setImmediate(done);
-        }
-        return this._consumer.close(done);
+        async.parallel([
+            next => {
+                if (this._consumer) {
+                    this.logger.debug('closing kafka consumer', {
+                        method: 'MongoQueueProcessor.stop',
+                        site: this.site,
+                    });
+                    return this._consumer.close(next);
+                }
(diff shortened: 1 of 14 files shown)


rahulreddy (approved):

JianqinWang (approved):

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -186,10 +186,20 @@ class MongoQueueProcessor {
      * @return {undefined}
      */
     stop(done) {
-        if (!this._consumer) {
-            return setImmediate(done);
-        }
-        return this._consumer.close(done);
+        async.parallel([
```
rahulreddy, line 189: Add a log here that the consumer is being stopped and another log if we are skipping because there was no consumer.

rahulreddy (changes requested):
The only thing I would suggest is adding logs in methods where there aren't any. It will help us debug easily when there is an issue. Make sure you add `method` param to the logs to give context.

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #658, opened 2019-03-22
feature: ZENKO-1585 configure replication group id

rahulreddy:
Replication group id is used in the calculation of version ids for objects
on buckets that have versioning enabled. It has many benefits including avoiding
collisions in an active-active scenario, identifying distinct entries from an
instance etc.
This allows commit allows configuring a custom replication group id for an instance,
ensuring that the version ids generated by the instance are unique.


philipyoo (approved):

JianqinWang (approved):

rahulreddy:
@bert-e approve

PR #659, opened 2019-03-22
bf: ZENKO-1605 fix null version obj md fetches

philipyoo:
Use the existing BackbeatClient `getMetadata` call vs `getObjectMetadata`.

The `getMetadata` route allows us to pass encoded version id's as a query param.
This solves a few issues for us:
- We no longer need to send a decoded version id as part of the url path
- Passing decoded version id in the path caused issues with nginx where it would see a null value in the url and fail
- For null versions, we fetch master which contains the null version id.

For null versions during snapshot phase, if the null version is the latest version, it is its own master key, so we send the single entry. If the null version is not the latest version, the master key contains a field called `nullVersionId` that holds the null versions version-id. In this case, we send both master and null version entries.

### DIFF
--- a/lib/clients/backbeat-2017-07-01.api.json
+++ b/lib/clients/backbeat-2017-07-01.api.json
@@ -1063,32 +1063,6 @@
                 "shape": "ObjectMDListResponse"
             }
         },
-        "GetObjectMetadata": {
-            "http": {
-                "method": "GET",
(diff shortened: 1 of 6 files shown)


JianqinWang (approved):

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #660, opened 2019-03-22
bugfix: ZENKO-1583 Non-versioned cloud transition

bennettbuchanan:
**Summary**
This change helps prevent the deletion of overwritten data on a non-versioned source during transition.

**Details**
In the action entry for a transition, set the `LastModified` value that was retrieved during the lifecycle processing task. Then, before garbage collecting the source object, perform an additional check that the object has not been modified.

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -812,6 +812,7 @@ class LifecycleTask extends BackbeatTask {
      * @param {string} params.objectKey - The object key name
      * @param {string} params.encodedVersionId - The object encoded version ID
      * @param {string} params.eTag - The object data ETag
+     * @param {string} params.lastModified - The last modified date of object
      * @param {string} params.site - The site name to transition the object to
      * @param {Werelogs.Logger} log - Logger object
      * @return {undefined}
@@ -828,7 +829,9 @@ class LifecycleTask extends BackbeatTask {
               .setAttribute('target.key', params.objectKey)
               .setAttribute('target.version', params.encodedVersionId)
               .setAttribute('target.eTag', params.eTag)
+              .setAttribute('target.lastModified', params.lastModified)
               .setAttribute('toLocation', params.site);
+
         this._sendDataMoverAction(entry, err => {
             if (err) {
                 log.error('could not send transition entry for consumption',
@@ -1024,6 +1027,7 @@ class LifecycleTask extends BackbeatTask {
                     bucket: bucketData.target.bucket,
                     objectKey: obj.Key,
                     eTag: obj.ETag,
+                    lastModified: obj.LastModified,
                     site: rules.Transition.StorageClass,
                 }, log);
                 return done();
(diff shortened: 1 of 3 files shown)


philipyoo (approved):

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -828,7 +829,9 @@ class LifecycleTask extends BackbeatTask {
               .setAttribute('target.key', params.objectKey)
               .setAttribute('target.version', params.encodedVersionId)
               .setAttribute('target.eTag', params.eTag)
-              .setAttribute('toLocation', params.site);
+              .setAttribute('toLocation', params.site)
+              .setAttribute('details.lastModified', params.lastModified);
```
jonathan-gramain: minor point: I know this have been derived from object processor code, but I think it's more logical to put this into `target` as it's directly related to defining target properties. It could make the message organization slightly clearer in case we have to add more stuff in the future.

extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
```diff
@@ -98,6 +115,20 @@ class LifecycleUpdateTransitionTask extends BackbeatTask {
         this.gcProducer.publishActionEntry(gcEntry, done);
     }
 
+    _checkAndGarbageCollectLocation(entry, locations, log, done) {
+        return async.series([
+            next => this._checkDate(entry, log, next),
+            next => this._garbageCollectLocation(entry, locations, log, next),
+        ], err => {
+            if (err && err.statusCode === 412) {
+                log.info('Object was modified after transition had begun ' +
```
jonathan-gramain: minor: Instead of "object was not deleted" I would refer to the data location, e.g. "data location was not deleted", not to confuse with deletion of the zenko object itself. "Object was modified" looks fine as it can apply to the zenko object as well as the underlying cloud object.

jonathan-gramain (commented):
Looks good, could you also add a pair of unit tests in tests/unit/lifecycle/LifecycleUpdateTransitionTask.spec.js where lastModified is specified and newer in one case, same in the other?

extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
```diff
@@ -98,6 +98,19 @@ class LifecycleUpdateTransitionTask extends BackbeatTask {
         this.gcProducer.publishActionEntry(gcEntry, done);
     }
 
+    _wasObjectModified(entry, objMD, log) {
+        const { lastModified } = entry.getAttribute('target');
+        const objectWasModified = lastModified !== objMD.getLastModified();
+        if (objectWasModified) {
+            log.info('object LastModified date changed during lifecycle ' +
+                     'transition processing');
+            Object.assign({
```
jonathan-gramain, line 107: This should be passed as second argument of `log.info`

bennettbuchanan: Thanks for that. Nice catch.

jonathan-gramain (approved):

bennettbuchanan:
@bert-e approve

PR #663, opened 2019-03-28
bugfix: ZENKO-1583 Add if-unmodified-since check

bennettbuchanan:
Functional tests forthcoming.

High-level description:

* Use the multiplebackend backbeat metadata route to fetch a cloud object `LastModified` value.
* Add conditional batch delete support using `IfUnmodifiedSince` logic.

The `IfUnmodifiedSince` value will only be sent by the backbeat garbage collector if the source object is on a non-versioned public cloud. If the object has changed since the transition was initiated, we do nothing. If it has not changed, we set a tag on the object (the tag key-value is sent by the garbage collector) indicating that the object has transitioned to a new location and should be deleted. Note: we do not carry out the delete operation ourselves due to a race condition which could result in lost data.

### DIFF
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -44,11 +44,27 @@ class GarbageCollectorTask extends BackbeatTask {
                 size: location.size,
                 dataStoreVersionId: location.dataStoreVersionId,
             })),
+            IfUnmodifiedSince: entry.getAttribute('source.lastModified'),
+            Bucket: entry.getAttribute('source.bucket'),
+            Key: entry.getAttribute('source.objectKey'),
+            StorageClass: entry.getAttribute('source.storageClass'),
+            Tags: JSON.stringify({
+                'scal-delete-marker': 'true',
+                'scal-delete-service': entry.getAttribute('serviceName'),
+            }),
         });
         attachReqUids(req, log);
         return req.send(err => {
             entry.setEnd(err);
             log.info('action execution ended', entry.getLogInfo());
+            if (err && err.statusCode === 412) {
+                log.info('precondition for garbage collection was not met',
+                    Object.assign({
+                        method: 'LifecycleObjectTask._executeDeleteData',
+                        lastModified: entry.getAttribute('source.lastModified'),
+                    }, entry.getLogInfo()));
+                return done();
+            }
             if (err) {
                 log.error('an error occurred on deleteData method to ' +
                           'backbeat route',
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -2,7 +2,9 @@
 
 const async = require('async');
 const { errors } = require('arsenal');
+const ObjectMD = require('arsenal').models.ObjectMD;
 
+const config = require('../../../conf/Config');
 const { attachReqUids } = require('../../../lib/clients/utils');
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
@@ -57,14 +59,29 @@ class LifecycleTask extends BackbeatTask {
     /**
      * Send entry to the data mover topic
      * @param {ActionQueueEntry} entry - The action entry to send to the topic
+     * @param {Logger.newRequestLogger} log - logger object
      * @param {Function} cb - The callback to call
      * @return {undefined}
      */
-    _sendDataMoverAction(entry, cb) {
+    _sendDataMoverAction(entry, log, cb) {
         const { bucket, key } = entry.getAttribute('target');
         const entries = [{ key: `${bucket}/${key}`,
                            message: entry.toKafkaMessage() }];
-        this.producer.sendToTopic(this.dataMoverTopic, entries, cb);
+        this.producer.sendToTopic(this.dataMoverTopic, entries, err => {
+            if (err) {
+                log.error('could not send entry for consumption',
+                    Object.assign({
+                        method: 'LifecycleTask._sendDataMoverAction',
+                        error: err,
+                    }, entry.getLogInfo()));
+                return cb(err);
+            }
+            log.debug('sent entry for consumption',
+                Object.assign({
+                    method: 'LifecycleTask._sendDataMoverAction',
+                }, entry.getLogInfo()));
+            return cb();
+        });
     }
 
     /**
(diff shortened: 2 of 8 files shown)


extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -802,6 +819,71 @@ class LifecycleTask extends BackbeatTask {
...
+            const { error, result } = ObjectMD.createFromBlob(blob.Body);
+            if (error) {
+                const msg = 'error parsing metadata blob';
+                return cb(errors.InternalError.customizeDescription(msg));
+            }
+            return cb(null, result);
+        });
+    }
+
+    _canSafelyGarbageCollectSource(objectMD) {
```
bennettbuchanan: Still working to validate this covers all of the cases.

lib/BackbeatMetadataProxy.js
```diff
@@ -105,6 +105,60 @@ class BackbeatMetadataProxy extends BackbeatTask {
...
+        log.debug('heading object', {
+            where: 'source',
+            bucket, objectKey, locations,
+            method: 'BackbeatMetadataProxy._headObjectOnce',
+        });
+        const cbOnce = jsutil.once(cb);
+        const req = this.backbeatSource.multipleBackendHeadObject({
+            Bucket: bucket,
+            Key: objectKey,
+            Locations: JSON.stringify(locations.map(location => ({
```
bennettbuchanan: Need to follow-up to ensure this cannot exceed a max header size for the request.

bennettbuchanan: Limited to just a single array entity as this serves the purpose in the multiple backend of cloudserver.

JianqinWang (approved):

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -44,6 +44,14 @@ class GarbageCollectorTask extends BackbeatTask {
                 size: location.size,
                 dataStoreVersionId: location.dataStoreVersionId,
             })),
+            IfUnmodifiedSince: entry.getAttribute('source.lastModified'),
+            Bucket: entry.getAttribute('source.bucket'),
+            Key: entry.getAttribute('source.objectKey'),
+            StorageClass: entry.getAttribute('source.storageClass'),
+            Tags: JSON.stringify({
+                'scal-delete-marker': 'true',
+                'scal-delete-service': 'lifecycle',
```
jonathan-gramain: I don't think we know it comes from lifecycle at this point. It would have to be passed as a separate attribute in the action message IMO (like an optional `"origin": "lifecycle"` attribute for example)

minor point: I would slightly prefer a naming such as `scal-delete-marker-created-by` or so, which looks more precise to me.

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -206,6 +206,58 @@
...
+                        "location": "header",
+                        "locationName": "X-Scal-Locations",
+                        "type": "string",
+                        "member": {
+                            "type": "structure",
+                            "required": [
+                                "key",
+                                "dataStoreName"
+                            ],
+                            "members": {
```
jonathan-gramain: Would it work to use the `LocationMDObj` shape instead of describing the structure?

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -818,31 +905,21 @@ class LifecycleTask extends BackbeatTask {
...
-              .setAttribute('target.key', params.objectKey)
-              .setAttribute('target.version', params.encodedVersionId)
-              .setAttribute('target.eTag', params.eTag)
-              .setAttribute('target.lastModified', params.lastModified)
-              .setAttribute('toLocation', params.site);
-
-        this._sendDataMoverAction(entry, err => {
+        async.waterfall([
+            next =>
+                this._getObjectMD(params, log, next),
```
jonathan-gramain, line 910: Can't we instead replace the `headObject` done earlier by this call? It sounds wasteful to do two metadata fetches per object (even three if we count the cloud one).

lib/BackbeatMetadataProxy.js
```diff
@@ -105,6 +105,58 @@ class BackbeatMetadataProxy extends BackbeatTask {
         });
     }
 
+    headObject(params, log, cb) {
```
jonathan-gramain: Why not `headLocation`, to disambiguate better between doing a HEAD on the zenko object and a HEAD on the object location itself?

lib/clients/backbeat-2017-07-01.api.json
bennettbuchanan: NIce! DIdn't notice that shape there. 😸 Will update.

extensions/gc/tasks/GarbageCollectorTask.js
bennettbuchanan: Nice catch. Will follow with that change. `lifecycle-transition`

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: As discussed, it looks like we can remove the `headObject` request altogether but requires further investigation wether we should instead just replace with the object MD request. Will follow with a separate PR for this specific change because it pertains to the expiration feature code path.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -885,6 +885,7 @@ class LifecycleTask extends BackbeatTask {
                     storageClass: objectMD.getDataStoreName(),
                     lastModified,
                 });
+                entry.setAttribute('serviceName', 'lifecycle-transition');
```
jonathan-gramain: Rather minor design comment: I did not think of this attribute as being passed around to all action messages during lifecycle processing, because the "context" attribute already fulfills this role for use in the lifecycle transition code.

I was thinking instead of an extra attribute passed to the GC API directly, expected and understood by the GC only, so in this regard it should simply be passed from the LifecycleUpdateTransitionTask code when creating the GC action, because this is the task that decides to GC locations (so akin to an extra parameter to a function call in other words).

jonathan-gramain (commented):
One comment, but otherwise looks good to me

extensions/lifecycle/tasks/LifecycleTask.js
bennettbuchanan: True, I had thought expiration went through that path as well, so this is a cleaner approach.

lib/BackbeatMetadataProxy.js
```diff
@@ -105,6 +105,58 @@ class BackbeatMetadataProxy extends BackbeatTask {
...
+
+    _headLocationOnce(params, log, cb) {
+        const { bucket, objectKey, locations } = params;
+        log.debug('heading object location', {
+            where: 'source',
+            bucket, objectKey, locations,
+            method: 'BackbeatMetadataProxy._headLocationOnce',
+        });
+        const cbOnce = jsutil.once(cb);
+        const req = this.backbeatSource.multipleBackendHeadObject({
```
jonathan-gramain, line 129: Ideally we would have called this route "multipleBackendHeadLocation" as well, we could fit this in a later refactor.

lib/BackbeatMetadataProxy.js
```diff
@@ -105,6 +105,58 @@ class BackbeatMetadataProxy extends BackbeatTask {
...
+        });
+        attachReqUids(req, log);
+        req.send((err, data) => {
+            if (err) {
+                // eslint-disable-next-line no-param-reassign
+                err.origin = 'source';
+                if (err.ObjNotFound || err.code === 'ObjNotFound') {
+                    return cbOnce(err);
+                }
+                log.error('an error occurred during head object request', {
```
jonathan-gramain: could log "head location" instead

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -897,13 +944,34 @@
         "BatchDelete": {
             "http": {
                 "method": "POST",
-                "requestUri": "/_/backbeat/batchdelete"
+                "requestUri": "/_/backbeat/batchdelete/{Bucket}/{Key+}"
```
jonathan-gramain, line 947: Are those new params required? If so, did we check that transient source complies with the change (the only other user of batchdelete route for now)?

jonathan-gramain (approved):
Looks good, left minor comments and one thing to double-check, but should be ready otherwise.

lib/clients/backbeat-2017-07-01.api.json
bennettbuchanan: It is my understanding that without explicitly stating them as required below, then they can be omitted. I.e. currently this route has no required params:

```
...
"required": [
],
...
```

Will validate with a transient source.

lib/BackbeatMetadataProxy.js
bennettbuchanan: I kept this as-is because in fact any error would originate from a head object request to the cloud. I think we can find a good medium if I update to say `head object location` as is done above.

lib/clients/backbeat-2017-07-01.api.json
jonathan-gramain: ok, so the API wrapper would automatically set the URL to `/_/backbeat/batchdelete` if none of Bucket or Key are provided? I don't know, but if it's the case, then we should be good.

Zenko E2E have a transient source test, it should catch any critical issue there.

lib/BackbeatMetadataProxy.js
jonathan-gramain: OK, indeed we could rephrase this way, or "head object on zenko location" or something like this.

lib/clients/backbeat-2017-07-01.api.json
bennettbuchanan: It seems to send placeholders for the bucket and key `/_/backbeat/batchdelete/{Bucket}/{Key+}`. After testing E2E and transient source it doesn't appear to affect behavior of the route.

bennettbuchanan:
@bert-e approve

PR #664, opened 2019-03-29
bugfix: ZENKO-1583 Check LastModified for versions

bennettbuchanan:
For versioned transitions, we also need to pass the `lastModified` value for our lifecycle updating check. Also, for safety, we should not perform the check if it is not provided in the target attributes.


jonathan-gramain (approved):

dora-korpar (approved):

JianqinWang (approved):

bennettbuchanan:
@bert-e approve

PR #667, opened 2019-04-01
bf: ZENKO-1683 ingestion location pause state fix

philipyoo:
Initial pause state for a new ingestion location is set as
paused only in-mem causing us to miss the init snapshot.


JianqinWang (approved):

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #671, opened 2019-04-02
bf: ZENKO-1587 fix ingestion pending metrics

philipyoo:
Changes in this PR:
- MongoQueueProcessor normalize pending method to decrement
  pending in cases of skipped entries or errors
- extend MetricsConsumer to handle these pending
  normalizations (currently only added pending decrements as
  I don't see a need for pending increments. But increments
  will be needed in replication)

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -20,7 +20,7 @@ const ObjectQueueEntry = require('../../lib/models/ObjectQueueEntry');
 const { getAccountCredentials } =
     require('../../lib/credentials/AccountCredentials');
 const MetricsProducer = require('../../lib/MetricsProducer');
-const { metricsExtension, metricsTypeCompleted } =
+const { metricsExtension, metricsTypeCompleted, metricsTypePendingOnly } =
     require('../ingestion/constants');
 
 // TODO - ADD PREFIX BASED ON SOURCE
@@ -374,6 +374,7 @@ class MongoQueueProcessor {
         return this._mongoClient.deleteObject(bucket, key, undefined,
             this.logger, err => {
                 if (err) {
+                    this._normalizePendingMetric(location);
                     this.logger.error('error deleting object metadata ' +
                     'from mongo', {
                         bucket,
@@ -407,6 +408,7 @@ class MongoQueueProcessor {
 
         this._getZenkoObjectMetadata(sourceEntry, (err, zenkoObjMd) => {
             if (err) {
+                this._normalizePendingMetric(location);
                 this.logger.error('error processing object queue entry', {
                     method: 'MongoQueueProcessor._processObjectQueueEntry',
                     entry: sourceEntry.getLogInfo(),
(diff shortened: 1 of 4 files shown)


lib/MetricsConsumer.js
```diff
@@ -95,37 +115,31 @@ class MetricsConsumer {
         }
         if (type === 'completed') {
             // Pending metrics
-            this._sendRequest('decrementKey', site, redisKeys, 'opsPending',
-                ops);
-            this._sendRequest('decrementKey', site, redisKeys, 'bytesPending',
-                bytes);
+            this._decrementPending(site, redisKeys, ops, bytes);
```
bennettbuchanan: ❤️

lib/MetricsConsumer.js
```diff
@@ -164,14 +178,14 @@ class MetricsConsumer {
...
-                type: 'processed'
+                type: 'queued'
             }
         */
         // filter metric entries by service, i.e. 'crr', 'ingestion'
         if (this._id !== data.extension) {
             return done();
         }
-        const operationTypes = ['completed', 'failed', 'queued'];
+        const operationTypes = MetricsConsumer.getValidMetricTypes();
```
bennettbuchanan: I really like the extraction of this array and the idea of containing the logic. It seems to me this idea can be extended even further here to determine `isValidType`. For example, you could keep `getValidMetricTypes` and add another method `MetricsConsumer.isValidMetricType` which calls `getValidMetricTypes` internally and returns a boolean. 😸

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -469,6 +473,19 @@ class MongoQueueProcessor {
...
 
+    /**
+     * For cases where we experience an error or skip an entry, we need to
+     * normalize pending metric. This means we will see pending metrics stuck
+     * above 0 and will need to bring those metrics down
+     * @param {string} location - location constraint name
+     * @return {undefined}
+     */
+    _normalizePendingMetric(location) {
+        const metric = { [location]: { ops: 1 } };
```
bennettbuchanan, line 484: 🤔 I wonder if it would be more intuitive and extensible to use a generic `metricsTypePending` which allows passing of a negative value for decrementing. That way we wouldn't need a special type just for decrementing. Maybe there is a factor that is preventing this though that I'm not aware of.

lib/MetricsConsumer.js
philipyoo: nice i like the idea of encapsulating both methods into one. So it will contain the list of valid types and return a boolean based on the type to check

lib/queuePopulator/IngestionProducer.js
```diff
@@ -166,7 +166,9 @@ class IngestionProducer {
                         if (err) {
                             return next(err);
                         }
-                        initialCseq = cseq;
+                        // we should perform snapshot phase up to this cseq, so
+                        // save bookmark at cseq + 1
+                        initialCseq = cseq + 1;
```
rahulreddy: This should be a commit of it's own as it's changing the behavior of QueuePopulator/IngestionProducer.

rahulreddy (changes requested):

extensions/mongoProcessor/MongoQueueProcessor.js
philipyoo: I changed it to have a type called "pendingOnly" and the caller can specify +/- values for ops and bytes. Since other metrics will report ops/bytes as positive values, its required to add negative values when reporting to the helper method. But other than that, it's a more flexible call. If normalizing requires a  positive ops value and a negative bytes value (if caller is batching), then thats possible with the change :+1:

jonathan-gramain (approved):

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #672, opened 2019-04-03
bf: ZENKO-1629 ingestion repInfo md only updates

philipyoo:
Identify md-only (i.e. PUT/DELETE tagging) updates for
setting the ReplicationInfo "Content" field. This is
necessary for replication-enabled ingestion buckets
in order to replicate md-only updates.

Idea is to fetch metadata from mongo and make comparisons
between the new entry and existing entry in mongo. If entry
does not exist in mongo, this is a newly ingested entry.
If entry exists in mongo, we only compare tagging between
the two entries as that is the only field we care about.

### DIFF
--- a/extensions/mongoProcessor/utils/contentTypeHelper.js
+++ b/extensions/mongoProcessor/utils/contentTypeHelper.js
@@ -1,0 +1,82 @@
+'use strict'; // eslint-disable-line
+
+function _getDataContent(entry) {
+    const contentLength = entry.getContentLength();
+    if (contentLength > 0) {
+        return ['DATA', 'METADATA'];
+    }
+    return ['METADATA'];
+}
+
+function _getMPUTagContent(entry) {
+    if (entry.isMultipartUpload()) {
+        return ['MPU'];
+    }
+    return [];
+}
+
+/**
+ * compares object tags between Zenko object and kafka object entry to see
+ * if entry contains a object tagging change (for replication Content field)
+ * @param {ObjectQueueEntry} entry - object metadata entry from Kafka entry
+ * @param {Object} zenkoObjMd - Zenko object metadata currently in Mongo
+ * @return {undefined}
+ */
+function _getObjectTagContent(entry, zenkoObjMd) {
+    const kafkaEntryTags = entry.getTags();
+    const zenkoTags = zenkoObjMd.tags;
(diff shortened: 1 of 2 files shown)


extensions/mongoProcessor/utils/contentTypeHelper.js
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Get the replicationInfo Content field for a newly ingested object entry
+ * @param {ObjectQueueEntry} entry - object queue entry object
+ * @param {Object|undefined} zenkoObjMd - object metadata currently saved in
+ *    mongo. Can be undefined if this entry has never been ingested before
+ * @return {Array} array of ReplicationInfo Content Type
+ */
+function getContentType(entry, zenkoObjMd) {
+    const content = [];
```
jonathan-gramain: minor style comment: since this variable is used only later, it could be created just before it's used.

extensions/mongoProcessor/utils/contentTypeHelper.js
```diff
@@ -0,0 +1,83 @@
+'use strict'; // eslint-disable-line
+
+function _getDataContent(entry) {
+    const contentLength = entry.getContentLength();
+    if (contentLength > 0) {
+        return ['DATA', 'METADATA'];
```
jonathan-gramain, line 6: Is it possible that we also need to do a `PUT_TAGGING` in case the kafka entry contains tags? Maybe it would not happen or it would always be a separate event, just wondering.

jonathan-gramain (approved):

extensions/mongoProcessor/utils/contentTypeHelper.js
philipyoo: edit:
i'll keep the content consistent with what i'm observing.
from some manual tests, it seems like tags aren't replicating when they are specified with  s3api put-object

JianqinWang (approved):

extensions/mongoProcessor/utils/contentTypeHelper.js
philipyoo: Seems like new objects w/ tagging will not replicate the tagging with it. Created a ticket (1707) for tracking

philipyoo:
@bert-e approve

PR #673, opened 2019-04-03
bf: ZENKO-1688 set location info for 0byte objs

philipyoo:
Ingested objects should have their location info set.
0 byte objects (incl delete markers) were not being set.

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -6,7 +6,7 @@ const http = require('http');
 
 const Logger = require('werelogs').Logger;
 const errors = require('arsenal').errors;
-const { replicationBackends } = require('arsenal').constants;
+const { replicationBackends, emptyFileMd5 } = require('arsenal').constants;
 const MongoClient = require('arsenal').storage
     .metadata.mongoclient.MongoClientInterface;
 const ObjectMD = require('arsenal').models.ObjectMD;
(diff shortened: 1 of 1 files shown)


extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -28,6 +28,9 @@ const { metricsExtension, metricsTypeCompleted } =
 // TODO - ADD PREFIX BASED ON SOURCE
 // april 6, 2018
 
+// Copied from cloudserver "emptyFileMd5"
+const ZERO_BYTE_ETAG = 'd41d8cd98f00b204e9800998ecf8427e';
```
bennettbuchanan: Looks like this is also available from Arsenal as a constant `emptyFileMd5`, and could be imported here. 😸

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -273,18 +276,33 @@ class MongoQueueProcessor {
...
+            const editLocations = locations.map(l => {
+                const newValues = {
+                    key: entry.getObjectKey(),
+                    dataStoreName: location,
+                    dataStoreType: 'aws_s3',
+                };
+                if (entry.getVersionId()) {
+                    newValues.dataStoreVersionId = entry.getEncodedVersionId();
+                }
+                return Object.assign({}, l, newValues);
```
bennettbuchanan: Minor: `l` seems a bit too vague here. Seems to only add benefit by spelling out `location`.

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -273,18 +276,33 @@ class MongoQueueProcessor {
      */
     _updateLocations(entry, location) {
         const locations = entry.getLocation();
-        const editLocations = locations.map(l => {
-            const newValues = {
+        if (!locations || locations.length === 0) {
+            const dataStoreVersionId = entry.getVersionId() ?
```
bennettbuchanan: Can't this be accessed with `entry.getDataStoreVersionId()`? That way we'd preserve the public cloud's version ID that's stored here, if I'm not mistaken.

philipyoo: For 0-byte objects and delete markers, the locations field will be empty. 
For non-0-byte objects, the locations field don't have dataStoreVersionId defined. I'm thinking since the objects I'm testing are created on s3c, it does not define this field itself, so we can't rely on this field to be set when ingesting these objects.

Do you know cases where this field will be set? I can only think of Zenko use-cases, but not sure of cases for s3c

JianqinWang (approved):

extensions/mongoProcessor/MongoQueueProcessor.js
bennettbuchanan: It looks like the `dataStoreVersionId` field is used in versions of S3C going back at least to 7.4.2. I'm not sure whether that affects this feature though.

bennettbuchanan: My understanding is that if S3C is using the multiple backend, you may end up with locations storing the value. @rahulreddy would probably have a better sense though!

rahulreddy: ~~S3C doesn't support any of the cloud backends even if the code is in there. So it's safe to ignore setting dataStoreVersionId.~~
I misunderstood the context. We need to set `dataStoreVersionId` to the encoded version id of S3C. This allows mapping of Zenko's version to S3C's version.

jonathan-gramain (approved):

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -268,23 +268,38 @@ class MongoQueueProcessor {
...
                 dataStoreType: 'aws_s3',
-            };
-            if (entry.getVersionId()) {
-                newValues.dataStoreVersionId = entry.getEncodedVersionId();
-            }
-            return Object.assign({}, l, newValues);
-        });
-        entry.setLocation(editLocations);
+                dataStoreETag: `1:${emptyFileMd5}`,
+                dataStoreVersionId,
```
jonathan-gramain, line 288: We don't seem to treat the same way the `dataStoreVersionId` here and below when there are locations which looks odd, setting it to the empty string here when no version is present, not setting it at when there are locations.

philipyoo: ah yea good catch, below I should set to empty string

philipyoo:
@bert-e approve

PR #674, opened 2019-04-05
bugfix: ZENKO-1606 Replicate tags from object PUT

bennettbuchanan:
Depends on https://github.com/scality/cloudserver/pull/1774.

### DIFF
--- a/extensions/replication/tasks/CopyLocationTask.js
+++ b/extensions/replication/tasks/CopyLocationTask.js
@@ -277,6 +277,7 @@ class CopyLocationTask extends BackbeatTask {
             ContentDisposition:
                 objMD.getContentDisposition() || undefined,
             ContentEncoding: objMD.getContentEncoding() || undefined,
+            Tags: JSON.stringify(objMD.getTags()),
             Body: incomingMsg,
         });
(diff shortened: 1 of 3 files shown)


rahulreddy (approved):

rahulreddy (changes requested):
Isn't passing the tags as a query param more efficient given that we don't have to parse it and query stringify again?

bennettbuchanan:
@rahulreddy It is more efficient, however we're using a query string to determine the type of multiple backend operation to perform i.e. `operation=putobject`. There wouldn't be a way to distinguish the query param `operation` from a tag set. For example, if a user put a tag key of `operation` and value `cats` it would break the request.

I think that it probably should have been designed with this in mind, but we've been using the tags passed as stringified JSON in headers so I've just stuck with the current approach in that regard.

rahulreddy (approved):

rahulreddy:
Ok makes sense, thanks. I think we need to revisit routing at some point.

jonathan-gramain (approved):

bennettbuchanan:
@bert-e approve

PR #675, opened 2019-04-09
bugfix: ZENKO-1606 MPU tagging during replication

bennettbuchanan:
Send any object tags to the route backbeat during an MPU operation.

Depends on:
* https://github.com/scality/cloudserver/pull/1779
* https://github.com/scality/Arsenal/pull/773

### DIFF
--- a/extensions/replication/tasks/CopyLocationTask.js
+++ b/extensions/replication/tasks/CopyLocationTask.js
@@ -444,6 +444,7 @@ class CopyLocationTask extends BackbeatTask {
             ContentEncoding: objMD.getContentEncoding() ||
                 undefined,
             UploadId: uploadId,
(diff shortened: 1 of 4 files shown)


dora-korpar (approved):

jonathan-gramain (commented):
Why do we need to pass tags both at init and complete MPU stages, is it because some backends need them at init and some other backends need them at complete, or some other reason?

bennettbuchanan:
@jonathan-gramain Yes, that's right. Specifically, Azure doesn't have a notion of initiating an MPU, so for that backend you set the tags during the completion.

jonathan-gramain (approved):

bennettbuchanan:
@bert-e approve

PR #676, opened 2019-04-10
bf: ZENKO-1718 ingestion master version

philipyoo:
Goes with https://github.com/scality/Arsenal/pull/774

Avoid handling of master version entries. Rely on pre-existing logic for handling this for us.

There are these single null versions that look like master keys. To create these, you create object(s) in a non-versioned bucket, then enable versioning on bucket, and try and ingest these null version objects. Their obj md specifies no "isNull" and they have no internal version id (the one we default for null versions in suspended buckets). I mention these in in-line comments usually as "single null versions"

Changes in this PR:
- Filter master keys on populator
- Update arsenal, and use `putObjectVerCase4`
- Always set `dataStoreVersionId` in the `locations` array on object md.
     - null version objects with no internal version id (single null versions). We encode the string "null" and save to mongo so we can reference these null versions from Zenko.

Note: moved out MongoQueueProcessor fxnl tests to own PR

### DIFF
--- a/extensions/ingestion/IngestionQueuePopulator.js
+++ b/extensions/ingestion/IngestionQueuePopulator.js
@@ -2,6 +2,7 @@ const { zenkoIDHeader } = require('arsenal').constants;
 
 const QueuePopulatorExtension =
           require('../../lib/queuePopulator/QueuePopulatorExtension');
+const { isMasterKey } = require('arsenal/lib/versioning/Version');
 const ObjectQueueEntry = require('../../lib/models/ObjectQueueEntry');
 
 class IngestionQueuePopulator extends QueuePopulatorExtension {
@@ -61,6 +62,10 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
                 this.log.trace('skipping retro-propagated entry');
                 return true;
             }
+            if (this._isMasterKeyEntry(queueEntry)) {
+                this.log.trace('skipping master key entry');
+                return true;
+            }
         }
         return false;
     }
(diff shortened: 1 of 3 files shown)


extensions/ingestion/IngestionQueuePopulator.js
```diff
@@ -19,15 +21,42 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
...
                 return;
             }
+            // Filter out any master key object entries
+            if (entry.type === 'put') {
+                const queueEntry = new ObjectQueueEntry(entry.bucket,
+                                                        entry.key,
+                                                        metadataVal);
+                const sanityCheckRes = queueEntry.checkSanity();
+                if (sanityCheckRes) {
+                    this.log.trace('entry malformed', {
```
bennettbuchanan: Minor: the log message might also benefit by including the `sanityCheckRes.message`.

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -475,19 +471,27 @@ class MongoQueueProcessor {
                 zenkoObjMd);
 
             const objVal = sourceEntry.getValue();
-            // Always call putObject with version params undefined so
-            // that mongoClient will use putObjectNoVer which just puts
-            // the object without further manipulation/actions.
-            // S3 takes care of the versioning logic so consuming the queue
-            // is sufficient to replay the version logic in the consumer.
-            return this._mongoClient.putObject(bucket, key, objVal, undefined,
+            const params = { usePHD: true };
```
bennettbuchanan: Minor: seems a bit cleaner to leave this `undefined` for the `putObjectNoVer` condition, and assign it as an object with those fields for the `putObjectVerCase4`.

bennettbuchanan (approved):

tmacro (approved):

bennettbuchanan (approved):

philipyoo:
@bert-e approve

PR #677, opened 2019-04-11
bf: ZENKO-1626 pause resume validation based on single source

philipyoo:
Pause/resume for crr unsubscribes/subscribes consumers to
a given topic. When a request comes in, a check was made
on kafka consumer subscriptions, then we fetch data from
zookeeper, update zookeeper, and then update kafka
consumer subscriptions.

There were cases that appeared where consumer subscription
and zookeeper state would differ. This meant that if in
zookeeper pause state and consumer active state, the user
would never be able to resume the given site.

This change relies only checking state in zookeeper as a
single source of state. If disparity occurs, the user will
be able to pause/resume again to try and fix the disparity.

If zookeeper shows paused and consumer is active and the
user wants paused state. The user can resume, and then
pause again. This will send another request to unsubscribe
the consumer.

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -1,6 +1,5 @@
 'use strict'; // eslint-disable-line
 
-const async = require('async');
 const http = require('http');
 const https = require('https');
 const { EventEmitter } = require('events');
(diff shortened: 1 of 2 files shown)


philipyoo:
@bert-e reset

alexanderchan-scality (approved):

JianqinWang (approved):

PR #680, opened 2019-04-11
bugfix: ZENKO-1729 Use if..else for error logging

bennettbuchanan:
During the error condition, we shouldn't log both the error message and success debug message.

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -918,8 +918,14 @@ class LifecycleTask extends BackbeatTask {
                     method: 'LifecycleTask._applyTransitionRule',
                     error: err,
                 });
+            } else {
+                log.debug('transition rule applied', {
+                    method: 'LifecycleTask._applyTransitionRule',
+                    bucket: params.bucket,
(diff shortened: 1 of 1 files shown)


jonathan-gramain (approved):

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -918,8 +918,9 @@ class LifecycleTask extends BackbeatTask {
                     method: 'LifecycleTask._applyTransitionRule',
                     error: err,
                 });
+            } else {
+                log.debug('transition rule applied');
```
jonathan-gramain: Not directly in the scope of this PR, though logs without any context tend to be useful only for the developer that actually put those logs in the first place while developing the code. It could be made more useful by adding at least the object info in the log.

bennettbuchanan: Good point. I've added the method, bucket name, object key, and site name in this debug log.

anurag4DSB (approved):

philipyoo (approved):

bennettbuchanan:
@bert-e approve

PR #681, opened 2019-04-19
ZENKO-1730 lifecycle checks data mover progress

jonathan-gramain:
**Summary**

To prevent lifecycle from running a batch again when data mover still
has some messages to process from the previous batch, introduce a new
topic offset snapshot mechanism to check progress of the data mover,
in addition to the existing bucket and object tasks consumer lag
checks.

Also refactored how checks are done: all the mechanisms to check
offsets have been moved to a common class KafkaBacklogMetrics for
clarity, that is shared between all components either publishing or
checking the offsets.

**Details**

Checking the consumer lag for the data mover topic, like for bucket
and object tasks would not be ideal, as the data mover might be used
by other processes which might introduce unrelated consumer lag.

Instead, we use a snapshot/check system, where:

- the bucket processor takes a snapshot of the data mover topic
  offsets after it has finished processing a bucket

- the conductor checks whether the latest data mover consumer offsets
  are beyond the latest snapshot offset values, which means that what
  the bucket processor published to the data mover has already been
  fully processed, but regardless of messages coming from other
  services. It also still checks the bucket processor and object
  processor consumer lag as before.


lib/KafkaBacklogMetrics.js
```diff
@@ -0,0 +1,467 @@
...
+const { EventEmitter } = require('events');
+const zookeeper = require('node-zookeeper-client');
+
+const Logger = require('werelogs').Logger;
+const errors = require('arsenal').errors;
+
+const zookeeperHelper = require('./clients/zookeeper');
+
+const CheckConditionError = {};
+const NoNodeError = {};
```
bennettbuchanan: Why not make these instances of `Error`?

bennettbuchanan (approved):

tests/functional/lifecycle/LifecycleTask.js
```diff
@@ -344,6 +344,12 @@ class ProducerMock {
     }
 }
 
+class KafkaBacklogMetricsMock {
```
philipyoo, line 351: looks like ProducerMock on L#305 just needs to add a mock of method `getKafkaProducer` to get these functional tests working again

lib/KafkaBacklogMetrics.js
jonathan-gramain: I could, but I don't know if that would bring any value, but I'm open to changing this if it does! I though since those errors never leave the inner class functions, they can be simple empty objects.

jonathan-gramain:
@philipyoo @bennettbuchanan I added a new commit to fix and test better the snapshot mechanism as there was a couple bugs there.

I also added another commit to fix a typo in the BackbeatMetadataProxy test mock of lifecycle tests (headObject -> headLocation), as I saw the tests failing when trying locally, but the tests are passing on the CI with the typo, that looks surprising to me and I don't have an explanation so far.

lib/KafkaBacklogMetrics.js
bennettbuchanan: Definitely it's functionally the same. In my mind I was expecting the objects to be populated with values in some way, so by making them an `Error` I would know they'd be used as such. It's a minor readability point though.

jonathan-gramain: done

bennettbuchanan (approved):

lib/KafkaBacklogMetrics.js
```diff
@@ -0,0 +1,476 @@
...
+    checkConsumerLag(topic, groupId, maxLag, cb) {
+        this._log.debug('checking consumer lag', {
+            topic, groupId, maxLag,
+        });
+        return this._checkConsumerOffsetsGeneric(
+            topic, groupId, maxLag, null, cb);
+    }
+
+    /**
+     * Check whether the given consumer group lags beyond the maximum
```
philipyoo: minor: rather than max lag, a check on consumer info

philipyoo (approved):

lib/KafkaBacklogMetrics.js
jonathan-gramain: Good catch, a cut&paste typo

jonathan-gramain:
@bert-e approve

PR #682, opened 2019-04-20
bf: ZENKO-1585 retro propagation filter for s3c ingestion

philipyoo:
Retro-propagation bug fix.

Filter any ingestion entry that has user metadata header "x-amz-meta-zenko-instance-id" defined.
If a hashed instance id is defined, compare to the current Zenko deployment instance id and only
filter if it matches with current deployment

### DIFF
--- a/conf/Config.js
+++ b/conf/Config.js
@@ -5,6 +5,7 @@ const { EventEmitter } = require('events');
 const fs = require('fs');
 const path = require('path');
 const joi = require('joi');
+const crypto = require('crypto');
 
 const extensions = require('../extensions');
 const backbeatConfigJoi = require('./config.joi.js');
@@ -212,6 +213,16 @@ class Config extends EventEmitter {
 
     getIsTransientLocation(locationName) {
         return this.transientLocations[locationName] || false;
+    }
+
+    getPublicInstanceId() {
+        return this.publicInstanceId;
+    }
+
+    setPublicInstanceId(instanceId) {
+        this.publicInstanceId = crypto.createHash('sha256')
+                                .update(instanceId)
+                                .digest('hex');
     }
 }
 
--- a/extensions/ingestion/IngestionQueuePopulator.js
+++ b/extensions/ingestion/IngestionQueuePopulator.js
@@ -8,6 +8,7 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
     constructor(params) {
         super(params);
         this.config = params.config;
+        this._instanceId = params.instanceId;
     }
 
     // called by _processLogEntry in lib/queuePopulator/LogReader.js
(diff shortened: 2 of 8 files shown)


conf/Config.js
```diff
@@ -215,6 +216,18 @@ class Config extends EventEmitter {
     getIsTransientLocation(locationName) {
         return this.transientLocations[locationName] || false;
     }
+
+    getInstanceId() {
+        return this.instanceId;
+    }
+
+    setInstanceId(instanceId) {
```
rahulreddy: This should be coming from the Cloudserver Manager. It should never be set by an Backbeat process.

rahulreddy (changes requested):
Instance ID can be retrieved from Penseive DB, it should not be generated by a Backbeat process.

conf/Config.js
```diff
@@ -215,6 +216,18 @@ class Config extends EventEmitter {
     getIsTransientLocation(locationName) {
         return this.transientLocations[locationName] || false;
     }
+
+    getPublicInstanceId() {
+        return this.publicInstanceId;
+    }
+
+    setPublicInstanceId(instanceId) {
+        if (!this.publicInstanceId) {
```
bennettbuchanan: I'd expect a setter to always set the value, leaving the caller responsible for determining when it should be set.

extensions/ingestion/IngestionQueuePopulator.js
```diff
@@ -19,15 +23,58 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
...
                 metadataVal.attributes) {
                 return;
             }
+            if (entry.type === 'put') {
+                const queueEntry = new ObjectQueueEntry(entry.bucket,
+                                                        entry.key,
+                                                        metadataVal);
+                const sanityCheckRes = queueEntry.checkSanity();
+                if (sanityCheckRes) {
+                    this.log.trace('entry malformed', {
```
bennettbuchanan: Could be helpful also to include the message in `sanityCheckRes.message`.

extensions/ingestion/IngestionQueuePopulator.js
```diff
@@ -19,15 +23,58 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
...
+                            bucket: entry.bucket,
+                            key: entry.key,
+                            type: entry.type,
+                        });
+                        return;
+                    }
+                    // if user metadata field of `constants.zenkoIDHeader`
+                    // exists and value is either 'zenko' (for delete markers)
+                    // or matches given hashed instance id
+                    if (existingIDHeader && (existingIDHeader === 'zenko' ||
```
bennettbuchanan: Is `'zenko'` meant to be `constants.zenkoIDHeader`?

extensions/ingestion/IngestionQueuePopulator.js
```diff
@@ -19,15 +23,58 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
...
+                    } catch (err) {
+                        this.log.trace('malformed user metadata', {
+                            method: 'IngestionQueuePopulator.filter',
+                            bucket: entry.bucket,
+                            key: entry.key,
+                            type: entry.type,
+                        });
+                        return;
+                    }
+                    // if user metadata field of `constants.zenkoIDHeader`
```
bennettbuchanan: I might suggest breaking this down into smaller functions because it's starting to get somewhat deep in conditionals. Doing so could help with readability and make stack traces more granular.

bennettbuchanan (approved):
👍 Nice turnaround on this! I've left some considerations.

extensions/ingestion/IngestionQueuePopulator.js
philipyoo: `constants.zenkoIDHeader` is the user metadata header name "x-amz-meta..."
The value for this header can either be the hashed zenko instance id or "zenko" though

extensions/ingestion/IngestionQueuePopulator.js
```diff
@@ -39,6 +58,83 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
         this._incrementMetrics(entry.bucket);
     }
 
+    _filterValueOp(entry) {
```
JianqinWang, line 39: This looks like it's a wrapper for some of the logic in the `filter` function, otherwise it looks to me like it's unused.

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #688, opened 2019-05-03
bf: ZENKO-1770 fix ingestion reader test

philipyoo:
A deep compare is made between kafka entry and expected
entry. The object MD field 'md-model-version' can and will
change in the future, so we should ignore comparing this
field.


philipyoo:
tested recent breaking eve here:
https://github.com/scality/backbeat/compare/bugfix/test1770changes?expand=1

jonathan-gramain (approved):

JianqinWang (approved):

philipyoo:
@bert-e approve

PR #691, opened 2019-05-08
ZENKO-1520 PromAPIClient helper

jonathan-gramain:
Create a helper class to query Prometheus API, with helper functions
to execute a query and build Prometheus label selector strings.

### DIFF
--- a/lib/clients/PromAPIClient.js
+++ b/lib/clients/PromAPIClient.js
@@ -1,0 +1,149 @@
+const http = require('http');
+const querystring = require('querystring');
+
+const { errors, jsutil } = require('arsenal');
+
+class PromAPIClient {
+    /**
+     * @constructor
+     * @param {string} promEndpoint - Prometheus API endpoint
+     * @param {http.Agent} [promAPIHTTPAgent] - HTTP Agent
+     */
+    constructor(promEndpoint, promAPIHTTPAgent) {
+        const [promHost, promPort] = promEndpoint.split(':');
+        this._promHost = promHost;
+        this._promPort = promPort;
+        this._promAPIHTTPAgent = promAPIHTTPAgent;
+    }
+
+    /**
+     * Convenience function to build a metric selector from a base
+     * metric name and optional label names and label regexps
+     *
+     * @param {string} metricName - base name of metric
+     * @param {object} [labels] - optional object to select a set of
+     * labels by name, as { labelType: labelName, ... } or {
+     * labelType: [labelName, labelName, ...], ... }
+     * @param {object} [labelRegexps] - optional object to select a
+     * set of labels by regular expression, as { labelType: regexp }
+     * @return {string} a PROMQL metric name with label selectors set
+     */
+    static buildMetricSelector(metricName, labels, labelRegexps) {
+        const selectors = [];
+        if (labels) {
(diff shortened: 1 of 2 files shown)


lib/clients/PromAPIClient.js
```diff
@@ -0,0 +1,149 @@
...
+        const cbOnce = jsutil.once(cb);
+        const qs = querystring.stringify(params);
+        const opts = {
+            hostname: this._promHost,
+            port: this._promPort,
+            path: `/api/v1/query?${qs}`,
+            agent: this._promAPIHTTPAgent,
+        };
+        log.debug('sending query to Prometheus API server', {
+            method: 'LifecycleMetrics._executePromQuery',
```
philipyoo: not sure if you have changes to this method name, but should this be changed to `PromAPIClient.executeQuery`?

lib/clients/PromAPIClient.js
```diff
@@ -0,0 +1,149 @@
...
+        const req = http.request(opts, res => {
+            const buffers = [];
+            res.on('data', chunk => {
+                buffers.push(chunk);
+            });
+            return res.on('end', () => {
+                const data = buffers.join('');
+                log.debug(
+                    'response received for query to Prometheus API server', {
+                        method: 'LifecycleMetrics._executePromQuery',
```
philipyoo: same as above about method name

lib/clients/PromAPIClient.js
```diff
@@ -0,0 +1,149 @@
...
+                        payloadLength: data.length,
+                        elapsedMs: Math.round(Date.now() - begin),
+                    });
+                let parsedResp;
+                try {
+                    parsedResp = JSON.parse(data);
+                } catch (err) {
+                    log.error(
+                        'error parsing response from Prometheus API server', {
+                            method: 'LifecycleMetrics._executePromQuery',
```
philipyoo: same as above about method name

lib/clients/PromAPIClient.js
```diff
@@ -0,0 +1,149 @@
...
+                            host: opts.hostname,
+                            port: opts.port,
+                            path: opts.path,
+                            error: err.message,
+                            rawData: data.slice(0, 200),
```
philipyoo, line 110: just curious why you chose to slice here to 200 chars?

lib/clients/PromAPIClient.js
```diff
@@ -0,0 +1,149 @@
...
+                        });
+                    return cbOnce(err);
+                }
+                if (res.statusCode !== 200 || parsedResp.status !== 'success') {
+                    log.error(
+                        'Prometheus API server returned an error status', {
+                            method: 'LifecycleMetrics._executePromQuery',
```
philipyoo: same as above about method name

philipyoo (approved):
lgtm :+1:

lib/clients/PromAPIClient.js
jonathan-gramain: This is to avoid making a huge log line (which could be truncated itself by the logger, but not sure), since we are logging some raw return data, but it can be useful to log the first characters to get a better sense of the error (since it is not valid JSON).

jonathan-gramain:
Will reopen a new PR because of a change in ticket number

PR #692, opened 2019-05-08
rf: ZENKO-1744 ingestion processor performance

philipyoo:
This PR is an attempt to improve performance of ingestion processor

Changes in this PR:
- Add MongoQueueProcessor functional tests
- Add use of request logger per kafka entry consumed
- Remove s3 client as not needed. Instead rely on info obtained from Mongo
- Memoize bucket info up to 1 minute expiry per ingestion processor
- Limit object md fetches from Mongo. This info is only required for determining content type for replication. We can fetch the object md only if the bucket has replication enabled


JianqinWang (approved):

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #693, opened 2019-05-09
feature: ZENKO-1520 expose data mover metrics

jonathan-gramain:
Add queued and processed metrics to the data mover.

In order to have the data mover completely manage its own metrics and
have a cleaner API, created a ReplicationAPI class containing methods
to create action entries and to send them to the data mover queue,
with a list of documented parameters.

### DIFF
--- a/extensions/replication/ReplicationAPI.js
+++ b/extensions/replication/ReplicationAPI.js
@@ -1,0 +1,100 @@
+const config = require('../../conf/Config');
+
+const ActionQueueEntry = require('../../lib/models/ActionQueueEntry');
+const ReplicationMetrics = require('./ReplicationMetrics');
+
+let { dataMoverTopic } = config.extensions.replication;
+
+class ReplicationAPI {
+    /**
+     * Create an action to copy an object's data to a new location.
+     *
+     * Pass the returned object to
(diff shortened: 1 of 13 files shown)


extensions/replication/ReplicationAPI.js
```diff
@@ -0,0 +1,100 @@
...
+                contentLength, reports[0].partition);
+            return cb();
+        });
+    }
+
+    static getDataMoverTopic() {
+        return dataMoverTopic;
+    }
+
+    static setDataMoverTopic(newTopicName) {
```
philipyoo, line 95: maybe being too picky, but I was thinking setters shouldn't be static methods because there is state that is changing

philipyoo (approved):
lgtm 👍

extensions/replication/ReplicationAPI.js
jonathan-gramain: Actually I added this setter/getter for lifecycle unit tests only as they set a specific topic name instead of the configured one. I think it's ok as it avoids to have to create a ReplicationAPI class instance which otherwise would not be of much use.

tmacro (approved):

jonathan-gramain:
@bert-e approve

PR #698, opened 2019-05-17
bugfix: ZENKO-1567 prevent overwriting a COMPLETED status

jonathan-gramain:
Add a check to prevent UpdateReplicationStatus task from overwriting
an existing COMPLETED status on a location. e.g. with a FAILED status.

That prevents setting spurious replication status because COMPLETED is
a final status, it does not make sense to change it afterwards.

There can be multiple reasons for this situation to occur, one is with
a transient source where replication kafka messages would be
reprocessed twice (e.g. due to node failures): it's then possible that
the second replication attempt fails if the garbage collector already
deleted the data after the first successful attempt. Note that we
already prevent triggering a new replication if the location status is
COMPLETED in the metadata, but it's possible that the new replication
is triggered before the first one completes.

### DIFF
--- a/extensions/replication/tasks/UpdateReplicationStatus.js
+++ b/extensions/replication/tasks/UpdateReplicationStatus.js
@@ -147,18 +147,38 @@ class UpdateReplicationStatus extends BackbeatTask {
         return undefined;
     }
 
-    _getUpdatedSourceEntry(params) {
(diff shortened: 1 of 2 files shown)


alexanderchan-scality (approved):

philipyoo (approved):

jonathan-gramain:
@bert-e approve

PR #708, opened 2019-05-29
bf: ZENKO-1744 reduce metrics kafka messages

philipyoo:
Accumulate metric entries for MongoQueueProcessor in batches of 5 seconds per consumer pod.

Ingestion consumers can potentially overload Kafka because a metric kafka entry will be created per consumed and processed kafka entry. For ingestion, consumers will typically consume and process an entry in under 100ms. This change will batch metric entries and publish the accumulated entries (per consumer and per location) in intervals of 5s (arbitrarily picked, but can be adjusted)

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -21,6 +21,9 @@ const { metricsExtension, metricsTypeCompleted, metricsTypePendingOnly } =
 const getContentType = require('./utils/contentTypeHelper');
 const BucketMemState = require('./utils/BucketMemState');
 
+// batch metrics by location and send to kafka metrics topic every 5 seconds
+const METRIC_REPORT_INTERVAL_MS = process.env.CI === 'true' ? 1000 : 5000;
+
 // TODO - ADD PREFIX BASED ON SOURCE
 // april 6, 2018
 
@@ -71,6 +74,14 @@ class MongoQueueProcessor {
         this.mongoClientConfig.logger = this.logger;
         this._mongoClient = new MongoClient(this.mongoClientConfig);
         this._bucketMemState = new BucketMemState(Config);
+
+        // in-mem batch of metrics, we only track total entry count by location
+        // this._accruedMetrics = { zenko-location: 10 }
+        this._accruedMetrics = {};
+
+        setInterval(() => {
+            this._sendMetrics();
+        }, METRIC_REPORT_INTERVAL_MS);
     }
 
     _setupMetricsClients(cb) {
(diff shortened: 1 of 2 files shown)


extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -21,6 +21,8 @@ const { metricsExtension, metricsTypeCompleted, metricsTypePendingOnly } =
 const getContentType = require('./utils/contentTypeHelper');
 const BucketMemState = require('./utils/BucketMemState');
 
+const METRIC_REPORT_TIMEOUT = process.env.CI === 'true' ? 1000 : 5000;
```
jonathan-gramain: Could be more accurately named `METRIC_REPORT_INTERVAL_MS`

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -476,10 +484,25 @@ class MongoQueueProcessor {
         });
     }
 
+    _sendMetrics() {
+        Object.keys(this._accruedMetrics).forEach(loc => {
+            const count = this._accruedMetrics[loc];
+
+            if (count > 0) {
+                this._accruedMetrics[loc] -= count;
```
jonathan-gramain: You can just reset it to 0 here.

jonathan-gramain (commented):
Ideally I would advise to move towards exposing metrics via the Prometheus endpoint, but we can think about this post 1.2.

extensions/mongoProcessor/MongoQueueProcessor.js
philipyoo: since this method is called on a separate schedule from the processing of kafka entries, I wanted to guarantee that I didn't reset possible new metrics accrued

tmacro (approved):

extensions/mongoProcessor/MongoQueueProcessor.js
jonathan-gramain: I was just thinking about the logic of the computation, since you substract `count` which is the original counter value, the only final value you can get is 0. It would be simpler to just reset the counter to 0 then.

jonathan-gramain: The possibility of the counter changing in between would only be valid if different threads could change this value in parallel, but it's not the case (all javascript code executes in a single thread).

philipyoo: ah! yea oops, will make this change, ty 👍

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -476,10 +484,25 @@ class MongoQueueProcessor {
         });
     }
 
+    _sendMetrics() {
+        Object.keys(this._accruedMetrics).forEach(loc => {
+            const count = this._accruedMetrics[loc];
+
+            if (count > 0) {
```
rahulreddy, line 499: Can you add a one liner saying what it means by `count > 0`? Something like `> 0 would mean some metrics were recorded`.

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #709, opened 2019-05-31
bf: ZENKO-1744 ingestion populator performance

philipyoo:
Changes in this PR:
- Each ingestion reader has an in-progress flag. This is to allow each reader to be able to process its next batch of entries on next cron job without waiting on other readers.
- Similar to how LogReader saves log offset in-mem, I did the same for init state for IngestionReader. Read from zookeeper only on startup. We still write to zookeeper each batch processed though unless we are done with snapshot phase
- Reduce kafka producer pollIntervalMs from 2s to 50ms. Ingestion can potentially have many parallel reads and they are all producing to kafka. I want to increase poll intervals to more frequently flush in-mem kafka stats.
- Limit max parallel IngestionReaders from 10 down to 5. To try and reduce pressure on kafka and also in-mem processing.
- Share a single BackbeatProducer instance across readers


bennettbuchanan (approved):

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -17,6 +18,9 @@ const {
     zkStateProperties,
 } = require('../../extensions/ingestion/constants');
 
+const POLL_INTERVAL_MS = 50;
+const MAX_PARALLEL_READERS = 5;
```
jonathan-gramain: Since this seems to be a sensitive configuration value, I suggest to expose it via the config so that it can be tuned if need be.

conf/config.json
```diff
@@ -47,6 +47,7 @@
             "topic": "backbeat-ingestion",
             "zookeeperPath": "/ingestion",
             "cronRule": "*/5 * * * * *",
+            "maxParallelReader": 5,
```
jonathan-gramain: typo, should be `maxParallelReaders`

jonathan-gramain (approved):

philipyoo:
ty @jonathan-gramain @bennettbuchanan !

squashing and will merge

philipyoo:
@bert-e approve

PR #725, opened 2019-06-11
ft: ZENKO-1449 validate ingestion locations

philipyoo:
Validate configs given for a new ingestion location,
validate credentials given, and make sure credentials
have admin access to use internal admin routes (i.e.
backbeat routes)

### DIFF
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -455,10 +455,9 @@ class IngestionPopulator {
 
     /**
      * Apply updates when ingestion buckets are updated in Config
-     * @param {Function} cb - callback(error)
      * @return {undefined}
      */
-    applyUpdates(cb) {
+    applyUpdates() {
         // Example buckets param object:
         // [
         //      {
@@ -475,7 +474,8 @@ class IngestionPopulator {
         const buckets = config.getIngestionBuckets();
 
         const configuredSources = Object.keys(this._ingestionSources);
-        return async.each(buckets, (bucket, next) => {
+
+        buckets.forEach(bucket => {
             const { zenkoBucket } = bucket;
             // if the zenko bucket has already been setup and is already
             // active, stop tracking bucket from `currentActiveSources`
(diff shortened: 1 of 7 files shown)


lib/queuePopulator/IngestionPopulator.js
```diff
@@ -590,14 +590,28 @@ class IngestionPopulator {
     }
 
     _setupUpdatedReaders(done) {
+        const readyReaders = [];
```
philipyoo: small change: rather than using this extra var, just push into `this.logReaders`

lib/queuePopulator/IngestionReader.js
```diff
@@ -445,21 +458,19 @@ class IngestionReader extends LogReader {
      * update relevant instance variables and reinstantiate any clients
```
philipyoo: update docs, no longer reinstantiates any clients

lib/queuePopulator/IngestionReader.js
```diff
@@ -71,22 +71,35 @@ class IngestionReader extends LogReader {
...
 
     setup(done) {
-        this._setupIngestionProducer(err => {
-            if (err) {
-                return done(err);
-            }
-            return this._iProducer.getRaftId(this.bucket, (err, data) => {
+        const retryableError = { retryable: true };
+        async.series([
+            next => this._setupIngestionProducer(next),
```
philipyoo, line 75: could use feedback on the idea of retryable errors

If an error occurs from `setupIngestionProducer`, this would mean the error occurred from mongo itself, so I decided this was an error that should crash pod as it can be a service down

lib/queuePopulator/IngestionReader.js
```diff
@@ -71,22 +71,35 @@ class IngestionReader extends LogReader {
...
                 if (err) {
-                    return done(err);
+                    if (err.code === 'AccessDenied') {
+                        // retryable
+                        this.logger.error('admin credentials are required', {
+                            error: err,
+                        });
+                        return next(retryableError);
+                    }
+                    return next(err);
```
philipyoo, line 80: same here, if an error occurred and it's not an 'AccessDenied' error, I crash the pod.

The potential error here could be that the source location is down and inaccessible.

lib/queuePopulator/IngestionReader.js
philipyoo: updated to log error and avoid crashing pod on any error during setup of IngestionReader

lib/queuePopulator/IngestionProducer.js
```diff
@@ -110,6 +110,56 @@ class IngestionProducer {
...
+            if (err) {
+                const logParams = {
+                    method: 'IngestionProducer.validateRemoteAccess',
+                    remoteBucket: bucket,
+                    zenkoBucket: this._targetZenkoBucket,
+                    error: err,
+                    errorType: err.code,
+                };
+                if (err.code === 'NoSuchBucket') {
+                    logger.error('remote bucket does not exist', logParams);
```
philipyoo: the error messages here aren't really providing any more info than the error itself. I could maybe log potential cause of why the user is facing these specific errors?
Or maybe just collapse everything into a single error log and let the error from aws speak for itself

PR #727, opened 2019-06-20
bugfix: ZENKO-1907 fix kafka backlog check

jonathan-gramain:
Fix kafka backlog check with removed locations or other situations
where nodes become unused (like with partitions waiting to be
consumed).

The fix consists of ignoring (and deleting) nodes that have been
updated since longer than twice the interval of the offset publishing
task. Since those updates normally occur regularly, nodes that did not
get updated are normally obsolete and should be ignored and deleted.

Also removed the check that forces the first batch to be executed
regardless of the backlog, it's now unnecessary since obsolete nodes
will be ignored, and it's better to enforce the check after a restart.

### DIFF
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -65,7 +65,6 @@ class LifecycleConductor {
         this._kafkaBacklogMetrics = null;
         this._started = false;
         this._cronJob = null;
-        this._totalProcessingCycles = 0;
 
         this.logger = new Logger('Backbeat:Lifecycle:Conductor');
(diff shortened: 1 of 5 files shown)


tmacro (approved):

bennettbuchanan (approved):

jonathan-gramain:
I amended the commit to remove the "totalProcessingCycles" check as it became unnecessary and could introduce annoying side-effects after a restart.

jonathan-gramain:
Did another cosmetic change to use a more readable special value than `null` for outdated offsets.

jonathan-gramain:
@bert-e approve

PR #729, opened 2019-07-03
Bugfix/zenko 1952 update putobject

rahulreddy:


tests/unit/replication/QueueEntry.spec.js
```diff
@@ -85,6 +85,7 @@ describe('QueueEntry helper class', () => {
                     dataStoreName: 'file',
                     dataStoreETag: '1:0e5a6f42662652d44fcf978399ef5709',
                     dataStoreVersionId: 'version1',
+                    blockId: undefined,
```
JianqinWang, line 88: when we don't explicitly set `blockId`, isn't it by default `undefined`?

anurag4DSB (approved):

jonathan-gramain (approved):

tests/unit/replication/QueueEntry.spec.js
rahulreddy: So unfortunately, it's all or nothing when you add a property to an Object. It's either set or not, if you set it with `undefined` it will still show up in the skeleton. I am trying to find ways to mitigate it but for now, I don't have much of an option.

rahulreddy:
@bert-e approve

PR #776, opened 2019-12-02
ZENKO-2239 post merge deploy to new registry

tcarmet:


JianqinWang (approved):

eve/main.yml
```diff
@@ -9,16 +9,37 @@ branches:
...
-  - SetProperty: &docker_image_name
-      name: Set docker image name property
-      property: docker_image_name
-      value:
-        "%(secret:private_registry_url)s/zenko/backbeat:\
-        %(prop:commit_short_revision)s"
+  - SetProperty: &registry_url
+      name: registry
+      property: registry_url
+      value: registry.scality.com/zenko-dev/backbeat
```
rahulreddy, line 15: Is there a reason you chose `zenko-dev` as the registry name and not just `zenko`?

eve/workers/release/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ARG DOCKER_VERSION=18.06.1~ce~3-0~ubuntu
+
+RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - \
+ && echo "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" \
+    | tee -a /etc/apt/sources.list.d/docker-ce.list \
+ && apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    docker-ce=${DOCKER_VERSION} \
+ && rm -rf /var/lib/apt/lists/* /var/cache/apt
+
+USER eve
```
rahulreddy: Please add a new line here

eve/workers/release/pod.yaml
```diff
@@ -0,0 +1,45 @@
...
+        mountPath: /home/eve/workspace
+    - name: dind-daemon
+      image: docker:18.06.1-dind
+      resources:
+        requests:
+          cpu: "500m"
+          memory: 1Gi
+        limits:
+          cpu: "1"
+          memory: 1Gi
```
rahulreddy, line 33: It was `2Gi` before, would we running into memory limits?

rahulreddy (approved):

eve/main.yml
JianqinWang: https://registry.scality.com/harbor/projects/7/repositories I think it's because that's how we have it set up here?

tcarmet: Well, my idea/state of mind, for now, is that, we're still playing with harbor. The images that we're publishing in our post-merge are for dev purposes. Therefore I don't want to give the idea of "this is production ready, please use it".

At some point we will publish the tagged and released images on harbor, and those will be on the registry.scality.com/zenko 

This is just an idea, and of course I'm open to any suggestion.

eve/workers/release/Dockerfile
tcarmet: woops

tcarmet: done

eve/workers/release/pod.yaml
tcarmet: I checked the metrics, and it is more than enough. The pods do consume more CPU then ram, but even that, the amount given is more than ok.

eve/main.yml
rahulreddy: It works for now. I had the idea that we can have nightlies/weeklies (tagged appropriately with the label nightly/weekly) in the same repo. Something like
`registry.scality.com/zenko/backbeat/nightly/nightly-20191211`, `registry.scality.com/zenko/backbeat/weekly/weekly-20191211` and `registry.scality.com/zenko/backbeat/ga/8.11.2` etc. Let me know what you think.

PR #785, opened 2020-01-09
improvement: ZENKO-2265 make ingestion not wait between batches

jonathan-gramain:
Instead of inconditionally triggering ingestion batches every 5
seconds, which becomes inefficient as the batch size gets small and
quick to process, do the following:

- check if the batch just processed was at the end of the log
- if so, trigger a new batch immediately

This way, we can process batches in series without waiting and
converge quicker when processing an existing raft log.


dora-korpar (approved):
lgtm :)

rahulreddy (approved):

jonathan-gramain:
@bert-e approve

PR #789, opened 2020-01-24
improvement:ZENKO-2262 Start OOB in paused state

naren-rajendran:
### Why is this change required? What problem does it solve?
**Problem**: OOB S3C/AWS is used in the context of D/R to a storage location which requires setting up replication to the  location. OOB starts ingesting objects as soon as it is setup and by the time replication is setup there a few objects that are ingested that will not be replicated (since replication only queues objects after the replication workflow is enabled). This shows the gap in the workflow where some objects are not replicated

**Solution**: To address the gap, start OOB in paused state by default when it is setup. This way the user can setup replication workflow and then enable OOB, there by having the control on when the objects are ingested.

### DIFF
--- a/bin/ingestion.js
+++ b/bin/ingestion.js
@@ -141,7 +141,8 @@ function checkAndApplyScheduleResume(zkClient, data, location, cb) {
  */
 function setupZkLocationNode(zkClient, location, done) {
     const path = `${getIngestionZkPath()}/${location}`;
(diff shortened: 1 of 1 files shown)


rahulreddy (approved):

naren-rajendran:
@bert-e reset

alexanderchan-scality (approved):

naren-rajendran:
@bert-e approve

PR #794, opened 2020-02-19
bugfix: S3C-2646 fix flaky tests caused by MultipleBackendTask unit

jonathan-gramain:
Some flaky tests occur because an exception is raised asynchronously
from the MultipleBackendTask unit tests where retry behavior is
checked.

The real fix is to always expect an error returned from
_getAndPutMultipartUpload to avoid an exception to be
raised. Additionally:

- improve the test coverage by catching cases where a retryable error
  would immediately return an error without retrying

- create a new instance of MultipleBackendTask for each test and
  inhibit further retries when the test completes to avoid the process
  being blocked by the lingering pending retries.

### DIFF
--- a/tests/unit/replication/MultipleBackendTask.js
+++ b/tests/unit/replication/MultipleBackendTask.js
@@ -7,17 +7,21 @@ const MultipleBackendTask =
 const log = require('../../utils/fakeLogger');
 const { sourceEntry, destEntry } = require('../../utils/mockEntries');
 
-const multipleBackendTask = new MultipleBackendTask({
-    getStateVars: () => ({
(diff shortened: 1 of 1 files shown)


tmacro (approved):

jonathan-gramain:
Actually, I'm realizing the test was not and still is not properly testing the retry case. If there was no retry, the test would pass. I'll fix this and push again to this PR.

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@bert-e approve

alexanderchan-scality (approved):

PR #579, opened 2018-12-20
bugfix: ZENKO-1404 Lifecycle replication status

bennettbuchanan:
We do not want to update the replication site status or replication status to 'REPLICA' for lifecycle transition entries being pushed to the replication status processor.

PR #824, opened 2020-04-08
bugfix: ZENKO-2447 pause resume corrections

naren-rajendran:
#### Why is this change required? What problem does it solve?

OOB ingestion pause via orbit did not pause the readers. Corrections are made to set readers to pause state when paused via orbit.

### DIFF
--- a/bin/ingestion.js
+++ b/bin/ingestion.js
@@ -25,8 +25,10 @@ const { connectionString, autoCreateNamespace } = zkConfig;
 const RESUME_NODE = 'scheduledResume';
 
 const log = new werelogs.Logger('Backbeat:IngestionPopulator');
-werelogs.configure({ level: config.log.logLevel,
-    dump: config.log.dumpLevel });
+werelogs.configure({
+    level: config.log.logLevel,
+    dump: config.log.dumpLevel
+});
 
 const healthServer = new HealthProbeServer({
     bindAddress: config.healthcheckServer.bindAddress,
@@ -38,35 +40,12 @@ let ingestionPopulator;
 // memoize
 const configuredLocations = {};
 
+const zkClient = zookeeperWrapper.createClient(connectionString, {
+    autoCreateNamespace,
+});
+
 function getIngestionZkPath() {
     return `${zookeeperNamespace}${zkStatePath}`;
-}
-
-function queueBatch(ingestionPopulator, log) {
-    log.debug('start queueing ingestion batch');
-    const maxRead = qpConfig.batchMaxRead;
-    // apply updates to Ingestion Readers
-    ingestionPopulator.applyUpdates(err => {
-        if (err) {
-            log.fatal('error occurred applying updates to ingestion readers', {
-                error: err,
-                method: 'bin::ingestion.queueBatch',
-            });
-            scheduler.cancel();
-            process.exit(1);
-        }
-        ingestionPopulator.processLogEntries({ maxRead }, err => {
-            if (err) {
-                log.fatal('an error occurred during ingestion', {
-                    method: 'bin::ingestion.queueBatch',
-                    error: err,
-                });
-                scheduler.cancel();
-                process.exit(1);
-            }
-        });
-    });
-    return undefined;
 }
 
 /**
@@ -148,14 +127,16 @@ function setupZkLocationNode(zkClient, location, done) {
             return zkClient.getData(path, (err, data) => {
                 if (err) {
                     log.fatal('could not check location status in zookeeper',
-                        { method: 'bin::ingestion.setupZkLocationNode',
-                          zookeeperPath: path,
-                          error: err.message });
+                        {
+                            method: 'bin::ingestion.setupZkLocationNode',
+                            zookeeperPath: path,
+                            error: err.message,
+                        });
                     return done(err);
                 }
-                let d;
+                let locationData;
                 try {
-                    d = JSON.parse(data.toString());
+                    locationData = JSON.parse(data.toString());
                 } catch (e) {
                     log.fatal('error setting state for queue processor', {
                         method: 'bin::ingestion.setupZkLocationNode',
@@ -164,11 +145,11 @@ function setupZkLocationNode(zkClient, location, done) {
                     });
                     return done(e);
                 }
-                if (d[RESUME_NODE]) {
-                    return checkAndApplyScheduleResume(zkClient, d, location,
-                        done);
+                if (locationData[RESUME_NODE]) {
+                    return checkAndApplyScheduleResume(zkClient, locationData,
+                        location, done);
                 }
-                if (data.paused === true) {
+                if (locationData.paused) {
                     ingestionPopulator.setPausedLocationState(location);
                 }
                 return done();
@@ -191,7 +172,7 @@ function setupZkLocationNode(zkClient, location, done) {
 function updateProcessors(zkClient, bootstrapList) {
     const active = Object.keys(configuredLocations);
     const update = bootstrapList.map(i => i.site);
-    const allLocations = [...new Set(active.concat(update))];
+    const allLocations = new Set(active.concat(update));
 
     async.each(allLocations, (location, next) => {
         if (!update.includes(location)) {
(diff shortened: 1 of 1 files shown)


bin/ingestion.js
```diff
@@ -225,6 +202,61 @@ function updateProcessors(zkClient, bootstrapList) {
...
+ * @param {node-zookeeper-client.Client} zkClient - zookeeper client
+ * @param {Logger.newRequestLogger} log - request logger object
+ * @param {Function} cb - callback(error)
+ * @return {undefined}
+ */
+function refreshProcessors(zkClient, log, cb) {
+    const bootstrapList = config.getBootstrapList();
+    const active = Object.keys(configuredLocations);
+    const update = bootstrapList.map(i => i.site);
+    const locations = [...new Set(active.concat(update))];
```
miniscruff: I do not think you need to turn this back into a list, async.each says it supports `Iterable` which Set is.

bin/ingestion.js
```diff
@@ -225,6 +202,61 @@ function updateProcessors(zkClient, bootstrapList) {
...
+                    method: 'bin::ingestion.refreshProcessors',
+                    location,
+                    error: reshapeExceptionError(e),
+                });
+                return next(e);
+            }
+            if (d[RESUME_NODE]) {
+                return checkAndApplyScheduleResume(zkClient, d, location, next);
+            }
+            if (d.paused === true) {
```
miniscruff: can just be `if (d.paused)`

bin/ingestion.js
```diff
@@ -225,6 +202,61 @@ function updateProcessors(zkClient, bootstrapList) {
...
+            if (err) {
+                log.fatal('could not check location status in zookeeper',
+                    { method: 'bin::ingestion.refreshProcessors',
+                      zookeeperPath: path,
+                      error: err.message });
+                return next(err);
+            }
+            let d;
+            try {
+                d = JSON.parse(data.toString());
```
miniscruff: Might want to consider a more descriptive variable name over just `d`

bin/ingestion.js
```diff
@@ -225,6 +202,61 @@ function updateProcessors(zkClient, bootstrapList) {
...
+    const update = bootstrapList.map(i => i.site);
+    const locations = [...new Set(active.concat(update))];
+    const ingestionZkPath = getIngestionZkPath();
+
+    return async.each(locations, (location, next) => {
+        const path = `${ingestionZkPath}/${location}`;
+        return zkClient.getData(path, (err, data) => {
+            if (err) {
+                log.fatal('could not check location status in zookeeper',
+                    { method: 'bin::ingestion.refreshProcessors',
```
miniscruff: Doesn't match the other formatting where the opening curly brace is on the previous line and the end curly is on a line by itself.

miniscruff:
Still new here, but this did not include any tests, not sure if I should mention that but yea.

bin/ingestion.js
naren-rajendran: Ah..my format plugin did not work in the editor..

bin/ingestion.js
naren-rajendran: we use async v2.3, I dont think `async.each` supports `Iterable` in v2.3.

bin/ingestion.js
naren-rajendran: copy-paste!

bin/ingestion.js
naren-rajendran: copy-paste!

bin/ingestion.js
naren-rajendran: Fixed it. Also, it wont be uniform at all times because max line length of 80 is followed.

naren-rajendran:
@bert-e reset

bin/ingestion.js
miniscruff: https://github.com/caolan/async/blob/master/CHANGELOG.md#new-features-1 seems like it was added in 2.0 unless I'm in the complete wrong package.

miniscruff: Specifically this comment:
Collection methods now accept ES2015 iterators. Maps, Sets, and anything that implements the iterator spec can now be passed directly to each, map, parallel, etc.. (#579, #839, #1074)

bin/ingestion.js
```diff
@@ -164,11 +145,11 @@ function setupZkLocationNode(zkClient, location, done) {
...
                 }
-                if (d[RESUME_NODE]) {
-                    return checkAndApplyScheduleResume(zkClient, d, location,
-                        done);
+                if (locationData[RESUME_NODE]) {
+                    return checkAndApplyScheduleResume(zkClient, locationData,
+                        location, done);
                 }
-                if (data.paused === true) {
+                if (data.paused) {
```
miniscruff: should this data be locationData?

bin/ingestion.js
miniscruff: Gotcha, makes sense. I like line limits ( although 80 is a bit low with 4 space tabs imo but meh ).

bin/ingestion.js
naren-rajendran: Ah..good catch.. I was redirected to 1.6 somehow..

bin/ingestion.js
naren-rajendran: Well `data` is converted to string, parsed as json again and assigned to `locationData`. Not sure why this was done, but going to keep and for sanity, changing it to `locationData`.

bin/ingestion.js
naren-rajendran: Went back in time and checked, v2.3 has been in use since `async` was added as a dependency. So it should not be an issue then.

bin/ingestion.js
miniscruff: Seems like zookeeper ( according to docs ) passes data as a raw Buffer, but that it is probably JSON data so that explains the toString + JSON.parse. Meaning data.paused would never evaluate to anything as a byte buffer has no paused attribute.

naren-rajendran: No, I meant in the original implementation, it was `data.paused`.

miniscruff: oh... that doesnt sound good. Are there specs or docs related to what the data comes back as? Data is defined by zookeeper as a Buffer, but what is actually in the buffer is defined somewhere else. Hard to say if this is correct without it. It could be isPaused or wasPaused for all I know.

naren-rajendran:
@bert-e  reset

naren-rajendran:
> Still new here, but this did not include any tests, not sure if I should mention that but yea.

Waiting for others to comment, as there could be a better way to handle this. I will add tests as needed.

bin/ingestion.js
naren-rajendran: well I corrected it.

bin/ingestion.js
```diff
@@ -248,6 +287,49 @@ function loadProcessors(zkClient) {
...
+        }),
+        next => ingestionPopulator.applyUpdates(err => {
+            if (err) {
+                log.fatal('error applying updates to ingestion readers', {
+                    error: err,
+                    method: 'bin::ingestion.queueBatch',
+                });
+                return next(new Error('apply updates failed'));
+            }
+            return ingestionPopulator.processLogEntries({ maxRead }, err => {
```
jonathan-gramain: minor: you could have this call in another step of the `async.series` block.

bin/ingestion.js
```diff
@@ -225,6 +206,64 @@ function updateProcessors(zkClient, bootstrapList) {
...
+                    location,
+                    error: reshapeExceptionError(e),
+                });
+                return next(e);
+            }
+            if (locationData[RESUME_NODE]) {
+                return checkAndApplyScheduleResume(zkClient, locationData,
+                    location, next);
+            }
+            if (locationData.paused) {
```
jonathan-gramain, line 251: I might be wrong, but I'm under the impression that to be complete, we should also call `ingestionPopulator.removePausedLocationState()` if the location is not paused.

naren-rajendran:
@bert-e reset

jonathan-gramain (approved):

tmacro (approved):

naren-rajendran:
@bert-e approve

rahulreddy:
@naren-scality Check @miniscruff 's comments. There was no test included with this bugfix, we should always have a two commits when we do a bugfix. The first commit which adds a test capturing the essence of the bug, which will fail and then a second commit with the fix showing that the test passes.

PR #845, opened 2020-06-06
bugfix: S3C-2899 strip vformat v1 key prefix in queue populator

jonathan-gramain:
Return entries with keys stripped of their v1 vformat key prefix to
the queue populator extensions

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -1,5 +1,8 @@
 const async = require('async');
 
+const { versioning } = require('arsenal');
+const { DbPrefixes } = versioning.VersioningConstants;
+
 const BackbeatProducer = require('../BackbeatProducer');
 const ReplicationQueuePopulator =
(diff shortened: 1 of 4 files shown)


miniscruff (approved):

naren-rajendran (approved):

lib/queuePopulator/LogReader.js
```diff
@@ -278,13 +281,22 @@ class LogReader {
     }
 
     _processLogEntry(batchState, record, entry) {
+        function _transformKey(key) {
```
naren-rajendran, line 284: Can `_transformKey` be used anywhere else? Should it be a util?

jonathan-gramain: I think it's too context-specific to deserve a generic function here, as backbeat relies on the version ID separator being in the key, I don't see any other possible use of this function. Also it's short so would not add much duplication if done elsewhere.

jonathan-gramain:
@bert-e approve

PR #894, opened 2020-07-14
improvement: S3C-2837 gather and log kafka consumer lag metrics

jonathan-gramain:
With an option to the BackbeatConsumer instance, periodically log the
total consumer lag, as the sum of the individual partitions lag as
returned by librdkafka in its statistics event.

It's not clear what is the behavior with multiple consumer groups, or
if the consumer instance retrieving the statistics has an influence on
the results. It looks to be returning the lag correctly for a single
consumer group and that is the case for the backbeat-replication topic
in S3C.

PR note: there is no change on Zenko branches, because Prometheus already provides lag metrics and is a more flexible approach for metrics.

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -50,6 +50,7 @@ const joiSchema = {
         groupId: joi.string().required(),
         retryTimeoutS: joi.number().default(300),
         concurrency: joi.number().greater(0).default(10),
+        logConsumerMetricsIntervalS: joi.number().greater(0).default(60),
     },
     replicationStatusProcessor: {
         groupId: joi.string().required(),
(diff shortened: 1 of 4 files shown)


jonathan-gramain:
@bert-e reset

rahulreddy (approved):
I think a Federation PR will also be nice that sets this config value so that it can be changed.

nicolas2bert (approved):

naren-rajendran (approved):

ilkescality (approved):

jonathan-gramain:
I added a check on "fetch_state" in the stats because the logs would otherwise show a lag for partitions that are not consumed anymore, making the aggregation of stats across consumers tricky. Now it only shows stats for partitions that are actively consumed.

PR #941, opened 2020-08-01
ft: S3C-3193 bucket notification config filter

naren-rajendran:
Validation and filter component for bucket notification populator and processor.

### DIFF
--- a/extensions/notification/utils/config.js
+++ b/extensions/notification/utils/config.js
@@ -1,0 +1,88 @@
+const constants = require('../constants');
+
+/**
+ * Transforms array of configurations to a Map
+ * @param  {Object[]} bnConfigs - Array of bucket notification configurations
+ * @return {Map} Map of bucket notification configurations
+ */
+function configArrayToMap(bnConfigs) {
+    const configMap = new Map();
+    bnConfigs.forEach(config => {
+        if (config.bucket && config.notificationConfiguration) {
+            const bucket = config.bucket;
+            const conf = config.notificationConfiguration.queueConfigurations;
+            if (configMap.has(bucket)) {
+                const val = configMap.get(bucket);
+                configMap.set(bucket, [...val, ...conf]);
+            } else {
+                configMap.set(bucket, [...conf]);
+            }
+        }
+    });
+    return configMap;
+}
+
+/**
+ * Validate object bucket notification configuration filters
+ * @param  {Object[]} filterRules - Filter rules from the bucket notification
+ * configuration.
+ * @param  {Object} entry - An entry from the log
+ * @return {boolean} true if filter matches the object name
+ */
+function validateEntryWithFilter(filterRules, entry) {
+    const { key } = entry;
+    return filterRules.every(rule => {
+        if (rule.name === constants.nameFilter.prefix) {
+            return key.startsWith(rule.value);
+        }
+        if (rule.name === constants.nameFilter.suffix) {
+            return key.endsWith(rule.value);
+        }
+        return false;
+    });
+}
+
+/**
+ * Validate event against bucket notification configuration event type
+ * @param  {Object[]} bnConfigs - Array of bucket specific notification
+ * configurations
+ * @param  {string} event - Type of event
+ * @return {undefined|Object} A bucket notification configuration if found
+ */
+function validateConfigsEvent(bnConfigs, event) {
+    const config = bnConfigs.filter(config => config.events.includes(event));
+    return config.length > 0 ? config[0] : undefined;
+}
+
+/**
+ * Validates an entry from log against bucket notification configurations to see
+ * if the entry has to be published. Validations include, bucket specific
+ * configuration check, event type check, object name specific filter checks
+ * @param  {Object[]} bnConfigs - Array of bucket notifications configuration
+ * @param  {Object} entry - An entry from the log
+ * @return {boolean} true if event qualifies for notification
+ */
+function validateEntry(bnConfigs, entry) {
+    const { bucket, type } = entry;
+    // check if the entry belongs to any bucket in the configuration
+    if (bnConfigs.has(bucket)) {
+        const bnConfig = bnConfigs.get(bucket);
+        // check if the event type matches
+        const qConfig = validateConfigsEvent(bnConfig, type);
+        if (qConfig) {
+            // check if there are filters and the object matches it
+            const configFilter = qConfig.filter;
+            if (configFilter && configFilter.filterRules) {
+                return validateEntryWithFilter(configFilter.filterRules, entry);
+            }
+            return true;
+        }
+        return false;
+    }
+    return false;
+}
+
+module.exports = {
+    configArrayToMap,
+    validateEntry,
+};
--- a/tests/unit/notification/utils/config.js
+++ b/tests/unit/notification/utils/config.js
@@ -1,0 +1,222 @@
+const assert = require('assert');
+const notifConfUtil
+    = require('../../../../extensions/notification/utils/config');
+
+const testConfigs = [
+    {
+        bucket: 'bucket1',
+        notificationConfiguration: {
+            queueConfigurations: [
+                {
+                    events: ['s3:ObjectCreated:Put'],
+                    queueArn: 'q1',
+                    filter: {
+                        filterRules: [
+                            {
+                                name: 'prefix',
+                                value: 'test',
+                            },
+                        ],
+                    },
+                    id: 'config1',
+                },
+            ],
+        },
+    },
+    {
+        bucket: 'bucket2',
+        notificationConfiguration: {
+            queueConfigurations: [
+                {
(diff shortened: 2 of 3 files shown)


extensions/notification/utils/config.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Validate event against bucket notification configuration event type
+ * @param  {Object[]} bnConfig - Array of bucket specific notification
+ * configurations
+ * @param  {string} event - Type of event
+ * @return {undefined|Object} A bucket notification configuration if found
+ */
+function validateEvent(bnConfig, event) {
+    if (bnConfig.length) {
```
tmacro: If `bnConfig` is always an Array is the check for length necessary? The `0` case would cause `.filter` to return an empty Array which is already handled by the ternary;

naren-rajendran: Ah..yes..forgot to remove this when I refactored.

tmacro (approved):

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+ * Transforms array of configurations to a Map
+ * @param  {Object[]} configs - Array of bucket notification configurations
+ * @return {Map} Map of bucket notification configurations
+ */
+function transform(configs) {
+    const configMap = new Map();
+    configs.forEach(config => {
+        if (config.bucket && config.notificationConfiguration) {
+            const bucket = config.bucket;
+            const conf = config.notificationConfiguration.queueConfigurations;
```
jonathan-gramain, line 13: similarly, `conf` seems a too generic name, `queueConf` could be more precise

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
+const constants = require('../constants');
+
+/**
+ * Transforms array of configurations to a Map
+ * @param  {Object[]} configs - Array of bucket notification configurations
+ * @return {Map} Map of bucket notification configurations
+ */
+function transform(configs) {
```
jonathan-gramain: `transform` seems too generic, the function name should describe somehow what the transform is about, like `configArrayToMap` for example

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Validate object name against bucket notification configuration filters
+ * @param  {Object[]} filterRules - Filter rules from the bucket notification
+ * configuration.
+ * @param  {string} objectName - Name of the object
+ * @return {boolean} true if filter matches the object name
+ */
+function validateFilter(filterRules, objectName) {
```
jonathan-gramain: A filter might have more than filtering by object name, right? If so I'd suggest to pass an object containing various attributes to filter against, even if only `{ objectName }` could be supported at first.

As a side note, we tend to use `objectKey` rather than `objectName` through the product, I'd recommend to use `objectKey` instead.

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+    return filterRules.some(rule => {
+        const name = rule.name;
```
jonathan-gramain: If you wish to use shorthands, you can also use the more compact form `const { name, value } = rule;`, but in that case I'm not convinced it helps readability, you could also directly use `rule.name` and `rule.value` to make it explicit.

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+        const value = rule.value;
+        if (name === constants.nameFilter.prefix) {
+            return objectName.startsWith(value);
+        }
+        if (name === constants.nameFilter.suffix) {
```
jonathan-gramain: Is it allowed to have both `prefix` and `suffix` in the same rule? If so, you would need to change the logic to test each in turn. To accept as many criteria as needed you can filter out by negating the test, so e..g `if (!objectName.startsWith(value)) { return false; }`

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Validate event against bucket notification configuration event type
+ * @param  {Object[]} bnConfig - Array of bucket specific notification
+ * configurations
+ * @param  {string} event - Type of event
+ * @return {undefined|Object} A bucket notification configuration if found
+ */
+function validateEvent(bnConfig, event) {
+    const config = bnConfig.filter(config => config.events.includes(event));
+    return config.length ? config[0] : undefined;
```
jonathan-gramain: I suggest an explicit check on the length rather than using the value as if it was a boolean: `config.length > 0`

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Validate event against bucket notification configuration event type
+ * @param  {Object[]} bnConfig - Array of bucket specific notification
+ * configurations
+ * @param  {string} event - Type of event
+ * @return {undefined|Object} A bucket notification configuration if found
+ */
+function validateEvent(bnConfig, event) {
+    const config = bnConfig.filter(config => config.events.includes(event));
```
jonathan-gramain: the name looks too generic here too, what about `configsWithEvent`

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+ * if the entry has to be published. Validations include, bucket specific
+ * configuration check, event type check, object name specific filter checks
+ * @param  {Object[]} bnConfigs - Array of bucket notifications configuration
+ * @param  {string} bucket - Bucket name
+ * @param  {string} key - Object name
+ * @param  {string} event - Event type
+ * @return {boolean} true if event qualifies for notification
+ */
+function validateEntry(bnConfigs, bucket, key, event) {
+    // check if the entry belongs to any bucket in the configuration!
```
jonathan-gramain: no need for `!` in comments :slightly_smiling_face:

extensions/notification/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+    if (bnConfigs.has(bucket)) {
+        const bnConfig = bnConfigs.get(bucket);
+        // check if the event type matches!
+        const qConfig = validateEvent(bnConfig, event);
+        if (qConfig) {
```
jonathan-gramain, line 72: to reduce the indentation level a bit you could reverse with `if (!qConfig) { return false }; ...`

tests/unit/notification/utils/config.js
```diff
@@ -0,0 +1,177 @@
...
+            });
+        });
+    });
+
+    describe('ValidateEntry', () => {
+        const configMap = notifConfigUtil.transform(testConfigs);
+
+        tests.forEach(test => {
+            it(`should ${test.desc}`, () => {
+                const bckt = test.entry.bucket;
```
jonathan-gramain: why not simply: `const { bucket, key, type } = test.entry;`

extensions/notification/utils/config.js
naren-rajendran: I kept it simple because these functions are part of the utility for configuration, will change it though.

extensions/notification/utils/config.js
naren-rajendran: I kept this simple, it was easier to read.

extensions/notification/utils/config.js
naren-rajendran: Yes, could be. I originally had passed the entire object, but wanted to keep the function simple and changed it.

extensions/notification/utils/config.js
naren-rajendran: Yes, both conditions are allowed in a configuration.

tests/unit/notification/utils/config.js
naren-rajendran: Was struggling to fit the variable names with column restrictions. Changed it.

extensions/notification/utils/config.js
jonathan-gramain: It's not about simplicity but readability :slightly_smiling_face:

extensions/notification/utils/config.js
jonathan-gramain: OK, up to you it's a minor point

tests/unit/notification/utils/config.js
jonathan-gramain: Don't hesitate to get rid of the current drastic column restriction by updating the dependency to eslint-config-scality, it will definitely help in the long run to avoid reducing readability to comply with the linter rules.

extensions/notification/utils/config.js
jonathan-gramain: You resolved this but don't see a change, did you intend to change it but forgot?

jonathan-gramain (approved):

naren-rajendran:
@bert-e approve

naren-rajendran:
@bert-e reset

PR #77, opened 2017-08-25
ft: add a readme section for setup script

philipyoo:
lmk if I should expand upon anything or remove things

hopefully, the credentials and mention about path to script wasn't too confusing

PR #980, opened 2020-08-20
bf: S3C-3277 notification filter corrections

naren-rajendran:
Config filter updated based on corrections to bucket notification configuration and tests for the config util are also updated based on the updated notification configuration.

### DIFF
--- a/extensions/notification/utils/config.js
+++ b/extensions/notification/utils/config.js
@@ -10,7 +10,7 @@ function configArrayToMap(bnConfigs) {
     bnConfigs.forEach(config => {
         if (config.bucket && config.notificationConfiguration) {
             const bucket = config.bucket;
-            const conf = config.notificationConfiguration.queueConfigurations;
+            const conf = config.notificationConfiguration.queueConfig;
             if (configMap.has(bucket)) {
                 const val = configMap.get(bucket);
                 configMap.set(bucket, [...val, ...conf]);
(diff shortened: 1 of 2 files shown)


ilkescality (approved):

ilkescality:
Seems like a backbeat test failing, but guess it might just be a flaky..

extensions/notification/utils/config.js
```diff
@@ -43,42 +43,45 @@ function validateEntryWithFilter(filterRules, entry) {
...
+    if (bucket !== bnConfig.bucket) {
         return false;
     }
+    // check if the event type matches
+    const qConfigs = validateConfigsEvent(notifConf.queueConfig, type);
+    if (qConfigs) {
+        // check if there are filters for any config and the object's key
+        // satisfies any filter
+        const qFilters
+            = qConfigs.filter(c => c.filterRules && c.filterRules.length > 0);
```
ilkescality, line 75: `c.filterRules.length > 0` seemed it might be redundant and just the `c.filterRules` might be sufficient alone but I might just be wrong.

naren-rajendran: Yeah, the filter rules are optional when you set bucket notification configuration, may be just checking the length would be correct here, I will change this.

tests/unit/notification/utils/config.js
```diff
@@ -210,11 +190,29 @@ describe('Notification configuration util', () => {
     });
 
     describe('ValidateEntry', () => {
-        const configMap = notifConfUtil.configArrayToMap(testConfigs);
+        let configMap = null;
+
+        function getBucketNotifConfig(bucket, bnConfigMap) {
+            const bnConfigs
```
jonathan-gramain, line 236: newline looks unneeded

extensions/notification/utils/config.js
```diff
@@ -43,42 +43,45 @@ function validateEntryWithFilter(filterRules, entry) {
...
  * configurations
  * @param  {string} event - Type of event
- * @return {undefined|Object} A bucket notification configuration if found
+ * @return {undefined|Object[]} Bucket queue configurations if found
  */
 function validateConfigsEvent(bnConfigs, event) {
-    const config = bnConfigs.filter(config => config.events.includes(event));
-    return config.length > 0 ? config[0] : undefined;
+    const configs = bnConfigs.filter(config => config.events.includes(event));
+    return configs.length > 0 ? configs : undefined;
```
jonathan-gramain: Maybe it's not needed to specialize the case where `configs` is an empty array, as it should be naturally handled by the rest of the code even if empty (to simplify the flow).

A minor naming comment: the function could be named `filterConfigsByEvent` for example to reflect the fact that it's a filter.

extensions/notification/utils/config.js
```diff
@@ -43,42 +43,45 @@ function validateEntryWithFilter(filterRules, entry) {
...
+    // check if the event type matches
+    const qConfigs = validateConfigsEvent(notifConf.queueConfig, type);
+    if (qConfigs) {
+        // check if there are filters for any config and the object's key
+        // satisfies any filter
+        const qFilters
+            = qConfigs.filter(c => c.filterRules && c.filterRules.length > 0);
+        if (qFilters.length > 0) {
+            return qConfigs.some(
+                c => validateEntryWithFilter(c.filterRules, entry));
```
jonathan-gramain: I don't think the logic is right as `c.filterRules` could be undefined here (does not seem handled by `validateEntryWithFilter`).

```suggestion
                return qConfigs.some(c.filterRules && c => validateEntryWithFilter(c.filterRules, entry));
```
And remove the previous `qFilter = qConfig...` + `if (qFilters.length > 0)` checks.

jonathan-gramain (changes requested):

extensions/notification/utils/config.js
naren-rajendran: The check here is to see if a bucket notification configuration has filters (filters are optional), if they don't (which is checked by `if (qFilters.length > 0)` ) `validateEntryWithFilter` will be skipped.

jonathan-gramain: What if you have `qConfig` contain two entries like:

```
[{ filterRules: [{ some rule }] }, { /* no filterRules here */ }]
```

I may be wrong but it seems to be this would lead to `qFilters.length > 0`, then `validateEntryWithFilter` would raise an exception on the second entry because there is no `filterRules` in the second entry.

naren-rajendran: `qFilters` will contain only configs with filters in them, ref line #77, so `validateEntryWithFilter` would not be validating a config without filters.

jonathan-gramain: But you're doing `qConfigs.some` not `qFilters.some`, maybe that is a typo then?

jonathan-gramain: In any case, make sure this situation is covered in unit tests.

naren-rajendran: New unit tests added to cover the condition discussed.

extensions/notification/utils/config.js
```diff
@@ -43,41 +43,42 @@ function validateEntryWithFilter(filterRules, entry) {
...
+    if (bucket !== bnConfig.bucket) {
+        return false;
+    }
+    // check if the event type matches
+    const qConfigs = filterConfigsByEvent(notifConf.queueConfig, type);
+    if (qConfigs.length > 0) {
+        const qConfigWithFilters
+            = qConfigs.filter(c => c.filterRules && c.filterRules.length > 0);
+        // if there are configs without filters, make the entry valid
+        if (qConfigs.length - qConfigWithFilters.length > 0) {
```
jonathan-gramain: Could also be written as `if (qConfigs.length > qConfigWithFilters.length)`

jonathan-gramain (approved):

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e approve

PR #1004, opened 2020-09-02
bf: S3C-3304 entry process corrections

naren-rajendran:
Bucket notifications queue populator fails to process bucket attributes change entry and object removed entry.

Corrections are made to log reader and queue populator  to process bucket attributes change and entries satisfying 'S3:ObjectRemoved:Delete' events.

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -211,11 +211,13 @@ class LogReader {
             },
             entriesToPublish: {},
             publishedEntries: {},
+            currentRecord: {},
         };
 
         async.waterfall([
             next => this._processReadRecords(params, batchState, next),
             next => this._processPrepareEntries(batchState, next),
+            next => this._processFilterEntries(batchState, next),
             next => this._processPublishEntries(batchState, next),
             next => this._processSaveLogOffset(batchState, next),
         ],
@@ -305,7 +307,7 @@ class LogReader {
         });
     }
 
-    _processLogEntry(batchState, record, entry) {
+    _processLogEntry(batchState, record, entry, cb) {
         function _transformKey(key) {
             if (!key) {
                 return undefined;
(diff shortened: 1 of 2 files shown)


lib/queuePopulator/LogReader.js
```diff
@@ -322,6 +307,9 @@ class LogReader {
 
     _processLogEntry(batchState, record, entry) {
         function _transformKey(key) {
+            if (!key) {
+                return key;
```
jonathan-gramain: I think it's more straightforward to `return null` or `return ''` depending on the convention used

lib/queuePopulator/LogReader.js
```diff
@@ -330,15 +318,29 @@ class LogReader {
             }
             return key;
         }
-        if (entry.key === undefined || entry.value === undefined) {
-            return;
+        // TODO: handle this better, this is too specific to notifications
```
jonathan-gramain: :+1: 

If you have time let's just fix it now, I think it's only a matter of doing the same special processing in the filter function of BN itself correct?

jonathan-gramain: Ah I think I see what you mean, because in the next PR you intend to add the `cb` arg to the filter function for BN? That's fine for me, but don't refrain from being more specific in the TODO then.

dora-korpar (approved):
lgtm

lib/queuePopulator/LogReader.js
```diff
@@ -330,15 +318,29 @@ class LogReader {
...
+        // temp fix
+        const notifExtension = this._extensions.find(ext => (
+            ext instanceof NotificationQueuePopulator
+        ));
+        if (notifExtension) {
+            notifExtension.filter({
```
jonathan-gramain: This section looks outdated now, you should merge the latest changes there.

jonathan-gramain (changes requested):

lib/queuePopulator/LogReader.js
naren-rajendran: This is the base pr for the other pr. I was thinking about merging this first.

jonathan-gramain: I see no point in merging a commit that we know is going to be changed immediately, it makes the history more confusing than it should be IMHO.

naren-rajendran: The commit now has working code from #1006. Not the latest changes from #1006

naren-rajendran:
The changes in the PR are merged via https://github.com/scality/backbeat/pull/1006/commits, so closing the pull request.

PR #1006, opened 2020-09-02
Bugfix/s3 c 3306 notification populator publish failure

naren-rajendran:
Bucket notification populator fails to publish messages to local queue if the publish is made asynchronously. Changes are made to make the call asynchronously.

### DIFF
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -129,7 +129,6 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
             if (error) {
                 return undefined;
             }
-
             return result.notificationConfiguration;
         }
         return undefined;
@@ -139,18 +138,24 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
         return key.split(VID_SEP)[0];
     }
 
-    filter(entry) {
+    /**
+     * Asynchronous filter
+     * @param {Object} entry - constructor params
+     * @param {function} cb - callback
+     * @return {undefined}
+     */
+    filterAsync(entry, cb) {
         const { bucket, key, value, type } = entry;
         const { error, result } = safeJsonParse(value);
         // ignore if entry's value is not valid
         if (error) {
             this.log.error('could not parse log entry', { value, error });
-            return undefined;
+            return cb();
         }
         // ignore bucket op, mpu's or if the entry has no bucket
         if (!bucket || bucket === usersBucket ||
             (key && key.startsWith(mpuBucketPrefix))) {
-            return undefined;
+            return cb();
         }
         // bucket notification configuration updates
         if (bucket && result && this._isBucketEntry(bucket, key)) {
@@ -184,7 +189,7 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                             + 'configuration';
                         this.log.error(errMsg, { error });
                     }
-                    return undefined;
+                    return cb();
                 });
             }
             // bucket notification conf has been removed, so remove zk node
@@ -203,7 +208,7 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                         + 'configuration';
                     this.log.err(errMsg, { error });
                 }
-                return undefined;
+                return cb();
             });
         }
         // object entry processing - filter and publish
@@ -220,12 +225,12 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                     if (config && Object.keys(config).length > 0) {
                         let eventType
                             = result[notifConstants.notificationEventPropName];
-                        const lastModified
-                            = result[notifConstants.eventTimePropName];
                         if (eventType === undefined &&
                             (type === 'delete' || type === 'del')) {
                             eventType = notifConstants.deleteEvent;
                         }
+                        const lastModified
+                            = result[notifConstants.eventTimePropName];
                         // TODO: publish necessary object properties, keeping it
                         // simple for the first iteration
                         const ent = {
@@ -239,7 +244,6 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                             this.publish(this.notificationConfig.topic,
                                 bucket,
                                 JSON.stringify(ent));
-                            return undefined;
                         }
                         return next();
                     }
@@ -254,10 +258,10 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                         error,
                     });
                 }
-                return undefined;
+                return cb();
             });
         }
-        return undefined;
+        return cb();
     }
 }
 
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -123,13 +123,29 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
             || key === undefined);
     }
 
-    _getBucketNotificationConfiguration(value) {
+    _getBucketAttributes(value) {
         if (value && value.attributes) {
             const { error, result } = safeJsonParse(value.attributes);
             if (error) {
                 return undefined;
             }
-            return result.notificationConfiguration;
+            return result;
+        }
+        return undefined;
+    }
+
+    _getBucketNameFromAttributes(value) {
+        const attributes = this._getBucketAttributes(value);
+        if (attributes && attributes.name) {
+            return attributes.name;
+        }
+        return undefined;
+    }
+
+    _getBucketNotificationConfiguration(value) {
+        const attributes = this._getBucketAttributes(value);
+        if (attributes && attributes.notificationConfiguration) {
+            return attributes.notificationConfiguration;
         }
         return undefined;
     }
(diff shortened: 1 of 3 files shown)


extensions/notification/NotificationQueuePopulator.js
```diff
@@ -154,7 +154,7 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
         return key.split(VID_SEP)[0];
     }
 
-    filter(entry) {
+    filter(entry, cb) {
```
jonathan-gramain: There are some missing `return cb()` in this function, github would not let me comment at the right places but I see the initial error cases not calling `cb`, it's important to be consistently calling it to avoid the processing being stuck.

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -268,7 +268,7 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                         error,
                     });
                 }
-                return undefined;
+                return cb();
             });
         }
         return undefined;
```
jonathan-gramain: This shoul also `return cb()` I think

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -253,7 +253,7 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                             this.publish(this.notificationConfig.topic,
                                 bucket,
                                 JSON.stringify(ent));
-                            return undefined;
+                            return next();
```
jonathan-gramain: you can just remove this line it will fallback to the same `return next()`

jonathan-gramain (changes requested):

extensions/notification/NotificationQueuePopulator.js
naren-rajendran: I think I missed while pushing the commit, I'll push again.

extensions/notification/NotificationQueuePopulator.js
naren-rajendran: I think I missed while pushing the commit, I'll push again.

dora-korpar (approved):
Good for me with Jonathan's changes

lib/queuePopulator/LogReader.js
```diff
@@ -318,33 +320,47 @@ class LogReader {
...
-        } else {
-            if (entry.key === undefined || entry.value === undefined) {
-                return;
+
+        function _executeFilter(ext, entry, cb) {
+            if (typeof ext.filterAsync === 'function') {
+                ext.filterAsync(entry, cb);
+            } else {
+                ext.filter(entry);
+                cb();
```
jonathan-gramain: Recommend `process.nextTick(cb)` to avoid the risk of stack overflow

lib/queuePopulator/LogReader.js
```diff
@@ -318,33 +320,47 @@ class LogReader {
...
+        }
+
+        async.eachSeries(this._extensions, (ext, next) => {
+            let entryValue = entry.value;
+            const notifExtension = ext instanceof NotificationQueuePopulator;
+            if (notifExtension) {
+                entryValue = entry.value || '{}';
             }
-            this._extensions.forEach(ext => ext.filter({
+            if (!notifExtension &&
```
jonathan-gramain, line 339: Why can't this logic be in the notification extension itself?

lib/queuePopulator/LogReader.js
```diff
@@ -381,6 +390,35 @@ class LogReader {
         return undefined;
     }
 
+    _processFilterEntries(batchState, done) {
+        const { entriesToPublish, currentRecord } = batchState;
+        if (!currentRecord || !currentRecord.entries) {
+            return done();
+        }
+        async.series([
```
jonathan-gramain: There is no need to use `async.series` as there's only one truly async call in the list (the 2nd one)

lib/queuePopulator/QueuePopulatorExtension.js
```diff
@@ -58,6 +58,8 @@ class QueuePopulatorExtension {
      * @return {undefined}
      */
     filter(entry) { // eslint-disable-line no-unused-vars
+        // sub classes can also implement 'filterAsync' for asynchronous
```
jonathan-gramain, line 62: I'd rather have `filter(entry)` default implementation do nothing then, it avoids to have to create a stub in the extension.

lib/queuePopulator/LogReader.js
```diff
@@ -330,19 +320,47 @@ class LogReader {
             }
             return key;
         }
-        if (entry.key === undefined || entry.value === undefined) {
-            return;
+
+        function _executeFilter(ext, entry, cb) {
+            if (typeof ext.filterAsync === 'function') {
```
jonathan-gramain, line 325: Also this change does not seem to be in the proper commit, as you modified the changes made in the first commit

naren-rajendran: I made this change after our discussion. Do you mean the other PR?

lib/queuePopulator/LogReader.js
naren-rajendran: If we don't handle this here, the entries with no value, typically `del` operation on an object (unversioned) will be filtered out here.

lib/queuePopulator/QueuePopulatorExtension.js
naren-rajendran: Should I remove the assert here? also should I create a default `filterAsync` here?

jonathan-gramain: That looks ok to me, the assert was here to tell that `filter` was needed but now it's not the case anymore because we also support `filterAsync`.

jonathan-gramain: If you create a default `filterAsync` the logic you put in checking the existence of the function will not work anymore

lib/queuePopulator/LogReader.js
jonathan-gramain: ah I see. In order not to have to change other existing extension filters, we can do this indeed, but maybe it should be reworked at some point (we only have a couple extensions to modify).

lib/queuePopulator/LogReader.js
jonathan-gramain: Yes, the other PR or commit that is included in the other PR

lib/queuePopulator/QueuePopulatorExtension.js
naren-scality: So, remove the assert?

lib/queuePopulator/LogReader.js
```diff
@@ -355,14 +369,7 @@ class LogReader {
...
             logStats.nbLogRecordsRead += 1;
-
-            this._setEntryBatch(entriesToPublish);
-            record.entries.forEach(entry => {
-                logStats.nbLogEntriesRead += 1;
-
-                this._processLogEntry(batchState, record, entry);
-            });
-            this._unsetEntryBatch();
+            batchState.currentRecord = record;
```
jonathan-gramain: `logRes.log` is a stream of record objects, so you may have more than one record (usually the configured batch size like 1000 at a time). If you set it here it will only keep the last record for `processFilterEntries` so the 999 previous ones will not be processed.

jonathan-gramain (changes requested):

lib/queuePopulator/LogReader.js
```diff
@@ -379,6 +390,47 @@ class LogReader {
...
+                });
+                return done(err);
+            }
+            return done();
+        });
+    }
+
+    _processFilterEntries(batchState, done) {
+        const { currentRecords } = batchState;
+        if (!currentRecords || !currentRecords.length > 0) {
```
jonathan-gramain: This contains a logic error due to priority of operators: `!currentRecords.length > 0` is equivalent to `(!currentRecords.length) > 0`. The end result may or not be affected in this particular case, but it should be fixed anyway.

lib/queuePopulator/LogReader.js
```diff
@@ -379,6 +390,47 @@ class LogReader {
...
+            return done();
+        }
+        async.eachSeries(currentRecords,
+            (rec, next) => this._processFilterEntry(batchState, rec, next),
+            err => {
+                batchState.currentRecords = [];
```
jonathan-gramain: This should not be needed I think as I don't think the batch is reused at this point before creating the next one

jonathan-gramain (changes requested):

lib/queuePopulator/LogReader.js
naren-rajendran: Ah I corrected this, didn't push it..will update.

lib/queuePopulator/LogReader.js
```diff
@@ -379,6 +390,46 @@ class LogReader {
...
+                });
+                return done(err);
+            }
+            return done();
+        });
+    }
+
+    _processFilterEntries(batchState, done) {
+        const { currentRecords } = batchState;
+        if (!currentRecords || !(currentRecords.length > 0)) {
```
jonathan-gramain: Could also be simply checked with `currentRecords.length === 0`, I will not block the PR for this though :wink:

jonathan-gramain (approved):

naren-rajendran:
@bert-e approve

naren-rajendran:
@bert-e approve

PR #1012, opened 2020-09-03
improvement: S3C-3316 modify notification message

naren-rajendran:
Bucket notification message structure is modified to match aws specifications.

### DIFF
--- a/extensions/notification/utils/message.js
+++ b/extensions/notification/utils/message.js
@@ -1,0 +1,79 @@
+const { versioning } = require('arsenal');
+const versionIdUtils = versioning.VersionID;
+
+const constants = require('../constants');
+
+/**
+ * Add attributes to an internal notification message,
+ * with its log entry counterpart.
+ * @param  {Object} logEntry - log entry that generated the message
+ * @param  {Object} message - initial message generated
+ * @return {Object} updated message object
+ */
+function addLogAttributes(logEntry, message) {
+    const { eventMessageProperty } = constants;
+    const msg = {};
+    Object.entries(eventMessageProperty).forEach(([key, value]) => {
+        let val = logEntry[value];
+        if (typeof val === 'number') {
+            val = val.toString();
+        }
+        msg[key] = val || null;
+    });
+    return Object.assign(msg, message);
+}
+
+/**
+ * Transforms an event object to specification.
+ * @param  {Object} event - event object
+ * @return {Object} updated event object
+ */
+function transformToSpec(event) {
+    const encodedVersionId
+        = event.versionId ? versionIdUtils.encode(event.versionId) : null;
+    const message = {
+        Records: [
+            {
+                eventVersion: constants.eventVersion,
+                eventSource: constants.eventSource,
+                awsRegion: event.region,
+                eventTime: event.dateTime,
+                eventName: event.eventType,
+                userIdentity: {
+                    principalId: null,
+                },
+                requestParameters: {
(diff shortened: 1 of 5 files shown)


dora-korpar (approved):

extensions/notification/utils/message.js
```diff
@@ -0,0 +1,79 @@
...
+ * @param  {Object} message - initial message generated
+ * @return {Object} updated message object
+ */
+function addLogAttributes(logEntry, message) {
+    const { eventMessageProperty } = constants;
+    const msg = {};
+    for (const [key, value] of Object.entries(eventMessageProperty)) {
+        let val = logEntry[value];
+        if (typeof val === 'number') {
+            val = val.toString();
```
jonathan-gramain, line 19: Is this required by the spec (convert numbers to strings)?

extensions/notification/utils/message.js
```diff
@@ -0,0 +1,79 @@
...
+                },
+                requestParameters: {
+                    sourceIPAddress: null,
+                },
+                responseElements: {
+                    'x-amz-request-id': null,
+                    'x-amz-id-2': null,
+                },
+                s3: {
+                    s3SchemaVersion: event.schemaVersion,
```
jonathan-gramain: Why is the schema version part of the `event` object? I imagined (but can be wrong) that the schema version is a AWS API version like `2006-07-01`

extensions/notification/utils/message.js
```diff
@@ -0,0 +1,79 @@
...
+    return Object.assign(msg, message);
+}
+
+/**
+ * Transforms an event object to specification.
+ * @param  {Object} event - event object
+ * @return {Object} updated event object
+ */
+function transformToSpec(event) {
+    const verId
```
jonathan-gramain: You may call this var `encodedVersionId` for clarity

extensions/notification/utils/message.js
naren-rajendran: yes everything appears to be string, https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html

extensions/notification/utils/message.js
naren-rajendran: I wasn't sure and the doc I referred wasn't clear either, https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html

jonathan-gramain: Ok, maybe it's  better to have it defined as a constant `1.0` then, if it refers to the version of this `message` object it should probably not be part of the `event`.

jonathan-gramain (approved):

extensions/notification/utils/message.js
naren-rajendran: @jonathan-gramain there is another property `eventVersion` in the message, this probably is  the one that relates to version of message object. Should I still make the `s3SchemaVersion` to be `1.0`?

jonathan-gramain: In doubt, I'd set `s3SchemaVersion` to `1.0` since this is what is shown as an example in the AWS page.

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e approve

naren-scality:
@bert-e resest

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

PR #1049, opened 2020-09-14
bugfix: S3C-2775 provide versionID to 'PUT /_/backbeat/metadata' request

jonathan-gramain:
Add the versionId in the query string on 'PUT /_/backbeat/metadata'
requests.

The goal is to be correct regarding the check done in cloudserver
backbeat routes: instead of checking the master version metadata, it
will check the target version object, resulting in a ObjNotFound error
if the specific version does not exist. This in turn allows to fix
UTAPI counting of replicated objects, by providing the information of
whether an existing object with the same version exists.

### DIFF
--- a/lib/clients/backbeat-2017-07-01.api.json
+++ b/lib/clients/backbeat-2017-07-01.api.json
@@ -549,6 +549,12 @@
                         "location": "uri",
                         "locationName": "Key"
                     },
+                    "VersionId": {
+                        "type": "string",
+                        "documentation": "VersionId used to reference a specific version of the object.",
+                        "location": "querystring",
+                        "locationName": "versionId"
+                    },
                     "ContentLength": {
                         "location": "header",
                         "locationName": "Content-Length",
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -418,6 +418,7 @@ class ReplicateObject extends BackbeatTask {
         const req = this.backbeatDest.putMetadata({
             Bucket: entry.getBucket(),
             Key: entry.getObjectKey(),
+            VersionId: entry.getEncodedVersionId(),
             ContentLength: Buffer.byteLength(mdBlob),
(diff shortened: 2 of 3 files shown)


naren-rajendran (approved):

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -549,6 +549,12 @@
                         "location": "uri",
```
miniscruff, line 549: just a little curios as to what this file does and how its updated. Cause the file has a date which sort of indicates a version as well as a version in the meta data. I would feel that if there is a version in some form that we couldn't just change it we would need to make a new version. As this looks like a fairly safe backwards compatable change a minor update would be fine but considering the file date is 3 years old and the version is 1.0 I would imagine that it has been updated more than once. 

Or is this used some other way and these dates / versions are not related to much or something?

jonathan-gramain: This file is a description for the AWS API generator to create the bindings automatically. The date version was set at the time we created this internal API, because the generator requires one. This being said, I never really considered updating this field for compatible changes, I fear it could bring incompatibility by preventing an updated client from talking to an older server, for example.

In doubt I'd leave it as it is, and also make sure an updated server will not cry if this field is missing because sent by an old client, and that it will just not have the benefit of the bugfix until the client is upgraded as well.

jonathan-gramain: As a matter of fact, I think the official AWS S3 client takes the same approach of updating regularly the file (to fix bugs or add some minor functionality) without changing the date when the API was settled.

miniscruff: So the date is when the original API was settled, maybe there is an internal version that is updated on minor fixes. Couldn't find any docs related maybe I can find it later.

tmacro (approved):

anurag4DSB (approved):

PR #1055, opened 2020-09-15
ZENKO-2810 upload docker image with short version

tcarmet:
With the following changes we will make available on merge docker images with the following tag: `latest-8.1`

This will allow us to call backbeat docker image that relates to the `development/*` branches

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -21,12 +21,21 @@ models:
       name: Set docker tag latest property
       property: docker_tag_latest
       value: "%(prop:registry_url)s:latest-%(prop:product_version)s"
+  - SetPropertyFromCommand: &short_version
+      name: Set short version
+      property: short_version
+      command: echo "%(prop:product_version)s" | grep -o '^[0-9]*\.[0-9]*'
+  - SetProperty: &docker_tag_short_version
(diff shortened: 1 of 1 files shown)


eve/main.yml
```diff
@@ -21,12 +21,21 @@ models:
...
       property: docker_tag_latest
       value: "%(prop:registry_url)s:latest-%(prop:product_version)s"
+  - SetPropertyFromCommand: &short_version
+      name: Set short version
+      property: short_version
+      command: echo "%(prop:product_version)s" | grep -o '^[0-9]*\.[0-9]*'
+  - SetProperty: &docker_tag_short_version
+      name: Set docker tag latest short version property
+      property: docker_tag_short_version
+      value: "%(prop:registry_url)s:latest-%(prop:short_version)s"
```
tcarmet, line 31: Here I'm wondering if I should rather replace the `docker_tag_latest` property to the `short_version` having this 3 digit latest tag may not be useful for us and we may only need to keep the one with the major and minor version.

jonathan-gramain: @rahulreddy

anurag4DSB (approved):

jonathan-gramain (approved):

PR #1061, opened 2020-09-18
improvement: S3C-3331 sync zookeeper calls

naren-rajendran:
A new configuration manager is introduced to handle zookeeper calls in the background and manage bucket notification configuration synchronously.

### DIFF
--- a/extensions/notification/NotificationConfigManager.js
+++ b/extensions/notification/NotificationConfigManager.js
@@ -1,0 +1,402 @@
+const async = require('async');
+const { EventEmitter } = require('events');
+const zookeeper = require('node-zookeeper-client');
+const joi = require('@hapi/joi');
+
+const safeJsonParse = require('./utils/safeJsonParse');
+const constants = require('./constants');
+
+const paramsJoi = joi.object({
+    zkClient: joi.object().required(),
+    parentNode: joi.string().optional(),
+    logger: joi.object().required(),
+}).required();
+
+/**
+ * @class NotificationConfigManager
+ *
+ * @classdesc Manages bucket notification configurations in zookeeper, maintains
+ * configurations across multiple managers.
+ */
+class NotificationConfigManager {
+    /**
+     * @constructor
+     * @param {Object} params - constructor params
+     * @param {Object} params.zkClient - zookeeper client
+     * @param {String} params.parentNode - parent node to be used
+     * @param {Logger} params.logger - logger object
+     */
+    constructor(params) {
+        joi.attempt(params, paramsJoi);
+        this._zkClient = params.zkClient;
+        this._parentNode = params.parentNode || '';
+        this.log = params.logger;
+        this._configs = new Map();
+        this._concurrency = constants.configManager.concurrency;
+        this._emitter = new EventEmitter();
+        this._setupEmitterListeners();
+    }
+
+    _emitterErrorListener(error, listener) {
+        this.log.error('BucketNotifConfigManager.emitter.error', {
+            listener,
+            error,
+        });
+        return undefined;
+    }
+
+    _setConfigListener(bucket, config) {
+        this.log.debug('BucketNotifConfigManager.emitter.setConfig', {
+            event: 'setConfig',
+            bucket,
+            config,
+        });
+        this._setBucketNotifConfig(bucket, JSON.stringify(config), err => {
+            if (err) {
+                this._emitter.emit('error', err, 'setConfigListener');
+            }
+            return undefined;
+        });
+    }
+
+    _getConfigListener(updatedBucket = '') {
+        this.log.debug('BucketNotifConfigManager.emitter.getConfig', {
+            event: 'getConfig',
+        });
+        this._listBucketsWithConfig((err, buckets) => {
+            if (err) {
+                this._emitter.emit('error', err, 'getConfigListener');
+                return undefined;
+            }
+            this.log.debug('bucket config to be updated in map', {
+                bucket: updatedBucket,
+            });
+            const newBuckets
+                = this._getNewBucketNodes(buckets);
+            this.log.debug('new bucket configs to be added to map', {
+                buckets: newBuckets,
+            });
+            const bucketsToMap
+                = updatedBucket ?
+                    [updatedBucket, ...newBuckets] : newBuckets;
+            this.log.debug('bucket configs to be added/updated to map', {
+                buckets: bucketsToMap,
+            });
+            if (bucketsToMap.length > 0) {
+                this._updateLocalStore(bucketsToMap);
+            }
+            return undefined;
+        });
+    }
+
+    _removeConfigListener(bucket) {
+        this.log.debug('BucketNotifConfigManager.emitter.removeConfig', {
+            event: 'removeConfig',
+            bucket,
+        });
+        this._removeBucketNotifConfigNode(bucket, err => {
+            if (err) {
+                this._emitter.emit('error', err, 'removeConfigListener');
+            }
+            return undefined;
+        });
+    }
+
+    _setupEmitterListeners() {
+        this._emitter.setMaxListeners(constants.configManager.maxListeners);
+        this._emitter
+            .on('error', error => this._emitterErrorListener(error))
+            .on('setConfig',
+                (bucket, config) => this._setConfigListener(bucket, config))
+            .on('getConfig', bucket => this._getConfigListener(bucket))
+            .on('removeConfig', bucket => this._removeConfigListener(bucket));
+    }
+
+    _callbackHandler(cb, err, result) {
+        if (cb && typeof cb === 'function') {
+            return cb(err, result);
+        }
+        return undefined;
+    }
+
+    _getBucketNodeZkPath(bucket) {
+        if (this._parentNode) {
+            return `/${this._parentNode}/${bucket}`;
+        }
+        return `/${bucket}`;
+    }
+
+    _getConfigDataFromBuffer(data) {
+        const { error, result } = safeJsonParse(data);
+        if (error) {
+            this.log.error('invalid config', { error, config: data });
+            return undefined;
+        }
+        return result;
+    }
+
+    _getBucketNotifConfig(bucket, cb) {
+        const method
+            = 'BucketNotificationConfigManager._getBucketNotifConfig';
+        const zkPath = this._getBucketNodeZkPath(bucket);
+        return this._zkClient.getData(zkPath, event => {
+            this.log.debug('zookeeper getData watcher triggered', {
+                zkPath,
+                method,
+                event,
+                bucket,
+            });
+            if (event.type === zookeeper.Event.NODE_DATA_CHANGED) {
+                this._emitter.emit('getConfig', bucket);
+            }
+            if (event.type === zookeeper.Event.NODE_DELETED) {
+                this.removeConfig(bucket, false);
+            }
+        }, (error, data) => {
+            if (error && error.name !== 'NO_NODE') {
+                const errMsg
+                    = 'error fetching bucket notification configuration';
+                this.log.error(errMsg, {
+                    method,
+                    error,
+                });
+                return this._callbackHandler(cb, error);
+            }
+            if (data) {
+                return this._callbackHandler(cb, null, data);
+            }
+            // no configuration
+            return this._callbackHandler(cb);
+        });
+    }
+
+    _checkNodeExists(zkPath, cb) {
+        const method
+            = 'BucketNotificationConfigManager._checkNodeExists';
+        return this._zkClient.exists(zkPath, (err, stat) => {
+            if (err) {
+                this.log.error('error checking node existence',
+                    { method, zkPath });
+                return this._callbackHandler(cb, err);
+            }
+            if (stat) {
+                this.log.debug('node exists', { method, zkPath });
+                return this._callbackHandler(cb, null, true);
+            }
+            this.log.debug('node does not exist', { method, zkPath });
+            return this._callbackHandler(cb, null, false);
+        });
+    }
+
+    _setBucketNotifConfig(bucket, data, cb) {
+        const method
+            = 'BucketNotificationConfigManager._setBucketNotifConfig';
+        const zkPath = this._getBucketNodeZkPath(bucket);
+        return async.waterfall([
+            next => this._checkNodeExists(zkPath, next),
+            (exists, next) => {
+                if (!exists) {
+                    return this._createBucketNotifConfigNode(bucket,
+                        err => next(err));
+                }
+                return next();
+            },
+            next => this._zkClient.setData(zkPath, Buffer.from(data), -1, next),
+        ], err => {
+            if (err) {
+                this.log.error('error saving config', { method, zkPath, data });
+            }
+            return this._callbackHandler(cb, err);
+        });
+    }
+
+    _createBucketNotifConfigNode(bucket, cb) {
+        const method
+            = 'BucketNotificationConfigManager._createBucketNotifConfigNode';
+        const zkPath = this._getBucketNodeZkPath(bucket);
+        return this._zkClient.mkdirp(zkPath, err => {
+            if (err) {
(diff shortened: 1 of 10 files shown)


extensions/notification/NotificationQueuePopulator.js
```diff
@@ -19,110 +17,35 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
 
+    /**
+     * Check if bucket entry based on bucket and key params
+     *
+     * @param {String} bucket - bucket
+     * @param {String} key - object key
+     * @return {boolean} - true if bucket entry
+     */
     _isBucketEntry(bucket, key) {
         return ((bucket.toLowerCase() === 'metastore' && key)
```
vrancurel: Where does it come from `metastore` ? Sometimes it is `__metastore` (e.g. Mongo). Shall it be in `constants.js` ?

lib/queuePopulator/QueuePopulator.js
```diff
@@ -74,7 +78,8 @@ class QueuePopulator {
         async.series([
             next => this._setupMetricsClients(next),
             next => this._setupFailedCRRClients(next),
-            next => this._setupZookeeper(() => {
+            next => this._setupZookeeper(next),
+            next => this._setupNotificationConfigManager(() => {
```
vrancurel, line 82: So finally the notification relies on the queuePopulator ? isnt'it what was discussed here https://github.com/scality/citadel/blob/development/1.0/docs/design/bucket-notification.md#alternatives ?
The inconvenient of this design is that the bucket notification queue populator will live in the CRR queue populator pod, and thus any change of the notification pod will restart the CRR (or vice versa).
Is it finally better to do so or am I missing something ?

vrancurel (changes requested):

extensions/notification/NotificationQueuePopulator.js
naren-rajendran: The current implementation only targets S3C, `notifications` is not implemented in zenko yet, so that's why only `metastore` is referenced. If your question is where `metastore` comes from, it is from the log, a sample entry,
`{ type: 'put',
  bucket: 'metastore',
  key: 'db/bucket1',
  value:
   '{"name":"bucket1","attributes":"{\\"name\\":\\"bucket1\\",\\"owner\\":\\"9abe0ec063308bc0c7cd5f99e1260f942afb7e8682a86edc807795fa652ee28a\\",\\"ownerDisplayName\\":\\"naren\\",\\"creationDate\\":\\"2020-08-30T02:42:48.224Z\\",\\"mdBucketModelVersion\\":8,\\"transient\\":true,\\"deleted\\":false,\\"serverSideEncryption\\":null,\\"versioningConfiguration\\":null,\\"locationConstraint\\":\\"us-east-1\\",\\"cors\\":null,\\"replicationConfiguration\\":null,\\"lifecycleConfiguration\\":null,\\"bucketPolicy\\":null,\\"objectLockEnabled\\":false,\\"objectLockConfiguration\\":null,\\"notificationConfiguration\\":null}","creating":true,"deleting":false,"id":"bucket1","raftSessionID":1,"version":2}' }`

lib/queuePopulator/QueuePopulator.js
naren-rajendran: Yes the current implementation of notifications relies on `queuePopulator` to read entries from log and yes `notification` is still an extension, but new containers are created to host only notification processes. The CRR containers will not have the extension `notification` and the bucket notification containers will only host `notification` processes. This was implemented after discussions with @jonathan-gramain.

extensions/notification/NotificationQueuePopulator.js
naren-scality: Ref, https://github.com/scality/Federation/commit/351b022151b52a73f5a6ac176031f5f767c69c44

lib/queuePopulator/QueuePopulator.js
naren-rajendran: Ref, https://github.com/scality/Federation/commit/351b022151b52a73f5a6ac176031f5f767c69c44

extensions/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,402 @@
...
+     * @param {Logger} params.logger - logger object
+     */
+    constructor(params) {
+        joi.attempt(params, paramsJoi);
+        this._zkClient = params.zkClient;
+        this._parentNode = params.parentNode || constants.zkConfigParentNode;
+        this.log = params.logger;
+        this._configs = new Map();
+        this._concurrency = constants.configManager.concurrency;
+        this._emitter = new EventEmitter();
```
jonathan-gramain, line 36: Instead of having an emitter object you could have the `NotificationConfigManager` class inherit from `EventEmitter`, it could simplify the code and logic a bit (on the class implementation as well as for the user of the class) and it is the model used for the main node.js event emitter classes.

extensions/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,402 @@
...
+            // no configuration
+            return this._callbackHandler(cb);
+        });
+    }
+
+    _checkNodeExists(zkPath, cb) {
+        const method
+            = 'BucketNotificationConfigManager._checkNodeExists';
+        return this._zkClient.exists(zkPath, (err, stat) => {
+            if (err) {
```
jonathan-gramain, line 177: Are you sure that `exists` does not return `NO_NODE` when a node does not exist?

extensions/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,402 @@
...
+            }
+            this._callbackHandler(cb, null, buckets);
+        });
+    }
+
+    _updateLocalStore(buckets, cb) {
+        async.eachSeries(buckets, (bucket, next) => {
+            this._getBucketNotifConfig(bucket, (err, data) => {
+                if (err) {
+                    next(err);
```
jonathan-gramain: You need a `return` here

extensions/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,402 @@
...
+    getConfig(bucket) {
+        return this._configs.get(bucket);
+    }
+
+    /**
+     * Add/update bucket notification configuration
+     *
+     * @param {String} bucket - bucket
+     * @param {Object} config - bucket notification configuration
+     * @return {boolean|undefined} - true if removed or undefined
```
jonathan-gramain: Could be more consistent to return `false` if there is an error setting the config so we always return a boolean value.

extensions/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,402 @@
...
+            return undefined;
+        }
+    }
+
+    /**
+     * Remove bucket notification configuration
+     *
+     * @param {String} bucket - bucket
+     * @param {boolean} removeNode - remove the zookeeper node (optional)
+     * @return {boolean|undefined} - true if removed or undefined
```
jonathan-gramain: same

extensions/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,402 @@
...
+    /**
+     * Init bucket notification configuration manager
+     *
+     * @param {function} [cb] - callback
+     * @return {undefined}
+     */
+    init(cb) {
+        this._listBucketsWithConfig((err, buckets) => {
+            if (err) {
+                cb(err);
```
jonathan-gramain: missing `return`

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -150,132 +85,112 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+            const bnConfig = {
+                bucket: bucketName,
+                notificationConfiguration,
+            };
+            // bucket notification config is available, update node
+            this.bnConfigManager.setConfig(bucketName, bnConfig);
+            return undefined;
+        }
+        // bucket notification conf has been removed, so remove zk node
+        return this.bnConfigManager.removeConfig(bucketName);
```
jonathan-gramain, line 119: Do you want to remove the config in case the config exists but has no entry (`Object.keys(notificationConfiguration).length === 0`)? Wondering if that is ok regarding the handling of zookeeper watchers etc., might be alright but worth checking.

tests/functional/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,161 @@
+const assert = require('assert');
+const async = require('async');
+const werelogs = require('werelogs');
+const ZookeeperMock = require('zookeeper-mock');
```
jonathan-gramain, line 4: So there is a zookeeper mock in the npm registry, good to know :+1:

tests/functional/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,161 @@
...
+}
+
+function populateTestConfigs(zkClient, numberOfConfigs, cb) {
+    async.timesLimit(numberOfConfigs, concurrency, (n, done) => {
+        const bucket = `${bucketPrefix}${n + 1}`;
+        const val = getTestConfigValue(bucket);
+        const strVal = JSON.stringify(val);
+        const node = `/${zkConfigParentNode}/${bucket}`;
+        async.series([
+            nxt => zkClient.mkdirp(node, nxt),
```
jonathan-gramain: minor: why `nxt` not `next` (if it comes from the line length limitation of the linter, I don't consider it a valid excuse ! :facepalm: )

tests/functional/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,161 @@
...
+            // nxt => zkClient.setData(node, strVal, nxt),
```
jonathan-gramain: looks like debug comment, should be removed

tests/functional/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,161 @@
...
+}
+
+function listBuckets(zkClient, cb) {
+    const node = `/${zkConfigParentNode}`;
+    zkClient.getChildren(node, cb);
+}
+
+function deleteTestConfigs(zkClient, cb) {
+    const node = `/${zkConfigParentNode}`;
+    listBuckets(zkClient, (e, children) => {
```
jonathan-gramain: Same idea, why `e` not `err` or `error` which is more descriptive while being reasonably short, and compels with 99% of our usage throughout the code

tests/unit/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,178 @@
...
+    it('should get bucket notification configuration', () => {
+        const manager = new NotificationConfigManager(params);
+        managerInit(manager, () => {
+            const bucket = 'bucket1';
+            const result = manager.getConfig(bucket);
+            assert.strictEqual(result.bucket, bucket);
+        });
+    });
+
+    it('should return undefined for an invalid bucket', () => {
```
jonathan-gramain: the description might say `for a bucket that does not have a notification config` or that the bucket itself does not exist, or something similar? Since the bucket name does not look invalid it looks confusing.

tests/unit/notification/NotificationConfigManager.js
```diff
@@ -0,0 +1,178 @@
...
+            }, timeoutMs);
+        });
+    });
+
+    it('config count should be zero when all configs are removed', done => {
+        const manager = new NotificationConfigManager(params);
+        async.series([
+            next => managerInit(manager, next),
+            next => deleteTestConfigs(zkClient, next),
+            next => setTimeout(next, 100),
```
jonathan-gramain: I assume you could use `timeoutMs` since you defined it

lib/queuePopulator/QueuePopulator.js
vrancurel: so is there a risk calling `_setupFailedCRRClients()` here ?

naren-rajendran: It will be called yes, but the setup will be made only when replication extension is available, https://github.com/scality/backbeat/blob/f17a1821b2604aed59e1ebcbe79336b5c0dce95c/lib/queuePopulator/QueuePopulator.js#L119

tests/functional/notification/NotificationConfigManager.js
naren-rajendran: Ha ha.. @vrancurel  made scality's `zookeeper-mock` available in npm. I am a contributor too :)

extensions/notification/NotificationConfigManager.js
naren-rajendran: Yes, `zookeeper-client` does not return `NO_NODE` when the path is invalid, https://github.com/alexguan/node-zookeeper-client/blob/master/index.js#L746

tests/unit/notification/NotificationConfigManager.js
naren-rajendran: :facepalm: my copy paste skills are not good enough.

tests/functional/notification/NotificationConfigManager.js
naren-rajendran: I am uninstalling my formatter plugin.. caused me enough pain :|

extensions/notification/NotificationConfigManager.js
naren-scality: Yes, I did consider it. Though, while implementing something made me not to. I can't remember what or why. Will change it.

naren-rajendran: The events are emitted only for internal(private) purposes of the class, not intended for the instances of the class to listen to directly. May be its not the cleanest implementation, do you recommend a different approach?

extensions/notification/NotificationQueuePopulator.js
naren-rajendran: Yes, I would, it keeps the logic simple and easy. Watchers are set for `getChildren` and `getData`, translates to `node children modified`, `node child deleted` and `node updated` (in that order).. I assume when you say `config exists but has no entry` it means a config exists but has `{}` as its value! If the config value is empty for a bucket, we remove it and set it in zookeeper when it is updated (watcher will be triggered and all the manager instances get the update).

extensions/notification/NotificationConfigManager.js
jonathan-gramain: My take: if the events are meaningful you don't necessarily need to prevent external users from listening to them. You can sure use them it it helps for the implementation and you can have "internal only" events as well if you need, I don't think it warrants to create a hidden object.

Events are a useful abstraction to decouple emitters and receivers, making the code more modular. Sometimes it's the good approach, sometimes it can be better to just have registered callbacks if the decoupling is not needed, for example if all you need is systematically trigger a single task when something completes. I don't have a special recommendation in this particular implementation.

extensions/notification/NotificationQueuePopulator.js
jonathan-gramain: OK, that works. Make sure this is consistent with the code that adds config too, i.e. it does not add it if the config is empty (otherwise, might still work but you would have an inconsistent handling of empty config, with zookeeper nodes sometimes present, sometimes not, which can lead to subtle bugs)

tests/functional/notification/NotificationConfigManager.js
jonathan-gramain: Awesome :tada:

jonathan-gramain (approved):

extensions/notification/NotificationQueuePopulator.js
naren-rajendran: yes, taken care of, https://github.com/scality/backbeat/blob/d8dec0ce11ce311785f16afee13a01c9fb497e88/extensions/notification/NotificationQueuePopulator.js#L108

vrancurel (approved):

naren-rajendran:
@bert-e approve

anurag4DSB (approved):

dora-korpar (approved):

PR #1142, opened 2020-10-23
bf: S3C-3455 notification config manager crash

naren-rajendran:
Notification config manager crashes if the default parent config zookeeper node does not exist. The parent node is checked and created if absent, on config manager initiation.

### DIFF
--- a/extensions/notification/NotificationConfigManager.js
+++ b/extensions/notification/NotificationConfigManager.js
@@ -210,6 +210,34 @@ class NotificationConfigManager {
         });
     }
 
+    _checkConfigurationParentNode(cb) {
+        const method
+            = 'BucketNotificationConfigManager._checkConfigurationParentNode';
+        const zkPath = `/${this._parentNode}`;
+        return async.waterfall([
+            next => this._checkNodeExists(zkPath, next),
+            (exists, next) => {
+                if (!exists) {
+                    this.log.debug('parent configuration zookeeper node does ' +
+                        'not exist', { method, zkPath });
+                    return this._zkClient.mkdirp(zkPath, err => next(err));
+                }
+                this.log.debug('parent configuration zookeeper node exists',
+                    { method, zkPath });
+                return next();
+            },
+        ], err => {
+            if (err) {
+                const errMsg
+                    = 'error checking configuration zookeeper parent node';
+                this.log.error(errMsg, { method, zkPath });
+            }
+            this.log.debug('parent configuration zookeeper checked/added',
+                { method, zkPath });
+            return this._callbackHandler(cb);
+        });
+    }
+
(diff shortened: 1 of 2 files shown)


naren-rajendran:
@bert-e reset

extensions/notification/NotificationConfigManager.js
```diff
@@ -210,6 +210,34 @@ class NotificationConfigManager {
...
+        const zkPath = `/${this._parentNode}`;
+        return async.waterfall([
+            next => this._checkNodeExists(zkPath, next),
+            (exists, next) => {
+                if (!exists) {
+                    this.log.info('parent configuration zookeeper node does ' +
+                        'not exist', { method, zkPath });
+                    return this._zkClient.mkdirp(zkPath, err => next(err));
+                }
+                this.log.info('parent configuration zookeeper node exists',
```
jonathan-gramain: not sure this deserves `info` level, should be `debug` IMHO

jonathan-gramain (approved):

extensions/notification/NotificationConfigManager.js
```diff
@@ -210,6 +210,34 @@ class NotificationConfigManager {
...
+                    { method, zkPath });
+                return next();
+            },
+        ], err => {
+            if (err) {
+                const errMsg
+                    = 'error checking configuration zookeeper parent node';
+                this.log.error(errMsg, { method, zkPath });
```
nicolas2bert, line 233: Should we return a callback with the "checking configuration" error?

extensions/notification/NotificationConfigManager.js
```diff
@@ -390,11 +418,13 @@ class NotificationConfigManager {
      * @return {undefined}
      */
     init(cb) {
-        return this._listBucketsWithConfig((err, buckets) => {
-            if (err) {
-                return cb(err);
-            }
-            return this._updateLocalStore(buckets, cb);
+        return this._checkConfigurationParentNode(() => {
```
nicolas2bert, line 421: Should we handle the case when the `_checkConfigurationParentNode` method fails? (related to the previous comment)

extensions/notification/NotificationConfigManager.js
naren-rajendran: This should not stop the process..there are multiple instances of managers running at any given moment, one of them will create the node, or the process would retry to create it again. So that's why we are just logging and returning the callback.

extensions/notification/NotificationConfigManager.js
naren-rajendran: Same as above.

naren-rajendran:
@bert-e reset

rahulreddy (approved):

naren-rajendran:
@bert-e approve

PR #1229, opened 2020-12-08
bf: S3C-3686 periodically recheck for ZK state

vrancurel:
This is a temporary fix

### DIFF
--- a/lib/provisioning/ProvisionDispatcher.js
+++ b/lib/provisioning/ProvisionDispatcher.js
@@ -19,6 +19,7 @@ const LEADERS = '/leaders';
 const LEADER = '/leader';
 const OWNERS = '/owners';
 const PROVISIONS = '/provisions';
+const PERIODIC_CHECK_TIMEOUT = 60000;
 
 class ProvisionDispatcher {
 
@@ -76,6 +77,21 @@ class ProvisionDispatcher {
         ], cb);
     }
 
+    checkState(cb, autoTrigger) {
+        this._onLeaderChange(autoTrigger);
+        this._onOwnerChange(autoTrigger);
+        this._onMyselfChange(cb, autoTrigger);
+        this._onProvisionChange(autoTrigger);
+        return undefined;
+    }
+
+    periodicCheckState(cb) {
+        setInterval(() => {
+            this._log.debug('periodic check state');
+            this.checkState(cb, false);
+        }, PERIODIC_CHECK_TIMEOUT);
+    }
+
     subscribe(cb) {
         this._waitConnected(() => {
             async.parallel([
@@ -87,10 +103,8 @@ class ProvisionDispatcher {
                 if (err) {
                     return cb(err);
                 }
-                this._onLeaderChange();
-                this._onOwnerChange();
-                this._onMyselfChange(cb);
-                return this._onProvisionChange();
+                this.periodicCheckState(cb);
+                return this.checkState(cb, true);
             });
         });
     }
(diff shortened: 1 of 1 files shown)


lib/provisioning/ProvisionDispatcher.js
```diff
@@ -76,6 +76,22 @@ class ProvisionDispatcher {
...
+    checkState(cb, autoTrigger) {
+        this._onLeaderChange(autoTrigger);
+        this._onOwnerChange(autoTrigger);
+        this._onMyselfChange(cb, autoTrigger);
+        this._onProvisionChange(autoTrigger);
+        return undefined;
+    }
+
+    periodicCheckState(cb) {
+        setTimeout(() => {
```
jonathan-gramain: Instead of re-calling `periodicCheckState` at the end, you could use `setInterval` in place of `setTimeout`

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -76,6 +76,22 @@ class ProvisionDispatcher {
...
+            this._log.debug('periodic check state');
+            this.checkState(cb, false);
+            this.periodicCheckState(cb);
+        }, 60000);
```
jonathan-gramain: Could be better to have this constant defined at the top of the file to make it more visible (ideally we would have it an optional parameter of the class, but it may need some refactor so not pushing for it).

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -76,6 +76,22 @@ class ProvisionDispatcher {
...
+        this._onOwnerChange(autoTrigger);
+        this._onMyselfChange(cb, autoTrigger);
+        this._onProvisionChange(autoTrigger);
+        return undefined;
+    }
+
+    periodicCheckState(cb) {
+        setTimeout(() => {
+            this._log.debug('periodic check state');
+            this.checkState(cb, false);
```
jonathan-gramain, line 91: I am under the impression that this call will accumulate watchers for each `getChildren` call that happen, but never release the watcher if there's no change. If that is correct, it is an important issue that needs a fix, maybe by having the watcher of the zookeeper calls only passed when `autoTrigger` is true, instead of re-calling the `onFoo` functions only when `autoTrigger` is true?

vrancurel: yes you are right

vrancurel: Fixed it

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -306,14 +327,14 @@ class ProvisionDispatcher {
...
+    _onOwnerChange(autoTrigger) {
         const zkPath = OWNERS;
         this._withRandomDelayIfTest(
-            () => this._client.getChildren(zkPath, event => {
+            () => this._myGetChildren(autoTrigger, zkPath, event => {
                 this._log.debug('got owner event',
                                 { zkPath: `${this._zkEndpoint}${zkPath}`,
                                   event });
-                this._onOwnerChange();
+                this._onOwnerChange(autoTrigger);
```
jonathan-gramain: if I understand correctly, here `autoTrigger` can only be true so for clarity we could pass `true` instead (same for other similar cases)

jonathan-gramain (approved):

yarn.lock
```diff
@@ -174,14 +174,14 @@ arsenal@scality/Arsenal#9f2e74e:
   version "7.4.3"
   resolved "https://codeload.github.com/scality/Arsenal/tar.gz/9f2e74ec6972527c2a9ca6ecb4155618f123fc19"
   dependencies:
+    "@hapi/joi" "^15.1.0"
```
rachedbenmustapha: I'm not sure this change is expected. (edit: seems to be the cause of unit test failure).

rachedbenmustapha (approved):

rahulreddy (approved):

PR #1234, opened 2020-12-09
bugfix: S3C-3685 stop reading new log entries after provisioning update

jonathan-gramain:
Do not trigger a new batch in case a provisioning update is pending: instead, stop after the current batch to give a chance to update the provisioning and consume from the new (set of) raft sessions that the
readers are responsible for.

Also improve logging by outputting a status log line after every batch, previously the log only occurred after a series of batches that completes reading the whole log and that were publishing at least one entry to a kafka queue.

### DIFF
--- a/lib/queuePopulator/QueuePopulator.js
+++ b/lib/queuePopulator/QueuePopulator.js
@@ -236,20 +236,21 @@ class QueuePopulator {
     }
 
     _processAllLogEntries(params, done) {
-        return async.map(
+        return async.each(
             this.logReaders,
-            (logReader, done) => logReader.processAllLogEntries(params, done),
-            (err, results) => {
-                if (err) {
-                    return done(err);
-                }
-                const annotatedResults = results.map(
-                    (result, i) => Object.assign(result, {
-                        logSource: this.logReaders[i].getLogInfo(),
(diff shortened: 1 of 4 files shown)


rachedbenmustapha (approved):
PR #1234 approved!

lib/queuePopulator/QueuePopulator.js
```diff
@@ -236,20 +236,21 @@ class QueuePopulator {
...
-                        logOffset: this.logReaders[i].getLogOffset(),
-                    }));
-                return done(null, annotatedResults);
-            });
+            (logReader, readerDone) => {
+                const batchCb = (err, processedAll) => {
+                    if (err) {
+                        return readerDone(err);
+                    }
+                    if (processedAll || this.logReadersUpdate) {
```
vrancurel, line 246: is it sufficient to know if an update is in progress ? is it the circuit breaker ?

jonathan-gramain: Yes it's the circuit breaker, because it's checked by the calling function `QueuePopulator.processAllLogEntries`

naren-rajendran (approved):

vrancurel (approved):

PR #1244, opened 2020-12-12
bf: S3C-3675 duplicate messages fixed

naren-rajendran:
Corrected duplicate event message production on versioned buckets.

### DIFF
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -129,11 +129,9 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
      * @return {undefined}
      */
     _processObjectEntry(bucket, key, value, type) {
-        let versionId = null;
-        let objectKey = key;
+        const versionId = value.versionId || null;
         if (!isMasterKey(key)) {
-            objectKey = this._extractVersionedBaseKey(key);
-            versionId = value.versionId;
+            return undefined;
         }
         const config = this.bnConfigManager.getConfig(bucket);
         if (config && Object.keys(config).length > 0) {
@@ -146,7 +144,7 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
             }
             const ent = {
                 bucket,
-                key: objectKey,
+                key,
                 eventType,
                 versionId,
             };


extensions/notification/NotificationQueuePopulator.js
```diff
@@ -129,11 +129,9 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
      * @return {undefined}
      */
     _processObjectEntry(bucket, key, value, type) {
-        let versionId = null;
-        let objectKey = key;
+        const versionId = value.versionId || null;
         if (!isMasterKey(key)) {
-            objectKey = this._extractVersionedBaseKey(key);
-            versionId = value.versionId;
+            return undefined;
```
jonathan-gramain, line 134: I don't understand the logic here: don't you want to **keep** the versioned keys, not skip them? Because if you only consider the master keys, you don't get the version ID associated with them, correct? Moreover you would skip any noncurrent version.

naren-rajendran: I was checking this, whenever a new version of an object is created, we get two entries from the log (via the log reader), one with the new version created and another with update on master with the newest version id in its metadata attributes `value.versionId`. Since I am getting the version info on master I am skipping the version log entry.

jonathan-gramain: OK, I understand you're using the master key so that it works both for versioned and nonversioned buckets.

It might be okay but keep in mind that in some cases, you may not have a master update when you have a new or updated version: for example on a tags or ACL update on a specific version, you will only see this particular version in the log. Also backbeat, when it replicates on the target, will write versions without updating the master key. That is to say that, even if no such case has an impact today on bucket notifications, it might in the future, so it would be cleaner to use the version keys for this reason.

Practically, I'm thinking it might be possible to achieve this by checking whether the key has a version ID in its metadata: if not or if it is a null version (and it's a master key), process it (nonversioned bucket case), if it has a non-null version ID, skip if it is a master key. This is just OTOH I haven't checked that it actually works.

I'm ok to keep as is for now if what I described above does not apply to bucket notifications, but I think we should keep track of it in a jira.

jonathan-gramain (approved):

ilkescality (approved):

tmacro (approved):

PR #1312, opened 2021-01-09
Bugfix/s3 c 3686 fix tests

vrancurel:
Fix with tests this time.

### DIFF
--- a/lib/provisioning/ProvisionDispatcher.js
+++ b/lib/provisioning/ProvisionDispatcher.js
@@ -7,8 +7,6 @@ const zookeeper = require('node-zookeeper-client');
 
 const Logger = require('werelogs').Logger;
 
-const runningCI = process.env.CI === 'true';
-
 function _genRandomHexChars(len) {
     return crypto.randomBytes(Math.ceil(len / 2))
         .toString('hex') // convert to hexadecimal format
@@ -29,19 +27,21 @@ class ProvisionDispatcher {
      * @constructor
      * @param {object} zkConfig - zookeeper config object
      * @param {string} zkConfig.connectionString - zookeeper connection string
-     * where provisions are managed
      * (e.g. "localhost:2181/shared-tasks")
+     * @param {string} [_zookeeper] - zookeeper module
      */
-    constructor(zkConfig) {
+    constructor(zkConfig, _zookeeper) {
         this._zkEndpoint = zkConfig.connectionString;
-        this._client = zookeeper.createClient(zkConfig.connectionString);
-        this._client.connect();
+        this._zookeeper = (_zookeeper !== undefined) ? _zookeeper : zookeeper;
+        this._doRandDelay = false;
+        this._client = this._zookeeper.createClient(zkConfig.connectionString);
         this._connectWaitList = [];
         this._client.once('connected', () => {
             this._log.debug('connected to the ZK server');
             this._connectWaitList.forEach(cb => cb());
             this._connectWaitList = null;
         });
+        this._client.connect();
         this._myName = null;
         this._myLeaderName = null;
         this._isLeader = false;
@@ -49,12 +49,24 @@ class ProvisionDispatcher {
         this._provisions = null;
         this._redispatchInProgress = false;
         this._redoRedispatch = false;
+        this._interval = -1;
 
         this._log = new Logger('Backbeat:ProvisionDispatcher');
     }
 
+    /**
+     * Create artificial random delays (for tests)
+     *
+     * @param {boolean} [_doRandDelay] - introduce random delays
+     * where provisions are managed
+     * @return {undefined}
+     */
+    _setDoRandDelay(_doRandDelay = false) {
+        this._doRandDelay = _doRandDelay;
+    }
+
     _withRandomDelayIfTest(cb) {
-        if (runningCI) {
+        if (this._doRandDelay) {
             setTimeout(cb, Math.random() * 100);
         } else {
             cb();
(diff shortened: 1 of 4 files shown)


vrancurel:
Tests are not triggered automatically because the ft_test:prov used to be a functional test. Now it is a unit test. What shall we do  @jonathan-gramain ?

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -3,12 +3,11 @@
 const async = require('async');
 const crypto = require('crypto');
 const path = require('path');
+// const werelogs = require('werelogs');
```
jonathan-gramain: Let's just remove this line

naren-rajendran (approved):

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -28,32 +28,39 @@ class ProvisionDispatcher {
      * @constructor
      * @param {object} zkConfig - zookeeper config object
      * @param {string} zkConfig.connectionString - zookeeper connection string
+     * @param {string} _zookeeper - zookeeper module
```
jonathan-gramain: minor: you can specify those two params are optional: `[_zookeeper]`, `[_runningCI]`

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -28,32 +28,39 @@ class ProvisionDispatcher {
...
         this._myName = null;
         this._myLeaderName = null;
         this._isLeader = false;
         this._owners = null;
         this._provisions = null;
         this._redispatchInProgress = false;
         this._redoRedispatch = false;
+        this._interval = -1;
+
+        // werelogs.configure({ level: 'debug', dump: 'error' });
```
jonathan-gramain: line to be removed

tests/functional/ProvisionDispatcher.spec.js
```diff
@@ -2,14 +2,17 @@
 
 const assert = require('assert');
 const async = require('async');
-const zookeeper = require('node-zookeeper-client');
+const ZookeeperMock = require('zookeeper-mock');
+// const zk = require('node-zookeeper-client');
```
jonathan-gramain: line to be removed

tests/functional/ProvisionDispatcher.spec.js
```diff
@@ -44,12 +50,48 @@ function testDispatch() {
...
+                // introduce a new client
+                clients[1] = new ProvisionDispatcher(
+                    zkConf, zk);
+                clients[1].subscribe(err => {
+                    assert.ifError(err);
+                }, false);
+            } else {
+                // there might be intermediate callbacks but after the timer
+                // we shall get half of it
+                if (items.length === 4) {
```
jonathan-gramain, line 92: I'd suggest adding the explanation of why we get 4 items (I assume this is because there are now 2 clients that get an equal number of provisions each, but was not obvious for me).

jonathan-gramain (approved):

tests/unit/ProvisionDispatcher.spec.js
```diff
@@ -2,14 +2,16 @@
...
-const zookeeper = require('node-zookeeper-client');
+const ZookeeperMock = require('zookeeper-mock');
 
 const jsutil = require('arsenal').jsutil;
 const ProvisionDispatcher =
           require('../../lib/provisioning/ProvisionDispatcher');
 
 const ZK_TEST_PATH = '/tests/prov-test';
 
+const RAND_DELAY = true;
```
miniscruff: rand delay feels to me like a number value and not a boolean, like a random delay in seconds to do something.

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -28,32 +27,37 @@ class ProvisionDispatcher {
      * @constructor
      * @param {object} zkConfig - zookeeper config object
      * @param {string} zkConfig.connectionString - zookeeper connection string
+     * @param {string} [_zookeeper] - zookeeper module
+     * @param {boolean} [_runningCI] - introduce random delays
      * where provisions are managed
      * (e.g. "localhost:2181/shared-tasks")
```
miniscruff: I think these two lines were misplaced and are related to the connection string

lib/provisioning/ProvisionDispatcher.js
```diff
@@ -28,32 +27,37 @@ class ProvisionDispatcher {
...
      */
-    constructor(zkConfig) {
+    constructor(zkConfig, _zookeeper, _runningCI = false) {
```
miniscruff: I think splitting functionality by running in a unit test or not kinda strange. Maybe rename this to `maxRandomDelay` or introduce a "private" method to add this function outside the constructor.

tests/unit/ProvisionDispatcher.spec.js
```diff
@@ -44,12 +49,52 @@ function testDispatch() {
     });
 
     it('should be given all provisions when alone', done => {
-        clients[0] = new ProvisionDispatcher(zkConf);
+        const cbOnce = jsutil.once(done);
```
miniscruff, line 52: I think done is already restricted to only being called once https://mochajs.org/#detects-multiple-calls-to-done
Edit: You may not care to fail if it is called more than once, hadn't considered that...

lib/provisioning/ProvisionDispatcher.js
vrancurel: I could not remove `_zookeeper` since it is used in the constructor.

tests/unit/ProvisionDispatcher.spec.js
vrancurel: Yes it will be called more than once in this case.

PR #1339, opened 2021-01-15
bf: S3C-3824 apply updates made to queue populator

naren-rajendran:
Notification populator crashes with the recent changes made to queue populator and log reader. Applying the changes made to generic populator on notification populator fixed this.

### DIFF
--- a/bin/notification.js
+++ b/bin/notification.js
@@ -24,25 +24,20 @@ werelogs.configure({
 /* eslint-disable no-param-reassign */
 function queueBatch(queuePopulator, taskState) {
     if (taskState.batchInProgress) {
-        log.warn('skipping notification batch: previous one still in progress');
+        log.debug('skipping notification batch: previous one still in progress');
         return undefined;
     }
     log.debug('start queueing notification batch');
     taskState.batchInProgress = true;
     const maxRead = qpConfig.batchMaxRead;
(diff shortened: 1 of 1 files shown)


ilkescality (approved):

bin/notification.js
```diff
@@ -24,25 +24,20 @@ werelogs.configure({
...
         if (err) {
             log.error('an error occurred during notification', {
                 method: 'QueuePopulator::task.queueBatch',
                 error: err,
             });
-            return undefined;
         }
-        const logFunc = (counters.some(counter => counter.readRecords > 0) ?
-            log.info : log.debug).bind(log);
-        logFunc('notification batch finished', { counters });
```
rahulreddy, line 44: Is this log being shown somewhere else or are we completely removing the log?

naren-rajendran: It is being shown on log reader. Introduced in https://github.com/scality/backbeat/pull/1234/files#diff-ffc77ab2f45340e1e9848a1814fb0040f0457ef4b53c851288313c7f6b213dd1R226 .. though it says `replication batch finished`..should be generic I think

jonathan-gramain (approved):

bin/notification.js
rahulreddy: Ok please change that

PR #1346, opened 2021-01-18
bf: S3C-3829 Ignore placeholder log entries

naren-rajendran:
Notification populator should not process placeholder log entries created for deletion or cleanup.

### DIFF
--- a/extensions/notification/utils/config.js
+++ b/extensions/notification/utils/config.js
@@ -69,6 +69,14 @@ function filterConfigsByEvent(bnConfigs, event) {
  */
 function validateEntry(bnConfig, entry) {
     const { bucket, eventType } = entry;
+    /**
+     * if the event type is unavailable, it is an entry that is a
+     * placeholder for deletion or cleanup, these entries should be ignored and
+     * not be processed.
+    */
+    if (!eventType) {
+        return false;
+    }
     const notifConf = bnConfig.notificationConfiguration;
     // check if the entry belongs to the bucket in the configuration
     if (bucket !== bnConfig.bucket) {
--- a/tests/unit/notification/utils/config.js
+++ b/tests/unit/notification/utils/config.js
@@ -247,6 +247,15 @@ const tests = [
         },
         pass: false,
     },
+    {
+        desc: 'fail if the event type is unavailable',
+        entry: {
+            eventType: undefined,
+            bucket: 'bucket8',
+            key: 'abcd.png',
+        },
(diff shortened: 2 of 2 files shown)


anurag4DSB (approved):

miniscruff:
Bugfix without a test change?

anurag4DSB:
> Bugfix without a test change?

I thought it will go to integration, right?

naren-rajendran:
> > Bugfix without a test change?
> 
> I thought it will go to integration, right?

Yes.

rahulreddy:
> > > Bugfix without a test change?
> > 
> > 
> > I thought it will go to integration, right?
> 
> Yes.

There should be a test here for this (unit/functional/both), we can't wait for a Integration test to give us feedback if this is fixed or not. I understand having a test setup here is probably hard but we should not use that as an excuse to defer testing in Integration.

naren-rajendran:
@bert-e reset

naren-rajendran:
> > > > Bugfix without a test change?
> > > 
> > > 
> > > I thought it will go to integration, right?
> > 
> > 
> > Yes.
> 
> There should be a test here for this (unit/functional/both), we can't wait for a Integration test to give us feedback if this is fixed or not. I understand having a test setup here is probably hard but we should not use that as an excuse to defer testing in Integration.

I refactored backbeat to handle this in util function and added a test to it.

naren-rajendran:
@bert-e reset

jonathan-gramain (approved):

extensions/notification/utils/config.js
```diff
@@ -69,6 +69,14 @@ function filterConfigsByEvent(bnConfigs, event) {
  */
 function validateEntry(bnConfig, entry) {
     const { bucket, eventType } = entry;
+    /**
+     * if the event type is undefined, it is most likely an entry that is a
+     * placeholder for deletion or cleanup, these entries should be ignored and
+     * not be processed.
```
miniscruff: * You say "is undefined" but use a falsey check below. Should this be `eventType === undefined`?
* "most likely" is not a very concrete description, it sounds like there is a chance it's something else. What that else is I have no idea.

naren-rajendran: corrected.

naren-rajendran:
@bert-e reset

miniscruff (approved):

rahulreddy (approved):

PR #1358, opened 2021-01-19
bugfix: S3C-3769 Timeout for populator batches

jonathan-gramain:
In case a populator batch takes too long to execute fully, trigger a
timeout, which both logs it along with the step at which the current
batch is, and allow a new batch to start.

This is a workaround to ensure the queue populator can make progress,
and should help investigation whenever a case of timeout happens, to
know what calls are causing the timeout.

Add debugging info about the current step of the LogReader, to help
investigating at what step it may be stuck.

Tested manually:

- introduced a random hook preventing a callback to be called during a batch processing, up to 10 times
- upgraded backbeat
- checked that backbeat logs the step at which it stopped, allows new batches to restart and keeps going

Note: Zenko does not have the workaround, the codebase is too different.

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -10,6 +10,8 @@ const ReplicationQueuePopulator =
 const { metricsExtension, metricsTypeQueued } =
     require('../../extensions/replication/constants');
 
+const BATCH_TIMEOUT_SECONDS = 300;
+
 class LogReader {
 
     /**
@@ -170,6 +172,8 @@ class LogReader {
      *   from the log. Records may contain multiple entries and all entries
      *   are not queued, so the number of queued entries is not directly
      *   related to this number.
+     * @param {function} [params.onTimeout] - optional callback,
+     *   called when a single batch times out
      * @param {function} done - callback when done processing the
      *   entries. Called with an error, or null and a statistics object as
      *   second argument. On success, the statistics contain the following:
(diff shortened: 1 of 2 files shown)


naren-rajendran:
Should the same be applied to bucket notification?

jonathan-gramain:
@naren-scality good question, I would say yes. But don't we have a plan to have bucket notifications use the current queue populator code eventually (mid-term)?

Edit: if you're talking about modifying the `bin/notification.js` binary, yes we may do this as well indeed. I'd rather create a separate PR since it only applies to development/7.9.

jonathan-gramain:
Meaning, I'd rather see this refactor done sooner than later, the more we wait the more the code bases will be slightly different in subtle ways, increasing maintenance cost and possibly leading to subtle bugs.

jonathan-gramain:
For the notification populator: #1362

lib/queuePopulator/LogReader.js
```diff
@@ -226,6 +241,7 @@ class LogReader {
                 this.log.info('replication batch finished', {
```
naren-rajendran, line 241: I think you need to rebase, I recently modified the message `replication batch finished` to `batch finished` or you could resolve conflict :)

jonathan-gramain: Ah, I think it will be alright the merge should be automatic. If need I'll do it for sure.

lib/queuePopulator/LogReader.js
```diff
@@ -191,15 +195,26 @@ class LogReader {
             },
             entriesToPublish: {},
             publishedEntries: {},
+            debugStep: '[INIT]',
```
naren-rajendran, line 198: Minor, why square brackets?

naren-rajendran (approved):

lib/queuePopulator/LogReader.js
jonathan-gramain: It's just to make the step more visible

alexanderchan-scality (approved):

PR #1370, opened 2021-01-22
bf: S3C-3783 notification manager crash

naren-rajendran:
Bucket notification specific zookeeper if not created via federation causes notification configuration manager to error out, resulting in a crash. This should ideally exit process but not with a crash. If this happens the process should exit out (without a crash).

### DIFF
--- a/extensions/notification/NotificationConfigManager.js
+++ b/extensions/notification/NotificationConfigManager.js
@@ -230,7 +230,8 @@ class NotificationConfigManager {
             if (err) {
                 const errMsg
                     = 'error checking configuration zookeeper parent node';
-                this.log.error(errMsg, { method, zkPath });
+                this.log.error(errMsg, { method, zkPath, error: err.message });
+                return this._callbackHandler(cb, err);
             }
             this.log.debug('parent configuration zookeeper checked/added',
                 { method, zkPath });
@@ -418,7 +419,10 @@ class NotificationConfigManager {
      * @return {undefined}
      */
     init(cb) {
-        return this._checkConfigurationParentNode(() => {
+        return this._checkConfigurationParentNode(err => {
+            if (err) {
+                return cb(err);
+            }
             return this._listBucketsWithConfig((err, buckets) => {
                 if (err) {
                     return cb(err);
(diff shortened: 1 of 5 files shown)


naren-rajendran:
@bert-e reset

extensions/notification/NotificationConfigManager.js
```diff
@@ -418,7 +418,10 @@ class NotificationConfigManager {
      * @return {undefined}
      */
     init(cb) {
-        return this._checkConfigurationParentNode(() => {
+        return this._checkConfigurationParentNode(err => {
+            if (err) {
+                return cb(err);
+            }
```
miniscruff, line 425: can this be tested? Same with the other.

naren-rajendran: I was thinking about it, the error was uncaught in the original case (the bug) and that was the reason for the crash.

miniscruff: It should be possible to add a test that causes an error and assert the error is called in the callback. That is usually how I test error checks like this.

naren-rajendran: .

miniscruff (approved):

jonathan-gramain (approved):
The PR will need to be retargeted to development/7.10.

tests/unit/notification/NotificationConfigManager.js
```diff
@@ -118,6 +124,17 @@ describe('NotificationConfigManager', () => {
...
             ], done);
         });
+
+        it('should return error if checkConfigParentNode fails', done => {
+            sinon.stub(NotificationConfigManager.prototype,
+                '_checkConfigurationParentNode')
+                .callsFake(checkConfigParentNodeStub);
+            const manager = new NotificationConfigManager(params);
+            manager.init(err => {
+                assert(err);
```
miniscruff, line 134:
```suggestion
                assert.ifError(err);
```
https://nodejs.org/api/assert.html#assert_assert_iferror_value

naren-rajendran: The test is expecting an error to be returned.

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e force_reset

naren-rajendran:
@bert-e approve

PR #1373, opened 2021-01-22
bg: S3C-3686 remove random delay from main code

vrancurel:
Instead use the feature from the mock. It makes the code more readable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,6 @@
   },
   "devDependencies": {
     "mocha": "^3.3.0",    
-    "zookeeper-mock": "^1.1.0"
+    "zookeeper-mock": "^1.2.0"
   }
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -49,6 +49,6 @@
   },
   "devDependencies": {
     "mocha": "^3.3.0",    
-    "zookeeper-mock": "^1.1.0"
+    "zookeeper-mock": "^1.1.1"
```
vrancurel: Note the `npm` is not yet created so tests won't pass.

miniscruff:
No issues on this side of things, will re-review when the dependency is approved.

miniscruff (changes requested):
Marking as needs work as it gives me the best notification when a change is made...

miniscruff (approved):
Approving even though the dependency isn't updated as I already reviewed that one too

naren-rajendran (approved):

jonathan-gramain (approved):

PR #1374, opened 2021-01-23
bugfix: S3C-3833 refresh role credentials earlier

jonathan-gramain:
Instead of refreshing credentials only after they expire, refresh them
a bit earlier (one minute earlier than the expiration time by
default).

This is to avoid a race condition where backbeat would use credentials
just about to expire, sending a request to the server with expired
credentials from the server point of view when the request arrives
(due to network delay and/or time skew between client and server).

### DIFF
--- a/lib/credentials/RoleCredentials.js
+++ b/lib/credentials/RoleCredentials.js
@@ -8,6 +8,7 @@ const configJoi = {
     vaultclient: joi.object().required(),
     extension: joi.string().required(),
     roleArn: joi.string().required(),
+    refreshCredsAnticipationSeconds: joi.number().greater(0).default(60),
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lib/credentials/RoleCredentials.js
```diff
@@ -8,6 +8,7 @@ const configJoi = {
     vaultclient: joi.object().required(),
     extension: joi.string().required(),
     roleArn: joi.string().required(),
+    refreshCredsAnticipationMs: joi.number().default(60000),
```
miniscruff: Not sure how long the creds last but I would set 1 minute as the minimum if JOI supports it. Also for times in the minute range or higher, I think using seconds or even minutes would work better than milliseconds.

jonathan-gramain: I think one minute is pretty conservative and looks to me a good default, but I'd rather not enforce it as a minimum, and also unit tests set their own lower value to avoid the test being too long.

I agree that using seconds instead of milliseconds could make more sense.

miniscruff: I am not sure the expected timing of refreshing but maybe a minimum of 1 or like 5 seconds to at least avoid someone incorrectly setting the time to 0 or somehow negative? But these types of validations might be atypical for our projects so it can be skipped.
Not sure how much we protect admins from self-destruction is all.

naren-rajendran (approved):

lib/credentials/RoleCredentials.js
alexandre-merle: I think a check for a value greater than 0 should be enough, still there is a potential case for a value > expiration time

jonathan-gramain: changed this, please review the latest commit

miniscruff (approved):

PR #1402, opened 2021-02-03
improvement: S3C-3835 start populating from latest cseq

jonathan-gramain:
Instead of starting populating entries from the beginning of each raft
session, start from the current cseq offset to skip existing data
written before the initial installation of the queue populator
(e.g. after enabling CRR or bucket notifications on an existing
installation).

This avoids having to go through an entire raft session log before
beginning to populate new entries.

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -116,7 +116,7 @@ class LogReader {
                                 error: err });
                         return done(err);
                     }
-                    return done(null, 1);
+                    return this._initializeLogOffset(done);
                 });
             }
             if (data) {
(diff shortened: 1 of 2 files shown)


lib/queuePopulator/LogReader.js
```diff
@@ -136,7 +136,31 @@ class LogReader {
                         logOffset });
                 return done(null, logOffset);
             }
-            return done(null, 1);
+            return this._initializeLogOffset(done);
```
naren-scality, line 139: Should this be specific to notifications or for all extensions using the `logReader`?

naren-scality: I assume this should also be applicable replication!

naren-rajendran (approved):

lib/queuePopulator/LogReader.js
naren-scality: LGTM.

alexandre-merle (approved):

PR #1410, opened 2021-02-04
ZENKO-3107 setting up release stage for backbeat

tcarmet:


docs/RELEASE.md
```diff
@@ -0,0 +1,37 @@
+# Backbeat Release Plan
```
miniscruff, line 1: No issues here but I do want to add that this is all very generic for JS projects and could also be defined in guidelines or citadel.

tcarmet: I agree, but what should we do?

Should we make a section on citadel explaining the release "template" for our JS projects. And then still copy paste the content and adapt it to this project.

Or Should I rather make a template on citadel, and say, to release backbeat follow this link on citadel.

I'm thinking about option 1, but additional advice is very welcome.

jonathan-gramain (approved):

docs/RELEASE.md
miniscruff: Because we are splitting up all our projects into separate repos I think we should either ( more or less your ideas )
A) Use citadel and create links for relevant files, eg release, contributing, getting help, etc
B) Create a more abstract process for not only copying these files to each project but making sure they remain in line with what is said.

The issues with A comes from potentially changing behavior without updating relevant projects.
The issues with B come from manually dealing with potentially a hundred or so seemingly connected documentation files that may or may not reflect our best practice at the time or the project itself.

( There is definitely the possibility of a generic tool to help with B like a dependabot for docs which I could see being a nice open-source thing )
This is one of the advantages of creating a monorepo as this problem does not exist.

I do prefer B as well as it would give us the option of updating or not if we choose to for outdated or deprecated projects, but I do believe this will be more work than A.

miniscruff: Either way, I think it is worth a wider discussion

anurag4DSB (approved):

docs/RELEASE.md
tcarmet: B is definitely more work, but I would be fine with either of the suggested solutions so it's very worth some thinking and a wider discussion.  I would suggest starting with A, and improve to B if A becomes an issue. 

I therefore created this ticket [OS-6](https://scality.atlassian.net/browse/OS-6) to address this issue. (Going towards implementing A) Feel free to edit the description or add more content to it.

I also don't want to consider this subject as a blocker to merge this PR. (if you agree) And tackle it in it's own time.

Let me know your thoughts!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "8.1.1",
+  "version": "8.1.21",
```
anurag4DSB, line 3: Should this be `8.1.2`?

tcarmet: Considering the tags made on backbeat, I went for `8.1.21` as the last `8.1` release is `8.1.20`.

Also weirdly the tag `8.1.2` does not exist, so your argument would be valid, but as we had `8.1.(2 + x)` and so on releases I suggest we go for `8.1.21`

anurag4DSB: cool. I just thought maybe this is a typo. :)

anurag4DSB (approved):

package.json
tcarmet: Knowing myself there was a 50% chance (at least) that this would be a typo.

docs/RELEASE.md
miniscruff: oh yea for sure not a blocker, I just wanted to make sure this sort of conversation developed before the ticket was merged so I didn't "lose" it.

miniscruff (approved):

PR #1439, opened 2021-02-12
improvement: S3C-3984 set default kinit command

naren-rajendran:
Set default kinit command for bucket notification kafka producer. This is an optional parameter for SASL mechanism. By providing the default command, it ensures usage of the principal and keytab provided and not use anything in cache on producer instantiation.

### DIFF
--- a/extensions/notification/utils/auth.js
+++ b/extensions/notification/utils/auth.js
@@ -24,8 +24,7 @@ function getAuthFilePath(fileName) {
 function generateKafkaAuthObject(auth) {
     const authObject = {};
     const { supportedAuthTypes } = constants;
-    // TODO: auth object is fluid at the moment, once a solid structure is
-    // defined, introduce a schema
+    // TODO: S3C-3985 auth object should be validated/checked
     const {
         type,
         ssl,
@@ -64,6 +63,9 @@ function generateKafkaAuthObject(auth) {
         const keytabPath = getAuthFilePath(keytab);
         if (keytabPath) {
             authObject['sasl.kerberos.keytab'] = keytabPath;
+            // default kinit command
+            const kinitCommand = `kinit -k ${principal} -t ${keytabPath}`;
+            authObject['sasl.kerberos.kinit.cmd'] = kinitCommand;
         }
     }
     return authObject;


extensions/notification/utils/auth.js
```diff
@@ -64,6 +64,10 @@ function generateKafkaAuthObject(auth) {
         const keytabPath = getAuthFilePath(keytab);
         if (keytabPath) {
             authObject['sasl.kerberos.keytab'] = keytabPath;
+            //optional parameter, default kinit command
+            const kinitCommand = `kinit -R -k ${principal} -t ${keytabPath} ||
```
jonathan-gramain: just double checking, is this command always run in a shell environment? in order to interpret the double command and `||`.

jonathan-gramain (approved):

extensions/notification/utils/auth.js
naren-rajendran: Yes.

naren-rajendran: Had to check it, I think using just `kinit -k -t...` should be fine. Changing it.

naren-rajendran:
@bert-e reset

tmacro (approved):

extensions/notification/utils/auth.js
```diff
@@ -64,6 +64,9 @@ function generateKafkaAuthObject(auth) {
         const keytabPath = getAuthFilePath(keytab);
         if (keytabPath) {
             authObject['sasl.kerberos.keytab'] = keytabPath;
+            // default kinit command
+            const kinitCommand = `kinit -k ${principal} -t ${keytabPath}`;
+            authObject['sasl.kerberos.kinit.cmd'] = kinitCommand;
```
rahulreddy, line 68: It's very important to address the TODO comment 
```
    // TODO: auth object is fluid at the moment, once a solid structure is
    // defined, introduce a schema
 ```
 , as I see there is no check to ensure principal or keytabPath are defined.

naren-rajendran: Agreed, it was an unknown when we started development, now that we have a good understanding of the parameters involved, we should enforce schema or checks. I will create a new JIRA ticket/PR for this.

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e approve

PR #1524, opened 2021-03-17
bugfix: ZENKO-2880 convert content length to integer

jonathan-gramain:
Accept content-length string values as well as integer values by
converting to a number before applying to replication metrics.

### DIFF
--- a/extensions/replication/ReplicationMetrics.js
+++ b/extensions/replication/ReplicationMetrics.js
@@ -93,22 +93,23 @@ class ReplicationMetrics extends ZenkoMetrics {
             origin: originLabel,
             fromLocation, fromLocationType,
             toLocation, toLocationType, partition,
-        }, contentLength);
+        }, Number.parseInt(contentLength, 10));
     }
 
     static onReplicationProcessed(originLabel, fromLocation, toLocation,
                                   contentLength, status, elapsedMs) {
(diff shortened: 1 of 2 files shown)


tmacro (approved):

alexanderchan-scality (approved):

miniscruff:
From the spec https://tools.ietf.org/html/rfc7230#section-3.3.2 I think content-length should be a number and instead of supporting strings we should find what sent the string and send a number instead. Assuming it is ours. Supporting more here doesn't seem like the right idea.

naren-rajendran (approved):

jonathan-gramain:
@miniscruff indeed, I agree with this, sorry for not giving all context, I opened https://scality.atlassian.net/browse/ZENKO-3298 to track this, but we need a quick fix for now for a POC in progress.

jonathan-gramain:
As far as I can tell, the stringified content length comes from object metadata (as returned by `ObjectMD.getContentLength()` it must already be a string), so the question would be why is there a stringified content length in metadata.

miniscruff:
> @miniscruff indeed, I agree with this, sorry for not giving all context, I opened https://scality.atlassian.net/browse/ZENKO-3298 to track this, but we need a quick fix for now for a POC in progress.

ok, creating a PR to revert this change but marking as hold will be good as well.

PR #1526, opened 2021-03-17
Add bucket scanner base data types

rachedbenmustapha:
First piece of the upcoming bucket scanner:

- data types for object metadata
- data types for atomic counters
- data types for CRR metrics groups per object status * per backend

### DIFF
--- a/bucket-scanner/pkg/counter/replication/counter_set.go
+++ b/bucket-scanner/pkg/counter/replication/counter_set.go
@@ -1,0 +1,119 @@
+package replication
+
+import (
+	"sync"
+
+	"github.com/scality/backbeat/bucket-scanner/pkg/types"
+)
+
+type (
+	CounterSet struct {
+		sync.Mutex
+		Global           *ByStatusCounterSet
+		ByTargetLocation map[types.SiteName]*ByStatusCounterSet
+
+		name string
+	}
+
+	SerializedCounterSet struct {
+		Global     *SerializedByStatusCounterSet                    `json:"global"`
+		ByLocation map[types.SiteName]*SerializedByStatusCounterSet `json:"byLocation"`
+	}
+)
+
+func New(name string) *CounterSet {
+	return &CounterSet{
+		Global:           NewByStatusCounterSet(name + ":global"),
+		ByTargetLocation: make(map[types.SiteName]*ByStatusCounterSet),
+
+		name: name,
+	}
+}
+
+func (cs *CounterSet) CopyTo(other *CounterSet) {
+	other.Lock()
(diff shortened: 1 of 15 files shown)


miniscruff (approved):

miniscruff (approved):
No changes with my approval as it was just split.

bucket-scanner/pkg/counter/replication/counter_set.go
```diff
@@ -0,0 +1,119 @@
...
+		byLocation[location] = bscs.Serialize()
+	}
+
+	return &SerializedCounterSet{
+		Global:     cs.Global.Serialize(),
+		ByLocation: byLocation,
+	}
+}
+
+func (cs *CounterSet) LoadSerialized(s *SerializedCounterSet) {
```
alexanderchan-scality, line 100: will `LoadSerialized` always be called in a safe manner?

Please correct me if I am wrong, with the lock claim/release being done per location, the result may be unexpected  if the `LoadSerialized` is called along other `CounterSet` write methods.

alexanderchan-scality (approved):

bucket-scanner/pkg/counter/replication/counter_set.go
```diff
@@ -0,0 +1,119 @@
...
+	cs.Global.LoadSerialized(s.Global)
+
+	for k, v := range s.ByLocation {
+		l := cs.GetTargetLocation(k)
+		l.LoadSerialized(v)
+	}
+}
```
alexanderchan-scality, line 107: please correct me if I am wrong.

With the lock being claimed/released per location as opposed on the method, it would seem like a race condition may occur if the `LoadSerialized` call is followed by a write method like `Add`, leading to an unexpected final state.

A read method after a `LoadSerialized` can also result in an invalid output as well? 

this seems to be a non-issue, if `LoadSerialized` is always called safely

rachedbenmustapha: I don't think it's an issue because:

- It's only used to deserialize values from storage at init time, before the `CounterSet` gets passed around.
- Actual modification is done in a thread-safe manner using atomic ops, so the risk is that global counters would be slightly off compared to the sum of the regions on a functional level, not actual unsafe memory access that could be seen/reported by the race detector. Since this whole counting system is eventually consistent and there can be some known (temporary) double-counting happening, no one would notice :)
- The `Add` method should, if that was a use case to support, use atomic addition ops instead of using `Set(Get()+Get())`, making the pattern you describe safe.

But thanks for noticing, it's a valid concern.

PR #1551, opened 2021-03-22
ft: S3C-4191 implement prom exporter over http

rachedbenmustapha:


bucket-scanner/pkg/exporter/exporter_test.go
```diff
@@ -0,0 +1,318 @@
...
+				# TYPE default_backbeat_replication_versions gauge
+				default_backbeat_replication_versions{status="completed"} 2
+				default_backbeat_replication_versions{status="failed"} 4
+				default_backbeat_replication_versions{status="nostatus"} 6
+				default_backbeat_replication_versions{status="pending"} 8
+				default_backbeat_replication_versions{status="processing"} 12
+				default_backbeat_replication_versions{status="replica"} 10
+			`
+
+			for _, s := range strings.Split(expectExported, "\n") {
```
miniscruff, line 295: Does the first and last line get included in the Split here? I know in JS if you use the multiline strings they do, not that it should matter.

miniscruff (approved):

bucket-scanner/pkg/exporter/exporter_test.go
rachedbenmustapha: The first and last line are indeed included by `strings.Split`, and `strings.TrimSpace` makes them empty strings that always match. Can be seen as sloppy but doesn't matter much, as you say :)

miniscruff: haha yea that is what I was thinking. The newlines make the text block easier to read so I am ok with it. I thought creating a string array would be better but then all the quote and commas... so it is what it is.

bucket-scanner/pkg/exporter/exporter.go
```diff
@@ -0,0 +1,197 @@
...
+		ctx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+		defer cancel()
+
+		err := s.Shutdown(ctx)
+		if err != nil && !errors.Is(err, ctx.Err()) {
+			log.Fatal(err)
+		}
+	}()
+
+	// mirror our metrics to the auto-exported prometheus metrincs
```
alexanderchan-scality:
```suggestion
	// mirror our metrics to the auto-exported prometheus metrics
```

bucket-scanner/pkg/exporter/exporter.go
```diff
@@ -0,0 +1,197 @@
...
+	go func() {
+		defer wg.Done()
+
+		err := s.ListenAndServe()
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
+			log.Fatal(err)
+		}
+	}()
+
+	// shutdown http server on ctx cancelation
```
alexanderchan-scality, line 174:
```suggestion
	// shutdown http server on ctx cancellation
```

alexanderchan-scality (approved):

bucket-scanner/pkg/exporter/exporter.go
rachedbenmustapha: Both are valid and "cancellation" is British-preferred. Since we usually favor American variants I'll leave this one :)

rachedbenmustapha:
@bert-e help

rachedbenmustapha:
@bert-e status

PR #1566, opened 2021-03-25
feature: S3C-4214 Update getting started

miniscruff:
Work on improving the readme:
* Incorrect script file path
* Instead of explaining how to install external software just link
* Bad import
* Remove broken CI links
* Use backbeatuser instead of Bart and Lisa

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,6 @@
 
 ![backbeat logo](res/backbeat-logo.png)
 
-[![CircleCI][badgepub]](https://circleci.com/gh/scality/backbeat)
-[![Scality CI][badgepriv]](http://ci.ironmann.io/gh/scality/backbeat)
-
 ## OVERVIEW
 
 Backbeat is an engine with a messaging system at its heart.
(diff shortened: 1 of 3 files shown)


extensions/replication/utils/SetupReplication.js
```diff
@@ -127,6 +127,7 @@ class SetupReplication extends BackbeatTask {
                 source.transport,
                 `${source.vault.host}:${source.vault.adminPort}`,
                 source.credentials, internalHttps),
+            // What is the proper way to use vault with the target?
             target: target.isExternal ? undefined : _setupIAMClient(
                 target.transport, `${destHost.host}:${destHost.port}`,
```
miniscruff, line 132: I had to change the host and port to use the source vault host and admin port in order to get this to work locally. I am not sure if there is a better method, or maybe it is not supported? I was maybe using a non-typical setup of replicating to the same cloud server on different buckets instead of two different instances.

conf/config.json
```diff
@@ -1,15 +1,15 @@
...
+        "autoCreateNamespace": true
     },
     "kafka": {
         "hosts": "127.0.0.1:9092"
     },
     "queuePopulator": {
         "cronRule": "*/5 * * * * *",
         "batchMaxRead": 10000,
-        "zookeeperPath": "/queue-populator",
+        "zookeeperPath": "/replication-populator",
```
jonathan-gramain, line 12: the populator can be used for other purposes than replication like lifecycle, hence may be better to leave it as is.

README.md
```diff
@@ -101,117 +64,45 @@ We may leave the zookeeper server now:
...
-```
-git clone https://github.com/scality/Vault ~/replication/vault && \
-cd ~/replication/vault && \
-yarn i && \
-chmod 400 ./tests/utils/keyfile && \
-VAULT_DB_BACKEND=MEMORY node vaultd.js
-```
-
-In a new shell, start the Scality S3 server:
+#### Start Vault and Scality Cloudserver
```
jonathan-gramain, line 69: By using the `"type":"account"` method of authentication, vault should not be needed. Moreover since Vault is a private component like stated, to me it does not make much sense in an open source guide, unless the user is already a Scality customer. I believe we could instead document a true open source setup without Vault, with backbeat using account credentials to access the source and target.

extensions/replication/utils/SetupReplication.js
```diff
@@ -127,6 +127,7 @@ class SetupReplication extends BackbeatTask {
                 source.transport,
                 `${source.vault.host}:${source.vault.adminPort}`,
                 source.credentials, internalHttps),
+            // What is the proper way to use vault with the target?
```
jonathan-gramain, line 130: On S3C, it goes to the nginx proxy listening on port 9080, and gets redirected internally on the target to the appropriate Vault port.

On an open source deployment, I believe SetupReplication is not appropriate currently due to its reliance on Vault (unless `isExternal` which is meant for Zenko public cloud use cases), but we could consider modifying SetupReplication for open source use cases, which I think would be simply setting up the buckets and replication configurations without setting up roles.

conf/config.json
miniscruff: oh good point, I can adjust the readme to use this name then.

extensions/replication/utils/SetupReplication.js
miniscruff: Hmm I am not sure that will work, the first parameter in the put bucket replication is the role https://docs.aws.amazon.com/cli/latest/reference/s3api/put-bucket-replication.html#examples

jonathan-gramain: I believe the role is not used in account credentials, or if it is, it's just to retrieve the account ID or something.

Also keep in mind that we hacked a little bit the role field to include both the source AND target replication roles (because we don't have a federated Vault, hence two different accounts on source and target). For this, we use two roles separated by a comma, instead of a single role.

In backbeat, the file extensions/replication/tasks/ReplicateObject.js contains some code related to parsing the role from the replication configuration.

extensions/replication/utils/SetupReplication.js
miniscruff, line 132:
```suggestion
                target.transport, `${source.vault.host}:${source.vault.adminPort}`,
```
Change this line to use your local vault instead of the destination host and port. Temp hack, do not merge.

PR #1643, opened 2021-04-26
feature: S3C-4073: backport probe servers

alexandre-merle:
backporting probe servers from 8.x

### DIFF
--- a/bin/queuePopulator.js
+++ b/bin/queuePopulator.js
@@ -12,7 +12,9 @@ const httpsConfig = config.internalHttps;
 const mConfig = config.metrics;
 const rConfig = config.redis;
 const QueuePopulator = require('../lib/queuePopulator/QueuePopulator');
+const zookeeper = require('node-zookeeper-client');
 
+const { HealthProbeServer } = require('arsenal').network.probe;
 const log = new werelogs.Logger('Backbeat:QueuePopulator');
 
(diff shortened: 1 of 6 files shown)


alexandre-merle:
@bert-e reset

bin/queuePopulator.js
```diff
@@ -59,6 +66,19 @@ async.waterfall([
         });
         done();
     },
+    done => {
+        healthServer.onReadyCheck(log => {
+            const state = queuePopulator.zkStatus();
+            if (state.code === zookeeper.State.SYNC_CONNECTED.code) {
+                return true;
+            }
+            log.error(`Zookeeper is not connected! ${state}`);
```
miniscruff, line 75: I have tried to debug backbeat and our other services multiple times where the liveness check that exists fails with no indication as to which part of the check actually failed. It is just some magic that you have to know when a liveness fails here it is probably related to X service. I think it would be much better to return some sort of error response indicating which check failed. The current HealthProbeServer doesn't really allow us to do that though. I did a quick redesign locally to see how we might go about it though and it does not seem too bad. 

See the example here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request
I think it would be especially important when we get multiple checks.

miniscruff:
Closing this as we moved to the ProbeServer I created in Arsenal

PR #1692, opened 2021-05-20
feature: S3C-4345

miniscruff:
Add prom-client and metrics for queue populator

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -31,6 +31,8 @@ class LogReader {
      *   queue populator extension modules
      * @param {MetricsProducer} params.metricsProducer - instance of metrics
      *   producer
+     * @param {MetricsHandler} params.metricsHandler - instance of metrics
+     *   handler
      */
     constructor(params) {
         this.zkClient = params.zkClient;
@@ -42,6 +44,7 @@ class LogReader {
         this._producers = {};
         this._extensions = params.extensions;
         this._mProducer = params.metricsProducer;
+        this._metricsHandler = params.metricsHandler;
     }
 
     _setEntryBatch(entryBatch) {
@@ -197,7 +200,8 @@ class LogReader {
                         zkPath: pathToLogOffset,
                         logOffset: this.logOffset });
                 return done();
-            });
+            }
+        );
     }
 
     /**
@@ -330,6 +334,7 @@ class LogReader {
             bucket: record.db,
             key: _transformKey(entry.key),
             value: entry.value,
+            logReader: this,
         }));
     }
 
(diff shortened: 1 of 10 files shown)


lib/queuePopulator/LogReader.js
```diff
@@ -423,16 +428,29 @@ class LogReader {
...
+                        {
+                            method: 'LogReader._processPublishEntries',
+                            topic,
+                            entryCount: topicEntries.length,
+                            error: err,
+                        }
+                    );
+                    const labels = Object.assign({},
+                        this.getMetricLabels(),
+                        { publishStatus: 'FAILURE' }
```
naren-scality, line 440: Do we have constants describing the status, if yes, we should use them instead!

naren-rajendran, line 440: Do we need constants for the status?

miniscruff: When we add status processor status values we should standardize terms, for now I think it is fine.

lib/queuePopulator/QueuePopulator.js
```diff
@@ -11,6 +9,102 @@ const MetricsProducer = require('../MetricsProducer');
...
+const logSizeMetric = new promClient.Gauge({
+    name: 'replication_log_size',
+    help: 'Current size of metadata journal',
+    labelNames: metricLabels,
+});
+
+const messageMetrics = new promClient.Counter({
+    name: 'replication_populator_messages',
+    help: 'Total number of Kafka messages produced by the queue populator',
+    labelNames: ['origin', 'logName', 'logId', 'publishStatus'],
```
naren-rajendran: Don't we need containerName here?

lib/queuePopulator/QueuePopulator.js
```diff
@@ -11,6 +9,102 @@ const MetricsProducer = require('../MetricsProducer');
...
+/**
+ * Contains methods to incrememt different metrics
+ * @typedef {Object} MetricsHandler
+ * @property {CounterInc} messages - Increments the message metric
+ * @property {CounterInc} objects - Increments the objects metric
+ * @property {CounterInc} bytes - Increments the bytes metric
+ * @property {GaugeSet} logReadOffset - Set the log read offset metric
+ * @property {GaugeSet} logSize - Set the log size metric
+ */
+const metricsHandler = {
```
naren-rajendran, line 101: Should metricsHandler be a class?

naren-rajendran:
Tests? and a minor comment on metricsHandler, otherwise LGTM.

lib/queuePopulator/QueuePopulator.js
miniscruff: Metrics handler is just a collection of metric methods, there is no internal state or parameters to pass so I feel a class is unnecessary.

naren-rajendran: are the functions reusable for processor, we dont want to duplicate!

miniscruff: the GaugeSet and CounterInc methods and types will be for sure, maybe a metrics util file will come in handy. metricsHandler itself will need to be per process but it should be relatively similar across the board.

naren-rajendran: Yes, I think a metrics util should be in place.

miniscruff:
ping

lib/tasks/BackbeatTask.js
```diff
@@ -31,33 +31,11 @@ class BackbeatTask {
         const startTime = Date.now();
         const self = this;
 
-        // FIXME workaround for S3C-4457:
```
naren-rajendran: Is this right? is the block really removed?

tests/unit/lib/tasks/BackbeatTask.spec.js
```diff
@@ -1,44 +0,0 @@
-const assert = require('assert');
```
naren-rajendran: Same here, why is the test removed?

lib/tasks/BackbeatTask.js
miniscruff: uhh.... well.... that was not supposed to happen. Let me try this again...

naren-rajendran (approved):

rahulreddy (approved):

PR #1700, opened 2021-05-25
bugfix: S3C-4457 workaround to catch duplicate callback

jonathan-gramain:
Add a workaround in BackbeatTask to catch duplicate callback calls and
avoid a process crash.

The suspicion is that the AWS SDK client may call its callback
multiple times from the send() method, but it was not reproduced so
far.

### DIFF
--- a/lib/tasks/BackbeatTask.js
+++ b/lib/tasks/BackbeatTask.js
@@ -31,11 +31,33 @@ class BackbeatTask {
         const startTime = Date.now();
         const self = this;
 
+        // FIXME workaround for S3C-4457:
+        //
+        // It seems the S3 client may call its callback multiple times
+        // in an unknown corner case (the callback passed to the
+        // send() function, like in ReplicateObject._setupRolesOnce()).
+        //
+        // Until we find the root cause, we catch duplicate calls and
+        // log them instead of crashing the process with an exception
+        // raised from the async module.
+        //
+        let cbCalled = false;
+        const doneOnce = function doneWrapper(...args) {
+            if (!cbCalled) {
+                cbCalled = true;
+                done.apply(done, args);
+            } else {
+                log.warn('callback was already called', Object.assign({
+                    method: 'BackbeatTask.retry',
+                }, logFields || {}));
+            }
+        };
+
         function _handleRes(...args) {
             const err = args[0];
             if (err) {
                 if (!shouldRetryFunc(err)) {
-                    return done(err);
+                    return doneOnce(err);
                 }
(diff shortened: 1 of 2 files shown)


lib/tasks/BackbeatTask.js
```diff
@@ -31,11 +31,33 @@ class BackbeatTask {
...
+        //
+        // It seems the S3 client may call its callback multiple times
+        // in an unknown corner case (the callback passed to the
+        // send() function, like in ReplicateObject._setupRolesOnce()).
+        //
+        // Until we find the root cause, we catch duplicate calls and
+        // log them instead of crashing the process with an exception
+        // raised from the async module.
+        //
+        let cbCalled = false;
```
anurag4DSB, line 44: Should we use `jsutil.once`?

https://github.com/scality/Arsenal/blob/a1c0dd24723be7f63a91056df98ccd8a256e4ef3/lib/jsutil.js#L20

jonathan-gramain: Not a good fit here, we want to log a warning when the duplicate happens, which jsutil doesn't do (it considers it a "normal" case).

tmacro (approved):

tests/unit/lib/tasks/BackbeatTask.spec.js
```diff
@@ -0,0 +1,42 @@
+const assert = require('assert');
+
+const werelogs = require('werelogs');
+
+const BackbeatTask = require('../../../../lib/tasks/BackbeatTask');
+
+const logger = new werelogs.Logger('BackbeatTask:test');
+
+describe('BackbeatTask', () => {
+    it('BackbeatTask.retry() should handle and trace double callback (S3C-4457)', done => {
```
naren-rajendran: why do we need a ref to the ticket here?

jonathan-gramain: Because the fix is actually a workaround, I anticipate that we would remove the workaround (and this test) once we find and fix the root cause, hence leaving the ticket number here can help to track this.

naren-rajendran: Would it make sense to have that information as a comment (TODO)? It looks a little odd in the test description!

jonathan-gramain: Indeed it could be better, will do

naren-rajendran (approved):

anurag4DSB (approved):

PR #1710, opened 2021-05-26
bugfix: S3C-4425 dont wait for failed queue delivery report

jonathan-gramain:
Note to reviewers: may be easier to review each commit individually.

Commits description:

* S3C-4425: improve BackbeatProducer.send() callback description

  Add info in the JSDoc about the fact that the callback is waiting for a delivery report from Kafka and should normally not be in the critical path.

* bugfix: S3C-4425 CRR failed: don't wait for delivery report

  Don't wait for the Kafka delivery reports to be received for the backbeat-replication-failed queue, instead, move on immediately with the processing of further items.

  This solves a low throughput issue when there are a significant number of FAILED entries because each of them held a queue worker for a few seconds, preventing new status updates from being done.

* bugfix: S3C-4425 fix configuration passing to FailedCRRProducer

  The FailedCRRProducer was fetching its Kafka config from the global config, but it was passed a param object. Change it to use the kafka config param object (apart from fixing the code, it helps with setting up local functional tests).

* test: S3C-4425 test for failed CRR mechanism

  Check that a correct message is sent to the `backbeat-replication-failed` queue when a failure to replicate occurs and a status FAILED is to be written, by extending a queue processor functional test.

  This adds some coverage for the FailedCRR mechanism, it does not specifically test the fix in S3C-4425 but adds regression testing.

### DIFF
--- a/extensions/replication/failedCRR/FailedCRRProducer.js
+++ b/extensions/replication/failedCRR/FailedCRRProducer.js
@@ -8,9 +8,11 @@ const config = require('../../../conf/Config');
 class FailedCRRProducer {
     /**
      * Create the retry producer.
+     *
+     * @param {object} kafkaConfig - kafka config param
      */
-    constructor() {
-        this._kafkaConfig = config.kafka;
+    constructor(kafkaConfig) {
+        this._kafkaConfig = kafkaConfig;
         this._topic = config.extensions.replication.replicationFailedTopic;
         this._producer = null;
         this._log = new Logger('Backbeat:FailedCRRProducer');
--- a/extensions/replication/failedCRR/FailedCRRProducer.js
+++ b/extensions/replication/failedCRR/FailedCRRProducer.js
@@ -45,15 +45,18 @@ class FailedCRRProducer {
     /**
      * Publish the given message to the retry Kafka topic.
      * @param {String} message - The message to publish
-     * @param {Function} cb - The callback function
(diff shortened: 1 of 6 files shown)


tmacro (approved):

extensions/replication/failedCRR/FailedCRRProducer.js
```diff
@@ -45,15 +47,18 @@ class FailedCRRProducer {
     /**
      * Publish the given message to the retry Kafka topic.
      * @param {String} message - The message to publish
-     * @param {Function} cb - The callback function
+     * @param {Function} [deliveryReportCb] - called when Kafka
+     * returns a delivery report
      * @return {undefined}
      */
-    publishFailedCRREntry(message, cb) {
+    publishFailedCRREntry(message, deliveryReportCb) {
```
naren-rajendran, line 54: I dont see the callback being used anywhere, do we need it?

naren-rajendran:
Looks good to me, just one question though :) posted it in the review comment.

naren-rajendran (approved):

extensions/replication/failedCRR/FailedCRRProducer.js
jonathan-gramain: We don't use it today, although I believe it could be used in a future improvement for error management, I though it would be good to keep it as part of the API to make it readily available then.

PR #1717, opened 2021-05-28
feature: S3C-4494 Add toggle for probe server

miniscruff:
Note that 7.4 is defaulted to OFF while 7.10 and onwards is defaulted to ON.
CRR_METRICS_PROBE=true/false


naren-rajendran (approved):

naren-rajendran:
Does it need a test?

miniscruff:
> Does it need a test?

I don't see any current or even the possibility of adding tests to the bin/* files. But if you have any ideas let me know.

alexanderchan-scality (approved):

PR #1721, opened 2021-05-28
feature: S3C-4500 Add probe server for queue processor

miniscruff:
Add probe server to queue processor with checks for producer and consumer

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -385,6 +385,65 @@ class QueueProcessor extends EventEmitter {
                           { entry: sourceEntry.getLogInfo() });
         return process.nextTick(done);
     }
+
+    /**
+     * Handle ProbeServer liveness check
+     *
+     * @param {http.HTTPServerResponse} res - HTTP Response to respond with
+     * @param {Logger} log - Logger
+     * @returns {string} Error response string or undefined
+     */
+    handleLiveness(res, log) {
+        const verboseLiveness = {};
+        // track and return all errors in one response
+        const responses = [];
+        if (this.replicationStatusProducer === undefined ||
+            this.replicationStatusProducer === null) {
+            verboseLiveness.replicationStatusProducer = 'undefined';
+            responses.push({
+                component: 'Replication Status Producer',
+                status: 'undefined',
+                site: this.site,
+            });
+        } else if (!this.replicationStatusProducer.isReady()) {
+            verboseLiveness.replicationStatusProducer = 'not ready';
+            responses.push({
+                component: 'Replication Status Producer',
+                status: 'not ready',
+                site: this.site,
+            });
+        } else {
+            verboseLiveness.replicationStatusProducer = 'ready';
+        }
+
+        if (this._consumer === undefined || this._consumer === null) {
+            verboseLiveness.consumer = 'undefined';
+            responses.push({
+                component: 'Consumer',
+                status: 'undefined',
+                site: this.site,
+            });
+        } else if (!this._consumer.isReady()) {
+            verboseLiveness.consumer = 'not ready';
+            responses.push({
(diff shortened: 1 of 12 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -385,6 +385,45 @@ class QueueProcessor extends EventEmitter {
...
+                component: 'Replication Status Producer',
+                status: 'not ready',
+                site: this.site,
+            });
+        }
+
+        log.debug('verbose liveness', verboseLiveness);
+
+        if (responses.length > 0) {
+            return JSON.stringify(responses);
```
rahulreddy, line 441: If you return here then no response will be send to the client, right?

extensions/replication/queueProcessor/task.js
```diff
@@ -29,16 +32,56 @@ werelogs.configure({ level: config.log.logLevel,
...
+            DEFAULT_LIVE_ROUTE,
+            (res, log) => queueProcessor.handleLiveness(res, log)
+        );
+        probeServer._cbOnListening = done;
+        probeServer.start();
+        return undefined;
+    },
+    done => {
+        metricsProducer.setupProducer(err => {
+            log.error('error starting metrics producer for queue processor', {
```
rahulreddy: You need a condition here so that it does not log when there is no error

rahulreddy (changes requested):

extensions/replication/queueProcessor/QueueProcessor.js
miniscruff: The probe server handles sending back bad responses from the string return value. https://github.com/scality/Arsenal/blob/e1a3b05330598a656806e888a760b65711fbfd76/lib/network/probe/ProbeServer.js#L94-L100

naren-rajendran:
Looks good to me.

rahulreddy (commented):
LGTM

naren-rajendran (approved):

lib/util/probeUtils.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
naren-rajendran: May be these should be constants like in other components! I see that we dont have that in backbeat 7.x!

miniscruff: I could move these into the constants.js if you think that is the better approach, personally not a fan of a 'constants' file as they can get quite a bit and messy over time ( reminds me of static "manager" classes from games... ).

naren-rajendran: Yes I'd prefer constants, seems consistent. Something about calling this a util seems a bit wrong.

miniscruff: yea that is fair, I sort of envisioned it having more stuff in it... but then it didn't. I will move it to the existing constants then.

tests/functional/replication/probe.spec.js
```diff
@@ -0,0 +1,72 @@
...
+        // return sample error for liveness
+        mockQp.handleLiveness.returns('error msg');
+        const config = {
+            bindAddress: 'localhost',
+            port: 52555,
+        };
+        startProbeServer(mockQp, config, probeServer => {
+            probeServer.onStop(done);
+            http.get(`http://localhost:52555${DEFAULT_LIVE_ROUTE}`, res => {
+                assert.strictEqual(500, res.statusCode);
```
rahulreddy: This should be `actual, expected` for the assertion check, I think its the other way around here.

rahulreddy: Is there any reason its 500 and not 200?

miniscruff: The liveness check returns an error so the status code is 500, the test confirms the server is up and running and correctly handles the liveness response. The handle liveness is more thoroughly tested in unit tests.

rahulreddy (approved):

PR #1722, opened 2021-05-29
bugfix: INTGR-22 Undefined metric value for log size

miniscruff:
* Bucket file log used info.end but raft log uses info.cseq
* The actual Jira ticket is INTGR-22...


miniscruff:
Closing to fix branch name

PR #1724, opened 2021-06-01
bugfix: INTGR-22 Undefined metric value for log size

miniscruff:
* Bucket file log used info.end but raft log uses info.cseq
* Should resolve the current issue of integration


miniscruff:
Closing this one too as INTGR is not meant for resolving issues, only reporting.

PR #1734, opened 2021-06-02
bugfix: S3C-4493 RoleCredentials: fix error handling

jonathan-gramain:
Fix by returning a copy of InternalError from arsenal errors.

Make sure we return a copy of the arsenal error to avoid side effects
with AWS SDK modifying the error properties.

Also fix the unit test to actually trigger a non-arsenal error like
described, as previously the "socket.destroy()" was caught as an
arsenal error in Vaultclient.

### DIFF
--- a/lib/credentials/RoleCredentials.js
+++ b/lib/credentials/RoleCredentials.js
@@ -90,15 +90,14 @@ class RoleCredentials extends AWS.Credentials {
                         error: err,
                         method: 'RoleCredentials.refresh',
                     });
-                    // We need to generate a new error instance
-                    // instead of passing a possibly global arsenal
-                    // error returned by vault client, because AWS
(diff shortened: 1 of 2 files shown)


naren-rajendran (approved):

tmacro (approved):

tmacro:
Vaguely related to this, I've been thinking about a way to evolve arsenal errors, especially the stack traces. ATM if you look at one the all originate at the line in arsenal where they're instantiated which can make finding the place where its actually "thrown" hard. Something like `cusomizeDescription` that doesn't require a message basically.

jonathan-gramain:
@tmacro I agree, this would be a good improvement, something similar to what `console.trace` gives but embedded in the error object where it is emitted. At least it could help debugging. Do you want to open an improvement ticket?

PR #1765, opened 2021-06-08
bugfix: S3C-4541 Return error and probe server for callback

miniscruff:
* Use onListening instead of directly assigning private var
* Add onError to handle errors

### DIFF
--- a/extensions/replication/queueProcessor/Probe.js
+++ b/extensions/replication/queueProcessor/Probe.js
@@ -9,8 +9,10 @@ const { ProbeServer, DEFAULT_LIVE_ROUTE } =
  */
 
 /**
- * Callback when Queue Processor Probe server is listening
+ * Callback when Queue Processor Probe server is listening.
+ * Note that a disabled probe server does not pass an error to the callback.
  * @callback DoneCallback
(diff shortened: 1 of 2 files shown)


extensions/replication/queueProcessor/Probe.js
```diff
@@ -31,7 +33,8 @@ function startProbeServer(queueProcessor, config, callback) {
         DEFAULT_LIVE_ROUTE,
         (res, log) => queueProcessor.handleLiveness(res, log)
     );
-    probeServer._cbOnListening = () => callback(probeServer);
+    probeServer.onListening(() => callback(undefined, probeServer));
```
jonathan-gramain: We usually follow a convention of passing `null` for the error

tests/functional/replication/probe.spec.js
```diff
@@ -14,7 +14,7 @@ function mockQueueProcessor() {
     };
 }
 
-describe('Probe server', () => {
+describe.only('Probe server', () => {
```
jonathan-gramain: remove `.only` :wink:

jonathan-gramain (approved):

naren-rajendran (approved):

PR #1777, opened 2021-06-10
feature: S3C-4556 Switch probe server to site specific mapping

miniscruff:
* probe server is now an array of objects
* add site to each object to map back to location

### DIFF
--- a/tests/functional/replication/probe.spec.js
+++ b/tests/functional/replication/probe.spec.js
@@ -1,6 +1,6 @@
 const assert = require('assert');
 const sinon = require('sinon');
-const { startProbeServer } =
+const { startProbeServer, getProbeConfig } =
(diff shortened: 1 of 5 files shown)


tests/functional/replication/probe.spec.js
```diff
@@ -14,7 +14,7 @@ function mockQueueProcessor() {
     };
 }
 
-describe('Probe server', () => {
+describe.only('Probe server', () => {
```
naren-rajendran: why `only`?

miniscruff: grrr sorry, really need to enable the eslint that checks that...

naren-rajendran (approved):

jonathan-gramain (approved):

PR #1797, opened 2021-06-25
Feature/S3C-4594/lifecycle authentication

alexanderchan-scality:


extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -42,36 +50,46 @@ class LifecycleBucketProcessor {
...
         this._transport = transport;
         this._producer = null;
         this._kafkaBacklogMetrics = null;
-        this.accountCredsCache = {};
+
+        this._producerReady = false;
+        this._consumerReady = false;
+
+        this.httpAgent = new http.Agent({ keepAlive: true });
+        this.httpsAgent = new https.Agent({ keepAlive: true });
```
miniscruff: This might follow a pattern somewhere else but, if you just create one agent here based on the transport we can save some trouble checking it again later on. That is, just create the one agent for the transport we are using and use it everywhere.

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -247,39 +317,49 @@ class LifecycleBucketProcessor {
         const producer = new BackbeatProducer({
             kafka: { hosts: this._kafkaConfig.hosts },
         });
-        producer.once('error', cb);
+        producer.once('error', err => {
+            this._log.error('error setting up kafka producer', {
+                error: err,
+                method: 'LifecycleBucketProcesso::_setupProducer',
```
miniscruff, line 323:
```suggestion
                method: 'LifecycleBucketProcessor::_setupProducer',
```

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -438,12 +406,44 @@ class LifecycleBucketProcessor {
...
+            const credentials = this.credentialsManager.getCredentials({
+                id: 'lifecycle:bucketProcessor:stsclient',
+                authConfig: this._lcConfig.auth.vault.serviceCredentials,
+            });
+
+            this._stsclient = createSTSClient({
+                transport: this._lcConfig.auth.transport,
+                host: this._lcConfig.auth.vault.host,
+                port: this._lcConfig.auth.vault.port,
+                agent: new http.Agent({ keepAlive: true }),
```
miniscruff: should this agent be http or https based on the transport?

extensions/lifecycle/bucketProcessor/task.js
```diff
@@ -17,55 +23,46 @@ werelogs.configure({
...
-            return;
+        if (!probeServer) {
+            logger.info('Skipping lifecycle bucket processor server setup');
+            return done();
         }
-        logger.info('management init done');
-        cb();
+
+        probeServer.addHandler(DEFAULT_LIVE_ROUTE, livenessCheck);
+        probeServer.addHandler(DEFAULT_READY_ROUTE, livenessCheck);
```
miniscruff, line 50: I have just been using live checks as from what I could tell from the k8s guidelines ready meant something else, but if we should just do the same for both I will update my uses.

alexanderchan-scality: the bucket/object processor and conductors won't need to handle requests, so there isn't a need for the ready check

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
alexanderchan-scality: this will be cleaned up after I have broken the PR into smaller PRs. still have some things to iron out

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
alexanderchan-scality: will be cleaned up when this PR is broken into smaller PRs

miniscruff: oh ok cool, I was doing an early review now to make it a little easier but if you are splitting it I will wait.

PR #1802, opened 2021-06-25
Feature/S3C-4599/add credentials manager

alexanderchan-scality:
* update lifecycle auth config structure
* renamed `vault` auth type to `assumeRole` auth type
* renamed VaultServiceCredentials to AssumeRoleCredentials 
* add CredentialsManager class


tests/unit/credentials/CredentialsManager.js
```diff
@@ -0,0 +1,191 @@
...
+
+const { VaultServiceCredentials } =
+    require('../../../lib/credentials/VaultServiceCredentials');
+const AccountCredentials = require('../../../lib/credentials/AccountCredentials');
+const CredentialsManager = require('../../../lib/credentials/CredentialsManager');
+const { authTypeAccount, authTypeVault } = require('../../../lib/constants');
+
+const extension = 'test-extension';
+const id = '5a4847c9d83443885282cf38a09fa1a6193de9ed970168376a7b10c438683da0';
+const accountId = '000000000000';
```
rachedbenmustapha: Having a second account id would be a good idea to ensure that the code always separates creds by account ids.

lib/credentials/CredentialsManager.js
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        switch (authConfig.type) {
+        case authTypeRole:
+            return this._addRoleCredentials(params);
+        case authTypeVault:
+            return this._addVaultServiceCredentials(params);
+        case authTypeAccount: // fallthrough
+        case authTypeService:
+            return this._addAccountCredentials(params);
```
rachedbenmustapha: Is there an immediate need to support multiple credentials types and all the overhead they incur? It seems like there is code that's not going to be used in the immediate future.

Edit: also if multiple types are going to be supported, they should probably be hashed by the (id, type) tuple rather than just id.

alexanderchan-scality: multiple credential types will be used. there is the sts client  that will be using the `account` type and the s3/backbeat clients that use the `vault` type. `role` type, however, won't be; the changes to adapt the existing replication service will require a bit more time.

as for the id, I had planned on the caller managing the id as opposed to the `CredentialsManager`. (open to changes if this won't work)

lib/credentials/CredentialsManager.js
```diff
@@ -0,0 +1,170 @@
...
+            getRoleArn(accountId, authConfig.vault.roleName),
+            this._logger.newRequestLogger());
+        return this._accountCredsCache[id];
+    }
+
+    _addAccountCredentials(params) {
+        const err = CredentialsManager.validateParams(params, accountParamsJoi);
+        if (err) {
+            this._logger.error('missing required params for account type credentials', {
+                method: 'CredentialsManager::_addVaultServiceCredentials',
```
nicolas2bert: NIT: `_addAccountCredentials`

lib/credentials/CredentialsManager.js
rachedbenmustapha: Perhaps it's just the naming that confuses me, but why would the sts client need account credentials? Any access key would do provided it's linked to an entity with enough privileges.

By role, do you mean assumed role (the temp credentials issued by AssumeRole that include a session token), or the actual role name and access/secret key that will be used to assume a role?

alexanderchan-scality: > Perhaps it's just the naming that confuses me, but why would the sts client need account credentials? Any access key would do provided it's linked to an entity with enough privileges. 

> By role, do you mean assumed role (the temp credentials issued by AssumeRole that include a session token), or the actual role name and access/secret key that will be used to assume a role?

* `account` / `service`: will be use for the credentials read from the `authdata.json` file. the sts client will be using this read the lifecycle credentials stored in the `authdata.json` (I can see the confusion here in this case. the lifecycle user credentials would not be an `AccountCredentials`) 

* ​`role`: at the moment, is the credential type using the vaultclient `assumeRoleBackbeat`

* `vault`:  will be the new credential type that uses the sts `assumeRole`

rachedbenmustapha: May I suggest having the credential manager handle only the type `vault` for now, and type `vault` renamed to `assumedRole` for clarity, given that:
- `role` is not used by lifecycle and not in scope of our current feature
- `account` / `service` do not need caching, refreshing or expiring and are readily available in configuration.

alexanderchan-scality: * renamed `vault` to `assumeRole`
* renamed `VaultServiceCredentials` to `AssumeRoleCredentials` to better describe the class

lib/credentials/CredentialsManager.js
```diff
@@ -0,0 +1,103 @@
+const joi = require('@hapi/joi');
+const EventEmitter = require('events');
+
+const { authTypeAssumeRole } = require('../constants');
+const { AssumeRoleCredentials, getRoleArn } = require('./AssumeRoleCredentials');
+
+const vaultParamsJoi = joi.object({
```
rachedbenmustapha: This becomes `assumeRoleParamsJoi`.

lib/credentials/CredentialsManager.js
```diff
@@ -0,0 +1,103 @@
...
+    id: joi.string().required(),
+    accountId: joi.string().required(),
+    stsclient: joi.object().required(),
```
rachedbenmustapha: I'm not sure why a direct sts client is needed. There is a `AWS.ChainableTemporaryCredentials`  class that takes a role arn and "origin credentials" and takes care of assuming the role and automatically refreshing the temp keys/tokens when they expire (provided callers use the `get` method).

lib/credentials/CredentialsManager.js
```diff
@@ -0,0 +1,103 @@
...
+            return null;
+        }
+
+        return this._addAssumeRoleCredentials(params);
+    }
+
+    /*
+     * removes inactive credentials
+     */
+    removeInactiveCredentials(maxInactiveDuration) {
```
rachedbenmustapha, line 89: I think this could be simpler, since the base `AWS.Credentials` class (that all supported credentials inherit from) has a `needsRefresh` method that is functionally equivalent (it could be testing for more than just expiration/time).

conf/config.json
```diff
@@ -92,17 +92,16 @@
...
+                "sts": {
                     "host": "127.0.0.1",
                     "port": 8800,
-                    "serviceCredentials": {
-                        "type": "account",
-                        "account": "lifecycle"
-                    },
-                    "adminPort": 8600,
-                    "adminCredentialsFile": "./admincredentials.json"
+                    "serviceUser": "lifecycle"
```
rachedbenmustapha: This seems to be taking advantage of `authdata.json`, which is made to hold accounts. I believe `authdata.json` was not meant to store credentials but rather be a "central" repository of accounts where Vault is not available (so typically Zenko or standalone open source cloudserver).

Its use here is a stretch because we're introducing a made up canonical id. Just adding an access key and secret key fields here would avoid subverting `authdata.json`'s initial purpose (which is less and less relevant now that Artesca uses Vault).

alexanderchan-scality: `accessKey` and `secretKey` in the `config.json` can work and avoid using the `authdata.json` for purposes it wasn't intended for

lib/credentials/CredentialsManager.js
alexanderchan-scality: wasn't aware of this. this would simplify what I had implemented (not needing to manage the stsclient for the credentials for the s3/backbeat clients)

*update*: using `AWS.ChainableTemporaryCredentials` will work (only an update of `aws-sdk` 
was needed)

lib/credentials/CredentialsManager.js
alexanderchan-scality: `removeInactivateCredentials` is meant to remove expired credentials that are older than the `maxInactiveDuration`; correct me if I am wrong, I don't believe `needRefresh` would help in that

rachedbenmustapha: I guess if the intent is to keep expired (thus inactive) creds around, or remove unexpired (thus "recently" active), then it won't help. It boils down to the difference between inactive and expired.

lib/config/configItems.joi.js
```diff
@@ -49,30 +54,27 @@ const adminCredsJoi = joi.object()
...
-        adminPort: joi.number().greater(0)
-            .when('adminCredentialsFile', {
-                is: joi.exist(),
-                then: joi.required(),
-            }),
-        adminCredentialsFile: joi.string().required(),
-    }).when('type', { is: 'vault', then: joi.required() }),
+        .when('type', { is: authTypeAccount, then: joi.required() })
+        .when('type', { is: authTypeService, then: joi.required() }),
+    role: joi.string()
```
rachedbenmustapha: Sorry to nitpick again, `roleName` was preferable, since it's clear what it is. With `role` I expect to find a role object, or wouldn't know if a full ARN is expected.

lib/credentials/AssumeRoleCredentials.js
```diff
@@ -0,0 +1,68 @@
...
+        this._log = new Logger('Backbeat').newRequestLogger(log.getUids());
+        this._extension = p.extension;
+        this._refreshCredsAnticipationSeconds = p.refreshCredsAnticipationSeconds;
+    }
+
+    /**
+     * check if credentials have expired
+     * @return {boolean} - true if expired, false otherwise
+     */
+    needsRefresh() {
```
rachedbenmustapha: Is overriding `needsRefresh` really necessary? It seems to be that it would be interfering with the builtin refresh mechanism with the only added value of taking into account the refresh anticipation window, which is taken care of automatically when using `.get(callback)` (even when the creds are already expired when get() is called, they are refreshed before being handed over to the callback).

alexanderchan-scality: the original issue that required the anticipation window (S3C-3833) sounds it might occur here as well, so the anticipation window will be needed.
with the clients set not to retry requests, objects can possibly be pushed out of their expected deletion windows.

rachedbenmustapha: I may be wrong, but I think that the code in question does not use `ChainableTemporaryCredentials` and uses the credentials ad-hoc instead of obtaining then throught `.get(cb)` before each operation, which indeed would result in intermittent failures.

alexanderchan-scality: **update**: looking at the `aws-sdk`'s `needsRefresh`, there is already a built in expiryWindow logic.

alexanderchan-scality: looking into `aws-sdk`, there is an existing mechanism in place to expire credentials 15seconds earlier, but it cannot be adjusted.
The logic here is also a copy of the aws-sdk's `needsRefresh` but with the adjustable expiry window

rachedbenmustapha: My point was that the adjustable expiry window did not bring any benefit on top of builtin aws-sdk behavior (which "always" works no matter the window), it just adds another layer of timekeeping.

alexanderchan-scality: > I may be wrong, but I think that the code in question does not use ChainableTemporaryCredentials and uses the credentials ad-hoc instead of obtaining them through .get(cb) before each operation, which indeed would result in intermittent failures.

`RoleCredentials` are retrieving credentials using `.get(cb)`

alexanderchan-scality: > My point was that the adjustable expiry window did not bring any benefit on top of builtin aws-sdk behavior (which "always" works no matter the window), it just adds another layer of timekeeping.

I can remove the `needsRefresh` overwrite for now. I have yet to see an error due to timing, so it may not be an issue. (Will remove the AssumeRoleCredentials type)

rachedbenmustapha (approved):

nicolas2bert (approved):

PR #1803, opened 2021-06-25
S3C-4599: implement VaultServiceCredentials class

alexanderchan-scality:


lib/credentials/VaultServiceCredentials.js
```diff
@@ -0,0 +1,135 @@
...
+            extension,
+            roleArn,
+            refreshCredsAnticipationSeconds,
+        }, configJoi);
+
+        this._stsclient = stsclient;
+        this._log = new Logger('Backbeat').newRequestLogger(log.getUids());
+        this._extension = extension;
+        this._roleArn = roleArn;
+        this._refreshCredsAnticipationSeconds = params.refreshCredsAnticipationSeconds;
```
miniscruff, line 46:
```suggestion
        this._refreshCredsAnticipationSeconds = refreshCredsAnticipationSeconds;
```
Already destructured the var

lib/credentials/VaultServiceCredentials.js
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+    * get credentials from cache or refresh credentials from vault
+    * @param {callback} cb - callback to be called with err or credentials obj
+    *   cb(null, { AccessKeyId, SecretAccessKey, SessionToken })
+    * @return {undefined}
+    */
+    refresh(cb) {
+        this._log.debug('refreshing credentials from vault', {
+            method: 'VaultServiceCredentials.refresh',
```
miniscruff:
```suggestion
            method: 'VaultServiceCredentials::refresh',
```
To be consistent

tests/unit/credentials/VaultServiceCredentials.js
```diff
@@ -0,0 +1,172 @@
...
+    SecretAccessKey: 'qscwdvefb1234567890',
+    SessionToken: '83bdcf990f5fe307db8f12ec1994f35f4022949f36b94655cc4474',
+};
+const credentialsSet2 = {
+    AccessKeyId: 'XXXXABCD1234567890',
+    SecretAccessKey: '1234567890qscwdvefb',
+    SessionToken: '1994f35f4022949f36b94655cc447483bdcf990f5fe307db8f12ec',
+};
+
+function assertCredentials(vaultCredentials, expectedCredentials) {
```
miniscruff, line 23: Null is only passed once and so I think we should remove the if statement here and just have the one set of asserts, maybe a deep equal if it makes sense but I am assuming there is some reason why it wasn't used. ( Usually happens one object has extra properties we don't care about whenever I suggest using deep )

miniscruff (approved):

lib/credentials/VaultServiceCredentials.js
alexanderchan-scality: `params.refreshCredsAnticipation` is preferred since it would have a default set by the `joi` validation

tests/unit/credentials/VaultServiceCredentials.js
alexanderchan-scality: as you had mentioned, a deep check would fail because of other properties in the test results.

lib/credentials/VaultServiceCredentials.js
miniscruff: oh nevermind yea, I see how it works now.

saitharunthodupunuru (approved):

PR #1824, opened 2021-07-02
feature: BB-1 Add comment for missing kafka lag

miniscruff:
The decision was made to not forward port kafka lag as a metric due to complications and other things, a ticket was made to address this as tech debt later. This PR is here to add an inline code comment to add another point to remember.


naren-rajendran (approved):

mrepop (approved):
LGTM

PR #1832, opened 2021-07-06
BB-4: add ClientCache class

alexanderchan-scality:
* add ClientCache class

### DIFF
--- a/lib/clients/ClientCache.js
+++ b/lib/clients/ClientCache.js
@@ -1,0 +1,223 @@
+const AWS = require('aws-sdk');
+const joi = require('@hapi/joi');
+
+const BackbeatClient = require('./BackbeatClient');
+const { transportJoi } = require('../config/configItems.joi');
+
+const s3ProfileJoi = {
+    transport: transportJoi,
+    host: joi.string().required(),
+    port: joi.number().greater(0).required(),
+    agent: joi.object().optional(),
+    credentials: joi.object().required(),
+};
+
+/**
+ * @class ClientCache
+ *
+ * @classdesc Helper class to manage a set of clients
+ *
+ * Clients share the same configuration based on the profile they
+ * belong to. A profile is automatically created when it does not
+ * exist and the first configuration is set for a given profile name.
+ *
+ * When getClient() is called, a new or cached client instance is
+ * returned, based on if the profile name, host and port provided
+ * match a cached instance. When client instances are created
+ * they are cached and kept indefinitely.
+ */
+class ClientCache {
+    constructor(clientCreateFn) {
+        this._profiles = {};
+        this._clients = {};
+        this._clientCreateFn = clientCreateFn;
+    }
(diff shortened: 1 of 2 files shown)


lib/clients/ClientCache.js
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * Set the target service auth credentials for the given profile
+     *
+     * @param {string} profileName - name of profile to bind to
+     * @param {number} credentials - auth credentials
+     * @return {ClientCache} this
+     */
+    setCredentials(profileName, credentials) {
```
nicolas2bert, line 95: Why do we need two levels: profiles and clients? 
Can not we create/retrieve/update client directly?

rachedbenmustapha (commented):
I don't get what this API brings for its code size, where can we see an actual client benefit from it?

lib/clients/ClientCache.js
```diff
@@ -0,0 +1,223 @@
+const AWS = require('aws-sdk');
+const joi = require('@hapi/joi');
+
+const BackbeatClient = require('./BackbeatClient');
+const { transportJoi } = require('../config/configItems.joi');
+
+const s3ProfileJoi = {
```
rachedbenmustapha, line 7: I'm not sure it's such a good idea to spread this pattern of using joi as a type system. It seems to be designed for the input data use case. I didn't comment on the last PR but it's not something I'm comfortable with seeing more and more.

For this type of code where we control all entry points, it's easy to know that all parameters will be compliant. For example the host and port come from the configuration where they will have already been validated.

Some typescript or flow annotations would be very useful (and urgent IMO) later on, but we're not achieving a lot of safety with joi, as by the time errors are caught at runtime, it's already too late.

lib/clients/ClientCache.js
alexanderchan-scality: > Why do we need two levels: profiles and clients?
> Can not we create/retrieve/update client directly?

@nicolas2bert taking a step back and looking at this. profiles and clients are essentially mapped 1-to-1, so there isn't a case where a single profile will be used by another client and is likely not helpful being managed.

alexanderchan-scality:
> I don't get what this API brings for its code size, where can we see an actual client benefit from it?

@RachedBenMustapha 
the intention with the `ClientCache` is to avoid allocating similar clients multiple times (as this would be called once for each task operation) and avoiding any overhead that comes with creating them. the original design was to also be similar that of the `VaultClientCache` in order to make transitioning easier and reducing duplicate code, but since that will be phased out, such a design may not be needed. 

as @nicolas2bert had pointed out as well, the management of profiles in addition to clients seems to be over-engineering (since profiles and clients for use case of lifecycle is 1-to-1)

with that said, a simple map + the existing credentials manager would just do.

alexanderchan-scality:
changes no longer needed

PR #1837, opened 2021-07-07
BB-7: update lifecycle object processor

alexanderchan-scality:


rachedbenmustapha (commented):
LGTM overall but I'm not seeing credentials being retrieved in an asynchronous manner, if we were to benefit from auto-refresh through `.get()` to avoid intermittent auth failures.

alexanderchan-scality:
@RachedBenMustapha please correct me if I am wrong. won't the service client object handle that logic at the time of performing the request? (I believe credentials are fetched for signing, and the `get()` async method would be called then)

rachedbenmustapha (approved):

nicolas2bert (approved):

PR #1855, opened 2021-07-19
improvement: S3C-4312 backbeat encryption support

jonathan-gramain:
Encryption support for CRR target on backbeat

### DIFF
--- a/lib/clients/backbeat-2017-07-01.api.json
+++ b/lib/clients/backbeat-2017-07-01.api.json
@@ -16,7 +16,7 @@
         "PutData": {
             "http": {
                 "method": "PUT",
-                "requestUri": "/_/backbeat/data/{Bucket}/{Key+}"
+                "requestUri": "/_/backbeat/data/{Bucket}/{Key+}?v2"
             },
             "input": {
                 "type": "structure",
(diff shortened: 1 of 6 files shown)


mrepop (approved):

lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -16,7 +16,7 @@
         "PutData": {
             "http": {
                 "method": "PUT",
-                "requestUri": "/_/backbeat/data/{Bucket}/{Key+}"
+                "requestUri": "/_/backbeat/data/{Bucket}/{Key+}?v2"
```
alexanderchan-scality, line 19: with cloudserver still supporting `v1` requests, should this be a different definition rather than as a replacement for the v1 request?

jonathan-gramain: We could have another definition, but the added benefit of reusing the same definition is that the unupgraded cloudserver would still accept requests from newer backbeats, just that they will naturally use "v1" (no encryption) because that's all they support, instead of returning an error for a route they don't know. This way we have two-way compatibility between 7.9 and 7.10 backbeat/cloudserver, but only if both are 7.10 will encryption be allowed.

Otherwise, just to clarify, the main reason we use a new route version is to make sure cloudserver does not encrypt the object when the requesting backbeat does not support it (< 7.10), that would make the data unreadable since the key would not be stored in metadata.

jonathan-gramain: If you meant to create another definition next to the existing one, it's not necessary because backbeat would always use the new definition so the old definition would not be used at all.

alexanderchan-scality (approved):

tmacro (approved):

PR #1857, opened 2021-07-19
BB-16: update lifecycle populator filter for metadata entries

alexanderchan-scality:


rachedbenmustapha (approved):

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -96,28 +96,70 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+        // raft log entries
+        // {
+        //   bucket: bucket name
+        //   key: no key value
+        //   value: bucket value
+        // }
+        return entry.key === undefined && entry.bucket !== METASTORE;
+    }
+
+    _isMongoEntry(entry) {
```
naren-rajendran: Mongo entry? Is this for 7.x branches?

jonathan-gramain: Yes, should not be needed normally as it's 8.x specific (and should already be compatible)

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -96,28 +96,70 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         });
     }
 
+    _isBucketdEntry(entry) {
```
jonathan-gramain: It seems you are specifically looking for bucket change entries, so maybe rename to `isBucketEntry` (assuming you also remove the `isMongoEntry` helper)

extensions/lifecycle/LifecycleQueuePopulator.js
alexanderchan-scality: `MongoEntry` may be a confusing name as `7.x` won't be supporting mongo. I can rename to `FileMDEntry`.

extensions/lifecycle/LifecycleQueuePopulator.js
alexanderchan-scality: I believe the `isMongoEntry` (will rename to `isFileMDEntry`) is still needed to handle the `file` MD backend

nicolas2bert (approved):

extensions/lifecycle/LifecycleQueuePopulator.js
jonathan-gramain: OK that works. But I believe the name should reflect that it's selecting bucket-update entries only, where the current name suggests it's filtering all entries coming from bucketd.

jonathan-gramain: What about: `isBucketEntryFromBucketd()` and `isBucketEntryFromFileMD`

PR #1866, opened 2021-07-20
BB-7: fix w/8.2 merge

alexanderchan-scality:


extensions/lifecycle/objectProcessor/task.js
```diff
@@ -3,14 +3,15 @@
 const async = require('async');
 const werelogs = require('werelogs');
 const { errors } = require('arsenal');
-const {
-    DEFAULT_LIVE_ROUTE,
-    DEFAULT_READY_ROUTE,
-} = require('arsenal').network.probe.ProbeServer;
+// const {
```
jonathan-gramain, line 6: Are the probe-related changes wanted? Just checking but good for me if that's the case

jonathan-gramain (approved):

extensions/lifecycle/objectProcessor/task.js
alexanderchan-scality: yes. I believe the plan is to move to the generic `ProbeServer` over the `HealthProbeServer`.
atm, `healthProbeServer` seems to also handle metrics which deviates from its original purpose.

anurag4DSB (approved):

PR #1905, opened 2021-08-11
feature: lc conductor bucket listing from bucketd

rachedbenmustapha:
Implements getting the list of buckets to lifecycle from a bucketd listing, instead of Zookeeper.

This keeps the Zookeeper dataset from growing linearly with the data plane's number of buckets, which would destabilize the quorum and worsen transient issues. The compromise is that more messages go through Kafka, which are then processed in parallel - buckets are either ignored or queued for listing by the bucket processor processes. The listing operation itself is relatively cheap.

Best case: all buckets have a lifecycle configuration; no change in the number of messages transiting through Kafka.
Worst case: no buckets have a lifecycle configuration; 1 message per bucket go through Kafka (instead of 0),  then are ignored by the bucket processor.

Note: had to enable object spread for eslint, as the `Object.assign` equivalent was unreadable.

### DIFF
--- a/extensions/lifecycle/LifecycleConfigValidator.js
+++ b/extensions/lifecycle/LifecycleConfigValidator.js
@@ -1,5 +1,5 @@
 const joi = require('@hapi/joi');
-const { retryParamsJoi, authJoi, probeServerJoi } =
+const { retryParamsJoi, authJoi, probeServerJoi, hostPortJoi } =
     require('../../lib/config/configItems.joi.js');
 
 const joiSchema = {
@@ -8,6 +8,10 @@ const joiSchema = {
     objectTasksTopic: joi.string().required(),
     auth: authJoi.required(),
     conductor: {
+        bucketSource: joi.string().
+            valid('bucketd', 'zookeeper').default('zookeeper'),
+        bucketd: hostPortJoi.
+            when('bucketSource', { is: 'bucketd', then: joi.required() }),
         cronRule: joi.string().required(),
         concurrency: joi.number().greater(0).default(10),
         backlogControl: joi.object({
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -4,14 +4,16 @@ const async = require('async');
 const schedule = require('node-schedule');
 const zookeeper = require('node-zookeeper-client');
 
-const errors = require('arsenal').errors;
+const { constants, errors } = require('arsenal');
 const Logger = require('werelogs').Logger;
+const BucketClient = require('bucketclient').RESTClient;
 
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
 const zookeeperHelper = require('../../../lib/clients/zookeeper');
 const KafkaBacklogMetrics = require('../../../lib/KafkaBacklogMetrics');
 const { authTypeAssumeRole } = require('../../../lib/constants');
 const VaultClientCache = require('../../../lib/clients/VaultClientCache');
+const safeJsonParse = require('../util/safeJsonParse');
 
 const DEFAULT_CRON_RULE = '* * * * *';
 const DEFAULT_CONCURRENCY = 10;
@@ -39,6 +41,9 @@ class LifecycleConductor {
      * metrics config object (see {@link BackbeatConsumer} constructor
      * for params)
      * @param {Object} lcConfig - lifecycle configuration object
+     * @param {String} lcConfig.bucketSource - whether to fetch buckets
+     * from zookeeper or bucketd
+     * @param {Object} lcConfig.bucketd - host:port bucketd configuration
      * @param {String} lcConfig.zookeeperPath - base path for
      * lifecycle nodes in zookeeper
      * @param {String} lcConfig.bucketTasksTopic - lifecycle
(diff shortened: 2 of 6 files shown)


extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -83,111 +92,157 @@ class LifecycleConductor {
...
+            });
+    }
+
+    listBucketdBuckets(queue, log, cb) {
+        let isTruncated = false;
+        let marker = null;
+        let nEnqueued = 0;
+
+        async.doWhilst(
+            next => this._bucketClient.listObject(
```
jonathan-gramain, line 218: In order to avoid taking too much memory (or running out of it!) in case there are a lot of buckets to list that cannot be processed timely by the cargo and queued, I suggest adding a primitive form of flow control, something along the lines of this, with tunables that may need adjustment:

```
if (cargo.length > 10 * concurrency) {
    setTimeout(next, 100);
}
```

tests/functional/lifecycle/LifecycleConductor.spec.js
```diff
@@ -17,152 +20,362 @@ const zkConfig = {
...
+            assert.strictEqual(req.method, 'GET');
+            assert.strictEqual(query.Action, 'GetAccounts');
+
+            const accountIds = query.canonicalIds.map(v => ({
+                id: v.replace('owner', 'account'),
+                canId: v,
+            }));
+
+            res.write(JSON.stringify(accountIds));
+            res.end();
```
jonathan-gramain:
```suggestion
            res.end(JSON.stringify(accountIds));
```

tests/functional/lifecycle/LifecycleConductor.spec.js
```diff
@@ -17,152 +20,362 @@ const zkConfig = {
...
-                        details: {},
+            res.write(JSON.stringify({
+                Contents: thisListing.map(key => ({
+                    key,
+                    value: {},
+                })),
+                IsTruncated: !!bucketdListing.length,
+            }));
+
+            res.end();
```
jonathan-gramain:
```suggestion
            res.end(JSON.stringify({
                Contents: thisListing.map(key => ({
                    key,
                    value: {},
                })),
                IsTruncated: !!bucketdListing.length,
            }));
```

tests/functional/lifecycle/LifecycleConductor.spec.js
```diff
@@ -17,152 +20,362 @@ const zkConfig = {
...
+        if (mockBucketd) {
+            lifecycleConfig.conductor.bucketd.port = bucketdPort;
+
+            bucketPopulatorStep1 = next => {
+                bucketdListing.push('owner1..|..bucket1', 'owner2..|..bucket2');
+                process.nextTick(next);
+            };
+
+            bucketPopulatorStep2 = next => {
+                bucketdListing.push('owner1..|..bucket1', 'owner2..|..bucket2');
```
jonathan-gramain: Might make sense to test with two of the buckets belonging to the same owner

alexanderchan-scality (approved):

extensions/lifecycle/conductor/LifecycleConductor.js
alexanderchan-scality: question: wouldn't this issue be covered by having `maxkeys` and the cargo size be the same value (as the input would equal the output, and cargo isn't doing any additional asynchronous processing)?

extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -8,6 +8,10 @@ const joiSchema = {
     objectTasksTopic: joi.string().required(),
     auth: authJoi.required(),
     conductor: {
+        bucketSource: joi.string().
+            valid('bucketd', 'zookeeper').default('zookeeper'),
+        bucketd: hostPortJoi.
```
alexanderchan-scality, line 13: noticed in https://github.com/scality/Federation/pull/3744/files#diff-eebf0415822c288651cc7eb695d029d6bfd390ac9bb4699be39a8cb76f59af67R126-R131 there is a reference to a `transport` field.

`hostPortJoi` does not define that, so there may be a validation issue

extensions/lifecycle/conductor/LifecycleConductor.js
jonathan-gramain: Since cargo is a queue, I think it would hold entries there until it can process them, taking memory in the meantime. (`maxKeys` is here the same as `concurrency` but it is only specifying how many keys we want to retrieve from a listing iteration, it does not seem to prevent queuing more than this number of items to the cargo queue)

nicolas2bert (approved):

extensions/lifecycle/conductor/LifecycleConductor.js
rachedbenmustapha: @jonathan-gramain cd8e9d14e0ec20e9cdd7c75bcf0e077217871660 is what I could come up with that wasn't too disruptive.

Getting the cargo to saturate was pretty difficult, but using extreme values (very small cargo size and very large listing `maxKeys`), any sort of throttling more aggressive than this actually had the adverse effect of making the bucket and object processors consume messages too fast, inhibiting the backlog control (could be worked around easily though).

alexanderchan-scality (approved):

PR #1921, opened 2021-09-08
BB-37 paused locations state is not set up properly

nicolas2bert:
`data` is not JSON parsed so `data.paused` condition was always false.
That would lead to the application "ingestion state" not reflecting the zookeeper "ingestion state".


jonathan-gramain (approved):

miniscruff (approved):

naren-rajendran:
Looks good to me, but I think it may not be enough. Approving this to see if the tests pass. We should check end 2 end tests with this fix also.

naren-rajendran (approved):

nicolas2bert:
@bert-e approve

PR #1923, opened 2021-09-09
bugfix: BB-31 skip metrics when using mongo log reader

miniscruff:
Mongo log reader has no value for the end of the log to use as the metric so just skip it for now


miniscruff:
Bad branch name...

PR #1934, opened 2021-09-16
improvement: BB-51 use default probe routes

naren-rajendran:
Use default probe routes in lifecycle processes.

### DIFF
--- a/extensions/lifecycle/conductor/service.js
+++ b/extensions/lifecycle/conductor/service.js
@@ -1,44 +1,86 @@
 'use strict'; // eslint-disable-line
 
+const async = require('async');
 const werelogs = require('werelogs');
+const { errors } = require('arsenal');
+const {
+    DEFAULT_LIVE_ROUTE,
+    DEFAULT_METRICS_ROUTE,
+    DEFAULT_READY_ROUTE,
+} = require('arsenal').network.probe.ProbeServer;
+const { ZenkoMetrics } = require('arsenal').metrics;
 
 const LifecycleConductor = require('./LifecycleConductor');
-const { HealthProbeServer } = require('arsenal').network.probe;
-
 const config = require('../../../lib/Config');
+const { sendSuccess, sendError, startProbeServer } = require('../../../lib/util/probe');
+
 const zkConfig = config.zookeeper;
 const kafkaConfig = config.kafka;
 const lcConfig = config.extensions.lifecycle;
 const repConfig = config.extensions.replication;
 
-const lcConductor = new LifecycleConductor(
-    zkConfig, kafkaConfig, lcConfig, repConfig);
 
-werelogs.configure({ level: config.log.logLevel,
-                     dump: config.log.dumpLevel });
-const logger = new werelogs.Logger('Backbeat:Lifecycle:Conductor:service');
-const healthServer = new HealthProbeServer({
-    bindAddress: config.healthcheckServer.bindAddress,
-    port: config.healthcheckServer.port,
+const lcConductor = new LifecycleConductor(zkConfig, kafkaConfig, lcConfig, repConfig);
+
+werelogs.configure({
+    level: config.log.logLevel,
+    dump: config.log.dumpLevel,
 });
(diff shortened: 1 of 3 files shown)


miniscruff (changes requested):
Separate discussion on the third lifecycle process being converted to the new probe server

extensions/lifecycle/conductor/service.js
```diff
@@ -1,44 +1,82 @@
...
+    done => startProbeServer(config.healthcheckServer, (err, probeServer) => {
+        if (err) {
+            logger.error('error starting probe server', { error: err });
+            return done(err);
+        }
+        if (probeServer !== undefined) {
+            // following the same pattern as other extensions, where liveness
+            // and readiness are handled by the same handler
+            probeServer.addHandler([DEFAULT_LIVE_ROUTE, DEFAULT_READY_ROUTE], handleLiveness);
+            probeServer.addHandler('/_/monitoring/metrics', handleMetrics);
```
miniscruff: There is a default metrics route as well

extensions/lifecycle/conductor/service.js
```diff
@@ -1,44 +1,82 @@
...
 
-lcConductor.start(err => {
+/**
+ * Handle ProbeServer metrics
+ *
+ * @param {http.HTTPServerResponse} res - HTTP Response to respond with
+ * @param {Logger} log - Logger
+ * @returns {string} response
+ */
+function handleMetrics(res, log) {
```
miniscruff, line 53: hmm on second through, we may want to skip this method for now as there are no prom metrics yet

extensions/lifecycle/conductor/service.js
```diff
@@ -1,44 +1,82 @@
...
+ *
+ * @param {http.HTTPServerResponse} res - HTTP Response to respond with
+ * @param {Logger} log - Logger
+ * @returns {string} response
+ */
+function handleLiveness(res, log) {
+    if (lcConductor.isReady()) {
+        sendSuccess(res, log);
+    } else {
+        log.error('lifecycle conductor is not ready!');
```
miniscruff: sendError will do a `log.debug` and because it should return a >300 response it might be enough. Not sure we want to overly log here is all. Open to debate for sure though...

miniscruff (changes requested):
some minor points

extensions/lifecycle/conductor/service.js
naren-rajendran: Yes, but the existing feature uses old way to get/set metrics, thus retaining the old route until we consolidate.

extensions/lifecycle/conductor/service.js
naren-rajendran: The existing feature does produce some metrics, so retaining the old route and the method.
Ref: https://github.com/scality/backbeat/blob/development/8.2/extensions/lifecycle/conductor/LifecycleConductor.js#L13
Ref: https://github.com/scality/backbeat/blob/development/8.2/lib/KafkaBacklogMetrics.js

extensions/lifecycle/conductor/service.js
miniscruff: you should add both routes

extensions/lifecycle/conductor/service.js
naren-rajendran: Ah, it was a copy-paste from the old implementation.

extensions/lifecycle/conductor/service.js
miniscruff: oh i did not know that

miniscruff (approved):

nicolas2bert (approved):
Looks good for the lifecycle probes. 
For Zenko 1.x to work, we will need to do the same for ingestion consumer, ingestion producer and garbage collector processes.

naren-rajendran:
> Looks good for the lifecycle probes.
> For Zenko 1.x to work, we will need to do the same for ingestion consumer, ingestion producer and garbage collector processes.

Yes, we have separate tickets for these. I will add them to the sprint (if it can be squeezed into the current sprint)

PR #1936, opened 2021-09-18
feature: BB-50 Probe server should return undefined

miniscruff:
* Use sendError and sendSuccess from Arsenal

### DIFF
--- a/bin/ingestion.js
+++ b/bin/ingestion.js
@@ -3,7 +3,14 @@ const schedule = require('node-schedule');
 const zookeeper = require('node-zookeeper-client');
 
 const werelogs = require('werelogs');
-const { HealthProbeServer } = require('arsenal').network.probe;
+const { errors } = require('arsenal');
+const {
+    DEFAULT_LIVE_ROUTE,
+    DEFAULT_METRICS_ROUTE,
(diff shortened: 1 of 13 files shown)


naren-rajendran:
Ingestion processes now use the new probe server, these need the same treatment, https://github.com/scality/backbeat/pull/1938/files

naren-rajendran (commented):
Ingestion processes now use the new probe server, these need the same treatment, https://github.com/scality/backbeat/pull/1938/files

miniscruff:
> Ingestion processes now use the new probe server, these need the same treatment, https://github.com/scality/backbeat/pull/1938/files

Yea, after my change is merged you can do similar changes to your PR. Unless you think I should handle them.

naren-rajendran:
> > Ingestion processes now use the new probe server, these need the same treatment, https://github.com/scality/backbeat/pull/1938/files
> 
> Yea, after my change is merged you can do similar changes to your PR. Unless you think I should handle them.

I think it should be with this PR because it would be easier to understand the changes, since they are all same across the processes.

naren-rajendran (approved):

nicolas2bert (approved):

PR #1959, opened 2021-09-24
BB-67 migrate to github actions

tcarmet:


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,115 @@
...
+    - name: Install ginkgo
+      run: go get github.com/onsi/ginkgo/ginkgo@${GINKGO_VERSION}
+    - name: Lint markdown
+      run: yarn run --silent lint_md
+    - name: Lint Javascript
+      run: yarn run --silent lint
+    - name: Run unit tests
+      run: yarn test
+    - name: Run bucket scanner unit tests
+      run: ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --cover --trace --race --progress -nodes 1 -stream -timeout 5m -slowSpecThreshold 60
```
miniscruff, line 98: small thing here that we can change later, the bucket scanner can be a separate workflow as it is the only part using go and ginkgo

miniscruff (approved):

.github/workflows/tests.yaml
tcarmet: ooh ok, yes indeed. I added here because I was mostly following what was being done in Eve.

I'll let you track this one and add it as soon as this PR is merged.

naren-rajendran (approved):

.github/workflows/tests.yaml
miniscruff: Made https://scality.atlassian.net/browse/BB-78

jonathan-gramain (approved):

PR #1965, opened 2021-09-27
improvement: BB-69 provisioning: no notify if provisions unchanged

jonathan-gramain:
Do not send a provisioning update notification if provision list is
unchanged from before the notification.

This is an optimization that can also avoid possible side effects:

- unnecessary reinitialization of state (e.g. queue populator log
  reader) that can lead to extra CPU usage, duplicate work (replays),
  memory leaks etc.

- unnecessary logging when provisions are updated


jonathan-gramain:
ping

mrepop (approved):

alexanderchan-scality (approved):

tmacro (approved):

PR #1974, opened 2021-09-30
bugfix: BB-60 close LogReader on provisioning update

jonathan-gramain:
Make sure we close the LogReader instance properly when a raft session
provisioning update occurs. This LogReader.close() call closes the
Kafka producer instance if it has been opened, to fix the Kafka
connection leak.

(cherry picked from commit 3d8c1c0f9ff9bca03341ee0da43256526da8fb02)

### DIFF
--- a/lib/queuePopulator/QueuePopulator.js
+++ b/lib/queuePopulator/QueuePopulator.js
@@ -219,14 +219,21 @@ class QueuePopulator {
     _setupUpdatedReaders(done) {
         const newReaders = this.logReadersUpdate;
         this.logReadersUpdate = null;
-        async.each(newReaders, (logReader, cb) => logReader.setup(cb),
-                   err => {
-                       if (err) {
-                           return done(err);
-                       }
-                       this.logReaders = newReaders;
-                       return done();
-                   });
+        async.each(
+            newReaders,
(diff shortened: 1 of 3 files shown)


mrepop (approved):

lib/queuePopulator/QueuePopulator.js
```diff
@@ -219,14 +219,21 @@ class QueuePopulator {
...
-                   });
+        async.each(
+            newReaders,
+            (logReader, cb) => logReader.setup(cb),
+            err => {
+                if (err) {
+                    return done(err);
+                }
+                return async.each(
+                    this.logReaders, (logReader, cb) => logReader.close(cb),
```
alexanderchan-scality, line 230: should the error returned from `logReader.close` be ignored (ex.. with `() => cb()`)?

jonathan-gramain: In theory, I think you're right, we should not prevent other log readers from being closed if one of them has an error. In practice though, I believe `async.each` will just trigger all of them in parallel so it should not make any difference, hence looks more of a code style issue to me.

naren-rajendran (approved):

PR #2011, opened 2021-10-21
feature: BB-66 Patch location constraints

miniscruff:


lib/management/patchConfiguration.js
```diff
@@ -18,6 +18,154 @@ function updateIngestionBuckets(locations, metadata, logger, cb) {
     });
 }
 
+function patchLocations(oldLocations, log) {
```
nicolas2bert: `oldLocation` to `overlayLocation`?

nicolas2bert: Add a comment with a ticket number: to be moved to Arsenal.

nicolas2bert (commented):
Looks good!
Ingestion info will have a new structure, so this needs to be updated as well.
https://github.com/scality/backbeat/compare/feature/ARTESCA-1849/poc#diff-37642996a2b69aa6a6249a74d929d3183248506555dc43a488879f1f9f2d348fR534

lib/management/patchConfiguration.js
```diff
@@ -47,27 +197,40 @@ function updateLocations(conf, logger, cb) {
...
         return async.waterfall([
+            next => metadata.getObjectMD(
+                managementDatabaseName,
+                tokenConfigurationKey,
+                {},
+                logger,
+                next,
+            ),
+            (creds, next) => {
+                conf.locations = patchLocations(conf.locations, creds, logger);
```
nicolas2bert: Nothing asynchronous here, we can move `patchLocations` to the next callback.

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -464,40 +464,40 @@ class IngestionPopulator {
         // Example buckets param object:
```
nicolas2bert, line 463: This `applyUpdates` seems to be synchronous.

lib/management/patchConfiguration.js
```diff
@@ -18,63 +21,213 @@ function updateIngestionBuckets(locations, metadata, logger, cb) {
...
-            config.setIsTransientLocation(
-                locName, locations[locName].isTransient);
-        });
-    } catch (error) {
-        logger.error('error updating locations', {
-            error: error.message,
-            method: 'patchConfiguration::updateLocations',
-        });
-        return cb(error);
+function patchLocations(overlayLocations, creds, log) {
```
nicolas2bert, line 25: Please, add a comment with ticket number: "Move this logic to Arsenal"

lib/management/patchConfiguration.js
```diff
@@ -18,63 +21,213 @@ function updateIngestionBuckets(locations, metadata, logger, cb) {
...
+    return async.waterfall([
+        next => metadata.getObjectMD(
+            managementDatabaseName,
+            tokenConfigurationKey,
+            {},
+            logger,
+            next,
+        ),
+        (creds, next) => {
+            // only update bootstraplist on new overlay version
```
nicolas2bert: This comment is for the second condition `updateLocations`. 
For  this `patchLocations` condition, we can say something like: `only update locations on new overlay version`

lib/management/patchConfiguration.js
```diff
@@ -18,63 +21,213 @@ function updateIngestionBuckets(locations, metadata, logger, cb) {
...
+            if (overlayVersion === undefined || conf.version > overlayVersion) {
```
nicolas2bert, line 214: Regrouping the two conditional statements with the same conditions might be more readable?
```suggestion
            // only update locations and bootstrap list on new overlay version
            if (overlayVersion === undefined || conf.version > overlayVersion) {
                patchedLocations = patchLocations(conf.locations, creds, logger);
                conf.locations = patchedLocations;
                config.setPublicInstanceId(conf.instanceId);
                updateLocations(conf);
            } else {
                conf.locations = patchedLocations;
            }
```

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -457,57 +457,49 @@ class IngestionPopulator {
...
+            const sourceInfo = this._getSourceInformation(bucket);
             // if the zenko bucket has already been setup and is already
             // active, stop tracking bucket from `currentActiveSources`
-            const isAlreadyActive =
-                configuredSources.indexOf(zenkoBucket) >= 0;
-            if (isAlreadyActive) {
-                // Get ingestion source information
-                const sourceInfo = this._getSourceInformation(bucket);
+            const bucketIndex = configuredSources.indexOf(zenkoBucket);
+            if (bucketIndex >= 0) {
```
nicolas2bert: Let's keep the `const isAlreadyActive = bucketIndex >= 0` to better understand the logic.

lib/queuePopulator/IngestionReader.js
```diff
@@ -51,41 +50,22 @@ class IngestionReader extends LogReader {
         return ['isStatusComplete', 'versionMarker', 'keyMarker'];
     }
 
-    _setupIngestionProducer(cb) {
-        const encryptedKey = this.bucketdConfig.auth.secretKey;
-        return decryptLocationSecret(encryptedKey, this.logger,
```
nicolas2bert, line 56: Should we remove `decryptLocationSecret` and `decryptSecret` functions if not used?

tests/unit/management/patchConfiguration.js
```diff
@@ -116,15 +139,25 @@ describe('patchConfiguration', () => {
...
-                }
-                return list;
-            }, []);
+        let configOverlay;
+        let defaultLocations;
+        let expectedLocations;
+        beforeEach(() => {
+            configOverlay = createConfig();
+            defaultLocations = configOverlay.locations;
+            expectedLocations = [
```
nicolas2bert, line 146: Don't this `expectedLocations` get overwritten by the following assignment?

nicolas2bert: Oh! was your idea to define the `expectedLocations` instead of building it? We should do that, and remove the next assignment if we can!

miniscruff: yea I thought it was weird it was being generated, but forgot to remove the generated part...

lib/queuePopulator/IngestionReader.js
miniscruff: done, decrypt secret was being used in one other place but is a copy of the arsenal version of the same name

lib/queuePopulator/IngestionPopulator.js
miniscruff: sure

lib/management/patchConfiguration.js
miniscruff: was trying to remove the else + conf.locations thing but yea it still looks better this way

lib/management/patchConfiguration.js
miniscruff: yea moving code around with comments can be tricky...

lib/management/patchConfiguration.js
miniscruff: done, will add more detail to the ticket soon

lib/management/patchConfiguration.js
```diff
@@ -18,6 +20,154 @@ function updateIngestionBuckets(locations, metadata, logger, cb) {
     });
 }
 
+function patchLocations(overlayLocations, creds, log) {
+    if (!overlayLocations) {
```
jonathan-gramain, line 26: I believe this check is not needed, since the empty case is already handled (https://github.com/scality/backbeat/pull/2011/commits/2fa06259f5fe158288ca449fc9bbe55bb725441f#diff-0cd58e4c043e9e37c0d1aadc14132245714a2dc764c65842ec05c05f0a6e9f5aR29)

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -457,57 +457,50 @@ class IngestionPopulator {
 
     /**
      * Apply updates when ingestion buckets are updated in Config
-     * @param {Function} cb - callback(error)
      * @return {undefined}
      */
-    applyUpdates(cb) {
+    applyUpdates() {
```
nicolas2bert, line 462: `applyUpdates` seems to be called in `bin/ingestion.js` with a callback.

lib/management/patchConfiguration.js
```diff
@@ -5,6 +5,7 @@ const config = require('../Config');
 const { decryptSecret } = require('arsenal').pensieve.credentialUtils;
 const managementDatabaseName = 'PENSIEVE';
 const tokenConfigurationKey = 'auth/zenko/remote-management-token';
+let patchedLocations = {};
```
jonathan-gramain, line 8: does this have to be a global, instead of a function-local variable in `patchConfiguration`?

jonathan-gramain (approved):

lib/management/patchConfiguration.js
miniscruff: it is global so it is cached between calls to patch location yes.

lib/management/patchConfiguration.js
miniscruff: true, I removed the other check as I think the early return is a little more obvious than `|| {}` inside the Object.keys

lib/queuePopulator/IngestionPopulator.js
miniscruff: sneaky sneaky

nicolas2bert (approved):
We might want to clean up some commits and merge it :)

PR #2031, opened 2021-12-03
feature: support mongodb sharding conf

rachedbenmustapha:
In Backbeat's 8.3 branch, backbeat-api now connects to mongodb on startup and fails if a replicaset name is not provided (where 8.2 worked). We now need to add the `shardCollections` config param so that connection to mongo works at startup.


jonathan-gramain (approved):
LGTM, just it could be useful to include the ticket number in the commit messages

tmacro (approved):

PR #2032, opened 2021-12-08
feature: BB-86 add metrics for ingestion populator

miniscruff:
image
image

### DIFF
--- a/lib/queuePopulator/IngestionPopulatorMetrics.js
+++ b/lib/queuePopulator/IngestionPopulatorMetrics.js
@@ -3,56 +3,52 @@ const { ZenkoMetrics } = require('arsenal').metrics;
 const ingestionObjectsQueued = ZenkoMetrics.createCounter({
     name: 'ingestion_objects_queued_total',
     help: 'Total number of Kafka messages produced by the ingestion populator',
-    labelNames: ['origin', 'locationName', 'raftID'],
+    labelNames: ['origin', 'locationNameRaftID'],
 });
 
 const ingestionJournalEntriesProcessed = ZenkoMetrics.createCounter({
     name: 'ingestion_journal_entries_processed_total',
     help: 'Total amount of journal entries processed by the ingestion populator',
-    labelNames: ['origin', 'locationName', 'raftID'],
+    labelNames: ['origin', 'locationNameRaftID'],
 });
 
 const ingestionLogReadOffset = ZenkoMetrics.createGauge({
     name: 'ingestion_read_offset',
     help: 'Current read offset of metadata journal',
-    labelNames: ['origin', 'locationName', 'raftID'],
+    labelNames: ['origin', 'locationNameRaftID'],
 });
 
 const ingestionLogSize = ZenkoMetrics.createGauge({
     name: 'metadata_journal_size',
     help: 'Current size of metadata journal',
-    labelNames: ['origin', 'locationName', 'raftID'],
+    labelNames: ['origin', 'locationNameRaftID'],
 });
 
 class IngestionPopulatorMetrics {
     static onIngestionJournalProcessed(recordsRead, locationName, raftID) {
         ingestionJournalEntriesProcessed.inc({
             origin: 'ingestion',
-            locationName,
-            raftID,
+            locationNameRaftID: `${locationName}-${raftID}`,
         }, recordsRead);
     }
 
     static onIngestionQueued(topicEntries, locationName, raftID) {
         ingestionObjectsQueued.inc({
             origin: 'ingestion',
-            locationName,
-            raftID,
+            locationNameRaftID: `${locationName}-${raftID}`,
         }, topicEntries.length);
     }
 
     static onIngestionLogSaved(logOffset, logSize, locationName, raftID) {
         ingestionLogReadOffset.set({
             origin: 'ingestion',
-            locationName,
-            raftID,
+            locationNameRaftID: `${locationName}-${raftID}`,
         }, logOffset);
 
         if (logSize) {
             ingestionLogSize.set({
                 origin: 'ingestion',
-                locationName,
-                raftID,
+                locationNameRaftID: `${locationName}-${raftID}`,
             }, logSize);
         }
     }
--- a/lib/queuePopulator/IngestionPopulatorMetrics.js
+++ b/lib/queuePopulator/IngestionPopulatorMetrics.js
@@ -25,12 +25,12 @@ const ingestionLogSize = ZenkoMetrics.createGauge({
 });
 
 class IngestionPopulatorMetrics {
-    static onIngestionJournalProcessed(locationName, raftID) {
+    static onIngestionJournalProcessed(recordsRead, locationName, raftID) {
         ingestionJournalEntriesProcessed.inc({
             origin: 'ingestion',
             locationName,
             raftID,
-        });
+        }, recordsRead);
     }
 
     static onIngestionQueued(topicEntries, locationName, raftID) {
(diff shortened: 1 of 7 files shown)


lib/queuePopulator/IngestionPopulatorMetrics.js
```diff
@@ -0,0 +1,53 @@
...
+});
+
+const ingestionJournalEntriesProcessed = ZenkoMetrics.createCounter({
+    name: 'ingestion_journal_entries_processed_total',
+    help: 'Total amount of journal entries processed by the ingestion populator',
+    labelNames: ['origin'],
+});
+
+const ingestionLogReadOffset = ZenkoMetrics.createGauge({
+    name: 'replication_read_offset',
```
rahulreddy: Shouldn't this be `ingestion_read_offset`

lib/queuePopulator/IngestionPopulatorMetrics.js
```diff
@@ -0,0 +1,53 @@
...
+    help: 'Current read offset of metadata journal',
+    labelNames: ['origin'],
+});
+
+const ingestionLogSize = ZenkoMetrics.createGauge({
+    name: 'replication_log_size',
```
rahulreddy: `metadata_journal_size` or something along those lines

monitoring/ingestion/populator-dashboard.json
```diff
@@ -0,0 +1,658 @@
...
+    ]
+  },
+  "time": {
+    "from": "now-5m",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "OOB Ingestion Populator",
+  "uid": "fNCu1Ch7k",
```
nicolas2bert: Let's set `uid` to `null` to avoid any conflict. 
Grafana will generate a new `uid` at the import.

https://grafana.com/docs/grafana/latest/http_api/dashboard/

monitoring/ingestion/populator-dashboard.json
```diff
@@ -0,0 +1,658 @@
...
+            "value": "${VAR_NAMESPACE}",
+            "text": "${VAR_NAMESPACE}",
+            "selected": false
+          }
+        ]
+      }
+    ]
+  },
+  "time": {
+    "from": "now-5m",
```
nicolas2bert: 5 minutes seems a bit low to me.
I set it to 24h by default. (can be discussed)

monitoring/ingestion/populator-dashboard.json
```diff
@@ -0,0 +1,658 @@
...
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": true,
+      "steppedLine": false,
+      "targets": [
+        {
+          "exemplar": true,
+          "expr": "sum by (locationName, raftID, pod) (rate(ingestion_journal_entries_processed_total[$__rate_interval]))",
```
nicolas2bert: Should it be `ingestion_objects_queued_total` metrics name instead?

monitoring/ingestion/populator-dashboard.json
```diff
@@ -0,0 +1,658 @@
...
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": true,
+      "steppedLine": false,
+      "targets": [
+        {
+          "exemplar": true,
+          "expr": "sum by (locationName, raftID, pod) (rate(ingestion_objects_queued_total[$__rate_interval]))",
```
nicolas2bert: `ingestion_journal_entries_processed_total`

lib/queuePopulator/IngestionReader.js
```diff
@@ -317,6 +318,10 @@ class IngestionReader extends LogReader {
                     logStats.nbLogRecordsRead += 1;
                 }
                 logStats.nbLogEntriesRead += 1;
+                IngestionPopulatorMetrics.onIngestionJournalProcessed(
```
nicolas2bert: You might not need to call onIngestionJournalProcessed() for every entry.
`logStats.nbLogEntriesRead` is doing it for you.
You can retreive this value first thing in `_processPublishEntries`.

lib/queuePopulator/IngestionReader.js
```diff
@@ -401,15 +415,22 @@ class IngestionReader extends LogReader {
...
 
         async.series([
             next => this._writeInitState(initState, logger, next),
             next => {
                 if (batchState.nextLogOffset !== undefined &&
                     batchState.nextLogOffset !== this.logOffset) {
                     if (batchState.nextLogOffset > this.logOffset) {
+                        const logSize = logRes.info.cseq || logRes.info.end;
                         this.logOffset = batchState.nextLogOffset;
+                        IngestionPopulatorMetrics.onIngestionLogSaved(
```
nicolas2bert: We can calculate the lag from: https://github.com/scality/backbeat/blob/development/8.3/lib/queuePopulator/IngestionReader.js#L344
Pros:
- We will get the data earlier in the process, 
- We won't have to deal with the `const logSize = logRes.info.cseq || logRes.info.end;` condition since cseq should always be set. 
- We don't need to measure the lag at the initial state, it should always be 0 since we set `nextLogOffset = initState.cseq` from: https://github.com/scality/backbeat/blob/development/8.3/lib/queuePopulator/IngestionReader.js#L365

monitoring/ingestion/populator-dashboard.json
miniscruff: I was using 5m while testing as it makes the best looking graphs, I can set it to 24h but also not sure my current "visual" settings really matter. I was hoping they would be trimmed out entirely.

monitoring/ingestion/populator-dashboard.json
miniscruff: I think you are right, changed.

miniscruff:
@nicolas2bert all updated

nicolas2bert (approved):

monitoring/ingestion/populator-dashboard.json
```diff
@@ -0,0 +1,660 @@
...
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    },
+    {
+      "name": "VAR_JOB",
+      "type": "constant",
+      "label": "job",
+      "value": "populator",
```
nicolas2bert: Would you mind changing the job's value for: `artesca-data-backbeat-ingestion-producer-headless`.
It seems to be hardcoded.

monitoring/ingestion/populator-dashboard.json
```diff
@@ -0,0 +1,660 @@
...
+      "description": ""
+    },
+    {
+      "name": "VAR_NAMESPACE",
+      "type": "constant",
+      "label": "namespace",
+      "value": "default",
```
nicolas2bert: and the namespace's value one to `zenko`.

lib/queuePopulator/IngestionPopulatorMetrics.js
```diff
@@ -0,0 +1,61 @@
+const { ZenkoMetrics } = require('arsenal').metrics;
+
+const ingestionObjectsQueued = ZenkoMetrics.createCounter({
+    name: 'ingestion_objects_queued_total',
+    help: 'Total number of Kafka messages produced by the ingestion populator',
+    labelNames: ['origin', 'locationName', 'raftID'],
```
rahulreddy: If you use `locationName` and `raftID` as labels, essentially it allows filtering by either of those labels right? If that's the case, I don't see a value of that. I would rather use `locationName-raftID` to decrease cardinality

rahulreddy (approved):

miniscruff:
ping

PR #2046, opened 2022-01-06
BB-102: fix lifecycle rules check issue from BB-83

alexanderchan-scality:


extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -76,6 +77,17 @@ class LifecycleBucketProcessor {
         this._producer = null;
         this._kafkaBacklogMetrics = null;
 
+        this._supportedRulesObject = {};
+        supportedLifecycleRules.forEach(rule => {
+            this._supportedRulesObject[rule] = { enabled: true };
+        });
+        this._hasSupportedRules = Array.isArray(supportedLifecycleRules) &&
```
jonathan-gramain, line 81: Wouldn't we need to also check that at least one of the supported rules has the `enabled: true` flag? Just to maintain the same behavior than before.

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -76,6 +77,17 @@ class LifecycleBucketProcessor {
...
+            supportedLifecycleRules.length > 0;
```
tmacro: If `supportedLifecycleRules` can be something other than an array I think we should guard the `forEach` as well
```suggestion
        this._hasSupportedRules = Array.isArray(supportedLifecycleRules) &&
            supportedLifecycleRules.length > 0;
        this._supportedRulesObject = {};
        if (this._hasSupportedRules) {
            supportedLifecycleRules.forEach(rule => {
                this._supportedRulesObject[rule] = { enabled: true };
            });
        }
```

alexanderchan-scality: `supportedLifecycleRules` is a constant from [arsenal](https://github.com/scality/Arsenal/blob/development/8.1/lib/constants.js#L146); so it safe to say that it is always an array, but I see you point

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
alexanderchan-scality: I don't believe so. the `supportedLifecycleRules` will be a constant from arsenal, and the generated object will always have `enabled: true` for each key.

jonathan-gramain (approved):

tmacro (approved):

PR #2051, opened 2022-01-07
ft: BB-95 oob alert rules

naren-rajendran:
OOB alert rules added for ingestion processor and producer.

### DIFF
--- a/monitoring/ingestion/ingestion-producer-alert.yaml
+++ b/monitoring/ingestion/ingestion-producer-alert.yaml
@@ -1,0 +1,71 @@
+x-inputs:
+  - name: namespace
+    type: constant
+  - name: job_ingestion_producer
+    type: constant
+  - name: replicas
+    type: constant
+
+groups:
+- name: IngestionProducer
+  rules:
+
+  - alert: IngestionProducerDown
+    Expr: sum(up{namespace="${namespace}", job="${job_ingestion_producer}"}) == 0
+    For:   "30s"
+    Labels:
+      severity: critical
+    Annotations:
+      zenko_service: backbeat-ingestion-producer
+      description: "Ingestion Producer pod has been down for 30 seconds"
+      summary: "Ingestion Producer service is down"
+
+  - alert: IngestionProducerLagIncreasedBy3Percent
+    Expr: |
+      ((sum(metadata_journal_size{namespace="${namespace}", job="${job_ingestion_processor}"} -
+        ingestion_read_offset{namespace="${namespace}", job="${job_ingestion_processor}"}+1) -
+      sum((metadata_journal_size{namespace="${namespace}", job="${job_ingestion_processor}"} offset 5m) -
+        (ingestion_read_offset{namespace="${namespace}", job="${job_ingestion_processor}"} offset 5m)+1))
+        /
+      sum((metadata_journal_size{namespace="${namespace}", job="${job_ingestion_processor}"} offset 5m) -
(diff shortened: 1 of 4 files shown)


miniscruff (approved):

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -0,0 +1,74 @@
...
+    Labels:
+      severity: warning
+    Annotations:
+      zenko_service: backbeat-ingestion-producer
+      description: "Ingestion Producer pod has been down for 30 seconds"
+      summary: "Ingestion Producer service is degraded"
+
+  - alert: IngestionProducerLagIncreasedBy3Percent
+    Expr: |
+      ((sum by (origin) (metadata_journal_size-ingestion_read_offset+1) -
```
nicolas2bert: What we do is specifying the job label to gather metrics from the same job. 
The job label name is the name of the Kubernetes services (e.g: dev-backbeat-ingestion-producer-headless).
The job label is set by Prometheus operator when the Service Monitor discovers the Service and its endpoints.
If we use job label, we won't need to sum by origin.

We also specify the namespace, in case one day we deal with metrics coming from different namespaces.

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -0,0 +1,74 @@
...
+    Labels:
+      severity: warning
+    Annotations:
+      zenko_service: backbeat-ingestion-producer
+      description: "Ingestion Producer pod has been down for 30 seconds"
+      summary: "Ingestion Producer service is degraded"
+
+  - alert: IngestionProducerLagIncreasedBy3Percent
+    Expr: |
+      ((sum(metadata_journal_size-ingestion_read_offset+1) -
```
nicolas2bert: Other jobs (services) might have the same metrics name. 
Please specify the `job_ingestion_producer` and `namespace` labels.

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -0,0 +1,81 @@
...
+    For:   "30s"
+    Labels:
+      severity: critical
+    Annotations:
+      zenko_service: backbeat-ingestion-producer
+      description: "Ingestion Producer pod has been down for 30 seconds"
+      summary: "Ingestion Producer service is down"
+
+  - alert: IngestionProducerDegraded
+    Expr: sum(up{namespace="${namespace}", job="${job_ingestion_producer}"}) < ${replicas}
```
nicolas2bert: Ingestion producer has replicas "hardcoded" to 1.

naren-rajendran: which is technically not an issue :)

naren-rajendran: But I will change it!

naren-rajendran: Removing this alert because the number of pods is always 1, the first alert should be fine.

nicolas2bert: It is an issue, since `replicas` value is the number of replicas for backbeat processes (expect producer ones) and can be set to a value > 1.
We will fire unwanted alerts, since the sum expression will always return 1.

naren-rajendran: Ah ok..I thought it was per pod.

nicolas2bert (approved):

PR #2070, opened 2022-02-16
improvement: BB-122 pass "client.rack" property to kafka consumer

jonathan-gramain:
To enable follower fetching on stretched configuration, pass the
"client.rack" property to kafka consumers from the "site" attribute
configured from Federation.

There is no unit test attached as this is barely passing a
configuration param to Kafka consumer client and would not check if
the param was applied all the way. The envisioned way to test this
properly (almost) end-to-end would be once Integration runs a 2-site
configuration to check if the backbeat and bucket notification logs
indeed contain the "follower fetching enabled" log line, asserting
that the configuration was applied for the relevant topics when on a
2-site config.

It does not look like an option either to check if the consumer is
always fetching from a local replica, because Kafka does not
systematically consume from a local replica but only when it sees fit
based on other considerations (load balancing etc.).

Note: to make it work, it also requires changes on Federation side to
set the replica selector class and to add the "site" property to
backbeat configuration.

### DIFF
--- a/conf/config.joi.js
+++ b/conf/config.joi.js
@@ -19,6 +19,7 @@ const joiSchema = {
             zkPath: joi.string().default('/backbeat/run/kafka-backlog-metrics'),
             intervalS: joi.number().default(60),
         },
+        site: joi.string(),
     },
     s3: hostPortJoi.optional(),
     queuePopulator: {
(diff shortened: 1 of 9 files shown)


jonathan-gramain:
ping

naren-rajendran (approved):

nicolas2bert (approved):

jonathan-gramain:
ping

PR #2094, opened 2022-03-07
BB-119: add lifecycle metrics

alexanderchan-scality:


miniscruff (approved):

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -0,0 +1,89 @@
+const { ZenkoMetrics } = require('arsenal').metrics;
+
+const LIFECYCLE_LABEL_ORIGIN =  'origin';
+const LIFECYCLE_LABEL_PROCESS = 'process';
+const LIFECYCLE_LABEL_OP = 'op';
+const LIFECYCLE_LABEL_STATUS = 'status';
+
+const conductorLastBatchTime = ZenkoMetrics.createGauge({
+    name: 'lifecycle_conductor_last_batch_time',
+    help: 'Timestamp of latest lifecycle batch',
```
rachedbenmustapha: Since it marks the (end of the) start of the batch, it  should be in the metric name and description to avoid confusion (like if the last timestamp was 1h ago, someone could expect to see some results when in fact it could still be running).

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -0,0 +1,89 @@
...
+    labelNames: [
+        LIFECYCLE_LABEL_ORIGIN
```
rachedbenmustapha: Is the origin label standard in artesca / already used by other artesca components? As usual, trying to minimize cardinality in prometheus, this would benefit from merging with process, using `origin=lifecycle-expiration` instead of `origin=lifecycle,process=expiration`.

Also the origin itself is in the counter name so no need for an extra label.

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -0,0 +1,89 @@
...
+        LIFECYCLE_LABEL_PROCESS,
+    ],
+});
+
+const lifecycleKafkaPublish = ZenkoMetrics.createCounter({
+    name: 'lifecycle_kafka_publish',
+    help: 'Total number messages published by lifecycle processes',
+    labelNames: [
+        LIFECYCLE_LABEL_ORIGIN,
+        LIFECYCLE_LABEL_STATUS,
```
rachedbenmustapha: Since status here is binary (success vs error), it could be 2 counters, as they are easy to sum, without the label tax.

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    if (statusCode >= 200) {
+        return '2xx';
+    }
+
+    return '1xx';
+}
+
+class LifecycleMetrics {
```
rachedbenmustapha, line 80: It would be wise to wrap all of these in try blocks, to avoid crashing the process in case anything (bug or runtime issue) throws, as this is unrelated to data flow.

rachedbenmustapha (dismissed):

extensions/lifecycle/LifecycleMetrics.js
alexanderchan-scality: ok. I can split the counter into separate counts for the different http code ranges to reduce overhead from label processing

extensions/lifecycle/LifecycleMetrics.js
alexanderchan-scality: will remove the redundancy, if label consistency is not required between the extensions, `origin` and `op` will be the only required labels

extensions/lifecycle/LifecycleMetrics.js
rachedbenmustapha: Here specifically for Kafka messages and only 2 outcomes possible it can be 2 counters. For HTTP status code ranges, a label is common practice and a reasonable use of cardinality.

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -0,0 +1,107 @@
+const { ZenkoMetrics } = require('arsenal').metrics;
+
+const LIFECYCLE_LABEL_ORIGIN =  'origin';
+const LIFECYCLE_LABEL_OP = 'op';
+const LIFECYCLE_LABEL_STATUS = 'status';
+
+const conductorLatestBatchTime = ZenkoMetrics.createGauge({
```
francoisferrand: time (usually) increases, is this true also for this timestamp?

If that is the case, we may use counter instead : so that prometheus gets the "intention" that this field increase monotonically, and handles reset appropriately.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -69,7 +70,24 @@ class LifecycleTask extends BackbeatTask {
...
+                    'error',
+                );
+            } else {
+                LifecycleMetrics.onKafkaPublish(
+                    null,
+                    'sendBucketEntry',
+                    'bucket',
+                    'success',
+                );
+            }
```
francoisferrand: May be better to simplify code :

```suggestion
            LifecycleMetrics.onKafkaPublish(
                null,
                'sendBucketEntry',
                'bucket',
                err ? 'error' : 'success',
            );
```

Or even to pass the `err` directly to onKafkaPublish() function : and let that function handle the "mapping" to whatever metrics is needed (thus keeping the `success` and `error` private implementation details of LifecycleMetrics kcass)

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -109,7 +127,24 @@ class LifecycleTask extends BackbeatTask {
      */
     _sendObjectAction(entry, cb) {
         const entries = [{ message: entry.toKafkaMessage() }];
-        this.producer.sendToTopic(this.objectTasksTopic, entries, cb);
+        this.producer.sendToTopic(this.objectTasksTopic, entries,  err => {
+            if (err) {
```
francoisferrand: same

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -134,13 +169,22 @@ class LifecycleTask extends BackbeatTask {
         async.waterfall([
             next => req.send((err, data) => {
                 if (err) {
+                    LifecycleMetrics.onS3Request(
+                        log,
+                        'ListObjects',
+                        'bucket',
+                        err.statusCode
```
francoisferrand: similarly, may be better to always pass err to `onS3Request` : and let the implementation decide to use statusCode field or otherwise.

extensions/lifecycle/LifecycleMetrics.js
alexanderchan-scality: the timestamp is a  unix timestamp and will be a monotonically increasing number.

naren-rajendran (approved):

alexanderchan-scality:
ping

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -0,0 +1,113 @@
+const { ZenkoMetrics } = require('arsenal').metrics;
+
+const LIFECYCLE_LABEL_ORIGIN =  'origin';
+const LIFECYCLE_LABEL_OP = 'op';
+const LIFECYCLE_LABEL_STATUS = 'status';
+
+const conductorLatestBatchStartTime = ZenkoMetrics.createCounter({
+    name: 'lifecycle_latest_batch_start_time',
+    help: 'Timestamp of latest lifecycle batch start time',
+    labelNames: [LIFECYCLE_LABEL_ORIGIN],
```
rachedbenmustapha: There can be only 1 origin for this counter.

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -21,6 +21,8 @@ const VaultClientCache = require('../../../lib/clients/VaultClientCache');
 const CredentialsManager = require('../../../lib/credentials/CredentialsManager');
 const safeJsonParse = require('../util/safeJsonParse');
 
+const { LifecycleMetrics } = require('../LifecycleMetrics');
```
rachedbenmustapha, line 24: There could be a few ops captured here as well, such as list buckets, get account ids, ...

rachedbenmustapha (changes requested):

extensions/lifecycle/LifecycleMetrics.js
alexanderchan-scality: this however will need to be a gauge in order to use the `set` method

rachedbenmustapha (approved):

PR #2100, opened 2022-03-10
Feature/bb 120/lifecycle dashboards

alexanderchan-scality:
overview:
 
Bucket process view:

Object processor view:

### DIFF
--- a/monitoring/lifecycle/dashboard.py
+++ b/monitoring/lifecycle/dashboard.py
@@ -1,0 +1,288 @@
+from grafanalib.core import (
+    BarGauge,
+    ConstantInput,
+    DataSourceInput,
+    GaugePanel,
+    Heatmap,
+    HeatmapColor,
+    RowPanel,
+    Stat,
+    Threshold,
+    YAxis,
+)
+
+from grafanalib import formatunits as UNITS
+from scalgrafanalib import layout, Tooltip, Target, TimeSeries, Dashboard
+
+
+STATUS_CODE_2XX = '2xx'
+STATUS_CODE_3XX = '3xx'
+STATUS_CODE_4XX = '4xx'
+STATUS_CODE_5XX = '5xx'
+
+
+def s3_request_timeseries_expr(process, job, code):
+    labelSelector = 'namespace="${namespace}"'
+    labelSelector += f',status="{code}"'
+
+    if job is not None:
+        labelSelector += f',job="{job}"'
+
+    if process is not None:
+        labelSelector += f',origin="{process}"'
+
+    return f'sum(increase(lifecycle_s3_operations{{{labelSelector}}}[$__interval]))'
+
+
+def s3_request_timeseries(title, process=None, job=None):
+    return TimeSeries(
+        title=title,
+        dataSource="${DS_PROMETHEUS}",
+        fillOpacity=5,
+        legendDisplayMode='table',
+        targets=[
+            Target(
+                expr=s3_request_timeseries_expr(process, job, STATUS_CODE_2XX),
+                legendFormat="HTTP 2xx",
+            ),
+            Target(
+                expr=s3_request_timeseries_expr(process, job, STATUS_CODE_3XX),
+                legendFormat="HTTP 3xx",
+            ),
+            Target(
+                expr=s3_request_timeseries_expr(process, job, STATUS_CODE_4XX),
+                legendFormat="HTTP 4xx",
+            ),
+            Target(
+                expr=s3_request_timeseries_expr(process, job, STATUS_CODE_5XX),
+                legendFormat="HTTP 5xx",
+            ),
+        ]
+    )
+
+
+def s3_request_error_rate_expr(process, job, code):
+    divdLabel = 'namespace="${namespace}"'
+    divsLabel = 'namespace="${namespace}"'
+
+    if code is not None:
+        divdLabel += f',status="{code}"'
+    else:
+        divdLabel += ',status!="2xx"'
+
+    if job is not None:
+        divdLabel += f',job="{job}"'
+        divsLabel += f',job="{job}"'
+
+    if process is not None:
+        divdLabel += f',origin="{process}"'
+        divsLabel += f',origin="{process}"'
+
+    divd = f'sum(rate(lifecycle_s3_operations{{{divdLabel}}}[$__rate_interval]))'
+    divs = f'sum(rate(lifecycle_s3_operations{{{divsLabel}}}[$__rate_interval]) > 0)'
+
+    return f'{divd}/{divs}'
+
+
+def s3_request_error_rate(title, process=None, job=None, code=None):
+    return Stat(
+        title=title,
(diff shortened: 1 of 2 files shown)


miniscruff (dismissed):

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,266 @@
...
+        ),
+    ],
+    thresholds=[
+        Threshold("red", 0, 0.0),
+        Threshold("green", 1, 1.0),
+    ],
+)
+
+lifecycle_batch = Stat(
+    title="Lastest Batch Time",
```
rachedbenmustapha: Latest

rachedbenmustapha (commented):
- Doesn't it need to be explicitly pushed with `oras` in release.yaml with the other dashboards?
- I don't see any reference to the kafka counters

alexanderchan-scality:
@RachedBenMustapha I did not include the kafka as the message counters would be monitored using the kafka dashboards. do correct me if I am wrong

rachedbenmustapha:
@alexanderchan-scality I don't know if these new counters would be automatically picked up by the kafka dashboard or if they would automatically be aggregated?

In any case, since the kafka topics represent input/outputs of the lifecycle components, a dashboard that does not include queue size/message rates specific to lifecycle does not give a complete picture and having to navigate around is not great UX.

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+        thresholds=[
+            Threshold("green", 0, 0.0),
+        ],
+    )
+
+
+up = Stat(
+    title="Up",
+    dataSource="${DS_PROMETHEUS}",
+    reduceCalc="lastNotNull",
```
francoisferrand: should be 'last' : we always want to display the very last value, esp. if there are no more up pods

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+    return TimeSeries(
+        title=f'{op} requests',
+        dataSource="${DS_PROMETHEUS}",
+        stacking={
+            "group": "A",
+            "mode": "normal",
+        },
+        targets=[
+            Target(
+                expr=f'sum(rate(lifecycle_s3_operations{{{successLabel}}}[$__range]))',
```
francoisferrand: should we not use `$__rate_interval` instead of `$__range` ?

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+                label="job lifecycle object processor",
+                description="Name of the lifecycle object processor job, used to filter only lifecycle object processor instances",
+                value="artesca-data-backbeat-lifecycle-object-processor-headless",
+            ),
+        ],
+        panels=layout.column([
+            layout.row([lifecycle_batch, up], height=4),
+            layout.row([lifecycle_global_s3_requests], height=10),
+            layout.row(lifecycle_global_s3_error_rates, height=4),
+            RowPanel(title="Kafka"),
```
francoisferrand, line 273: do we really want an empty row?

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+                expr=f'sum(increase(lifecycle_kafka_publish_success{{op="{op}",namespace="{"${namesapce}"}}[$__range]))',
+            ),
+        ],
+    )
+
+
+def kafka_topic_rate(title, op):
+    return Stat(
+        title=title,
+        dataSource="${DS_PROMETHEUS}",
```
francoisferrand: should specify a unit

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+
+    if job is not None:
+        divdLabel += f',job="{job}"'
+        divsLabel += f',job="{job}"'
+
+    if process is not None:
+        divdLabel += f',origin="{process}"'
+        divsLabel += f',origin="{process}"'
+
+    return f'sum(rate(lifecycle_s3_operations{{{divdLabel}}}[$__range]))/sum(rate(lifecycle_s3_operations{{{divsLabel}}}[$__range]))'
```
francoisferrand: to avoid "not a number", you need to filter the expression `>0` :

```
sum(rate(lifecycle_s3_operations{...}[$__range]))
  /  sum(rate(lifecycle_s3_operations{..}[$__range]) > 0)
```

Just found the issue here: https://scality.atlassian.net/browse/ARTESCA-3935

francoisferrand: (and use `$__rate_interval` ?)

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def s3_request_error_rate(title, process=None, job=None, code=None):
+    return Stat(
+        title=title,
+        dataSource="${DS_PROMETHEUS}",
+        format=UNITS.PERCENT_UNIT,
```
francoisferrand, line 91: the `calc` should be set to `mean` : to get the average error rate over the current time frame

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+            ),
+        ],
+    )
+
+
+def kafka_topic_rate(title, op):
+    return Stat(
+        title=title,
+        dataSource="${DS_PROMETHEUS}",
+        targets=[
```
francoisferrand, line 142: not sure what `reduceCalc` should be used here: `mean` i guess?
may be better to explicitely set it, even if it is the default

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+            Target(
+                expr=f'sum(rate(lifecycle_kafka_publish_success{{something="{}",some}}[$__range]))',
```
francoisferrand: `{something="{}",some}` seems wrong or incomplete

monitoring/lifecycle/dashboard.py
```diff
@@ -0,0 +1,283 @@
...
+        ],
+    )
+
+
+def kafka_topic_rate(title, op):
+    return Stat(
+        title=title,
+        dataSource="${DS_PROMETHEUS}",
+        targets=[
+            Target(
```
francoisferrand, line 143: adding a unit (possibly custom, like "msg/s") would make this panel clearer

francoisferrand (dismissed):
I think the correct way to use rate() is passing `$__rate_interval` : `$__range` is documented as `to - from`, and thus would only be the "duration" of the time frame...

Also, adding all "details" (units, "noValue" when appropriate, different colors, shaded graph backgrounds, smooth lines...) where appropriate would make the dashboard more polished.

miniscruff (approved):

rachedbenmustapha (approved):

PR #2101, opened 2022-03-11
Feature/bb 121/lifecycle alerts

alexanderchan-scality:


monitoring/lifecycle/alerts.yaml
```diff
@@ -0,0 +1,129 @@
...
+    Expr: |
+      sum(rate(lifecycle_s3_operations{namespace="${namespace}", job="${job_lifecycle_bucket_processor}", status!="2xx", process="bucket"}}[1m]))
+        / sum(rate(lifecycle_s3_operations{namespace="${namespace}", job="${job_lifecycle_bucket_processor}", process="bucket"}}[1m]))
+        >= ${systemErrorsWarningThreshold}
+    For: "5m"
+    Labels:
+     severity: warning 
+    Annotations:
+      zenko_service: backbeat-lifecycle-bucket-processor
+      description: "More than 3% of S3 requests by bucket processors resulting in errors"
```
rachedbenmustapha, line 64: Are interpolations possible in descriptions? If not maybe it'd be best not to write numbers in the descriptions as they would not be inaccurate if the x-inputs change.

monitoring/lifecycle/alerts.yaml
```diff
@@ -0,0 +1,129 @@
...
+    Labels:
+     severity: critical 
+    Annotations:
+      zenko_service: backbeat-lifecycle-bucket-processor
+      description: "Less than 50% of lifecycle bucket processors are up and healthy"
+      summary: "Degraded lifecycle bucket processor"
+
+  - alert: LifecycleBucketProcessorRequestWarning
+    Expr: |
+      sum(rate(lifecycle_s3_operations{namespace="${namespace}", job="${job_lifecycle_bucket_processor}", status!="2xx", process="bucket"}}[1m]))
```
miniscruff: I think it has the same result, but I am not sure using rate here is accurate. The reason it probably doesn't matter is that you are using rate for both the numerator and denominator so it is still a percentage. But since rate does a difference over time I think this is more saying, the rate of failures is more then 3% the rate of success. Which I mean also sounds bad but even if the rate was decreasing it could still be more then 3% of the requests.

alexanderchan-scality: late reply to clarify:
* the math would still work out the same way when compared to the alternative of using `increase()`.
* in the decreasing case as mentioned, the error % would be an accurate % over the timerange; the divisor is all request.

monitoring/lifecycle/alerts.yaml
alexanderchan-scality: removed as a configurable option.

miniscruff (approved):

rachedbenmustapha (approved):

alexanderchan-scality:
ping

PR #2110, opened 2022-03-16
feature: BB-133 add replay specific backlog metrics

miniscruff:
image

Replay specific backlog of objects and bytes waiting to be replicated. This will be a subset of the replication backlog as it only contains items that have failed once and are waiting to be replayed. Items that are waiting there first attempt will not be counted here. 

Also a file size breakdown for objects in the queue and also objects that failed. Graph coming soon.

Also handles BB-132 and BB-136

### DIFF
--- a/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
+++ b/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
@@ -18,7 +18,11 @@ const FailedCRRProducer = require('../failedCRR/FailedCRRProducer');
 const ReplayProducer = require('../replay/ReplayProducer');
 const promClient = require('prom-client');
 const constants = require('../../../lib/constants');
-const { wrapCounterInc, wrapGaugeSet } = require('../../../lib/util/metrics');
+const {
+    wrapCounterInc,
+    wrapGaugeSet,
+    wrapHistogramObserve,
+} = require('../../../lib/util/metrics');
 
 promClient.register.setDefaultLabels({
     origin: 'replication',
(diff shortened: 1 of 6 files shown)


extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -35,44 +39,101 @@ promClient.register.setDefaultLabels({
...
+ * @param {Object} repConfig - Replication configuration
+ * @returns {ReplicationStatusMetricsHandler} Metric handlers
  */
-const metricsHandler = {
-    status: wrapCounterInc(replicationStatusMetric),
-    lag: wrapGaugeSet(kafkaLagMetric),
-    replayAttempts: wrapCounterInc(replayAttempts),
-    replaySuccess: wrapCounterInc(replaySuccess),
-};
+function loadMetricHandlers(repConfig) {
```
miniscruff, line 61: Metrics need to be loaded post constructor since we need some configs

nicolas2bert: This file is pretty large, would it be a good idea to move the metrics logic to its own file?

nicolas2bert (approved):

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
miniscruff: Generally I would say yes, but with having to keep forward ports relatively easy it might be easier to keep it here even though it is quite large. I am a little concerned on switching to a function instead of top level consts in the general scheme. I do like this pattern as it allows a lot more configurations and testing. But they are still tied to the global prometheus registry, which is not ideal. Maybe we could find a pattern that fits our work and adapt it moving forward.

naren-rajendran (approved):

PR #2116, opened 2022-03-18
BB-134 - Set zstd compression in producers

jbertran:
Setting compression to `zstd` in producer configuration tackles one part of ZENKO-3223 - the other part is explicitly setting `producer` compression at the broker level.

Compression settings in Kafka are transparent: messages are sent in a compressed form by producers, or compressed at rest according to broker or topic configuration, but consumers receive the uncompressed message.

Additionally, brokers in Kafka are configured by default to `producer` compression, that is to say they match the compression setting of the producer sending the data and make no additional compression on their side.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -81,7 +81,7 @@ jobs:
     - name: Install build dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y build-essential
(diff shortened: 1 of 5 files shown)


jbertran:
You're drunk Bert-E, refresh pls

tmacro (approved):

francoisferrand (approved):

PR #2117, opened 2022-03-23
BB-139: throttle conductor depending on bucket queue

rachedbenmustapha:
- Fixes crash of lifecycle conductor when kafka is committing messages much slower than they're pulled from bucketd.
- Adds test that batches run in constant memory no matter the number of buckets (removing the circuit breaker causes the crash almost immediately).

### DIFF
--- a/.github/dockerfiles/syntheticbucketd/Dockerfile
+++ b/.github/dockerfiles/syntheticbucketd/Dockerfile
@@ -1,0 +1,10 @@
+FROM node:10
+
+RUN mkdir /app
+COPY package.json /app
+COPY yarn.lock /app
+RUN cd /app && yarn install
+
+COPY tests/utils /app
+
+CMD ["node", "/app/syntheticbucketd.js", "200000", "1000"]
--- a/tests/utils/syntheticbucketd.js
+++ b/tests/utils/syntheticbucketd.js
@@ -1,0 +1,90 @@
+const url = require('url');
+const http = require('http');
+const assert = require('assert');
+
+const werelogs = require('werelogs');
+const logger = new werelogs.Logger('Backbeat:syntheticbucketd');
+
+const bucketdPort = 9001;
+
+const bucketNumber = Number.parseInt(process.argv[2], 10);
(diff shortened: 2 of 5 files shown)


.github/dockerfiles/syntheticbucketd/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:10
```
miniscruff, line 1: Should we use node 16 here

tests/utils/syntheticbucketd.js
```diff
@@ -0,0 +1,91 @@
...
+    if (marker) {
+        return Number.parseInt(marker.split('..|..')[1], 10) + 1;
+    }
+
+    return 0;
+}
+
+function generateAccountIds(accountsNumber) {
+    const accountPadding16 = '0000000000000000';
+    const accountPadding = accountPadding16 + accountPadding16 +
```
miniscruff: You can use `.repeat(4);`

tests/utils/syntheticbucketd.js
```diff
@@ -0,0 +1,91 @@
...
+function makeListing(marker, maxKeys) {
+    const accountIds = generateAccountIds(accountsNumber);
+    const index = getMarkerIndex(marker);
+    const ret = [];
+
+    const upperBound = Math.min(index + maxKeys, bucketNumber);
+
+    for (let i = index; i < upperBound; i += 1) {
+        const account = accountIds[i % accountsNumber];
+        const name = (`000000000000000000000000${i}`.slice(-24));
```
miniscruff: can do `'0'.repeat(N)`

.github/dockerfiles/syntheticbucketd/Dockerfile
rachedbenmustapha: It's a hotfix branch, we can't change the nodejs version.

miniscruff: Fair

miniscruff (approved):

tmacro (approved):

PR #2119, opened 2022-03-24
bugfix: BB-141-remove v1 prefix in IngestionReader

welansari:


francoisferrand (approved):
maybe we should add another test with "v0" buckets (check that the key is not affected)

tmacro (approved):

PR #2121, opened 2022-03-24
Throttle conductor depending on bucket queue

rachedbenmustapha:
Same as #2117:
- Fixes crash of lifecycle conductor when kafka is committing messages much slower than they're pulled from bucketd.
- Adds test that batches run in constant memory no matter the number of buckets (removing the circuit breaker causes the crash almost immediately).

### DIFF
--- a/.github/dockerfiles/syntheticbucketd/Dockerfile
+++ b/.github/dockerfiles/syntheticbucketd/Dockerfile
@@ -1,0 +1,10 @@
+FROM node:16
+
+RUN mkdir /app
+COPY package.json /app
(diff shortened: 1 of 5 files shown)


.github/dockerfiles/syntheticbucketd/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:16
```
miniscruff, line 1: ✨

miniscruff (approved):

nicolas2bert (approved):

PR #2133, opened 2022-04-02
BB-148: add processBuckets callback function type check

alexanderchan-scality:
add callback type check before invoking in processBuckets method.
cron scheduler does not pass a function type as a parameter to
processBuckets.


naren-rajendran (approved):

miniscruff (approved):

rachedbenmustapha (changes requested):
If the signature of cron callbacks changed, the it should be updated at the signature level and in the tests, rather than checking the type at runtime.

miniscruff (approved):

rachedbenmustapha (approved):

alexanderchan-scality:
ping

PR #2136, opened 2022-04-12
feature: BB-149 support s3:Replication:OperationFailedReplication notification

welansari:
[ZENKO-3885](https://scality.atlassian.net/browse/ZENKO-3885)

### DIFF
--- a/tests/unit/replication/ReplicateObject.spec.js
+++ b/tests/unit/replication/ReplicateObject.spec.js
@@ -1,0 +1,139 @@
+const assert = require('assert');
+const errors = require('arsenal').errors;
+const sinon = require('sinon');
+
+const config = require('../../config.json');
+const notificationConfig = require('../../config.notification.json');
+const ReplicateObject =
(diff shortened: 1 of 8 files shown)


tmacro (approved):

tests/unit/replication/ReplicateObject.spec.js
```diff
@@ -0,0 +1,131 @@
...
+                'originOp': 'origin-operation',
+                'dataStoreName': 'metastore',
+                'md-model-version': '1',
+                'content-length': '6',
+                'versionId': '1234',
+            });
+            task._publishFailedReplicationStatusNotification(sourceEntry, fakeLogger);
+            const entryPublished = {
+                key: 'example-bucket',
+                message: '{\"dateTime\":\"0000\",\"eventType\":\"s3:Replication:OperationFailedReplication\"' +
```
jbertran: Not sure you need to escape the double quotes here. This may also be more readable as a `JSON.stringify`d object, even if it feels too close to implementation? Up to you

jbertran (approved):

PR #2140, opened 2022-04-18
BB-152 re-queuing batch only once for Lifecycle Expiration process

nicolas2bert:


nicolas2bert:
ping

miniscruff (approved):

anurag4DSB (approved):

rachedbenmustapha (approved):
LGTM, this will avoid the snowballing of bucket tasks when S3 error rates are greater than 0.1% (with 1000 objects per listing).

This changes the queueing from at-least-once to at-most-once, when we need exactly-once, so I've created S3C-5714 for the case when the kafka publisher intermittent error rate is greater than `(100*1000/NObjectsPerBucket)%`, which would cause buckets to never finish scanning ever.

nicolas2bert:
@bert-e approve

PR #2144, opened 2022-04-20
BB-156 objectSizeMetrics config property is undefined by default

nicolas2bert:


nicolas2bert:
ping

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -4,6 +4,7 @@ const { hostPortJoi, transportJoi, bootstrapListJoi, adminCredsJoi } =
     require('../../lib/config/configItems.joi.js');
 
 const CRR_FAILURE_EXPIRY = 24 * 60 * 60; // Expire Redis keys after 24 hours.
+const OBJECT_SIZE_METRICS = [66560, 8388608, 68157440];
```
francoisferrand, line 7: Did you mean this?

```suggestion
const OBJECT_SIZE_METRICS = [66536, 8388608, 68157440];
```

(not sure how you came up with the last number though, it is quite different from 64*1024*1024...)

Generally, where do these values come from?

nicolas2bert: These values had been defined by @miniscruff and @cmrh . 
Knowledge from the field I assumed.
This fix is about defining a default value, I just reused values that were intended to be used.

miniscruff: under 65k is COS keys, 8MB is default MPU, 65MB+ is SPLIT keys ( from Cyrus )
I may have messed up my math though.
Here is what I did: `65*1024, 8*1024*1024 and 65*1024*1024`

anurag4DSB (approved):

extensions/replication/ReplicationConfigValidator.js
nicolas2bert: Not sure to understand what are the 3 values, you guys decided. @miniscruff 
What I see from your commit are those 3 values: https://github.com/scality/backbeat/blob/development/7.10/conf/config.json#L117

francoisferrand: @miniscruff your math is good, thanks for the clarifications!

francoisferrand (approved):

extensions/replication/ReplicationConfigValidator.js
miniscruff: Note that this is just what Cyrus thought was a good idea and that better values probably exist, hence why there is a config option. Maybe down the road when we have more info from the field we can update the default values.

nicolas2bert:
@bert-e approve

PR #2151, opened 2022-04-25
Feature: BB-124 replace zookeeper by mongo GET in notification extension

welansari:
Issue: [BB-124](https://scality.atlassian.net/browse/BB-124)

### DIFF
--- a/extensions/notification/NotificationConfigManager.js
+++ b/extensions/notification/NotificationConfigManager.js
@@ -1,335 +1,221 @@
-const async = require('async');
-const { EventEmitter } = require('events');
-const zookeeper = require('node-zookeeper-client');
 const joi = require('joi');
 
-const safeJsonParse = require('./utils/safeJsonParse');
+const LRUCache = require('arsenal').algorithms
+    .cache.LRUCache;
+const MongoClient = require('mongodb').MongoClient;
+
 const constants = require('./constants');
 
 const paramsJoi = joi.object({
-    zkClient: joi.object().required(),
-    parentNode: joi.string().optional(),
+    mongoConfig: joi.object().required(),
     logger: joi.object().required(),
 }).required();
+
+const MAX_CACHED_ENTRIES = Number(process.env.MAX_CACHED_BUCKET_NOTIFICATION_CONFIGS)
+    || 1000;
 
 /**
  * @class NotificationConfigManager
  *
- * @classdesc Manages bucket notification configurations in zookeeper, maintains
- * configurations across multiple managers.
+ * @classdesc Manages bucket notification configurations, the configurations
+ * are directly retrieved from the metastore, and are locally cached. Cache
+ * is invalidated using MongoDB change streams.
  */
 class NotificationConfigManager {
     /**
      * @constructor
      * @param {Object} params - constructor params
-     * @param {Object} params.zkClient - zookeeper client
-     * @param {String} params.parentNode - parent node to be used
+     * @param {String} params.mongoConfig - config for connecting to mongo
      * @param {Logger} params.logger - logger object
      */
     constructor(params) {
         joi.attempt(params, paramsJoi);
-        this._zkClient = params.zkClient;
-        this._parentNode = params.parentNode || constants.zkConfigParentNode;
-        this.log = params.logger;
-        this._configs = new Map();
-        this._concurrency = constants.configManager.concurrency;
-        this._emitter = new EventEmitter();
-        this._setupEventListeners();
+        this._logger = params.logger;
+        this._mongoConfig = params.mongoConfig;
+        this._cachedConfigs = new LRUCache(MAX_CACHED_ENTRIES);
+        this._mongoClient = null;
+        this._metastore = null;
+        this._metastoreChangeStream = null;
+        this._changeStreamResumeToken = null;
     }
 
-    _errorListener(error, listener) {
-        this.log.error('BucketNotifConfigManager.emitter.error', {
-            listener,
-            error,
-        });
-        return undefined;
-    }
+    /**
+     * Connects to MongoDB using the MongoClientInterface
+     * and retreives the metastore collection
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _setupMongoClient(cb) {
+        const { authCredentials, replicaSetHosts, replicaSet, database }
+            = this._mongoConfig;
 
-    _setConfigListener(bucket, config) {
-        this.log.debug('BucketNotifConfigManager.emitter.setConfig', {
-            event: 'setConfig',
-            bucket,
-            config,
-        });
-        this._setBucketNotifConfig(bucket, JSON.stringify(config), err => {
-            if (err) {
-                this._emitter.emit('error', err, 'setConfigListener');
-            }
-            return undefined;
-        });
-    }
-
-    _getConfigListener(updatedBucket = '') {
-        this.log.debug('BucketNotifConfigManager.emitter.getConfig', {
-            event: 'getConfig',
-        });
-        this._listBucketsWithConfig((err, buckets) => {
-            if (err) {
-                this._emitter.emit('error', err, 'getConfigListener');
-                return undefined;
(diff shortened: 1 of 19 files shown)


extensions/notification/NotificationConfigManager.js
```diff
@@ -1,436 +1,216 @@
...
-        return async.waterfall([
-            next => this._checkNodeExists(zkPath, next),
-            (exists, next) => {
-                if (!exists) {
-                    return this._createBucketNotifConfigNode(bucket,
-                        err => next(err));
+    /**
+     * Handler for the change stream "change" event.
+     * Invalidates cached bucket configs based on the change.
+     * @param {Object} change Change stream change object
```
jbertran: We may as well leverage the partial typing provided by the mongodb driver - this object is a [ChangeStreamDocument](https://mongodb.github.io/node-mongodb-native/4.5/interfaces/ChangeStreamDocument.html).

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,436 +1,216 @@
...
-            },
-            next => this._zkClient.setData(zkPath, Buffer.from(data), -1, next),
-        ], err => {
-            if (err) {
-                this.log.error('error saving config', { method, zkPath, data });
-            }
-            return this._callbackHandler(cb, err);
+                break;
+            default:
+                this._logger.debug('Unsupported change stream event', {
```
jbertran: Nitpick: `Skipping unsupported ... ` so we understand that this is an expected non-error case.

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,436 +1,216 @@
...
-            this._getBucketNotifConfig(bucket, (err, data) => {
-                if (err) {
-                    return next(err);
-                }
-                const configObject = this._getConfigDataFromBuffer(data);
-                if (configObject) {
-                    this._configs.set(bucket, configObject);
-                }
-                return next();
+                this._setMetastoreChangeStream(() => {});
```
jbertran: Shouldn't we just clear listeners on the changeStream, rather than change the callback?

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,436 +1,216 @@
...
+     * Sets up the NotificationConfigManager by
+     * connecting to mongo and initializing the
+     * change stream
+     * @param {Function} cb callback
+     * @returns {undefined}
      */
-    getConfig(bucket) {
-        return this._configs.get(bucket);
+    setup(cb) {
+        async.series([
```
jbertran: Does this need to be async? Could just be setup/setmetastore wrapped in a try/except

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -36,89 +34,10 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
      * @return {boolean} - true if bucket entry
      */
     _isBucketEntry(bucket, key) {
-        return ((bucket.toLowerCase() === notifConstants.bucketMetastore && key)
+        return ((bucket.toLowerCase() === notifConstants.bucketMetastore && !!key)
             || key === undefined);
```
jbertran, line 38: Wish we had TS here - that's a good candidate for a literal type

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -128,18 +47,18 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
         }
-        const config = this.bnConfigManager.getConfig(bucket);
+        const config = await this.bnConfigManager.getConfig(bucket);
         if (config && Object.keys(config).length > 0) {
             const { eventMessageProperty }
                 = notifConstants;
             let eventType
                 = value[eventMessageProperty.eventType];
-            if (eventType === undefined && type === 'del') {
+            if (eventType === undefined && type === 'delete') {
```
jbertran, line 61: Where is this change from? (hey, another TS literal type waiting to happen)

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -178,9 +97,9 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
      * filter
      *
      * @param {Object} entry - log entry
-     * @return {undefined}
+     * @return {Promise|undefined} promise|undefiend
```
jbertran: Typo

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -189,15 +108,12 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
             this.log.error('could not parse log entry', { value, error });
             return undefined;
         }
-        // ignore bucket op, mpu's or if the entry has no bucket
+        // ignore bucket operations, mpu's or if the entry has no bucket
         if (!bucket || bucket === usersBucket ||
-            (key && key.startsWith(mpuBucketPrefix))) {
+            (key && key.startsWith(mpuBucketPrefix)) ||
```
jbertran: For legibility, this could become a `.some` call

jbertran (commented):
A few nitpicks. Let's sync tomorrow to set my brain back on track

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,436 +1,216 @@
...
 const constants = require('./constants');
 
 const paramsJoi = joi.object({
-    zkClient: joi.object().required(),
-    parentNode: joi.string().optional(),
+    mongoConfig: joi.object().required(),
     logger: joi.object().required(),
 }).required();
 
+const MAX_CACHED_ENTRIES = 1000;
```
jonathan-gramain: This should be configurable IMHO since it's a tunable, maybe via an environment variable

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,436 +1,216 @@
...
+
 /**
  * @class NotificationConfigManager
  *
- * @classdesc Manages bucket notification configurations in zookeeper, maintains
- * configurations across multiple managers.
+ * @classdesc Manages bucket notification configurations, the configurations
+ * are directly retreived from the metastore, and are locally cached. Cache
```
jonathan-gramain:
```suggestion
 * are directly retrieved from the metastore, and are locally cached. Cache
```

extensions/notification/NotificationConfigManager.js
welansari: i agree that we should clear listeners, but i think we should also try restarting the stream if possible
An error at the change stream level will not cause the NotificationConfigManager to malfunction in the sense that the `getConfig()` function will still work. So if we just close the change stream when getting an error, the config in cache will never get invalidated, and the extension would still receive and process notification events

extensions/notification/NotificationQueuePopulator.js
welansari: I'm not actually sure on when or why it got changed, but you can see the part of code that sets the type of the entries  here 

https://github.com/scality/Arsenal/blob/019907e2ab841acc70a32331fec77d99efc4265c/lib/storage/metadata/mongoclient/ListRecordStream.js#L120-L126

jbertran (approved):
LGTM!

tmacro (approved):

lib/queuePopulator/LogReader.js
```diff
@@ -407,25 +416,29 @@ class LogReader {
             logStats.nbLogRecordsRead += 1;
 
             this._setEntryBatch(entriesToPublish);
-            record.entries.forEach(entry => {
+            /**
+             * waiting for all sync and async extentions
+             * to finish processing before emptying the batch
+             */
+            async.each(record.entries, async entry => {
```
jonathan-gramain: It may be important to process the entries in series (as is, only call the next async `_processLogEntry` when the previous one completes), in which case using `async.eachSeries` is needed. Otherwise, it could still be nicer to use `async.eachLimit` to limit the parallelism.

PR #2158, opened 2022-05-03
Feature: BB-163 add kafka connect api

welansari:
Issue: [BB-163](https://scality.atlassian.net/browse/BB-163)

- Added wrapper around kafka connect REST API

### DIFF
--- a/extensions/oplogPopulator/KafkaConnectWrapper.js
+++ b/extensions/oplogPopulator/KafkaConnectWrapper.js
@@ -1,0 +1,307 @@
+const { errors } = require('arsenal');
+const joi = require('joi');
+const http = require('http');
+
+const paramsJoi = joi.object({
+    kafkaConnectHost: joi.string().required(),
+    kafkaConnectPort: joi.number().required(),
+    logger: joi.object().required(),
+}).required();
+
+/**
+ * @class KafkaConnectWrapper
+ *
+ * @classdesc Wrapper arround the Kafka Connect REST API.
+ */
+class KafkaConnectWrapper {
+
+    /**
+     * @constructor
+     * @param {Object} params constructor params
+     * @param {string} params.kafkaConnectHost kafka connect host
+     * @param {number} params.kafkaConnectPort kafka connect port
+     * @param {Logger} params.logger - logger object
+     */
+    constructor(params) {
+        joi.attempt(params, paramsJoi);
+        this._kafkaConnectHost = params.kafkaConnectHost;
+        this._kafkaConnectPort = params.kafkaConnectPort;
+        this._logger = params.logger;
+    }
+
+    /**
+     * makes http request
+     * @param {Object} params request params
+     * @param {string} params.method http method
+     * @param {string} params.path path to resource
+     * @param {string} params.headers request headers (optional)
+     * @param {Object} data request data
+     * @returns {Promise|Object} http response
+     */
+    makeRequest(params, data) {
+        return new Promise((resolve, reject) => {
+            const req = http.request({
+                host: this._kafkaConnectHost,
+                port: this._kafkaConnectPort,
+                ...params,
+            }, res => {
+                // cumulate data
+                let body = [];
+                res.on('data', chunk => {
+                    body.push(chunk);
+                });
+                // resolve on end
+                res.on('end', () => {
+                    try {
+                        body = JSON.parse(Buffer.concat(body).toString());
+                    } catch (e) {
+                        this._logger.error('Error occured while parsing request body', {
+                            method: 'KafkaConnectManager.makeRequest',
+                            error: e.message,
+                            params,
+                        });
+                        return reject(errors.InternalError.customizeDescription(
+                            `Error occured while parsing request body, ${e.message}`));
+                    }
+                    // reject on bad status
+                    if (res.statusCode < 200 || res.statusCode >= 300) {
+                        this._logger.error('Got bad status when making request', {
+                            method: 'KafkaConnectManager.makeRequest',
+                            statusCode: res.statusCode,
+                            statusMessage: res.message,
+                            bodyErrorCode: body.error_code,
+                            bodyMessage: body.message,
+                            params,
+                        });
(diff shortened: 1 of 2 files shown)


francoisferrand:
This PR includes all commits from #2151, please rebase this branch or stack PRs.

francoisferrand:
Also, not that this should target 8.4 branch

extensions/oplogPopulator/KafkaConnectWrapper.js
```diff
@@ -0,0 +1,80 @@
...
+     * @param {Logger} params.logger - logger object
+     */
+    constructor(params) {
+        joi.attempt(params, paramsJoi);
+        this._kafkaConnectHost = params.kafkaConnectHost;
+        this._logger = params.logger;
+    }
+
+    /**
+     * Returns a list of active connectors
```
francoisferrand: This function does not just _returns_, and neither _any_ list → "Get the list of active connectors"

extensions/oplogPopulator/KafkaConnectWrapper.js
```diff
@@ -0,0 +1,80 @@
...
+     * @returns {Array<string>} a list of active connectors
+     */
+    async getConnectors() {
+        const resp = await axios.get(`http://${this._kafkaConnectHost}/connectors`);
+        return resp.data;
+    }
+
+    /**
+     * Returns the configuration object of a connector
```
francoisferrand: same

extensions/oplogPopulator/KafkaConnectWrapper.js
```diff
@@ -0,0 +1,80 @@
...
+        joi.attempt(params, paramsJoi);
+        this._kafkaConnectHost = params.kafkaConnectHost;
+        this._logger = params.logger;
+    }
+
+    /**
+     * Returns a list of active connectors
+     * @returns {Array<string>} a list of active connectors
+     */
+    async getConnectors() {
```
francoisferrand, line 106: async : while I welcome use of async (and esp. here, since we are developing a brand new extension), I wonder if we won't have issues introducing it here.

e.g. if we need to mix kafka-connect client code with the rest of backbeat code (e.g. logreader, ...) which is still using callbacks

Can you try to double check, to estimate if this will not be a problem to mix-and-match?

yarn.lock
```diff
@@ -830,6 +830,14 @@ axios@^0.18.0:
     follow-redirects "1.5.10"
     is-buffer "^2.0.2"
 
+axios@^0.27.2:
```
francoisferrand: is `axios` already used, directly or in dependencies?
To keep the code maintainable, it is best to avoid introduced "redundant" deps, and use the same approach for similar uses (e.g. what is used when connecting to other APIs)

welansari: replaced axios with native node http module

extensions/oplogPopulator/KafkaConnectWrapper.js
welansari: I switched the code to callbacks to avoid any future complications as it is hard to guess where we might use this part of the code

extensions/oplogPopulator/KafkaConnectWrapper.js
```diff
@@ -0,0 +1,322 @@
...
+     * @param {string} params.host host name
+     * @param {string} params.port port
+     * @param {string} params.method http method
+     * @param {string} params.path path to resource
+     * @param {string} params.headers request headers
+     * @param {Object} data request data
+     * @returns {Promise} http response
+     */
+    makeRequest(params, data) {
+        return new Promise((resolve, reject) => {
```
francoisferrand, line 42: i am wondering why use a Promise here, when you always call it with the same pattern : `makeRequest().then(res -> cb(null, res))`

it may make transition to actual async/promise easier in the future... You may want to discuss this with @jbertran , I think he talked about an approach to "transition" API to async/promise...

extensions/oplogPopulator/KafkaConnectWrapper.js
```diff
@@ -0,0 +1,322 @@
...
+
+    /**
+     * Get the list of active connectors
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    getConnectors(cb) {
+        this.makeRequest({
+            host: this._kafkaConnectHost,
+            port: this._kafkaConnectPort,
```
francoisferrand: these params may be set by default in makeRequest : so we don't need to add host/port everytime, and method when it is a GET...

e.g. something like this in `makeRequest` :

```
http.request([host: this._kafkaConnectHost, port: this._kafkaConnectPort, method: 'GET'] + params, res => {
   ...
```

francoisferrand (approved):

extensions/oplogPopulator/KafkaConnectWrapper.js
jbertran: Late to the party, but the following describes the idea: https://scality.atlassian.net/wiki/spaces/OS/pages/1869774969/TypeScript+package+maintenance+strategies#The-use-of-%40deprecated

Essentially, this means exposing 2 overloads of the same function - one returning a promise, the other a callback - and making TS connect the dots, with a deprecation on the function call we want to remove.

This doesn't work in pure JS, because JS doesn't have overloading.

jbertran (approved):

extensions/oplogPopulator/KafkaConnectWrapper.js
```diff
@@ -0,0 +1,307 @@
...
+    /**
+     * makes http request
+     * @param {Object} params request params
+     * @param {string} params.method http method
+     * @param {string} params.path path to resource
+     * @param {string} params.headers request headers (optional)
+     * @param {Object} data request data
+     * @returns {Promise|Object} http response
+     */
+    makeRequest(params, data) {
```
francoisferrand, line 41: in preparation for the coming typescript migration, we may want to start "typing" (or pre-typing in comments) these kind of parameters....

(or maybe we should have added this file as typescript already, btw? maybe we'll try that next time...)

PR #2165, opened 2022-05-04
Bugfix: BB-159 version delete not producing notification

welansari:
Issue: [BB-159](https://scality.atlassian.net/browse/BB-159)

**Original Issue:**
Version delete events not producing notifications as we only process master object events.
Simply processing both master and version events does fix the issue but introduces another, where now we have two notifications for the same event in certain cases, like when we delete the latest version of an object.

After the work done [in this PR](https://github.com/scality/Arsenal/pull/1944), we now update the objects' metadata before deleting them, adding a new event to the oplog containing the latest metadata of the object. The update adds a deletion flag to the object making it simple to distinguish objects that are in the process of deletion.

We now totally ignore actual deletion events and only process the update event that was added, as we do need the object metadata for the notifications.

The following are the cases that are processed:
- In non versioned buckets, we process the master object events.
- In versioned buckets, we only process version events and ignore all master events, this makes us avoid dealing with PHD cases.
- In versioning suspended buckets, we process both version and master events. As the master object is considered an independent version.

**Other Changes:**
- Removed version id from the object key in the notifications
- Set version id field to null in the notification when the object is a null version


extensions/notification/NotificationQueuePopulator.js
```diff
@@ -38,6 +39,45 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+     * @param {*} versioningConfig bucket versioning configuration
+     * @returns {Boolean} should the entry be processed
+     */
+    _shouldProcessEntry(key, versioningConfig) {
+        // bucket versioning suspended condition (accept all)
+        const versionSuspendedCond = versioningConfig &&
+            versioningConfig.Status === 'Suspended';
+        // bucket versioning active condition (accept versions)
+        const versionActiveCond = versioningConfig &&
+            versioningConfig.Status === 'Enabled' && !isMasterKey(key);
```
jonathan-gramain: Beware, as versioned buckets can contain objects with only a master key and no version key, for objects written before versioning was enabled. For notification purpose it may not matter much but we may lose notifications around the time a bucket's versioning is enabled due to this. Can you check this to make sure this is wanted?

Maybe we would need to process all keys and ignore versions when we see a master instead.

francoisferrand: is there an easy way to identify the "last" version of an object?
just ignoring changes to this last version (and processing master instead) would deal with all situations I guess...

welansari: Not that i know of, i checked the metadata we store for versions and none contain that information

jonathan-gramain: the "last" version is the first version key listed for a particular object key. The latest version ID is also stored in the master key if it helps.

tmacro (approved):

lib/queuePopulator/LogReader.js
```diff
@@ -379,20 +379,8 @@ class LogReader {
...
-                entry.value = JSON.stringify({
-                    'last-modified': record.timestamp
-                });
-            // skipping entry for other extensions
-            } else if (entry.value === undefined) {
+            // skipping delete entry for all extensions
+            if (entry.value === undefined) {
                 return null;
             }
             return ext.filter({
```
francoisferrand: if we detect this 'update' is actually a 'delete', should we not update the record, so that downstream extensions receive a 'delete' as before?

jonathan-gramain (approved):

naren-rajendran (approved):

PR #2166, opened 2022-05-04
BB-169: add transition role policies

alexanderchan-scality:


naren-rajendran (approved):

extensions/lifecycle/objectProcessor/policy_transition.json
```diff
@@ -0,0 +1,15 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Sid": "LifecycleExpirationTransitionProcessor",
```
rachedbenmustapha: The name seems off

rachedbenmustapha (changes requested):
I'm unclear on why the GC and transition workers need `s3:ReplicateObject`

extensions/lifecycle/objectProcessor/policy_transition.json
alexanderchan-scality: yes, it is. changed to `LifecycleObjectTransitionProcessor`

alexanderchan-scality:
@RachedBenMustapha 
the `routeBackbeat` handler uses a single `apiMethod`, `objectReplicate`,   for all backbeat requests (https://github.com/scality/cloudserver/blob/2a80c583a6eee8daf55c567a3d371f5afa5a034c/lib/routes/routeBackbeat.js#L1104


ideally, the permissions should be made more granular, but in the interest of not breaking replication this was left unchanged until replication has been updated to use internal service accounts.

I have also found that the authorization results are not verified in the routeBackbeat handler,

rachedbenmustapha (approved):
@alexanderchan-scality let's please create a tech debt ticket in the cold storage epic to address this, it's not viable to use `s3:ReplicateObject` in "new" deployments for tasks that don't have anything to do with replication (quite the opposite, GC is about deleting source data).

PR #2167, opened 2022-05-05
Bugfix: BB-161 add timestamp to object deletion oplog entry in notification

welansari:
Issue: [BB-161](https://scality.atlassian.net/browse/BB-161)

Issue:
Oplog delete events don't have a value field, which makes the object deletion notifications have a `null` timestamp, as its value is directly extracted from the object metadata field `last-modified`.

Solution:
Manually add the required field in the entry value when filtering the oplog entries
The` last-modified` value used is the oplog record timestamp

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -378,15 +378,21 @@ class LogReader {
     }
 
     async _processLogEntry(batchState, record, entry) {
-        // some extension's filter functions are async (notification)
         return Promise.all(this._extensions.map(ext => {
             /**
-             * object delete entries don't have a value field,
-             * those should not be skipped for the notification
-             * extension
+             * Delete entries don't have a value field.
+             * Value field normally contains the timestamp
+             * of the event, which is needed for the notification
+             * extension.
+             * That gets added here, we use the oplog event timestamp
              */
-            if (entry.value === undefined &&
-                ext.constructor.name !== 'NotificationQueuePopulator') {
+            if (entry.type === 'delete' &&
+                ext.constructor.name === 'NotificationQueuePopulator') {
+                entry.value = JSON.stringify({
+                    'last-modified': record.timestamp
+                });
+            // skipping entry for other extensions
+            } else if (entry.value === undefined) {
                 return null;
             }
             return ext.filter({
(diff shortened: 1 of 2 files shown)


lib/queuePopulator/LogReader.js
```diff
@@ -378,17 +378,31 @@ class LogReader {
...
-            logReader: this,
-        }));
+        this._extensions.forEach(ext => {
+            /**
+             * object delete entries don't have a value field,
+             * those should not be skipped for the notification
+             * extension
+             */
+            if (entry.value === undefined) {
+                if (ext.constructor.name !== 'NotificationQueuePopulator') {
```
francoisferrand: what is this case?
a comment would be helpful to understand how/when this happens...

lib/queuePopulator/LogReader.js
```diff
@@ -378,17 +378,31 @@ class LogReader {
...
-            type: entry.type,
-            bucket: record.db,
-            // removing the v1 metadata prefix if it exists
-            key: transformKey(entry.key),
-            value: entry.value,
-            logReader: this,
-        }));
+        this._extensions.forEach(ext => {
+            /**
+             * object delete entries don't have a value field,
```
francoisferrand: maybe this comment should be the 'else' block?
i.e. are undefined values only for "delete" events? is there a more explicit way to detect these?

e.g. something like

```
if (entry.type == "delete") {....}                        // Add the missing processing for this case
else if entry.value === undefined) { return; } // Old code, was already here...
```

jbertran (approved):

lib/queuePopulator/LogReader.js
```diff
@@ -378,15 +378,21 @@ class LogReader {
...
+             * That gets added here, we use the oplog event timestamp
              */
-            if (entry.value === undefined &&
-                ext.constructor.name !== 'NotificationQueuePopulator') {
+            if (entry.type === 'delete' &&
+                ext.constructor.name === 'NotificationQueuePopulator') {
+                entry.value = JSON.stringify({
+                    'last-modified': record.timestamp
+                });
+            // skipping entry for other extensions
```
francoisferrand, line 394: nit : not indented correctly (inside the brace)

lib/queuePopulator/LogReader.js
```diff
@@ -378,15 +378,21 @@ class LogReader {
...
+            } else if (entry.value === undefined) {
```
francoisferrand, line 395: don't we need `ext.constructor.name !== 'NotificationQueuePopulator'` anymore, like before?

welansari: no that case is already handled in the if statement above

francoisferrand (approved):

PR #2175, opened 2022-05-12
Feature: BB-167 inject oplog for specific buckets

welansari:
Issue: [BB-167](https://scality.atlassian.net/browse/BB-167)

**Changes:**

NotificationConfigManager:
- Mongo changeStream delete events don't have the `fullDocument` field, which throws an error when trying to get the object id from it. Switched to the `documentKey` field instead as it is always provided

OplogPopulator:
- Now configures Kafka-connect mongo connector to only listen to specific buckets (collections) that have at least one backbeat extension active.
- A Mongo change stream is established to listen to `__metastore` changes and reconfigure the Kafka-connect connector accordingly.

ChangeStreamWrapper:
- Created simple class to handle change streams

KafkaConnectWrapper:
- Changed location of file

### DIFF
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -1,10 +1,15 @@
 const joi = require('joi');
+const { errors } = require('arsenal');
+const { MongoClient } = require('mongodb');
 const constants = require('./constants');
-const KafkaConnectWrapper = require('./KafkaConnectWrapper');
+const { constructConnectionString } = require('../utils/MongoUtils');
+const KafkaConnectWrapper = require('../../lib/wrappers/KafkaConnectWrapper');
+const ChangeStream = require('../../lib/wrappers/ChangeStream');
 
 const paramsJoi = joi.object({
     config: joi.object().required(),
     mongoConfig: joi.object().required(),
+    activeExtensions: joi.array().required(),
     logger: joi.object().required(),
 }).required();
 
@@ -27,49 +32,75 @@ class OplogPopulator {
      * @param {Object} params.mongoConfig.replicaSet - mongo replica set
      * @param {Object} params.mongoConfig.readPreference - mongo read preference
      * @param {Object} params.mongoConfig.database - metadata database
+     * @param {string[]} params.activeExtensions - list of all active extension names
      * @param {Object} params.logger - logger
      */
     constructor(params) {
         joi.attempt(params, paramsJoi);
         this._config = params.config;
         this._mongoConfig = params.mongoConfig;
+        this._activeExtensions = params.activeExtensions;
         this._logger = params.logger;
         this._connectWrapper = new KafkaConnectWrapper({
             kafkaConnectHost: this._config.kafkaConnectHost,
             kafkaConnectPort: this._config.kafkaConnectPort,
             logger: this._logger,
         });
+        this._changeStreamWrapper = null;
+        // stores the set of buckets assigned to connector
+        this._bucketsForConnector = new Set();
+        // MongoDB related
+        this._mongoClient = null;
+        this._metastore = null;
+        // setup mongo connection data
+        this._mongoUrl = constructConnectionString(this._mongoConfig);
+        this._replicaSet = this._mongoConfig.replicaSet;
+        this._database = this._mongoConfig.database;
     }
 
     /**
-     * Sets default connector configuration
-     * @param {Object} connectorConfig custom connector configuration
-     * @param {string} connectorConfig.name connector name
+     * Connects to MongoDB using the MongoClientInterface
+     * and retreives the metastore collection
+     * @returns {Promise|undefined} undefined
+     * @throws {InternalError}
+     */
+    async _setupMongoClient() {
+        try {
+            const client = await MongoClient.connect(this._mongoUrl, {
+                 replicaSet: this._replicaSet,
+                 useNewUrlParser: true,
+            });
+            // connect to metadata DB
+            this._mongoClient = client.db(this._database, {
+                ignoreUndefined: true,
+            });
+            // get metastore collection
+            this._metastore = this._mongoClient.collection(constants.bucketMetastore);
+            this._logger.debug('Connected to MongoDB', {
+                method: 'OplogPopulator._setupMongoClient',
+            });
+            return undefined;
+        } catch (err) {
+            this._logger.error('Could not connect to MongoDB', {
+                method: 'OplogPopulator._setupMongoClient',
+                error: err.message,
+            });
+            throw errors.InternalError.customizeDescription(err.message);
+        }
+    }
+
+    /**
+     * get default connector configuration
+     * @param {string} connectorName connector name
      * @returns {Object} connector configuration
      */
-    getDefaultConnectorConfiguration(connectorConfig) {
-        const { authCredentials, replicaSetHosts, writeConcern, replicaSet,
-            readPreference, database } = this._mongoConfig;
+    _getDefaultConnectorConfiguration(connectorName) {
         // getting default connector config values
         const config = constants.defaultConnectorConfig;
         // connection and source configuration
-        config.name = connectorConfig.name;
-        config.database = database;
-        // get credentials
-        let cred = '';
-        if (authCredentials &&
-            authCredentials.username &&
-            authCredentials.password) {
-            const username = encodeURIComponent(authCredentials.username);
-            const password = encodeURIComponent(authCredentials.password);
-            cred = `${username}:${password}@`;
-        }
-        let connectionUrl = `mongodb://${cred}${replicaSetHosts}/` +
-            `?w=${writeConcern}&readPreference=${readPreference}`;
-        if (replicaSet) {
-            connectionUrl += `&replicaSet=${replicaSet}`;
-        }
-        config['connection.uri'] = connectionUrl;
+        config.name = connectorName;
+        config.database = this._database;
+        config['connection.uri'] = this._mongoUrl;
         // destination topic configuration
         config['topic.namespace.map'] = JSON.stringify({
             '*': this._config.topic,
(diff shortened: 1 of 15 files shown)


jbertran (approved):

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -64,10 +63,61 @@ class OplogPopulator {
...
+     * and retreives the metastore collection
+     * @returns {Promise|undefined} undefined
+     * @throws {InternalError}
+     */
+    async _setupMongoClient() {
+        try {
+            const client = await this._MongoClientConnect(this._mongoUrl, {
+                 replicaSet: this._replicaSet,
+                 useNewUrlParser: true,
+            });
```
francoisferrand, line 72: can't we should pass `params.mongoConfig` ?
the param in backbeat config is the same format used by MongoClient, so this is usually enough I think

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -64,10 +63,61 @@ class OplogPopulator {
...
+        }
+    }
+
+    /**
+     * Sets default connector configuration
+     * @param {Object} connectorConfig custom connector configuration
+     * @param {string} connectorConfig.name connector name
+     * @returns {Object} connector configuration
+     */
+    _getDefaultConnectorConfiguration(connectorConfig) {
```
francoisferrand: is it a getter (as the name of the function) or a setter (as in the comment) ?
seems awfully similar to some mongodb connection code, also; don't we have it somewhere already, where we can use it?

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -77,30 +127,262 @@ class OplogPopulator {
...
+     * @returns {string[]} list of buckets to listen to
+     * @throws {InternalError}
+     */
+    async _getBackbeatEnabledBuckets() {
+        const filter = {
+            $or: [],
+        };
+        // setup filters
+        this._activeExtensions.forEach(extName => {
+            const field = constants.extensionNameToBucketMetadataConfigField[extName];
```
francoisferrand: `extensionNameToBucketMetadataConfigField` : may want to try and find a better name...

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -77,30 +127,262 @@ class OplogPopulator {
...
+                return;
+            }
+            areExtensionsEnabled.push(bucketMetadata[field] !== null);
+        });
+        return areExtensionsEnabled.some(ext => ext);
+    }
+
+    /**
+     * Handler for the change stream "change" event.
+     * Updated connector pipeline when change occurs
```
francoisferrand:
```suggestion
     * Updates connector pipeline when change occurs
```

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -77,30 +127,262 @@ class OplogPopulator {
...
+     * @param {ChangeStreamDocument} change Change stream change object
+     * @returns {undefined}
+     */
+    _handleChangeStreamChangeEvent(change) {
+        // caching change stream resume token
+        this._changeStreamResumeToken = change._id;
```
francoisferrand: I think there was some (generic) code from Vianney to implement this "scan & watch" pattern : can't we use it?

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -77,30 +127,262 @@ class OplogPopulator {
...
+        // updating list of buckets to listen to
+        const idx = this._bucketsToListenTo.indexOf(change.documentKey._id);
+        // no fullDocument field in delete events
+        const isBackbeatEnabled = change.fullDocument ? this._isBucketBackbeatEnabled(change.fullDocument.value)
```
francoisferrand: instead of receiving all events, we may be able to add the same filter we use to "read" the current list (e.g. `_getBackbeatEnabledBuckets`) direclty in the change stream ?

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -77,30 +127,262 @@ class OplogPopulator {
...
+     * Handler for the change stream "change" event.
+     * Updated connector pipeline when change occurs
+     * @param {ChangeStreamDocument} change Change stream change object
+     * @returns {undefined}
+     */
+    _handleChangeStreamChangeEvent(change) {
+        // caching change stream resume token
+        this._changeStreamResumeToken = change._id;
+        // updating list of buckets to listen to
+        const idx = this._bucketsToListenTo.indexOf(change.documentKey._id);
```
francoisferrand: searching in an array should not be done : unless we can be sure the array size will always be small (not the case here)

may be acceptable for an initial implementation (we know things will change) ; but if there is a chance this code stays (e.g. we may still need this kind of lookup, even in next step(s) of implementation) : then we need to come up with a more efficient approach.

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -77,30 +127,262 @@ class OplogPopulator {
...
+                method: 'OplogPopulator._setMetastoreChangeStream',
+            });
+        } catch (err) {
+            this._logger.error('Error while setting the change stream', {
+                method: 'OplogPopulator._setMetastoreChangeStream',
+                error: err.message,
+            });
+            throw errors.InternalError.customizeDescription(err.message);
+        }
+    }
```
francoisferrand, line 331: looking at the code here, I think we may benefit from extracting a generic "change stream" wrapper class : e.g. to handle errors, resume token, ... (which is done here, and in a few other places)

(unless we can use/extend existing implementation of the scan&watch pattern)

extensions/oplogPopulator/OplogPopulatorTask.js
```diff
@@ -1,34 +1,53 @@
...
     config: oplogPopulatorConfig,
     mongoConfig,
+    activeExtensions,
     logger,
 });
 
 (async () => {
     try {
         await oplogPopulator.setup();
+        schedule.scheduleJob(qpConfig.cronRule, async () => {
```
francoisferrand: why do we need to do this periodically?
update is required only :
- on startup
- when the list of buckets to watch has changed

extensions/oplogPopulator/OplogPopulatorTask.js
```diff
@@ -1,34 +1,53 @@
...
+            try {
+                // Update each connector's config
+                await Promise.all(oplogPopulatorConfig.connectors.map(connConf =>
```
francoisferrand: is this tacking more than BB-167 ? i think we should have a single connector for now...
dispatching onto multiple connectors is next ticket (e.g. BB-168) ...

extensions/oplogPopulator/constants.js
```diff
@@ -5,7 +5,13 @@ const constants = {
         'change.stream.full.document': 'updateLookup',
         'pipeline': '[]',
         'collection': '',
-    }
+    },
+    extensionNameToBucketMetadataConfigField: {
+        notification: 'notificationConfiguration',
```
francoisferrand, line 11: is it enough to check the presence of these field in the bucket metadata?
e.g. can we have the field with `enabled=false` inside?

francoisferrand (commented):
some cleanup to do, and I am really concerned about the "multiple connectors" code

extensions/oplogPopulator/OplogPopulator.js
welansari: This will cause an issue if a bucket becomes invalid (no extension active), as the change stream event that invalidates it will be ignored and so the connector pipeline won't be updated

extensions/oplogPopulator/OplogPopulator.js
welansari: the code is not yet merged, so it cannot be used
Will use a simple wrapper class for now, for better code readability

extensions/oplogPopulator/OplogPopulator.js
welansari: We are not using arsenal's MongoClient here so params need to be adjusted
we are using the MongoClient provided by the mongodb library, this is done because we won't be using any of the arsenal implementation functions (not needed)

extensions/oplogPopulator/OplogPopulator.js
welansari: ~~The mongodb connection code you are referring to i'm assuming is in arsenal's MongoClientInterface ? this part of the code is directly in the constructor so it can't be used~~
Fixed

extensions/oplogPopulator/OplogPopulatorTask.js
```diff
@@ -23,12 +25,13 @@ const oplogPopulator = new OplogPopulator({
...
         logger.error('Error when starting up the oplog populator', {
             method: 'OplogPopulatorTask.setup',
-            error,
+            error: error.description,
         });
         process.emit('SIGTERM');
     }
 })();
 
+
```
ghost: necessity?

ghost (approved):
for what it's worth, seems ok to me

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -27,80 +34,390 @@ class OplogPopulator {
...
+        // MongoDB related
+        this._mongoClient = null;
+        this._metastore = null;
+        // setup mongo connection data
+        this._mongoUrl = constructConnectionString(this._mongoConfig);
+        this._replicaSet = this._mongoConfig.replicaSet;
+        this._database = this._mongoConfig.database;
+
+        // Promisifying callback functions
+        this._MongoClientConnect = util.promisify(MongoClient.connect);
```
jonathan-gramain: is there a reason to capitalize `Mongo` instead of `_mongoClientConnect`

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -27,80 +34,390 @@ class OplogPopulator {
...
+                    enabled: true,
+                },
+            };
+            filter.$or.push(extFilter);
+        }
+        // lifecycle filter
+        if (this._activeExtensions.includes('lifecycle')) {
+            const field = constants.extensionConfigField.lifecycle;
+            const extFilter = {};
+            // processing buckets where at least one replication
```
jonathan-gramain: replication -> lifecycle

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -27,80 +34,390 @@ class OplogPopulator {
...
+            filter.$or.push(extFilter);
+        }
+        try {
+            const buckets = await this._metastore.find(filter)
+                .project({ _id: 1 })
+                .map(bucket => bucket._id)
+                .toArray();
+            return buckets;
+        } catch (err) {
+            this._logger.error('Could not connect to MongoDB', {
```
jonathan-gramain: Maybe a more generic message like `error querying MongoDB` could be more appropriate, in case the error is not about connectivity

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -27,80 +34,390 @@ class OplogPopulator {
...
+                method: 'OplogPopulator._setupMongoClient',
```
jonathan-gramain: `OplogPopulator._getBackbeatEnabledBuckets`

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -27,80 +34,390 @@ class OplogPopulator {
...
+        }
+        // lifecycle extension
+        if (this._activeExtensions.includes('lifecycle')) {
+            const field = constants.extensionConfigField.lifecycle;
+            const rules = (bucketMetadata[field] && bucketMetadata[field].rules)
+                || null;
+            if (!rules || rules.length === 0) {
+                areExtensionsEnabled.push(false);
+            } else {
+                const areRulesActive = bucketMetadata[field].rules.some(rule =>
```
jonathan-gramain:
```
const areRulesActive = rules.some(rule =>
```

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -27,80 +34,390 @@ class OplogPopulator {
...
+            default:
+                this._logger.debug('Skipping unsupported change stream event', {
+                    method: 'OplogPopulator._handleChangeStreamChange',
+                    type: change.operationType,
+                    key: change.documentKey._id,
+                });
+                break;
+        }
+        this._logger.debug('Change stream event processed', {
+            method: 'OplogPopulator._handleChangeStreamChange',
```
jonathan-gramain, line 300: could add

```
type: change.operationType
```

tests/unit/lib/wrappers/ChangeStreamWrapper.js
```diff
@@ -0,0 +1,200 @@
...
+                assert(closeStub.notCalled);
+                assert(removeListenerStub.getCall(0).calledWithMatch('change',
+                    changehandler));
+                assert(removeListenerStub.getCall(1).calledWithMatch('error',
+                    wrapper._handleChangeStreamErrorEvent.bind(wrapper)));
+            })
+            .catch(err => assert.ifError(err));
+        });
+
+        it('should close then reset the change steam on error', async () => {
```
jonathan-gramain: steam -> stream

jonathan-gramain (commented):
I did a partial review only.

This PR is too large, it probably explains why it's hard to get timely reviews from NA ;)

Try to see if you can split it into smaller meaningful chunks, or for the next ones if it's too hard or late for this one.

lib/wrappers/ChangeStreamWrapper.js
```diff
@@ -0,0 +1,95 @@
...
+    pipeline: joi.array().required(),
+    throwOnError: joi.boolean().required(),
+}).required();
+
+/**
+ * @class ChangeStreamWrapper
+ *
+ * @classdesc Wrapper arround MongoDB change stream API
+ */
+class ChangeStreamWrapper {
```
francoisferrand: just `MongoChangeStream` would be enough?
The "Wrapper" suffix does not add any informatin...

extensions/oplogPopulator/OplogPopulator.js
welansari: Forgot `Mongo` supported async 😅, fixed it

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -1,10 +1,15 @@
 const joi = require('joi');
+const { errors } = require('arsenal');
+const MongoClient = require('mongodb').MongoClient;
```
francoisferrand: nit: couldn't we do `const { MongoClient } = require('mongodb')` instead?

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -27,80 +32,388 @@ class OplogPopulator {
...
+            };
+            filter.$or.push(extFilter);
+        }
+        // ingestion filter
+        if (this._activeExtensions.includes('ingestion')) {
+            const field = constants.extensionConfigField.ingestion;
+            const extFilter = {};
+            extFilter[`value.${field}.status`] = 'enabled';
+            filter.$or.push(extFilter);
+        }
```
francoisferrand, line 160: ideally, I would prefer this logic (e.g. the mapping from "extension" to associated filter) to be outside of this code : possibly be part of each extension itself.

So that we don't need to modify oplogPopulator itself, and simply need to add the required "support" in the extension.

(to be clear: the part which should not be here IMO is the "business logic" of each extension, i.e. knowing that _for ingestion we need whatever field to be set to enabled_. There may however be code here to loop over "something" and generically call that to build the actual mongo filter)

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -27,80 +32,388 @@ class OplogPopulator {
...
+    _isBucketBackbeatEnabled(bucketMetadata) {
+        const areExtensionsEnabled = [];
+        // notification extension
+        if (this._activeExtensions.includes('notification')) {
+            const field = constants.extensionConfigField.notification;
+            areExtensionsEnabled.push(Boolean(bucketMetadata[field] !== null));
+        }
+        // replication extension
+        if (this._activeExtensions.includes('replication')) {
+            const field = constants.extensionConfigField.replication;
```
francoisferrand, line 229: same as above, the business logic should be moved out of this (generic) code, somehow ⬆️

Note that this may be done in a separate techdebt ticket, to keep momentum and avoid conflicts/merges (→ please create it, in that case)

extensions/oplogPopulator/OplogPopulatorTask.js
```diff
@@ -10,10 +10,12 @@ werelogs.configure({ level: config.log.logLevel,
 
 const mongoConfig = config.queuePopulator.mongo;
 const oplogPopulatorConfig = config.extensions.oplogPopulator;
+const activeExtensions = Object.keys(config.extensions);
```
francoisferrand: not correct : should only include the extensions which are actually using oplogPopulator...

e.g. nothing for now :-/ ....though we may enable it for bucket notification already anyway, since we are transitioning it and it is not yet used in 8.x. This will make testing easier, with no product impact in the meantime.

francoisferrand (changes requested):
oplog populator should not be enabled for all "active" extensions yet : only for these which require it (e.g. nothing yet, or possibly bucket-notification to make ongoing dev easier)

extensions/oplogPopulator/OplogPopulatorTask.js
francoisferrand: In addition, maybe we may need to check a bit deeper?
I am thinking of the case (which may not be possible, to check) where the config contains some details about an extension, but it is not fully enabled (e.g. something like `config = { "extensions": [ "replication": { "enabled": false } ] }`)

extensions/oplogPopulator/OplogPopulator.js
welansari: Techdebt ticket created

extensions/oplogPopulator/OplogPopulator.js
welansari: Techdebt ticket created : [BB-194](https://scality.atlassian.net/browse/BB-194)

francoisferrand (approved):

PR #2180, opened 2022-05-13
Feature: BB-164 support multiple connectors in oplog populator

welansari:
Issue: [BB-164](https://scality.atlassian.net/browse/BB-164)

Added support for the use of multiple connectors by the oplogPopulator

--------
**Connector Class**
This class manages the state of a kafka-connect mongodb source connectors.
The class contains the connector's config and can assign or unassign buckets from it

--------
**ConnectorsManager Class**
A ConnectorsManager keeps a list of active connectors

It is also used to spawn and configure connectors.

At startup, old connectors are first retreived and reconfigured*, then if more
connectors are required it creates them.

*Old connectors are reconfigured where buckets that are no longer valid
are removed from the connectors' config

--------
**Allocator Class**
The Allocator handles assigning buckets to connectors using
an allocation strategy.

--------
**Allocation Strategy Class**
Allocation strategy is what decides what connector should be assigned
to a bucket next

The allocation strategy implemented for now, returns the connector with the
fewest buckets assigned

--------
**OplogPopulator**
The OplogPopulator dynamically configures kafka-connect MongoDB source connectors to watch over changes in relevant buckets in MongoDB

At startup, the OplogPopulator first initializes connectors following its config.
If we have old connectors, these are reconfigured as explained above. More connectors are created if needed
The allocator is then instantiated, and used to start listening to valid buckets that don't have an assigned connector yet
A change stream is then established with the `__metastore` db to watch for changes of bucket configs, this is used to decide if we should listen to object operations in a bucket or not.
We start listening to a bucket if it uses at least one of the active backbeat extensions (ex. replication, notification...)

### DIFF
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -3,8 +3,9 @@ const { errors } = require('arsenal');
 const { MongoClient } = require('mongodb');
 const constants = require('./constants');
 const { constructConnectionString } = require('../utils/MongoUtils');
-const KafkaConnectWrapper = require('../../lib/wrappers/KafkaConnectWrapper');
 const ChangeStream = require('../../lib/wrappers/ChangeStream');
+const Allocator = require('./modules/Allocator');
+const ConnectorsManager = require('./modules/ConnectorsManager');
 
 const paramsJoi = joi.object({
     config: joi.object().required(),
@@ -41,14 +42,9 @@ class OplogPopulator {
         this._mongoConfig = params.mongoConfig;
         this._activeExtensions = params.activeExtensions;
         this._logger = params.logger;
-        this._connectWrapper = new KafkaConnectWrapper({
-            kafkaConnectHost: this._config.kafkaConnectHost,
-            kafkaConnectPort: this._config.kafkaConnectPort,
-            logger: this._logger,
-        });
         this._changeStreamWrapper = null;
-        // stores the set of buckets assigned to connector
-        this._bucketsForConnector = new Set();
+        this._allocator = null;
+        this._connectorsManager  = null;
         // MongoDB related
         this._mongoClient = null;
         this._metastore = null;
@@ -87,25 +83,6 @@ class OplogPopulator {
             });
             throw errors.InternalError.customizeDescription(err.message);
         }
-    }
-
-    /**
-     * get default connector configuration
-     * @param {string} connectorName connector name
-     * @returns {Object} connector configuration
-     */
-    _getDefaultConnectorConfiguration(connectorName) {
-        // getting default connector config values
-        const config = constants.defaultConnectorConfig;
-        // connection and source configuration
-        config.name = connectorName;
-        config.database = this._database;
-        config['connection.uri'] = this._mongoUrl;
-        // destination topic configuration
-        config['topic.namespace.map'] = JSON.stringify({
-            '*': this._config.topic,
-        });
-        return config;
     }
 
     /**
@@ -158,6 +135,10 @@ class OplogPopulator {
             extFilter[`value.${field}.status`] = 'enabled';
             filter.$or.push(extFilter);
         }
+        // return empty list if no extension active
+        if (filter.$or.length === 0) {
+            return [];
+        }
         try {
             const buckets = await this._metastore.find(filter)
                 .project({ _id: 1 })
@@ -174,46 +155,7 @@ class OplogPopulator {
     }
 
     /**
-     * Makes new connector pipeline to listen to
-     * specified buckets
-     * @param {string[]} bucketNames list of buckets to listen to
-     * @returns {Object} new connector pipeline
-     */
-    _generateNewConnectorPipeline(bucketNames) {
-        const pipeline = [
-            {
-                $match: {
-                    'ns.coll': {
-                        $in: bucketNames,
-                    }
-                }
-            }
-        ];
-        return JSON.stringify(pipeline);
-    }
-
-    /**
-     * start listening to bucket
-     * @param {string} bucketName bucket name
-     * @returns {undefined}
-     */
-    _listenToBucket(bucketName) {
-        this._bucketsForConnector.add(bucketName);
-        this.updateConnectorPipeline(constants.defaultConnectorName, [...this._bucketsForConnector]);
-    }
-
-    /**
-     * stops listening to bucket
-     * @param {string} bucketName bucket name
-     * @returns {undefined}
-     */
-    _stopListeningToBucket(bucketName) {
-        this._bucketsForConnector.delete(bucketName);
-        this.updateConnectorPipeline(constants.defaultConnectorName, [...this._bucketsForConnector]);
-    }
-
-    /**
-     * Check if buckets has at least one backbeat extension active
+     * Checks if buckets has at least one backbeat extension active
      * @param {BucketInfo} bucketMetadata bucket metadata
      * @returns {boolean} is bucket backbeat enabled
      */
@@ -266,15 +208,15 @@ class OplogPopulator {
      * @param {ChangeStreamDocument} change Change stream change object
      * @returns {undefined}
      */
-    _handleChangeStreamChangeEvent(change) {
-        const isListeningToBucket = this._bucketsForConnector.has(change.documentKey._id);
+    async _handleChangeStreamChangeEvent(change) {
+        const isListeningToBucket = this._allocator.has(change.documentKey._id);
         // no fullDocument field in delete events
         const isBackbeatEnabled = change.fullDocument ?
             this._isBucketBackbeatEnabled(change.fullDocument.value) : null;
         switch (change.operationType) {
             case 'delete':
                 if (isListeningToBucket) {
-                    this._stopListeningToBucket(change.documentKey._id);
+                    await this._allocator.stopListeningToBucket(change.documentKey._id);
                 }
                 break;
             case 'replace':
@@ -282,10 +224,10 @@ class OplogPopulator {
             case 'insert':
                 // remove bucket if no longer backbeat enabled
                 if (isListeningToBucket && !isBackbeatEnabled) {
-                    this._stopListeningToBucket(change.documentKey._id);
-                // add bucket if it is backbeat enabled
+                    await this._allocator.stopListeningToBucket(change.documentKey._id);
+                // add bucket if it became backbeat enabled
                 } else if (!isListeningToBucket && isBackbeatEnabled) {
-                    this._listenToBucket(change.documentKey._id);
+                    await this._allocator.listenToBucket(change.documentKey._id);
                 }
                 break;
             default:
@@ -331,85 +273,44 @@ class OplogPopulator {
     }
 
     /**
-     * Updates connector pipeline
-     * @param {string} connectorName connector to update
-     * @param {array} buckets connector assigned buckets
-     * @returns {Promise|Object} connector config
-     * @throws {InternalError}
-     */
-    async updateConnectorPipeline(connectorName, buckets) {
-        // TODO: Add support for multiple connectors
-        const pipeline = this._generateNewConnectorPipeline(buckets);
-        try {
-            const config = await this._connectWrapper.updateConnectorPipeline(connectorName, pipeline);
-            return config;
-        } catch (err) {
-            this._logger.error('Error while updating connector', {
-                method: 'OplogPopulator.updateConnectorPipeline',
-                connector: connectorName,
-                buckets,
-                error: err.description,
-            });
-            throw errors.InternalError.customizeDescription(err.description);
-        }
-    }
-
-    /**
-     * Creates and configures kafka connect
-     * connectors based on the config
-     * @param {Object} connectorConfig Config of connector
-     * @returns {Promise|Object} Updated connector configuration
-     * @throws {InternalError}
-     */
-    async _configureConnector(connectorConfig) {
-        try {
-            const activeConnectors = await this._connectWrapper.getConnectors();
-            const connectorName = connectorConfig.name;
-            // update the connector config if it already exist
-            if (activeConnectors.includes(connectorName)) {
-                return this._connectWrapper.updateConnectorConfig(connectorName,
-                    connectorConfig.config);
-            } else {
-                // otherwise create a new connector
-                return this._connectWrapper.createConnector({
-                    name: connectorName,
-                    config: connectorConfig.config,
-                });
-            }
-        } catch (err) {
-            this._logger.error('Error while initially configuring connectors', {
-                method: 'OplogPopulator._configureConnectors',
-                error: err.description,
-                config: connectorConfig,
-            });
-            throw errors.InternalError.customizeDescription(err.description);
-        }
-    }
-
-    /**
-     * Sets up the OplogPopulator
+     * Sets the OplogPopulator
      * @returns {Promise|undefined} undefined
      * @throws {InternalError}
      */
     async setup() {
        try {
-           await this._setupMongoClient();
-           // getting buckets with at least one extension active
-           const backbeatEnabledBuckets = await this._getBackbeatEnabledBuckets();
-           this._bucketsForConnector = new Set(backbeatEnabledBuckets);
-           // TODO: add support for multiple connectors
-            const defaultConnectorConfig = this._getDefaultConnectorConfiguration(constants.defaultConnectorName);
-            defaultConnectorConfig.pipeline = this._generateNewConnectorPipeline(backbeatEnabledBuckets);
-            const connectorConfigs = {
-                name: constants.defaultConnectorName,
-                config: defaultConnectorConfig
-            };
-            await this._configureConnector(connectorConfigs);
-           this._setMetastoreChangeStream();
+           this._connectorsManager = new ConnectorsManager({
+               nbConnectors: this._config.numberOfConnectors,
+               database: this._database,
+               mongoUrl: this._mongoUrl,
+               oplogTopic: this._config.topic,
+               prefix: this._config.prefix,
+               kafkaConnectHost: this._config.kafkaConnectHost,
+               kafkaConnectPort: this._config.kafkaConnectPort,
+               logger: this._logger,
+            });
+            await this._connectorsManager.initializeConnectors();
+            this._allocator = new Allocator({
+                connectorsManager: this._connectorsManager,
+                logger: this._logger,
+            });
+            // initialize mongo client
+            await this._setupMongoClient();
+            // get currently valid buckets from mongo
+            const validBuckets = await this._getBackbeatEnabledBuckets();
+            // listen to valid buckets
+            await Promise.all(validBuckets.map(bucket => this._allocator.listenToBucket(bucket)));
+            // establish change stream
+            this._setMetastoreChangeStream();
+            // remove no longer valid buckets from old connectors
+            await this._connectorsManager.removeInvalidBuckets(validBuckets);
+            this._logger.debug('OplogPopulator setup complete', {
+                method: 'OplogPopulator.setup',
+            });
        } catch (err) {
             this._logger.error('An error occured when setting up the OplogPopulator', {
                 method: 'OplogPopulator.setup',
-                error: err.description,
+                error: err.description || err.message,
             });
             throw errors.InternalError.customizeDescription(err.description);
        }
--- a/extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
+++ b/extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
@@ -1,0 +1,32 @@
+const AllocationStrategy = require('./AllocationStrategy');
+
+/**
+ * @class LeastFullConnector
+ *
+ * @classdesc LeastFullConnector is an allocation
+ * strategy that assigns buckets to connectors based
+ * on the number of buckets assigned to each connector.
+ * Connectors with the fewest buckets are filled first
+ */
+class LeastFullConnector extends AllocationStrategy {
+
+    /**
+     * @constructor
+     * @param {Object} params params
+     * @param {Logger} params.logger logger object
+     */
+    constructor(params) {
+        super(params.logger);
+    }
+
+    /**
+     * Get best connector for assigning a bucket
+     * @param {Connector[]} connectors available connectors
+     * @returns {Connector} connector
+     */
+    getConnector(connectors) {
+        return connectors.reduce((prev, elt) => (elt.bucketCount < prev.bucketCount ? elt : prev));
+    }
+}
+
+module.exports = LeastFullConnector;
--- a/extensions/oplogPopulator/modules/Allocator.js
+++ b/extensions/oplogPopulator/modules/Allocator.js
@@ -1,0 +1,124 @@
+const joi = require('joi');
+const async = require('async');
+const { errors } = require('arsenal');
+const LeastFullConnector = require('../allocationStrategy/LeastFullConnector');
+
+const paramsJoi = joi.object({
+    connectorsManager: joi.object().required(),
+    logger: joi.object().required(),
+}).required();
+
+// Number of times the allocator should try to update
+// a connector
+const MAX_RETRIES = process.env.OPLOG_POPULATOR_ALLOCATOR_RETRY || 3;
+
+/**
+ * @class Allocator
+ *
+ * @classdesc Allocator handles listening to buckets by assigning
+ * a connector to them.
+ */
+class Allocator {
+
+    /**
+     * @constructor
+     * @param {Object} params Allocator param
+     * @param {ConnectorsManager} params.connectorsManager connectorsManager
+     * @param {Logger} params.logger logger object
+     */
+    constructor(params) {
+        joi.attempt(params, paramsJoi);
+        this._connectorsManager = params.connectorsManager;
+        this._logger = params.logger;
+        this._allocationStrategy = new LeastFullConnector({
+            logger: params.logger,
+        });
+        // Stores connector assigned for each bucket
+        this._bucketsToConnectors = new Map();
+        this._initConnectorToBucketMap();
+    }
+
+    /**
+     * Initializes connectorsManager map
+     * @returns {undefined}
+     */
+    _initConnectorToBucketMap() {
+        const connectors = this._connectorsManager.connectors;
+        connectors.forEach(connector => {
+            connector.buckets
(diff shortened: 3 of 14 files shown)


extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -49,8 +49,12 @@ class OplogPopulator {
...
         });
         this._changeStreamWrapper = null;
-        // stores the set of buckets assigned to connector
-        this._bucketsForConnector = new Set();
+        // contains created connector names
+        this._configuredConnectors = [];
+        // stores connector assigned for each bucket
+        this._connectorToBucket = new Map();
+        // stores sets of buckets assigned to each connector
+        this._bucketsForConnectors = {};
```
francoisferrand: should be a field inside each `Connector` (i.e. each entry of the `_configuredConnectors` list)

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -198,27 +202,57 @@ class OplogPopulator {
...
+     * @returns {string} assigned connector for bucket
+     */
+     _assignConnectorForBucket(bucketName) {
+        const assignedConnector = this._connectorToBucket.get(bucketName);
+        // return previously assigned bucket if it exists
+        if (assignedConnector) {
+            return assignedConnector;
+        } else {
+            // else return the connector with the fewest buckets assinged to it
+            const bucketPerConnector = this._configuredConnectors.map(connector =>
```
francoisferrand: this algorithm will not scale (`O(N) * 3`), e.g. if we need to manage *many* connectors : we may need keep this list ordered somehow.

If the number of connectors stays low, this may be acceptable ; but at the moment we don't know how we will (need to) scale : e.g. if it will be better/needed to increase the number of connectors (and keep a low bucket count per connector) OR to increase the number of buckets per connector.

→ either we can make an easy algo upgrade (and be somewhat future-proof), or we need to create a TechDebt ticket (and address in BucketNotif or next EPIC)

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -198,27 +202,57 @@ class OplogPopulator {
...
      * @param {string} bucketName bucket name
+     * @param {boolean} updateConnector should connector get updated
      * @returns {undefined}
      */
-    _listenToBucket(bucketName) {
-        this._bucketsForConnector.add(bucketName);
-        this.updateConnectorPipeline(constants.defaultConnectorName, [...this._bucketsForConnector]);
+    _listenToBucket(bucketName, updateConnector = true) {
+        const assignedConnector = this._assignConnectorForBucket(bucketName);
+        this._connectorToBucket.set(bucketName, assignedConnector);
```
francoisferrand: the scope/role of the assignConnectorForBucket, and the followup "assignation", is not very consistent : either it should do indeed the assignation (and update the data), only simply handle the "retrieval" (and we must assign here if not already done)

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -198,27 +202,57 @@ class OplogPopulator {
...
+            const min = Math.min(...bucketPerConnector);
+            const idx = bucketPerConnector.indexOf(min);
+            return this._configuredConnectors[idx];
+        }
+    }
+
+    /**
+     * start listening to a specific bucket's events
      * @param {string} bucketName bucket name
+     * @param {boolean} updateConnector should connector get updated
```
francoisferrand: not clear this does (e.g. looking just at this function) : maybe this should not be part of this function, and/or the function prototype should be updated to better convey the intent of the function

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -391,25 +437,146 @@ class OplogPopulator {
...
+     */
+    async _spawnConnectors() {
+        try {
+            await Promise.all(this._configuredConnectors.map(async connectorName => {
+                // setting up config for connector
+                const config = this._getDefaultConnectorConfiguration(connectorName);
+                config.pipeline = this._generateNewConnectorPipeline(
+                    [...this._bucketsForConnectors[connectorName]]
+                );
+                // creating or updating connector
```
francoisferrand: nit: "create or update connector"

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -391,25 +437,146 @@ class OplogPopulator {
...
+    /**
+     * Restores the previous state of the connectors by
+     * extracting previously assigned buckets from their configs
+     * @param {string[]} backbeatEnabledBuckets backbeat enabled bucket names
+     * @returns {Promise|undefined} undefined
+     * @throws {InternalError}
+     */
+    async _restoreConnectorBuckets(backbeatEnabledBuckets) {
+        try {
+            const activeConnectors = await this._connectWrapper.getConnectors();
```
francoisferrand: `_connectWrapper` is not very representative: may be better to rename this to `_kafkaConnect`, so that code is easier to read/follow/understand

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -391,25 +437,146 @@ class OplogPopulator {
...
+                .filter(connectorName => activeConnectors.includes(connectorName));
+            await Promise.all(connectorNames.map(async connectorName => {
+                // getting previous connector config
+                const connectorConfig = await this._connectWrapper
+                    .getConnectorConfig(connectorName);
+                // extracting buckets from config pipeline
+                const bucketList = this._extractBucketsFromConfig(connectorConfig)
+                    .filter(bucket => backbeatEnabledBuckets.includes(bucket));
+                // storing the previous state
+                this._bucketsForConnectors[connectorName] = new Set([...bucketList]);
```
francoisferrand: should split the code, I can see a few concerns (classes) we can isolate:
* the `Connector` itself (which has a number of buckets)
* the `allocation strategy` : not much for now, but mostly an interface with a contract, e.g. return the connector to use (or possibly create, in a future strategy : already now this could be used to delay the creation of connector until we need one) → will be easier to test if isolated
* the `Allocator` : e.g. handles the glue between the data (connectors) and (abstarct) allocation strategy → not sure if it should include the actual kafka-connect management, or move this elsewhere?

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -391,25 +437,146 @@ class OplogPopulator {
...
+            const activeConnectors = await this._connectWrapper.getConnectors();
+            const connectorNames = this._configuredConnectors
+                .filter(connectorName => activeConnectors.includes(connectorName));
+            await Promise.all(connectorNames.map(async connectorName => {
+                // getting previous connector config
+                const connectorConfig = await this._connectWrapper
+                    .getConnectorConfig(connectorName);
+                // extracting buckets from config pipeline
+                const bucketList = this._extractBucketsFromConfig(connectorConfig)
+                    .filter(bucket => backbeatEnabledBuckets.includes(bucket));
```
francoisferrand: this filtering is somewhat weird:
* need to actually "remove" the buckets from the connector, if this is not needed anymore
* i am concerned there may be a race condition : e.g. if a bucket is reconfigured after the initial list is read
* the scope of the function is somewhat confusing, seems to rely very strongly on the way it is called inside setup() instead of having a clear purpose (i.e. imagining it could be reused / checking if the function provides the "right" level of abstraction / more pragmatically: 1. can we review this function without knowing where it is called ; 2. can we then understand the calling code [setup] without looking at the imeplementation of this function)

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -391,25 +437,146 @@ class OplogPopulator {
...
+            // generating & saving connector names
+            this._setConnectorNames();
+            // initializing data store
+            this._configuredConnectors.forEach(connector => {
+                this._bucketsForConnectors[connector] = new Set();
+            });
+            // getting buckets with at least one backbeat extension active
+            const backbeatEnabledBuckets = await this._getBackbeatEnabledBuckets();
+            // saving previous state of connectors
+            await this._restoreConnectorBuckets(backbeatEnabledBuckets);
```
francoisferrand: regarding "race condition", I guess the general logic here should be:

1. get the existing configuration from kafka-connect : e.g. see what is currently setup
2. do a "scan & watch" of mongo, to ensure we somewhat safely get the list of buckets to watch and all subsequent updates → update the connectors accordingly.

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -391,25 +437,146 @@ class OplogPopulator {
...
+            // assigning a connector to bucket not in previous state
+            this._handleBucketsWithNoConnectors(backbeatEnabledBuckets);
+            // creating/updating the connectors
+            await this._spawnConnectors();
+            // deleting unused connectors
+            await this._deleteUnusedConnectors();
```
francoisferrand: not sure there is really a specific case here?
it could (should) happen after we stop listening to any bucket (though we may indeed want to have some mecanism to delay/bundle the check)

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -391,25 +437,146 @@ class OplogPopulator {
...
+                this._bucketsForConnectors[connector] = new Set();
+            });
+            // getting buckets with at least one backbeat extension active
+            const backbeatEnabledBuckets = await this._getBackbeatEnabledBuckets();
+            // saving previous state of connectors
+            await this._restoreConnectorBuckets(backbeatEnabledBuckets);
+            // assigning a connector to bucket not in previous state
+            this._handleBucketsWithNoConnectors(backbeatEnabledBuckets);
+            // creating/updating the connectors
+            await this._spawnConnectors();
```
francoisferrand: may be created on demand, which would provide more future-proof interface with the "allocation strategy" (i.e. if/when we want to support a strategy with max-number-of-buckets-per-connector)

francoisferrand (commented):
There is a corner case where we may on startup discover more connectors than the expected value in settings.

This is a somewhat valid use case (e.g. when "scaling down" the number of connectors), and The absolute invariant we must maintain is that a bucket must *never* be moved from one connector to another (since the ChangeStream offset is maintained by the connector) : thus we must not try to "reallocate" the bucket to another connector, and we must be able to live with that: e.g. the value in settings is a goal, not the "real" limit, and we may indeed have more connectors.

This also implies that extra connectors may empty-up afterwards (when un-listening to a bucket), get deleted later... which can lead to a "random" list of connector names (i.e. not just `[0-ConenctorsCount[`)

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    /**
+     * Get number of buckets assigned to this
+     * connector
+     * @returns {Number} number of buckets
+     */
+    getBucketCount() {
+        return this._buckets.size;
+    }
```
jbertran: We can make these "standard" getters:

```js
get name() { return this.name; }    // get with instance.name
```

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,278 @@
...
+        const config = constants.defaultConnectorConfig;
+        // connection and source configuration
+        config.name = connectorName;
+        config.database = this._database;
+        config['connection.uri'] = this._mongoUrl;
+        // destination topic configuration
+        config['topic.namespace.map'] = JSON.stringify({
+            '*': this._oplogTopic,
+        });
+        return config;
```
jbertran: In the spirit of avoiding side-effects:

```js
const connectorConfig = {
    name: connectorName,
    database: this._database,
    connection.uri: this._mongoUrl,
    topic.namespace.map: JSON.stringify({
        '*': this._oplogTopic,
    }),
}
return {
    ...constants.defaultConnectorConfig,
    ...connectorConfig
}
```

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    /**
+     * generates a random connector name
+     * @returns {string} generated connector name
+     */
+    _generateConnectorName() {
+        let connectorName = `${this._prefix}${constants.defaultConnectorName}`;
```
jbertran: We have `uuid` installed in backbeat already, might be interesting instead of hand-rolling the naming algorithm.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,278 @@
...
+        try {
+            // generate connector name
+            let connectorName = this._generateConnectorName();
+            while (this._isConnectorNameUsed(connectorName, connectors)) {
+                connectorName = this._generateConnectorName();
+            }
+            // get connector config
+            const config = this._getDefaultConnectorConfiguration(connectorName);
+            // initialize connector
+            const connector = new Connector({
```
jbertran: This introduces a time of check/time of use race condition on the connector name. If we use UUIDs, we largely avoid these (i.e. we stop checking because UUID is good enough). I'm leaning towaards deferring the name-_generating_ logic to the connector itself, and let the `ConnectorManager` decide whether or not it is valid.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    /**
+     * Creates a connector
+     * @param {Connector[]} connectors previous connectors
+     * @param {boolean} spawn should connector be spawned
+     * @returns {Promise|Connector} created connector
+     * @throws {InternalError}
+     */
+    async addConnector(connectors, spawn = true) {
```
jbertran: We have access to `this._connectors` here, why are we not using it?

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,278 @@
...
+            });
+            throw errors.InternalError.customizeDescription(err.description);
+        }
+    }
+
+    /**
+     * Get currently active connectors
+     * @returns {Connectors[]} list of connectors
+     */
+    getConnectors() {
```
jbertran: Ditto for `get`

```js
get connectors() {
    return this._connectors;
}
```

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Starts listening to bucket by
+     * adding and assigning a connector to it
+     * @param {string} bucket bucket name
+     * @param {boolean} isretry is function call a retry
+     * @returns {Promise|undefined} undefined
+     * @throws {InternalError}
+     */
+    async listenToBucket(bucket, isretry = false) {
```
jbertran: minor: `isRetry` for consistent snake-case

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -0,0 +1,124 @@
...
+     * @constructor
+     * @param {Object} params Allocator param
+     * @param {ConnectorsManager} params.connectorsManager connectorsManager
+     * @param {Logger} params.logger logger object
+     */
+    constructor(params) {
+        joi.attempt(params, paramsJoi);
+        this._connectorsManager = params.connectorsManager;
+        this._logger = params.logger;
+        this._allocationStrategy = new LeastFullConnector({
```
jbertran, line 33: So the connector list is frozen on startup? Is this expected?

extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
```diff
@@ -0,0 +1,41 @@
...
+        super(params.logger);
+    }
+
+    /**
+     * Get best connector for assigning a bucket
+     * @param {Connector[]} connectors available connectors
+     * @returns {Connector} connector
+     */
+    getConnector(connectors) {
+        let candidat = null;
```
jbertran: minor: `candidate`

extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
```diff
@@ -0,0 +1,41 @@
...
+        connectors.forEach(connector => {
```
jbertran: If you want to avoid side-effects, you can `reduce` instead:

```js
return connectors.reduce((prev, elt) => elt.getBucketCount() < prev.getBucketCount() ? elt : prev);
```

jbertran (changes requested):
Looks good, most of the comments are cosmetic

extensions/oplogPopulator/modules/Allocator.js
miniscruff: That is not snake-case but camal casing. But I do agree.

extensions/oplogPopulator/modules/Allocator.js
welansari: Good catch, it's not even used actually by the allocationStrategy
Just some old code that i forgot to remove while working on it

jbertran (approved):

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * @constructor
+     * @param {Object} params connector config
+     * @param {string} params.name connector name
+     * @param {Object} params.config Kafka-connect MongoDB source
+     * connector config
+     * @param {Object} params.buckets buckets assigned to this connector
+     * @param {Logger} params.logger logger object
+     * @param {Object} params.kafkaConnectHost kafka connect host
```
jonathan-gramain: `{string}`?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -0,0 +1,202 @@
...
+     * @param {Object} params.kafkaConnectPort kafka connect port
```
jonathan-gramain: `{number}`?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -0,0 +1,202 @@
...
+            });
+            throw errors.InternalError.customizeDescription(err.description);
+        }
+    }
+
+    /**
+     * Remove bucket from this connector
+     * Connector is updated with new bucket list
+     * @param {string} bucket bucket to add
+     * @param {boolean} doUpdate updates connector if true
```
jonathan-gramain: may use `[doUpdate=true]` syntax

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+
+    /**
+     * @constructor
+     * @param {Object} params params
+     * @param {Number} params.nbConnectors number of connectors to have
+     * @param {string} params.database MongoDB database to use (for connector)
+     * @param {string} params.mongoUrl MongoDB connection url
+     * @param {string} params.oplogTopic topic to use for oplog
+     * @param {string} params.kafkaConnectHost kafka connect host
+     * @param {string} params.kafkaConnectPort kafka connect port
```
jonathan-gramain: `{number}`?

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+                logger: this._logger,
+                kafkaConnectHost: this._kafkaConnectHost,
+                kafkaConnectPort: this._kafkaConnectPort,
+            });
+            if (spawn) {
+                await connector.spawn();
+            }
+            return connector;
+        } catch (err) {
+            // allowing one retry for when a connector name
```
jonathan-gramain: in which case would a name be already taken? Apparently we use uuid so there should be virtually no chance of a conflict, hence no need to deal with this issue, unless I'm misreading it.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+                    buckets,
+                    logger: this._logger,
+                    kafkaConnectHost: this._kafkaConnectHost,
+                    kafkaConnectPort: this._kafkaConnectPort,
+                });
+                return connector;
+            }));
+            return connectors;
+        } catch (err) {
+            this._logger.error('An error occured while getting old connectors', {
```
jonathan-gramain: occured -> occurred (same for the other log messages around)

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+     */
+    async _getOldConnectors(connectorNames) {
+        try {
+            const connectors = Promise.all(connectorNames.map(async connectorName => {
+                // get old connector config
+                const config = await this._kafkaConnect.getConnectorConfig(connectorName);
+                // extract buckets from old connector config and filter them
+                // only leaving currently valid buckets
+                const buckets = this._extractBucketsFromConfig(config);
+                // removing pipeline from config
```
jonathan-gramain: why removing the pipeline? Also it could be better if we have to update the config, to do it in a more obvious place, like in the main function, to avoid unintended side-effects from a function starting with `_get` (that ideally should not have side effects), unless it would just be a cache for the get function but it does not look so.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+            const oldConnectorNames = await this._kafkaConnect.getConnectors();
+            if (oldConnectorNames) {
+                const oldConnectors = await this._getOldConnectors(oldConnectorNames);
+                this._connectors.push(...oldConnectors);
+                this._oldConnectors.push(...oldConnectors);
+            }
+            // Add connectors if required number of connectors not reached
+            const nbConnectorsToAdd = this._nbConnectors - this._connectors.length;
+            if (nbConnectorsToAdd > 0) {
+                for (let i = 0; i < nbConnectorsToAdd; i++) {
```
jonathan-gramain: I tend to prefer using `async` module functions such as `async.timesSeries`, it's more explicit and more flexible on what kind of concurrency we're getting, but I'm fine with how it is as a loop.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+            // get and initialize old connectors
+            const oldConnectorNames = await this._kafkaConnect.getConnectors();
+            if (oldConnectorNames) {
+                const oldConnectors = await this._getOldConnectors(oldConnectorNames);
+                this._connectors.push(...oldConnectors);
+                this._oldConnectors.push(...oldConnectors);
+            }
+            // Add connectors if required number of connectors not reached
+            const nbConnectorsToAdd = this._nbConnectors - this._connectors.length;
+            if (nbConnectorsToAdd > 0) {
```
jonathan-gramain: This test looks redundant, could be removed

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+     * @param {string[]} buckets valid bucket names
+     * @returns {Promise|undefined} undefined
+     * @throws {InternalError}
+     */
+    async removeConnectorInvalidBuckets(connector, buckets) {
+        try {
+            // getting connector's invalid buckets
+            const invalidBuckets = connector.buckets.filter(bucket =>
+                !buckets.includes(bucket));
+            if (invalidBuckets.length > 0) {
```
jonathan-gramain: looks like we can remove this check, as `Promise.all` should run fine on an empty list

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+                // removing invalid buckets
+                await Promise.all(invalidBuckets.map(bucket =>
```
jonathan-gramain: How many invalid buckets could there be at worst? Asking to check if there could be so many (say more than 1000) that it might be a bad idea to do them all in parallel, for all connectors, and instead using `async.eachSeries`, `async.eachLimit` or equivalent could be better.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -0,0 +1,267 @@
...
+                    connector.removeBucket(bucket, false)));
+                // updating connector pipeline
+                await connector.updatePipeline();
+            }
+        } catch (err) {
+            this._logger.error('An error occured while removing invalid buckets from a connector', {
+                method: 'ConnectorsManager.removeConnectorInvalidBuckets',
+                error: err.description || err.message,
+                connector,
```
jonathan-gramain: How is the connector going to be logged? I think it's better to just log its name for example, to avoid potentially a giant log line with a lot of internal fields logged.

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -0,0 +1,123 @@
...
+            if (!this._bucketsToConnectors.has(bucket)) {
+                const connectors = this._connectorsManager.connectors;
+                const connector = this._allocationStrategy.getConnector(connectors);
+                await connector.addBucket(bucket);
+                this._bucketsToConnectors.set(bucket, connector);
+            }
+        } catch (err) {
+            // We allow one retry as in initial setup
+            // the connector might not be fully ready
+            if (!isRetry) {
```
jonathan-gramain: One retry may not always be the best value, we can simply make it more flexible without extra complexity by passing `remainingRetries` instead, decrementing it at each call and stopping when reaching 0 (and then, choosing a reasonable initial value that could be 1 but also something else configurable).

Alternatively, `async.retry` provides support for such pattern and may also be tuned to do things like exponential wait times between retries to cope with transient issues - maybe not useful here but for future reference.

jonathan-gramain (commented):
Partial review, will continue tomorrow

tests/unit/oplogPopulator/oplogPopulator.js
```diff
@@ -629,202 +529,113 @@ describe('OplogPopulator', () => {
...
-                assert(generatePipelineStub.calledOnceWithMatch([]));
-                assert(configureConnectorsStub.calledOnceWithMatch({
-                    name: 'source-connector',
-                    config: connectorConfig,
-                }));
-                assert(setupMetastore.calledOnce);
-            })
-            .catch(err => assert.ifError(err));
+    describe('_handleChangeStreamChangeEvent', () => {
+        it('should stop listening to bucket if got deleted', done => {
```
jonathan-gramain: This series of tests may not require a `done` callback since they look to be synchronous.

jonathan-gramain (approved):

PR #2190, opened 2022-05-19
feature: BB-166 change dmf location from aws s3

miniscruff:


lib/Config.js
```diff
@@ -16,11 +16,11 @@ const locationTypeMatch = {
     'location-do-spaces-v1': 'aws_s3',
     'location-aws-s3-v1': 'aws_s3',
     'location-wasabi-v1': 'aws_s3',
-    'location-dmf-v1': 'aws_s3',
     'location-gcp-v1': 'gcp',
     'location-scality-ring-s3-v1': 'aws_s3',
     'location-scality-artesca-s3-v1': 'aws_s3',
     'location-ceph-radosgw-s3-v1': 'aws_s3',
+    'location-dmf-v1': 'location-dmf-v1',
```
rachedbenmustapha: The value `location-dmf-v1` is not consistent with the value `dmf` you added in configuration files generated by zenko-operator.

rachedbenmustapha (changes requested):

lib/Config.js
miniscruff: oh it matches the `type` value not the `locationType` value? So it should be just `dmf`?

rachedbenmustapha: Yes that's what this mapping does. Maybe there would be a way to deduplicate and remove this mapping (this pre-dates the `locationConfig.json` file) but would be out of scope.

rachedbenmustapha (approved):

nicolas2bert (approved):

PR #2205, opened 2022-06-01
Feature: BB-190 implement kafka log reader

welansari:
Issue: [BB-190](https://scality.atlassian.net/browse/BB-190)

The Kafka LogReader consumes messages from the kafka topic used by the kafka-connect mongo connector to mongo publish change stream events.

The Kafka LogReader uses consumer groups and the partition offsets are managed by kafka.

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -37,6 +37,10 @@
             "readPreference": "primary",
             "database": "metadata"
         },
+        "kafka": {
+            "topic": "backbeat-oplog",
+            "consumerGroupId": "backbeat-qp-oplog-group"
+        },
         "probeServer": {
             "bindAddress": "0.0.0.0",
             "port": 4042
--- a/lib/config/configItems.joi.js
+++ b/lib/config/configItems.joi.js
@@ -132,6 +132,12 @@ const mongoJoi = joi.object({
     }),
 });
 
+const qpKafkaJoi = joi.object({
+    topic: joi.string().default('backbeat-oplog'),
+    consumerGroupId: joi.string().default('backbeat-oplog-consumer-id'),
+    logName: joi.string().default('s3-recordlog'),
+});
+
 module.exports = {
     hostPortJoi,
     transportJoi,
@@ -145,4 +151,5 @@ module.exports = {
     probeServerJoi,
     stsConfigJoi,
     mongoJoi,
+    qpKafkaJoi,
 };
--- a/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
+++ b/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
@@ -1,0 +1,85 @@
+const stream = require('stream');
+
+class ListRecordStream extends stream.Transform {
+    /**
+     * @constructor
+     * @param {Logger} logger logger
+     */
+    constructor(logger) {
+        super({ objectMode: true });
+        this._logger = logger;
+    }
+
+    /**
+     * Maps change stream operation type to
+     * backbeat supported types
+     * @param {string} operationType mongo opetation type
+     * @returns {string|undefined} supported operation type
+     */
+    _getType(operationType) {
+        switch (operationType) {
+            case 'insert':
+            case 'update':
+            case 'replace':
+                return 'put';
+            case 'delete':
+                return 'delete';
+            default:
+                this._logger.warn('Got unsupported operation', {
+                    method: 'ListRecordStream._getType',
+                    operationType,
+                });
+                return undefined;
+        }
+    }
+
+    /**
+     * Parse kafka message's value field twice, as the Kafka-connect MongoDB
+     * source connector will double stringify the value field
+     * @param {string} blob sringified json
+     * @returns {Object} parsed json
+     */
+    _parseKafkaMessageValue(blob) {
+        try {
+            const parsed = JSON.parse(JSON.parse(blob));
+            return parsed;
+        } catch (err) {
+            this._logger.warn('Got invalid kafka message value field format', {
+                method: 'ListRecordStream._parseJson',
+                value: blob,
+            });
+            return {};
+        }
+     }
+
+    /**
+     * Formats change stream entries
+     * @param {Object} data chunk of data
+     * @param {Buffer} data.value message contents as a Buffer
+     * In our case this contains the changeStreamDocument data
+     * @param {Number} data.size size of the message, in bytes
+     * @param {string} data.topic topic the message comes from
+     * @param {Number} data.offset  offset the message was read from
+     * @param {Number} data.partition partition the message was on
+     * @param {string} data.key key of the message if present
+     * @param {Number} data.timestamp timestamp of message creation
+     * @param {string} encoding enconding of data
+     * @param {Function} callback callback
+     * @returns {undefined}
+     */
+    _transform(data, encoding, callback) {
+        const changeStreamDocument = this._parseKafkaMessageValue(data.value);
+        const streamObject = {
+            timestamp: new Date(data.timestamp),
+            db: changeStreamDocument.ns && changeStreamDocument.ns.coll,
+            entries: [{
+                key: changeStreamDocument.documentKey && changeStreamDocument.documentKey._id,
+                type: this._getType(changeStreamDocument.operationType),
+                value: changeStreamDocument.fullDocument && changeStreamDocument.fullDocument.value,
+            }],
+        };
+        callback(null, streamObject);
+    }
+}
+
+module.exports = ListRecordStream;
--- a/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
+++ b/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
@@ -1,0 +1,169 @@
+const kafka = require('node-rdkafka');
+const async = require('async');
+const ListRecordStream = require('./ListRecordStream');
+
+// maximum time to wait for consumer group to rebalance (ms)
+const maxDelayRebalanceMs = 120000;
+
+class LogConsumer {
+
+    /**
+     * @constructor
+     * @param {Object} kafkaConfig queue populator kafka config
+     * @param {string} kafkaConfig.hosts kafka hosts
+     * @param {string} kafkaConfig.topic kafka oplog topic
+     * @param {string} kafkaConfig.consumerGroupId consumer group id
+     * @param {Logger} logger logger
+     */
+    constructor(kafkaConfig, logger) {
+        const { hosts, topic, consumerGroupId } = kafkaConfig;
+        this._kafkaHosts = hosts;
+        this._topic = topic;
+        this._consumerGroupId = consumerGroupId;
+        this._log = logger;
+        this._offsets = {};
+    }
+
+    /**
+     * Get partition offsets
+     * @returns {string} stored partition offsets
+     */
+    _getOffset() {
+        // Format: [{ topic: 'oplog-topic', partition: 0, offset: 0 }]
+        return JSON.stringify(this._offsets);
+    }
+
+    /**
+     * Connects consumer to kafka and subscribes
+     * to oplog topic
+     * @param {Function} done callback
+     * @returns {undefined}
+     */
+    setup(done) {
+        // partition offsets will be managed by kafka
+        const consumerParams = {
+            'enable.auto.offset.store': true,
+            'metadata.broker.list': this._kafkaHosts,
+            'group.id': this._consumerGroupId,
+        };
+        const topicParams = {
+            'auto.offset.reset': 'earliest',
+            'enable.auto.commit': true,
+        };
+        this._consumer = new kafka.KafkaConsumer(consumerParams, topicParams);
+        this._consumer.connect();
+        this._consumer.on('ready', () => {
+            this._consumer.subscribe([this._topic]);
+            return done();
+        });
+    }
+
+    /**
+     * Waits for consumer group to rebalance
+     * @param {number} wait amount of time to wait already waited for, in milliseconds
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _waitForAssignment(wait, cb) {
+        setTimeout(() => {
+            // assignements contain the partitions
+            // assigned to this consumer, they are only
+            // set once consumer group is balanced
+            const assignments = this._consumer.assignments();
+            if (assignments.length === 0) {
+                if (wait > maxDelayRebalanceMs) {
+                    this._log.error('Timeout waiting for consumer to be assigned to partitions', {
+                        method: 'LogConsumer._waitForAssignment',
+                        topic: this._topic,
+                        consumerGroupId: this._consumerGroupId,
+                    });
+                    return cb(true);
+                }
+                return this._waitForAssignment(wait + 2000, cb);
+            }
+            return cb();
+        }, 2000);
+    }
+
+    /**
+     * Queries last commited partition offsets
+     * and stores them
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _storeCurrentOffsets(cb) {
+        this._consumer.committed(5000, (err, topicPartitions) => {
+            if (err) {
+                this._log.error('Error while getting offsets', {
+                    method: 'LogConsumer._storeCurrentOffsets',
+                    topic: this._topic,
+                    consumerGroupId: this._consumerGroupId,
+                });
+                return cb(err);
+            }
+            // saving offsets
+            this._offsets = topicPartitions;
+            return cb();
+        });
+    }
+
+    /**
+     * Inintializes record stream
+     * @returns {undefined}
+     */
+    _resetRecordStream() {
+        this._listRecordStream = new ListRecordStream(this._log);
+        this._listRecordStream.getOffset = this._getOffset.bind(this);
+    }
+
+    /**
+     * Consumes kafka messages and writes them to record
+     * stream
+     * @param {Number} limit maximum messages to consume
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _consumeKafkaMessages(limit, cb) {
+        this._resetRecordStream();
+        this._consumer.consume(limit, (err, messages) => {
+            if (err) {
+                this._log.error('An error occured while consuming messages', {
+                    method: 'LogConsumer.readRecords',
+                    topic: this._topic,
+                    consumerGroupId: this._consumerGroupId,
+                });
+                return cb(err);
+            }
+            // writing consumed messages to the stream
+            messages.forEach(message => this._listRecordStream.write(message));
+            return cb();
+        });
+    }
+
+    /**
+     * Reads a certain number of messages from oplog kafka topic
+     * The caller of this function expects a stream to be returned
+     * in the callback
+     * @param {Object} params reading params
+     * @param {number} params.limit maximum number of elements to fetch
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    readRecords(params, cb) {
+        async.series([
+            // waiting for the consumer group to rebalance
+            next => this._waitForAssignment(0, next),
+            // saving last commited offset
+            next => this._storeCurrentOffsets(next),
+            // consuming the desired number of messages at most
+            next => this._consumeKafkaMessages(params.limit, next),
+            next => {
+                // ending and returning the stream
+                this._listRecordStream.end();
+                return next(null, { log: this._listRecordStream, tailable: false });
+            }
+        ], (err, res) => cb(err, res[3]));
+    }
+}
+
+module.exports = LogConsumer;
--- a/lib/queuePopulator/KafkaLogReader.js
+++ b/lib/queuePopulator/KafkaLogReader.js
@@ -1,0 +1,76 @@
+const LogConsumer = require('./KafkaLogConsumer/LogConsumer');
+const LogReader = require('./LogReader');
+
+class KafkaLogReader extends LogReader {
+    /**
+     * @constructor
+     * @param {Object} params - constructor params
+     * @param {Object} params.zkClient - zookeeper client object
+     * @param {Object} params.zkConfig - zookeeper config
+     * @param {Object} params.kafkaConfig - kafka configuration object
+     * @param {Logger} params.logger - logger object
+     * @param {Object} params.qpKafkaConfig - queue populator kafka configuration
+     * @param {QueuePopulatorExtension[]} params.extensions - array of
+     *   queue populator extension modules
+     * @param {MetricsProducer} params.metricsProducer - instance of metrics
+     *   producer
+     * @param {MetricsHandler} params.metricsHandler - instance of metrics
+     *   handler
+     */
+    constructor(params) {
+        const { zkClient, kafkaConfig, zkConfig, qpKafkaConfig,
+            logger, extensions, metricsProducer, metricsHandler } = params;
+        // conf contains global kafka and queuePoplator kafka configs
+        const conf = {
+            hosts: kafkaConfig.hosts,
+            ...qpKafkaConfig
+        };
+        logger.info('initializing kafka log reader',
+            { method: 'KafkaLogReader.constructor',
+            kafkaConfig: conf });
+        const logConsumer = new LogConsumer(conf, logger);
+        super({ zkClient, kafkaConfig, zkConfig, logConsumer,
+                logId: `kafka_${qpKafkaConfig.logName}`, logger, extensions,
+                metricsProducer, metricsHandler });
+        this._kafkaConfig = conf;
+    }
+
+    /**
+    * Setup log consumer
+    * @param {object} done callback function
+    * @return {undefined}
+    */
+     setup(done) {
+        this.logConsumer.setup(err => {
+            if (err) {
+                this.log.error('error setting up log consumer', {
+                    method: 'KafkaLogReader.setup',
+                    kafkaConfig: this._kafkaConfig,
+                });
+                return done(err);
+            }
+            return super.setup(done);
+        });
+    }
+
+    /**
+     * Get kakfa log info
+     * @return {object} logName
+     */
+    getLogInfo() {
+        return { logName: this._kafkaConfig.logName };
+    }
+
+    /**
+     * Get metric label
+     * @returns {object} metric labels
+     */
+    getMetricLabels() {
+        return {
+            logName: 'kafka-log',
+            logId: this._kafkaConfig.logName,
+        };
+    }
+}
+
+module.exports = KafkaLogReader;
--- a/tests/unit/lib/queuePopulator/kafkaLogConsumer/ListRecordStream.js
+++ b/tests/unit/lib/queuePopulator/kafkaLogConsumer/ListRecordStream.js
@@ -1,0 +1,126 @@
+const assert = require('assert');
+const werelogs = require('werelogs');
+const logger = new werelogs.Logger('ListRecordStream');
+const ListRecordStream =
+    require('../../../../../lib/queuePopulator/KafkaLogConsumer/ListRecordStream');
+
+const changeStreamDocument = {
+    ns: {
+        db: 'metadata',
+        coll: 'example-bucket',
+    },
+    documentKey: {
+        _id: 'example-key',
+    },
+    operationType: 'insert',
+    fullDocument: {
+        value: {
+            field: 'value'
+        }
+    }
+};
+const kafkaMessage = {
+    // kafka-connect double stringifies the message
+    value: Buffer.from(JSON.stringify(JSON.stringify(changeStreamDocument))),
+    timestamp: Date.now(),
+    size: 2,
+    topic: 'oplog-topic',
+    offset: 1337,
+    partition: 0,
+    key: Buffer.from('key'),
+};
+
+const InvalidKafkaMessage = {
+    value: Buffer.from(''),
+    timestamp: Date.now(),
+    size: 2,
+    topic: 'oplog-topic',
+    offset: 1337,
+    partition: 0,
+    key: Buffer.from('key'),
+};
+
+describe('ListRecordStream', () => {
+    let listRecordStream;
+    beforeEach(() => {
+        listRecordStream = new ListRecordStream(logger);
+    });
+
+    describe('_getType', () => {
+        [
+            {
+                opType: 'insert',
+                expected: 'put'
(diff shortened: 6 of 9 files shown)


conf/config.json
```diff
@@ -37,6 +37,11 @@
             "readPreference": "primary",
             "database": "metadata"
         },
+        "kafka": {
```
francoisferrand, line 40: this is not (IMHO) a separate kafka host : we should use the one configured "globally" for backbeat.
The only thing we need here is indeed the `topic` & `groupId` (which may be renamed `consumerGroupId`, for clarity ?)

lib/config/configItems.joi.js
```diff
@@ -132,6 +132,13 @@ const mongoJoi = joi.object({
     }),
 });
 
+const qpKafkaJoi = joi.object({
+    hosts: joi.string().required(),
+    topic: joi.string().required(),
+    groupId: joi.string().required(),
```
francoisferrand: should these be `required`?
it may seem appropriate to have some defaults here, instead of just in zkop (and could help this code gets partially integrated, e.g. not all parts are done)

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * parses json strings
+     * @param {string} blob sringified json
+     * @returns {Object} parsed json
+     */
+    _jsonParser(blob) {
+        let parsed = JSON.parse(blob);
+        if (typeof parsed === 'string') {
+            parsed = this._jsonParser(parsed);
+        }
```
francoisferrand: why do we need to (sometimes?) parse Json twice?
this should be predictable, and we should now what to expect in all cases... and adding some explanation in the code what/why happens would probably help readability

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,72 @@
...
+                return 'none';
+        }
+    }
+
+    /**
+     * parses json strings
+     * @param {string} blob sringified json
+     * @returns {Object} parsed json
+     */
+    _jsonParser(blob) {
```
francoisferrand: nit : this is not a "parser", but actually does the parsing → `_parseJson`

An object represents some 'data', so variables and types names should be a **noun** (or nominal expression)
A function represents an 'operation', so functon names should be a **verb** (or sentence)

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,166 @@
+const kafka = require('node-rdkafka');
+const async = require('async');
+const ListRecordStream = require('./ListRecordStream');
+
+// maximum time to wait for consumer group to rebalance (ms)
+const maxDelayRebalance = 120000;
```
francoisferrand: why "2 minute", is this some common/standard value?
* If we rely on some default value, can't we just leave the field empty and let kafka use whatever value it finds appropriate (or has in its config_
* otherwise, should this be added to the config?

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,166 @@
...
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _waitForAssignment(wait, cb) {
+        setTimeout(() => {
+            /**
+             * assignements contain the partitions
+             * assigned to this consumer, they are only
+             * set once consumer group is balanced
+             */
```
francoisferrand: Should not be a "doc/api" comment?

```suggestion
             // assignements contain the partitions
             // assigned to this consumer, they are only
             // set once consumer group is balanced
```

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,166 @@
...
+            if (err) {
+                this._log.error('Error while getting offsets', {
+                    method: 'LogConsumer._storeCurrentOffsets',
+                    topic: this._topic,
+                    groupId: this._groupId,
+                });
+                return cb(err);
+            }
+            // saving offsets
+            this._offsets = topicPartitions;
```
francoisferrand, line 105: why do we need to store the offset? (and where does it get stored / how to ensure we get back on track if we crash/lose it)

storing/restoring the offset should be handled already by kafka consumer api, as i understand it?

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,166 @@
...
+     * @param {number} params.limit maximum number of elements to fetch
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    readRecords(params, cb) {
+        async.series([
+            // waiting for the consumer group to rebalance
+            next => this._waitForAssignment(0, next),
+            // retreiving last stored offset
+            next => this._storeCurrentOffsets(next),
```
francoisferrand: nit: type in comment, should be `retrieving`

this is hard to read, the comment saying `retrieving` while the function is calls to `store`, which is kind of the opposite :-/

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,166 @@
...
+            return cb();
+        });
+    }
+
+    /**
+     * Reads a certain number of messages from oplog kafka topic
+     * The caller of this function expects a stream to be returned
+     * in the callback
+     * @param {Object} params reading params
+     * @param {number} params.limit maximum number of elements to fetch
```
francoisferrand, line 148: what is this "limit" : the size of a batch, or the total number of elements?

generally, we should not limit the (total) number of elements i guess : we just setup the stream, and process messages until stopped, no?

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
welansari: The offsets are stored and handled entirely by Kafka, we don't manage them by hand

This was done to be compatible with how the logReader works as it requires getting the last offsets from the logConsumer and stores it when entries are published

https://github.com/scality/backbeat/blob/020213d16178fb9d2a1ae0fb41763d7bc3a7f7ea/lib/queuePopulator/LogReader.js#L553-L554

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
welansari: I tried using a tailable stream first, but that didn't work as expected

The issue i got was that the offsets stored in Kafka weren't necessarily where the logReader stopped, as even after consuming a message from a topic, it might not get processed by the logReader

This might of not been an issue if we could start the tailable stream from an offset that we determine, however as far as i know (and tried) the Kafka library used in backbeat do not support that

The second point is that even when using a tailable stream the logReader still uses a limited batch size. The stream either gets closed after the limit is reached, or after a timer sets off (this prevents infinite waiting time when the max batch size is not reached)

https://github.com/scality/backbeat/blob/020213d16178fb9d2a1ae0fb41763d7bc3a7f7ea/lib/queuePopulator/LogReader.js#L438-L446

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
welansari: 2 minutes is just a safe value that i found while testing, it usually takes less time than that to rebalance.
I don't think there is a use for this being in the config, and the value is used in a custom function that waits for rebalance, so there is no default Kafka value for that

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: the Kafka messages we consume contain over-stringified ChangeStreamDocuments, that's why they need to be parsed twice (it is always the case btw)

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,74 @@
...
+     * @param {Object} data chunk of data
+     * @param {string} encoding enconding of data
+     * @param {Function} callback callback
+     * @returns {undefined}
+     */
+    _transform(data, encoding, callback) {
+        // data.value contains an over-stringified string which'll
+        // neeed to be parsed twice to get the object
+        const changeStreamDocument = this._parseJson(data.value);
+        const streamObject = {
```
jonathan-gramain, line 72: We may want to be resilient against missing attributes in the parsed value to avoid crashes and log errors instead, e.g. if an admin decides to push a "test" message without the required fields.

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,169 @@
...
+        this._consumer.connect();
+        this._consumer.on('ready', () => {
+            this._consumer.subscribe([this._topic]);
+            return done();
+        });
+    }
+
+    /**
+     * Waits for consumer group to rebalance
+     * @param {number} wait amount of time to wait
```
jonathan-gramain:
```suggestion
     * @param {number} wait amount of time already waited for, in milliseconds
```

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,169 @@
+const kafka = require('node-rdkafka');
+const async = require('async');
+const ListRecordStream = require('./ListRecordStream');
+
+// maximum time to wait for consumer group to rebalance (ms)
+const maxDelayRebalance = 120000;
```
jonathan-gramain: It can improve readability to attach the unit to the variable name:

```suggestion
const maxDelayRebalanceMs = 120000;
```

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,169 @@
...
+     * stream
+     * @param {Number} limit maximum messages to consume
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _consumeKafkaMessages(limit, cb) {
+        this._resetRecordStream();
+        this._consumer.consume(limit, (err, messages) => {
+            if (err) {
+                this._log.error('Error occured while consuming messages', {
```
jonathan-gramain:
```suggestion
                this._log.error('An error occurred while consuming messages', {
```

lib/queuePopulator/KafkaLogReader.js
```diff
@@ -0,0 +1,76 @@
...
+     *   queue populator extension modules
+     * @param {MetricsProducer} params.metricsProducer - instance of metrics
+     *   producer
+     * @param {MetricsHandler} params.metricsHandler - instance of metrics
+     *   handler
+     */
+    constructor(params) {
+        const { zkClient, kafkaConfig, zkConfig, qpKafkaConfig,
+            logger, extensions, metricsProducer, metricsHandler } = params;
+        logger.info('initializing kafka log reader',
```
jonathan-gramain: it might be useful to log some or all of `qpKafkaConfig` as well

jonathan-gramain (approved):

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,82 @@
...
+            case 'replace':
+                return 'put';
+            case 'delete':
+                return 'delete';
+            default:
+                this._logger.warn('Got unsupported operation', {
+                    method: 'ListRecordStream._getType',
+                    operationType,
+                });
+                return 'none';
```
miniscruff: Are we meant to use this value later on to compare? I feel like `null` or `undefined` would be a better value. Maybe even an error to be honest.

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * parses json strings
+     * @param {string} blob sringified json
+     * @returns {Object} parsed json
+     */
+    _parseJson(blob) {
+        try {
+            let parsed = JSON.parse(blob);
+            if (typeof parsed === 'string') {
```
miniscruff: How do you get this condition? I could not make it happen.

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,82 @@
...
+                return 'none';
+        }
+    }
+
+    /**
+     * parses json strings
+     * @param {string} blob sringified json
+     * @returns {Object} parsed json
+     */
+    _parseJson(blob) {
```
miniscruff: Either this function needs new name or the error should be more generic, also it doesn't accurately represent it does a recursive parse.

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,82 @@
...
+        // data.value contains an over-stringified string which'll
+        // need to be parsed twice to get the object
+        const changeStreamDocument = this._parseJson(data.value);
+        const streamObject = {
+            timestamp: new Date(data.timestamp),
+            db: changeStreamDocument.ns && changeStreamDocument.ns.coll,
+            entries: [{
+                key: changeStreamDocument.documentKey && changeStreamDocument.documentKey._id,
+                type: this._getType(changeStreamDocument.operationType),
+                value: changeStreamDocument.fullDocument && changeStreamDocument.fullDocument.value,
```
miniscruff, line 78:
```suggestion
                key: changeStreamDocument.documentKey?._id,
                type: this._getType(changeStreamDocument.operationType),
                value: changeStreamDocument.fullDocument?.value,
```

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,82 @@
...
+     * @param {Function} callback callback
+     * @returns {undefined}
+     */
+    _transform(data, encoding, callback) {
+        // data.value contains an over-stringified string which'll
+        // need to be parsed twice to get the object
+        const changeStreamDocument = this._parseJson(data.value);
+        const streamObject = {
+            timestamp: new Date(data.timestamp),
+            db: changeStreamDocument.ns && changeStreamDocument.ns.coll,
```
miniscruff, line 74:
```suggestion
            db: changeStreamDocument.ns?.coll,
```

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,82 @@
...
+                method: 'ListRecordStream._parseJson',
+                value: blob,
+            });
+            return {};
+        }
+     }
+
+    /**
+     * Formats change stream entries
+     * @param {Object} data chunk of data
```
miniscruff, line 57: Would really prefer a typedef for objects like this, I simply have no idea what is actually in data unless I analyze line by line if you use all parts of it.

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,82 @@
...
+            timestamp: new Date(data.timestamp),
+            db: changeStreamDocument.ns && changeStreamDocument.ns.coll,
+            entries: [{
+                key: changeStreamDocument.documentKey && changeStreamDocument.documentKey._id,
+                type: this._getType(changeStreamDocument.operationType),
+                value: changeStreamDocument.fullDocument && changeStreamDocument.fullDocument.value,
+            }],
+        };
+        this.push(streamObject);
+        callback();
```
miniscruff: I read that you can do this, ( I found it weird we were pushing in a transform function but that is what you need to do so I guess its right... )
```suggestion
        callback(null, streamObject);
```

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,169 @@
...
+        });
+    }
+
+    /**
+     * Waits for consumer group to rebalance
+     * @param {number} wait amount of time to wait already waited for, in milliseconds
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _waitForAssignment(wait, cb) {
```
miniscruff, line 67: I won't block this PR or anything cause of it, but i really think we should use async await instead of callbacks for all new code and I can see a few uses here. If you need to convert a callback to a promise you can use https://nodejs.org/api/util.html#utilpromisifyoriginal

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,169 @@
...
+     * Reads a certain number of messages from oplog kafka topic
+     * The caller of this function expects a stream to be returned
+     * in the callback
+     * @param {Object} params reading params
+     * @param {number} params.limit maximum number of elements to fetch
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    readRecords(params, cb) {
+        async.series([
```
miniscruff, line 153: Similar to my comment above, async series can just be a sequence of awaits.

miniscruff (changes requested):

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: Kafka-Connect MongoDB Source Connector is the one that'll be pushing messages into the topic used by this logReader. And so it is the one that over-stringifies the messages

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: linter doesn't like that apparently

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: this too

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
welansari: Although i completely agree with your point (we did this with new OplogPopulator extension), IMHO this part of the code should use callbacks as that is what the LogReader uses, unless we change that. I'm not really a fan of mixing between the two styles in the same place

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
miniscruff: Do you have an example of what an over-stringified value looks like? I couldn't make one myself.

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
miniscruff: What is the linting error, this should be fine it means the same thing

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: when consuming from kafka this is what i get:
```
"{\"_id\": {\"_data\": {\"$binary\": \"gmKorHgAAAABRjxfaWQAPGEtYnVja2V0AABaEAR3m6BArY5JKYbPqzz2tpQPBA==\", \"$type\": \"00\"}}, \"operationType\": \"delete\", \"ns\": {\"db\": \"metadata\", \"coll\": \"__metastore\"}, \"documentKey\": {\"_id\": \"a-bucket\"}}"
```

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * @constructor
+     * @param {Object} kafkaConfig queue populator kafka config
+     * @param {string} kafkaConfig.hosts kafka hosts
+     * @param {string} kafkaConfig.topic kafka oplog topic
+     * @param {string} kafkaConfig.consumerGroupId consumer group id
+     * @param {Logger} logger logger
+     */
+    constructor(kafkaConfig, logger) {
```
miniscruff, line 18: @RachedBenMustapha @jonathan-gramain @alexanderchan-scality I know we follow this pattern all over the place but I was wondering if it was possible to instead of passing in configs, in this case for kafka but a similar pattern is used for mongo and zookeeper I believe, we pass in an already connected client. 
This would remove the need for the setup method as well as removing a responsibility we have. We also can reduce some of our code duplication.

Wouldn't block this PR cause it is following what we do but just mentioning it.

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: I get this :
```
/Users/walidelansari/Desktop/backbeat/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
  79:41  error  Parsing error: Unexpected token .
```

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
miniscruff: That was parsed just fine in one go.

tests/unit/lib/queuePopulator/kafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,133 @@
...
+            const objStr = JSON.stringify(JSON.stringify(changeStreamDocument));
+            const parsed = listRecordStream._recursiveParseKafkaMessageValue(objStr);
+            assert.deepEqual(parsed, changeStreamDocument);
+            return done();
+        });
+        it('Should return empty object when format is invalid', done => {
+            const parsed = listRecordStream._recursiveParseKafkaMessageValue('');
+            assert.deepEqual(parsed, {});
+            return done();
+        });
```
miniscruff: None of these seem to need done.
```suggestion
        it('Should parse stringified object', () => {
            const objStr = JSON.stringify(changeStreamDocument);
            const parsed = listRecordStream._recursiveParseKafkaMessageValue(objStr);
            assert.deepEqual(parsed, changeStreamDocument);
        });
        it('Should parse doubly stringified object', () => {
            const objStr = JSON.stringify(JSON.stringify(changeStreamDocument));
            const parsed = listRecordStream._recursiveParseKafkaMessageValue(objStr);
            assert.deepEqual(parsed, changeStreamDocument);
        });
        it('Should return empty object when format is invalid', () => {
            const parsed = listRecordStream._recursiveParseKafkaMessageValue('');
            assert.deepEqual(parsed, {});
        });
```

tests/unit/lib/queuePopulator/kafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,219 @@
...
+            });
+        });
+    });
+
+    describe('_resetRecordStream', () => {
+        it('Should initialize record stream', done => {
+            logConsumer._resetRecordStream();
+            assert(logConsumer._listRecordStream instanceof ListRecordStream);
+            assert.strictEqual(typeof(logConsumer._listRecordStream.getOffset), 'function');
+            return done();
```
miniscruff: done not required

tests/unit/lib/queuePopulator/kafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,219 @@
...
+        });
+
+        it('Should fail if it can\'t consume kafka messages', done => {
+            const waitAssignementStub = sinon.stub(logConsumer, '_waitForAssignment')
+                .callsArg(1);
+            const storeOffsetsStub = sinon.stub(logConsumer, '_storeCurrentOffsets')
+                .callsArg(0);
+            const consumeKafkaStub = sinon.stub(logConsumer, '_consumeKafkaMessages')
+                .callsArgWith(1, errors.InternalError);
+            logConsumer._resetRecordStream();
```
miniscruff: do we need to call reset with each test, isn't log consumer re-instantiated?

tests/unit/lib/queuePopulator/kafkaLogConsumer/LogConsumer.js
```diff
@@ -0,0 +1,219 @@
...
+    key: Buffer.from('key'),
+};
+
+describe('LogConsumer', () => {
+    let logConsumer;
+    beforeEach(() => {
+        logConsumer = new LogConsumer(kafkaConfig, logger);
+    });
+
+    describe('_getOffset', () => {
```
miniscruff: This test is not really necessary, testing a getter gets the value you just set doesn't really help us much. As long as you use the getter in some other test it should be fine. I know it does stringify it but we don't need to test that javascript can stringify an object, also not entirely sure we should be stringifying a value like this. I would let our callee stringify it if they need it as a string.

miniscruff (changes requested):

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
miniscruff: Talked on slack, comment above is fine.

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
miniscruff: Eslint needs to be upgraded for optional chaining, or we add parser

tests/unit/lib/queuePopulator/kafkaLogConsumer/LogConsumer.js
welansari: I think the reason i did that was because i stubbed the `_consumeKafkaMessages` function, and that's the only place where we do stream initialization.
But looking at it again, we don't even need the stream in some of those test cases as we throw an error before reaching the code that uses the stream in the `readRecords` function

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
miniscruff: Ok, this is the last thing left for me, I think the code is fine ( although I would use a while loop over recursion but whatever )  my only concern is why we need it in the first place. I have not seen this sort of double stringified before. If anything if it is a specific structure just doing parse twice would make more sense as its a fixed requirement.

miniscruff: @vrancurel ^

vrancurel: It is the JS coding of a simple stringification:
```
> let b="{\"_id\": {\"_data\": {\"$binary\": \"gmKorHgAAAABRjxfaWQAPGEtYnVja2V0AABaEAR3m6BArY5JKYbPqzz2tpQPBA==\", \"$type\": \"00\"}}, \"operationType\": \"delete\", \"ns\": {\"db\": \"metadata\", \"coll\": \"__metastore\"}, \"documentKey\": {\"_id\": \"a-bucket\"}}"
undefined
> b
'{"_id": {"_data": {"$binary": "gmKorHgAAAABRjxfaWQAPGEtYnVja2V0AABaEAR3m6BArY5JKYbPqzz2tpQPBA==", "$type": "00"}}, "operationType": "delete", "ns": {"db": "metadata", "coll": "__metastore"}, "documentKey": {"_id": "a-bucket"}}'
> JSON.parse(b)
{
  _id: {
    _data: {
      '$binary': 'gmKorHgAAAABRjxfaWQAPGEtYnVja2V0AABaEAR3m6BArY5JKYbPqzz2tpQPBA==',
      '$type': '00'
    }
  },
  operationType: 'delete',
  ns: { db: 'metadata', coll: '__metastore' },
  documentKey: { _id: 'a-bucket' }
}
```

welansari: @miniscruff I had a call with @vrancurel yesterday, and we found that the value we get from the kafka message is actually doubly stringified, Kafka-connect is the one doing that and as far as i know that can't be changed.

So we can either do as we do now and parse it twice, or we could switch to BSON format and configure Kafka-connect to output in that format, in which case we'll only need to parse once. 

The issue with BSON though is that it's going to make debugging a bit more complicated if we want to see what get's pushed into the topic.

Side Note: i removed the recursion from the code, so now we parse twice at most. In case of debugging if we want to push a message manually the code can handle single parsing as well.

welansari: i changed that actually
https://github.com/scality/backbeat/compare/4f42471470be74a7f4f1d1f507866157921ce65a..b9a85abe356a0594401c72a7e7bb61e26680903c

miniscruff: Ok I see the new code, one moment

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,90 @@
...
+                });
+                return undefined;
+        }
+    }
+
+    /**
+     * Parse kafka message's
+     * value field twice, as it can be over stringified
+     * when reading messages pushed by Kafka-connect
+     * MongoDB source connector
```
miniscruff:
```suggestion
     * Parse kafka message's value field twice, as the Kafka-connect MongoDB
     * source connector will double stringify the value field
```
I would make this comment more of an absolute not a "can be"

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -0,0 +1,90 @@
...
+     * @param {string} blob sringified json
+     * @returns {Object} parsed json
+     */
+    _parseKafkaMessageValue(blob) {
+        try {
+            let parsed = JSON.parse(blob);
+            if (typeof parsed === 'string') {
+                parsed = JSON.parse(parsed);
+            }
+            return parsed;
```
miniscruff:
```suggestion
            return JSON.parse(JSON.parse(blob));
```
Just do a double parse immediately, we are parsing from a defined rule so the string check is redundant

miniscruff (changes requested):
Small cosmetic changes, should be good after this

miniscruff (approved):

PR #2206, opened 2022-06-01
Feature: BB-146 support different internal topics for notification targets

welansari:
Issue: [BB-146](https://scality.atlassian.net/browse/BB-146)

**Changes**:

NotificationQueuePopulator:
- Now pushes notification messages onto specific internal Kafka topics of destinations for which the object validates their notification config.

NotificationQueueProcessor:
- Now consumes messages form internal Kafka topic specific to their destination
- Made processKafkaEntry use callbacks to fix an async bug where we get the following error `done is not a function`

Config
- Added fields in the notification config to specify the internal topic to use for each destination

ReplicationProcessor:
- Now supports pushing replication failed messages to specific internal topic of destinations

### DIFF
--- a/extensions/notification/NotificationConfigValidator.js
+++ b/extensions/notification/NotificationConfigValidator.js
@@ -1,5 +1,15 @@
 const joi = require('joi');
 
+const destinationSchema = joi.object({
+    resource: joi.string().required(),
+    type: joi.string().required(),
+    host: joi.string().required(),
+    port: joi.number().required(),
+    internalTopic: joi.string(),
+    topic: joi.string().required(),
+    auth: joi.object().default({}),
+});
+
 const joiSchema = joi.object({
     topic: joi.string(),
     monitorNotificationFailures: joi.boolean().default(true),
@@ -8,6 +18,7 @@ const joiSchema = joi.object({
         groupId: joi.string().required(),
         concurrency: joi.number().greater(0).default(10),
     },
+    destinations: joi.array().items(destinationSchema).default([]),
 });
 
 function configValidator(backbeatConfig, extConfig) {
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -73,21 +73,40 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                 key,
                 eventType,
             });
-            if (configUtil.validateEntry(config, ent)) {
-                const message
-                    = messageUtil.addLogAttributes(value, ent);
-                this.log.info('publishing message', {
-                    method: 'NotificationQueuePopulator._processObjectEntry',
-                    bucket,
-                    key: message.key,
-                    eventType,
-                    eventTime: message.dateTime,
-                });
-                this.publish(this.notificationConfig.topic,
-                    `${bucket}/${key}`,
-                    JSON.stringify(message));
-            }
-            return undefined;
+            // validate and push kafka message foreach destination topic
+            this.notificationConfig.destinations.forEach(destination => {
+                // get destination specific notification config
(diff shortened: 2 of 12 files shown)


extensions/notification/NotificationConfigValidator.js
```diff
@@ -1,7 +1,12 @@
 const joi = require('joi');
 
+const internalTopic = joi.object({
+    resource: joi.string().required(),
+    topic: joi.string().required(),
+});
+
 const joiSchema = joi.object({
-    topic: joi.string(),
+    topics: joi.array().items(internalTopic).required(),
```
francoisferrand: ~should still support the "old" format with just a single stream?~

or actually, the topic should be part of each "target"; and this global `topic` config may be used as the default or prefix
→ i.e. allow specifying the name of the "internal" topic to use inside each target, else use some default

basically it seems quite redundant and risky to require "matching" different parts of the config (and not very intuitive)

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -73,21 +73,38 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
-                    eventType,
-                    eventTime: message.dateTime,
-                });
-                this.publish(this.notificationConfig.topic,
-                    bucket,
-                    JSON.stringify(message));
-            }
-            return undefined;
+            // validate and push kafka message foreach destination topic
+            this.notificationConfig.topics.forEach(destination => {
```
francoisferrand: seems this is not "topics" but "destinations" ?

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -45,6 +47,12 @@ class QueueProcessor extends EventEmitter {
         this.bnConfigManager = null;
         this._consumer = null;
         this._destination = null;
+        /**
+         * Once the notification manager is initialized
+         * this will hold the callback version of the getConfig
+         * function of the notification config manager
+         */
```
francoisferrand: should not be a doc comment?

```suggestion
         // Once the notification manager is initialized
         // this will hold the callback version of the getConfig
         // function of the notification config manager
```

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -28,23 +30,28 @@ class QueueProcessor extends EventEmitter {
...
+    constructor(mongoConfig, kafkaConfig, notifConfig, destinationId) {
         super();
         this.mongoConfig = mongoConfig;
         this.kafkaConfig = kafkaConfig;
         this.notifConfig = notifConfig;
-        this.destinationConfig = destinationConfig;
         this.destinationId = destinationId;
+        this.destinationConfig
+            = notifConfig.destinations.find(dest => dest.resource === destinationId);
+        assert(this.destinationConfig, 'Invalid destination argument. Destination ' +
```
jonathan-gramain: It's minor as this is only to debug a config issue, but we could help the admin by logging the `destinationId` with a formatted string

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -158,59 +170,69 @@ class QueueProcessor extends EventEmitter {
...
+                            method: 'QueueProcessor.processKafkaEntry',
+                            bucket,
+                            key,
+                            eventType: eventRecord.eventName,
+                            eventTime: eventRecord.eventTime,
+                            destination: this.destinationId,
+                        });
+                        return this._destination.send([msg], done);
+                    }
+                    return done();
```
jonathan-gramain: This line looks unnecessary, as the callback will be called at 234

jonathan-gramain (approved):

naren-rajendran (approved):

PR #2215, opened 2022-06-08
Bugfix/bb 191 lc robustness workarounds

rachedbenmustapha:
Workarounds for robustness:
- ghost buckets
- orphan buckets
- bucketd errors
- throttling
- vault errors

### DIFF
--- a/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
+++ b/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
@@ -301,6 +301,11 @@ class LifecycleBucketProcessor {
                     return cb();
                 }
 
+                if (err.code === 'NoSuchBucket') {
+                    this._log.error('skipping non-existent bucket', { bucket });
+                    return cb();
+                }
+
                 this._log.error('error getting bucket lifecycle config', {
                     method: 'LifecycleBucketProcessor._processBucketEntry',
                     bucket,
--- a/extensions/lifecycle/util/AccountIdCache.js
+++ b/extensions/lifecycle/util/AccountIdCache.js
@@ -1,0 +1,41 @@
+'user strict';
+
+class AccountIdCache extends Map {
+    constructor(maxSize) {
+        super();
+        this.maxSize = maxSize;
+        this.misses = new Set();
+    }
+
+    expireOldest() {
+        let overflowKeysNum = this.size - this.maxSize;
+        const keyIterator = this.keys();
+        let current = keyIterator.next();
+
+        while (overflowKeysNum > 0 && !current.done) {
+            this.delete(current.value);
+            current = keyIterator.next();
+            overflowKeysNum--;
+        }
+    }
+
+    miss(key) {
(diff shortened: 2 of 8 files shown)


tests/functional/lifecycle/LifecycleBucketProcessor.spec.js
```diff
@@ -162,6 +167,16 @@ describe('Lifecycle Bucket Processor', function lifecycleBucketProcessor() {
         });
     });
 
+    it.only('should not retry for NoSuchBucket errors', () => {
```
miniscruff: remove only

rachedbenmustapha: Oops, done. thanks

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -301,6 +301,11 @@ class LifecycleBucketProcessor {
                     return cb();
                 }
 
+                if (err.code === 'NoSuchBucket') {
+                    this._log.error('skipping non-existent bucket', { bucket });
```
rahulreddy, line 305: Should this be `info` unless you see a reason to dump the logs when it reaches this code path.

rachedbenmustapha: It effectively is an error, not caused by the user (if we were to return a status code it would be 5xx, not 4xx). In theory it should not dump, but werelogs doesn't have an `errorButDontDump` api. `this._log` is not a request logger so I don't think it would dump.

extensions/lifecycle/util/AccountIdCache.js
```diff
@@ -0,0 +1,41 @@
+'user strict';
+
+class AccountIdCache extends Map {
+    constructor(maxSize) {
+        super();
+        this.maxSize = maxSize;
+        this.misses = new Set();
+    }
+
+    expireOldest() {
```
alexanderchan-scality, line 10: would it be better to use an LRU cache here?
there is an implementation in arsenal: https://github.com/scality/Arsenal/blob/development/8.1/lib/algos/cache/LRUCache.js

rachedbenmustapha: I haven't compared run-time with the LRU cache but this Map-based implementation has close to 0 memory/cpu overhead (no extra list) and the implementation is trivial in comparison, while allowing control of when to purge which is useful in this particular use case.

alexanderchan-scality (approved):

tests/utils/S3ClientMock.js
```diff
@@ -87,6 +90,13 @@ class S3ClientMock {
         });
     }
 
+    verifyNoRetries() {
+        Object.keys(this.failures).forEach(f => {
```
miniscruff, line 94: you could use Object.entries here, might look slightly better. Maybe.

miniscruff (approved):

PR #2218, opened 2022-06-09
Feature: BB-194 move extension specific code out from oplog populator

welansari:
Moving extension specific code from the OplogPopulator to the extensions
The code moved is used for :
- Checking a bucket's metadata to see if it uses a backbeat extension
- Getting MongoDB filters that get buckets that use the extensions

### DIFF
--- a/extensions/lifecycle/LifecycleOplogPopulatorUtils.js
+++ b/extensions/lifecycle/LifecycleOplogPopulatorUtils.js
@@ -1,0 +1,42 @@
+const OplogPopulatorUtils = require('../../lib/util/OplogPopulatorUtils');
+
+class LifecycleOplogPopulatorUtils extends OplogPopulatorUtils {
+    /**
+     * Get extension specific MongoDB filter
+     * to get buckets with that have the lifecycle
+     * extension enabled
+     * @returns {Object} MongoDB filter
+     */
+     static getExtensionMongoDBFilter() {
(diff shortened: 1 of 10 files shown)


Kerkesni:
ping

jbertran (approved):

extensions/lifecycle/LifecycleOplogPopulatorUtils.js
```diff
@@ -0,0 +1,43 @@
...
+     * active
+     * @param {Object} bucketMD bucket metadata
+     * @returns {boolean} true if the bucket has the
+     * current extension enabled
+     */
+    static isBucketExtensionEnabled(bucketMD) {
+        const rules = (bucketMD.lifecycleConfiguration
+            && bucketMD.lifecycleConfiguration.rules)
+            || null;
+        if (!rules || rules.length === 0) {
```
ghost: If the test fails it should be already undefined or null, right?
```suggestion
        const rules = bucketMD.lifecycleConfiguration
            && bucketMD.lifecycleConfiguration.rules;
        if (!rules || rules.length === 0) {
```

ghost (approved):
LGTM, just a minor comment

extensions/lifecycle/LifecycleOplogPopulatorUtils.js
welansari: yes you're right, fixing that

PR #2228, opened 2022-06-15
BB-138 - deduplicate producer code

jbertran:
Producer code was more or less replicated between the canonical `BackbeatProducer` code and the bucket notification extension `KafkaProducer` code.

This PR concentrates logic in `BackbeatProducer`, of which `KafkaProducer` is essentially a Kafka configuration superset at startup.

Since the 2 had diverged between the creation of `KafkaProducer` as a duplicate with slightly different configuration, we get the following enhancements made in `BackbeatProducer` that `KafkaProducer` did not receive:

* 265cffe5 (ZENKO-1420) canary message filtering
* f1daedc4 (ZENKO-1520) receivedReport counting
* 51a9c539 (ZENKO-1643) filtering out broker transport failures

Additionally, we finish the non-keyed partitioner update (`7e863860` - S3C-1154) by removing `keyedPartitioner` from joi configuration schema in `BackbeatProducer`, since it's not used anywhere any more.

Note: this code would be better served by using `static` members for a lot of the configuration "variables", but this means pulling on a big ball of yarn until reaching the fact that the linter doesn't accept this if we remain in default (i.e. ES6) parser mode for eslint.

:heavy_check_mark: Green Zenko smoke tests - https://eve.devsca.com/github/scality/zenko/#/builders/4/builds/22418

### DIFF
--- a/lib/BackbeatProducer.js
+++ b/lib/BackbeatProducer.js
@@ -9,107 +9,73 @@ const { withTopicPrefix } = require('./util/topic');
 const KafkaBacklogMetrics = require('./KafkaBacklogMetrics');
 const Constants = require('./constants');
 
-// waits for an ack for messages
-const REQUIRE_ACKS = 'all';
-// time in ms. to wait for acks from Kafka
-const ACK_TIMEOUT = 5000;
-// max time in ms. to wait for flush to complete
+const DEFAULT_POLL_INTERVAL = 2000;
+const PRODUCER_MESSAGE_MAX_BYTES = 5000020;
 const FLUSH_TIMEOUT = 5000;
 
-/**
-* Backbeat replication topic is currently setup with a config
-* max.message.bytes of 5MB. Producers need to update their
-* messageMaxBytes to get at least 5MB put in the Kafka topic, adding a
-* little extra bytes of padding for approximation.
-*/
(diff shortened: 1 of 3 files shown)


lib/BackbeatProducer.js
```diff
@@ -9,110 +9,75 @@ const { withTopicPrefix } = require('./util/topic');
...
-            topic: joi.string(),
-            keyedPartitioner: joi.boolean().default(true),
-            pollIntervalMs: joi.number().default(2000),
-            messageMaxBytes: joi.number().default(PRODUCER_MESSAGE_MAX_BYTES),
-        });
-        const validConfig = joi.attempt(config, configJoi,
-                                        'invalid config params');
-        const { kafka, topic, pollIntervalMs, messageMaxBytes } = validConfig;
+        const validConfig = joi.attempt(
+            config, this.getConfigJoi(),
```
miniscruff:
```suggestion
            config, 
            this.getConfigJoi(),
```
Kinda hard to see there are 3 params here instead of two when two are on one line

miniscruff: Was this missed?

jbertran: Sorry, I haven't been able to get back to this until today :sweat:

KillianG (approved):

ghost (approved):

PR #2233, opened 2022-06-16
improvement: BB-203 cold storage archive topic

naren-rajendran:
publish messages to a location specific archive topic if the target is a cold location.

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -254,7 +254,8 @@
                     "bindAddress": "0.0.0.0",
                     "port": 8554
                 }
-            }
+            },
+            "coldStorageArchiveTopicPrefix": "cold-archive-req-"
         },
         "gc": {
             "topic": "backbeat-gc",
--- a/extensions/replication/ReplicationAPI.js
+++ b/extensions/replication/ReplicationAPI.js
@@ -1,9 +1,11 @@
 const config = require('../../lib/Config');
+const locations = require('../../conf/locationConfig.json') || {};
 
 const ActionQueueEntry = require('../../lib/models/ActionQueueEntry');
 const ReplicationMetrics = require('./ReplicationMetrics');
 
 let { dataMoverTopic } = config.extensions.replication;
+const { coldStorageArchiveTopicPrefix } = config.extensions.lifecycle;
 
 class ReplicationAPI {
     /**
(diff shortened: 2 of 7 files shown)


naren-rajendran:
Adding tests separately, if necessary.

extensions/replication/ReplicationAPI.js
```diff
@@ -67,23 +69,31 @@ class ReplicationAPI {
...
-        const kafkaEntries = [{ key: `${bucket}/${key}`,
-                                message: action.toKafkaMessage() }];
-        producer.sendToTopic(dataMoverTopic, kafkaEntries, (err, reports) => {
+        const toLocation = action.getAttribute('toLocation');
+        const locationConfig = locations[toLocation];
+        let topic = dataMoverTopic;
+        if (locationConfig && locationConfig.isCold) {
+            topic = `${constants.coldStorageArchiveTopicPrefix}-${toLocation}`;
+        }
+        const kafkaEntries = [{
```
alexanderchan-scality: the `archive` request message should have a different format.
looking at https://github.com/scality/sorbet/blob/a017505e8e52f18403d24af20ef86761b64c0c72/forwarder/kafka_forwarder.go#L46


it should be something like: 
```
key: 
value: {
  bucketName,
  objectKey,
  objectVersion,
  size,
  requestId,
}
```

@RachedBenMustapha can you please confirm this

alexanderchan-scality (changes requested):

extensions/replication/ReplicationAPI.js
rachedbenmustapha: Confirmed. And to be future proof (I'm currently working on AssumeRole during the processing of these messages) we'll need the `accountId` as well.

extensions/replication/ReplicationAPI.js
```diff
@@ -67,23 +69,31 @@ class ReplicationAPI {
...
     static sendDataMoverAction(producer, action, log, cb) {
         const { bucket, key } = action.getAttribute('target');
-        const kafkaEntries = [{ key: `${bucket}/${key}`,
-                                message: action.toKafkaMessage() }];
-        producer.sendToTopic(dataMoverTopic, kafkaEntries, (err, reports) => {
+        const toLocation = action.getAttribute('toLocation');
+        const locationConfig = locations[toLocation];
+        let topic = dataMoverTopic;
+        if (locationConfig && locationConfig.isCold) {
+            topic = `${constants.coldStorageArchiveTopicPrefix}-${toLocation}`;
```
rachedbenmustapha: I'd be more comfortable having the `-` part of the prefix, and the prefix in configuration instead of constant.

extensions/replication/ReplicationAPI.js
```diff
@@ -67,23 +69,31 @@ class ReplicationAPI {
      */
     static sendDataMoverAction(producer, action, log, cb) {
         const { bucket, key } = action.getAttribute('target');
-        const kafkaEntries = [{ key: `${bucket}/${key}`,
-                                message: action.toKafkaMessage() }];
-        producer.sendToTopic(dataMoverTopic, kafkaEntries, (err, reports) => {
+        const toLocation = action.getAttribute('toLocation');
+        const locationConfig = locations[toLocation];
+        let topic = dataMoverTopic;
+        if (locationConfig && locationConfig.isCold) {
```
rachedbenmustapha: An error should be logged if we don't know about the location to transfer to.

extensions/replication/ReplicationAPI.js
```diff
@@ -66,24 +68,46 @@ class ReplicationAPI {
...
+        if (locationConfig.isCold) {
+            topic = `${coldStorageArchiveTopicPrefix}${toLocation}`;
+            const { reqId } = action.getContext();
+            const message = {
+                accountId,
+                bucketName: bucket,
+                objectKey: key,
+                objectVersion: version,
+                requestId: reqId,
+                size: contentLength,
```
alexanderchan-scality, line 95: `size` could be removed. it is not expected by `sorbet`

conf/config.json
```diff
@@ -201,7 +201,8 @@
                     "bindAddress": "localhost",
                     "port": 4045
                 }
-            }
+            },
+            "coldStorageArchiveTopicPrefix": "cold-archive-req-"
```
alexanderchan-scality: this should be set in `lifecycle`?
or it could be defined in constants instead

naren-rajendran: I had it in constants, Rached suggested to have this in config. At the moment, it is being used by ReplicationAPI, should this still be in lifecycle?

alexanderchan-scality: it could possibly be set at the root of the config object. 
I believe the `joi` validator will need to be updated as well

naren-rajendran: moved it under `lifecycle`. should this property be validated?

alexanderchan-scality: yes. it should be validated.

extensions/replication/ReplicationAPI.js
rachedbenmustapha: Actually yes it's expected, the TLP server behind sorbet needs it for scheduling.

extensions/replication/ReplicationAPI.js
```diff
@@ -66,24 +68,46 @@ class ReplicationAPI {
...
+            topic = `${coldStorageArchiveTopicPrefix}${toLocation}`;
+            const { reqId } = action.getContext();
+            const message = {
+                accountId,
+                bucketName: bucket,
+                objectKey: key,
+                objectVersion: version,
+                requestId: reqId,
+            };
+            kafkaEntries[0].message = JSON.stringify(message);
```
rachedbenmustapha: Instead of reaching inside the array and assuming its size, just create `kafkaEntry` instead and send `[kafkaEntry]`.

extensions/replication/ReplicationAPI.js
```diff
@@ -66,24 +68,46 @@ class ReplicationAPI {
...
+        }
+        if (locationConfig.isCold) {
+            topic = `${coldStorageArchiveTopicPrefix}${toLocation}`;
+            const { reqId } = action.getContext();
+            const message = {
+                accountId,
+                bucketName: bucket,
+                objectKey: key,
+                objectVersion: version,
+                requestId: reqId,
```
rachedbenmustapha, line 93: Please add `size` back in.

rachedbenmustapha (changes requested):
Tests would be good, although maybe in another PR as this needs to move forward.

extensions/replication/ReplicationAPI.js
alexanderchan-scality: yes. I was wrong and missed the size in the `archiveRequest` struct

rachedbenmustapha (approved):
@naren-rajendran please create a tech debt ticket (with jira label `debt`) to not get `contentLength` from the metrics. We don't want any changes in metrics impacting the request path here, as the size is a first class parameter, not just an "observation to be exposed".

alexanderchan-scality (approved):

alexanderchan-scality:
@RachedBenMustapha 
question: as the `contentLength` from metrics are obtained from `objectMD`, would  it still  be considered a `debt`?
* https://github.com/scality/backbeat/blob/c997d12c51bfb1d188642dce5da5652f5f6ae77c/extensions/replication/ReplicationAPI.js#L50
* https://github.com/scality/backbeat/blob/c997d12c51bfb1d188642dce5da5652f5f6ae77c/extensions/lifecycle/tasks/LifecycleTask.js#L826

rachedbenmustapha:
@alexanderchan-scality yes it's still debt, it's part of the actual request data, not derived for metrics so we have an inversion of control here.

PR #2234, opened 2022-06-16
feature: BB-204 update expiration task

miniscruff:


extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
```diff
@@ -0,0 +1,125 @@
...
+     */
+    constructor(proc) {
+        const procState = proc.getStateVars();
+        super();
+        Object.assign(this, procState);
+    }
+
+    _getMetadata(entry, log, done) {
+        const { bucket, key, version } = entry.getAttribute('target');
+        this.backbeatClient.getMetadata({
```
alexanderchan-scality: there should be a call here to get an instance of the client. ex. https://github.com/scality/backbeat/blob/development/8.4/extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js#L25-L26

alexanderchan-scality (changes requested):

extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
```diff
@@ -0,0 +1,143 @@
...
+            log.error('failed to get backbeat client', {
+                canonicalId,
+                accountId,
+            });
+            return done(errors.InternalError.customizeDescription(
+                'Unable to obtain client',
+            ));
+        }
+
+        this.backbeatClient.getMetadata({
```
alexanderchan-scality:
```suggestion
        backbeatClient.getMetadata({
```

extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
```diff
@@ -0,0 +1,143 @@
...
+                    'error parsing metadata blob'
+                ));
+            }
+            return done(null, res.result);
+        });
+    }
+
+    _putMetadata(entry, objMD, log, done) {
+        const { bucket, key, version } = entry.getAttribute('target');
+        this.backbeatClient.putMetadata({
```
alexanderchan-scality: same as the previous change. the backbeatClient will need in declared

alexanderchan-scality (approved):

extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
```diff
@@ -0,0 +1,169 @@
...
+            versionId: 'target.version',
+        });
+
+        async.waterfall([
+            next => this._getMetadata(entry, log, next),
+            (objMD, next) => {
+                const archive = objMD.getArchive();
+                objMD.setArchive({
+                    archiveInfo: archive.archiveInfo,
+                });
```
francoisferrand, line 159: this code looks/reads like a NOOP, while it does the reset all the "time" fields : which I imagine is what is actually useful here....

so i thing this code should be made more explicit about what it does : at minimum, adding a callback expliciting that we are resetting these fields ; or better yet introduce a function whose goal is exactly this...

For exemple, something like this (where `resetRestoration()` may actually have the same implementation you put here) would make it much easier to read IMHO:

```suggestion
                const archive = objMD.getArchive();
                archive.resetRestoration();
```

francoisferrand (approved):

extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
miniscruff: I could do that in a second pass as the object MD is part of Arsenal. 
It could be possible to do `objMD.resetRestoration();` directly instead of getting the archive first.

PR #2235, opened 2022-06-17
Feature: BB-205 add probe server to bucket notification

welansari:
- Start probe server for notification processor & populator
- Add liveness and readiness handler
- Fix notification queuePopulatorExtension export

### DIFF
--- a/bin/notification.js
+++ b/bin/notification.js
@@ -12,9 +12,10 @@ const rConfig = config.redis;
 const vConfig = config.vaultAdmin;
 const qpConfig = config.queuePopulator;
 const extConfigs = config.extensions;
-
 const QueuePopulator = require('../lib/queuePopulator/QueuePopulator');
-
+const { startProbeServer } = require('../lib/util/probe');
+const { DEFAULT_LIVE_ROUTE, DEFAULT_METRICS_ROUTE, DEFAULT_READY_ROUTE } =
+    require('arsenal').network.probe.ProbeServer;
 const log = new werelogs.Logger('Backbeat:NotificationQueuePopulator');
 
 werelogs.configure({
@@ -49,6 +50,28 @@ const queuePopulator = new QueuePopulator(
     zkConfig, kafkaConfig, qpConfig, httpsConfig, mConfig, rConfig, vConfig, extConfigs);
 
 async.waterfall([
+    done => startProbeServer(qpConfig.probeServer, (err, probeServer) => {
+        if (err) {
+            log.error('error starting probe server', {
+                error: err,
+                method: 'NotificationQueuePopulator::startProbeServer',
+            });
+            done(err);
+            return;
+        }
+        if (probeServer !== undefined) {
+            probeServer.addHandler([DEFAULT_LIVE_ROUTE, DEFAULT_READY_ROUTE],
+                (res, log) => queuePopulator.handleLiveness(res, log)
+            );
+            if (process.env.ENABLE_METRICS_PROBE === 'true') {
+                probeServer.addHandler(
+                    DEFAULT_METRICS_ROUTE,
+                    (res, log) => queuePopulator.handleMetrics(res, log)
+                );
+            }
+        }
+        done();
+    }),
     done => queuePopulator.open(done),
     done => {
         const taskState = {
--- a/extensions/notification/queueProcessor/QueueProcessor.js
+++ b/extensions/notification/queueProcessor/QueueProcessor.js
@@ -107,16 +107,17 @@ class QueueProcessor extends EventEmitter {
      * @param {boolean} [options.disableConsumer] - true to disable
      *   startup of consumer (for testing: one has to call
      *   processQueueEntry() explicitly)
+     * @param {function} done callback
      * @return {undefined}
      */
-    start(options) {
+    start(options, done) {
         async.series([
             next => this._setupNotificationConfigManager(next),
             next => this._setupDestination(this.destinationConfig.type, next),
             next => this._destination.init(() => {
                 if (options && options.disableConsumer) {
                     this.emit('ready');
-                    return undefined;
+                    return next();
                 }
                 const { groupId, concurrency }
                     = this.notifConfig.queueProcessor;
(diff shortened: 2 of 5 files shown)


extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -139,19 +140,20 @@ class QueueProcessor extends EventEmitter {
                     this.logger.info('queue processor is ready to consume ' +
                         'notification entries');
                     this.emit('ready');
+                    return next();
```
ghost, line 151: In case of error above (line 137), I don't see any call to `next(err)`, is it expected?

welansari: ~~good catch ! yes we do need to call the callback when getting an error~~

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -133,25 +134,29 @@ class QueueProcessor extends EventEmitter {
                     concurrency,
                     queueProcessor: this.processKafkaEntry.bind(this),
                 });
-                this._consumer.on('error', () => { });
+                this._consumer.on('error', err => {
+                    this.logger.error('error starting notification consumer',
+                    { method: 'QueueProcessor.start', error: err.message });
+                });
```
ghost: I still cannot see the call to the `next` function :slightly_smiling_face:

extensions/notification/queueProcessor/QueueProcessor.js
welansari: Actually we might not want to do that, as the `error` event is also emitted when an error occurs while reading a Kafka message, and we don't want to crash the whole thing for just one message that we couldn't read.

And in the case where the backbeatConsumer fails to start, the pod is going to crash and will be restarted anyway as it is never going to get ready

extensions/notification/queueProcessor/QueueProcessor.js
welansari: yep commented above ^

extensions/notification/queueProcessor/QueueProcessor.js
ghost: But in this case, how are we supposed to end the function, if this event is called? Do you mean that the `ready` event will be always called, or, in case of error, the service will crash until a stable state is reached?

ghost: I mean, if we get the `error` event and `ready` is never called, you will never reach the end of your async.series, as the `next` will never be called.

welansari: Yes i meant in case of error the service will crash, but fair enough the function has to end if it fails at startup to make things cleaner

bin/notification.js
```diff
@@ -48,6 +49,28 @@ const queuePopulator = new QueuePopulator(
...
 
 async.waterfall([
+    done => startProbeServer(qpConfig.probeServer, (err, probeServer) => {
+        if (err) {
+            log.error('error starting probe server', {
+                error: err,
+                method: 'NotificationQueuePopulator::startProbeServer',
+            });
+            done(err);
+            return;
```
ghost, line 60: Nit:
```suggestion
            return done(err);
```

bin/notification.js
```diff
@@ -48,6 +49,28 @@ const queuePopulator = new QueuePopulator(
...
+                (res, log) => queuePopulator.handleLiveness(res, log)
+            );
+            if (process.env.ENABLE_METRICS_PROBE === 'true') {
+                probeServer.addHandler(
+                    DEFAULT_METRICS_ROUTE,
+                    (res, log) => queuePopulator.handleMetrics(res, log)
+                );
+            }
+        }
+        done();
```
ghost, line 73: To be consistent and return something in both cases (err or not err)
```suggestion
        return done();
```

ghost (approved):
LGTM, just two small code improvements you can do

bin/notification.js
miniscruff: I disagree on this one. Returning a callback can have weird side affects where we return a value when no value is expected.

ghost: I am curious what's the difference between this and when we usually return the `next` function in `async.waterfall` calls? Why can we return next in the latter but not here?

miniscruff: Well in the async waterfall or any other of the async chains, async handles the return value for us and likely just ignores it completely anyway making it safer. But when we return the result of the async waterfall or any other callback, we are leaking a return value that we have no clue where it is going or used. It may get assigned to a value, it may be incorrectly used as a normal response or anything else. It is misleading by saying our functions return undefined, when they don't. And the normal expectation that when we pass in a callback, we have no return value is not followed, so debugging would be harder if you didn't know we return callback values all over.

For the record, I don't think we should return next either, but since the immediate caller is going to ignore it anyway its not as bad. Or in cases where the function itself is guaranteed to return undefined anyway its not as bad. However, those two additional questions are two more steps to making sure our code does what it is supposed to do over just using two lines making it 100% clear without extra steps. I tend not to comment the suggestion at this point though just cause it is used so much in our code base that it would feel weird going against that now. But I do still disagree with it.

ghost: Thanks for the detailed answer, very interesting! One more argument to switch to async/await :smirk:

miniscruff: Yea, async await would remove this option as we need the return values of course.

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -250,6 +260,15 @@ class QueueProcessor extends EventEmitter {
...
         }
     }
+
+    /**
+     * Checks if queue processor is ready to consume
+     *
+     * @returns {boolean} is queue processor ready
+     */
+    isReady() {
+        return this._consumer && this._consumer.isReady();
```
francoisferrand, line 270: could use the shortcut operator:

```suggestion
        return this._consumer?.isReady();
```

extensions/notification/queueProcessor/task.js
```diff
@@ -14,33 +22,79 @@ werelogs.configure({
...
 }
+
+async.series([
+    next => queueProcessor.start(null, next),
+    next => startProbeServer(notifConfig.probeServer, (err, probeServer) => {
+        if (err) {
+            log.error('error starting probe server', { error: err });
+            return next(err);
+        }
+        if (probeServer !== undefined) {
```
francoisferrand, line 71: can the opposite (probeServer == undefined) happen: it seems to me that should be an error? or is there a way to "disable" probe server from the config?

nicolas2bert (approved):
LGTM

extensions/notification/queueProcessor/task.js
miniscruff: This was a valid path due to 7.x having probe servers being opt-in env var values. In 8.x it is default on but iirc can be disabled with an env var. If it is disabled undefined is returned that is why this is here.

extensions/notification/queueProcessor/QueueProcessor.js
welansari: the linter version used in backbeat is a bit old, it doesn't like when i write in that syntax

PR #2245, opened 2022-06-20
Feature: BB-210 handle readiness liveness oplogPopulator

welansari:


extensions/oplogPopulator/OplogPopulatorTask.js
```diff
@@ -1,9 +1,21 @@
...
 const OplogPopulator = require('./OplogPopulator');
 
+const {
+    DEFAULT_LIVE_ROUTE,
+    DEFAULT_READY_ROUTE,
+} = require('arsenal').network.probe.ProbeServer;
+const { sendSuccess, sendError } = require('arsenal').network.probe.Utils;
+const { startProbeServer } = require('../../lib/util/probe');
+
+const startProbeServerAsync = util.promisify(startProbeServer);
```
miniscruff: Would you mind adding this util to the /lib/util/probe file so other probe servers can use it?
Note: Not sure if we should use the async suffix or a promise suffix, but picking one and using it across the board is ideal

extensions/oplogPopulator/OplogPopulatorConfigValidator.js
```diff
@@ -1,11 +1,13 @@
 const joi = require('joi');
+const { probeServerJoi } = require('../../lib/config/configItems.joi.js');
```
jbertran: We can let node work out the extension.

```suggestion
const { probeServerJoi } = require('../../lib/config/configItems.joi');
```

jbertran (approved):

lib/util/probe.js
```diff
@@ -33,6 +34,14 @@ function startProbeServer(config, callback) {
     probeServer.start();
 }
 
+/**
+ * Start probe server for Queue Processor
+ * @param {ProbeServerConfig} config - Configuration for probe server
+ * @returns {Promise|ProbeServer} Probe server or undefined if disabled
+ */
+const startProbeServerPromise = util.promisify(startProbeServer);
```
francoisferrand, line 42: not sure introducing this method makes the code more readable...

A better way may be to update the `startProbeServer` function to return a promise when no callback is provided:

```
// @return {Promise|undefined}
function startProbeServer(config, callback) {
   if (callback == undefined)
     return util.promisify(startProbeServer)(config);

  // else return legacy code
  [...]
}

francoisferrand (approved):

PR #2246, opened 2022-06-20
Feature/bb 193/cold storage transitioned garbage collection

alexanderchan-scality:


extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -64,6 +95,33 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
+                task = new LifecycleColdStatusArchiveTask(this);
+                break;
+            default:
+                return process.nextTick(done);
+        }
+
+        return this.retryWrapper.retry({
+            actionDesc: 'process cold storage status entry',
+            logFields: entry.getLogInfo(),
+            actionFunc: done => task.processColdStorageStatusEntry(entry, done),
```
nicolas2bert:
```suggestion
            actionFunc: done => task.processEntry(entry, done),
```

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -64,6 +95,33 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
         return new LifecycleUpdateTransitionTask(this);
     }
+
+    processColdStorageStatusEntry(kafkaEntry, done) {
+        const entry = ColdStorageQueueEntry.createFromKafkaEntry(kafkaEntry);
+        if (entry.error) {
+            this._log.error('malformed status entry', kafkaEntry.value);
+            return process.nextTick(done);
+        }
+        this._log.debug('processing cold storage entry entry', entry.getLogInfo());
```
nicolas2bert:
```suggestion
        this._log.debug('processing cold storage entry', entry.getLogInfo());
```

nicolas2bert (approved):

lib/models/ColdStorageActionQueueEntry.js
```diff
@@ -0,0 +1,91 @@
...
+const entrySchema = joi.object({
+    op: joi.string().valid('restore', 'delete', 'archive').required(),
+    bucketName: joi.string().required(),
+    objectKey: joi.string().required(),
+    objectVersion: joi.string().required(),
+    accountId: joi.string().required(),
+    archiveInfo: joi.object({
+        archiveId: joi.string().required(),
+        archiveVersion: joi.string().required(),
+    }).required(),
```
rachedbenmustapha: Please make `archiveId` opaque. It should never be looked into, as it's handled by the TLP implementation and could change depending on backend (tape, azure archive, glacier, ...).

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -0,0 +1,75 @@
...
+        return {
+            bucket: bucketName,
+            objectKey,
+            version: objectVersion,
+            storageClass: entry.get('sourceStorageClass'),
+            lastModified: entry.get('sourceLastModified'),
+        };
+    }
+
+    _updateMdWithArchiveInfo(entry, objMD) {
```
rachedbenmustapha: Should probably mention in the name that it clears all archive attributes except archiveinfo

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -0,0 +1,75 @@
...
+        return async.waterfall([
+            next => this._getMetadata(entry, log, next),
+            (objMD, next) => {
+                locationToGC = objMD.getLocation();
+                entry.set('sourceStorageClass', objMD.getDataStoreName());
+                entry.set('sourceLastModified', objMD.getLastModified());
+                this._updateMdWithArchiveInfo(entry, objMD);
+                return this._putMetadata(entry, objMD, log, next);
+            },
+            next => {
```
rachedbenmustapha, line 109: This sequence can create orphans if `_garbageCollectLocation` fails (retry won't find any location to gc), and `_garbageCollectLocation` seems to ignore errors.

rachedbenmustapha (changes requested):

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
alexanderchan-scality: update:
* add delete method without relying on existing GC processor
* split the update md into 2 parts; archive update and location removal update
* new sequence:
  * get objectMD
  * update objectMD archive info and write to db
  * perform location delete
    * on success, unset objectMD location and write to db
    * on error, return error and retry

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -0,0 +1,147 @@
...
+            objectKey,
+            objectVersion,
+            accountId
+        } = entry.getTarget();
+        return {
+            bucket: bucketName,
+            key: objectKey,
+            version: objectVersion,
+            accountId
+        };
```
miniscruff: Quite small change but I would do this:
```suggestion
        const {
            bucketName: bucket,
            objectKey: key,
            objectVersion: version,
            accountId
        } = entry.getTarget();
        return { bucket, key, version, accountId };
```

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    getSourceAttribute(entry) {
+        const { bucketName, objectKey, objectVersion } = entry.getTarget();
```
miniscruff: could spread bucket name and object version to bucket and version variable names directly here

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -0,0 +1,147 @@
...
+            objectKey,
+            version: objectVersion,
+            storageClass: entry.get('sourceStorageClass'),
+            lastModified: entry.get('sourceLastModified'),
+        };
+    }
+
+    /**
+     * delete data in set in ObjectMD location
+     * @param {object} entry - ColdStorageActionQueueEntry
```
miniscruff: I am not sure if JSDoc will properly handle the variable type but at least until we switch to type script this should be close enough.
```suggestion
     * @param {ColdStorageActionQueueEntry} entry - what entry to delete
```
( feel free to fix my description :smirk: )

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -0,0 +1,147 @@
...
+            }),
+            next => {
+                const archiveInfo = entry.getArchiveInfo();
+                // set new ObjectMDArchive to ObjectMD
+                objectMD.setArchive(new ObjectMDArchive(archiveInfo));
+                this._putMetadata(entry, objectMD, log, next);
+            },
+            next => {
+                const locationToGC = objectMD.getLocation();
+                if (!locationToGC || (Array.isArray(locationToGC) && locationToGC.length === 0)) {
```
miniscruff: I don't think the second half of this condition can be true if the first half is not already true since an empty array is also falsy.

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -0,0 +1,147 @@
...
+                    return process.nextTick(next);
+                }
+
+                return this._executeDeleteData(entry, objectMD, this._log, err => {
+                    if (err) {
+                        return next(err);
+                    }
+
+                    log.debug('successfully deleted location data');
```
miniscruff: should we also log some sort of key or id here considering its async it might come out of order and we would have no clue which one was deleted

lib/BackbeatConsumerManager.js
```diff
@@ -0,0 +1,85 @@
...
+     * @param {Object} consumerParams - object containing BackbeatConsumer params
+     * @param {Object} logger - logger
+     */
+    constructor(id, consumerParams, logger) {
+        super();
+        this.id = id;
+        this._log = logger;
+        this._consumerParams = consumerParams;
+        this._consumers = {};
+        // this._consumersCount = {};
```
miniscruff: commented code

lib/BackbeatConsumerManager.js
```diff
@@ -0,0 +1,85 @@
...
+    setupConsumers(cb) {
+        async.each(
+            Object.keys(this._consumerParams),
+            (id, done) => this._setupConsumer(id, this._consumerParams[id], done),
+            err => {
+                if (err) {
+                    return cb(err);
+                }
+                return cb();
+            });
```
miniscruff:
```suggestion
            cb,
        );
```
We can just call cb directly as we effectively already have this if err check in the setup consumer method

lib/BackbeatConsumerManager.js
```diff
@@ -0,0 +1,85 @@
...
+        return Object.keys(this._consumers).every(id => this._consumers[id].isReady());
+    }
+
+    _setupConsumer(consumerId, params, cb) {
+        let consumerReady = false;
+        const consumer = new BackbeatConsumer(params);
+        this._consumers[consumerId] = consumer;
+
+        consumer.on('error', err => {
+            if (!consumerReady) {
```
miniscruff: personally, I would use a guard clause here but its not too big a difference

lib/models/ColdStorageActionQueueEntry.js
```diff
@@ -0,0 +1,88 @@
...
+
+    get(key) {
+        return this._attributes[key];
+    }
+
+    getOp() {
+        return this._attributes.op;
+    }
+
+    getArchiveInfo() {
```
miniscruff: I feel like we should use getters for this sort of stuff https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

lib/models/ColdStorageActionQueueEntry.js
```diff
@@ -0,0 +1,88 @@
...
+        return this._attributes.objectKey;
+    }
+
+    getObjectVersion() {
+        return this._attributes.objectVersion;
+    }
+
+    getTarget() {
+        return {
+            bucketName: this._attributes.bucketName,
```
miniscruff: should we use the get functions here instead of the attributes directly?

tests/functional/lifecycle/LifecycleObjectProcessor.spec.js
```diff
@@ -34,13 +34,16 @@ describe('Lifecycle Object Processor', function lifecycleObjectProcessor() {
                     return done(err);
                 }
 
-                lop._consumer.onEntryCommittable = () => {
+                const objectTaskConsumer =
+                    lop._consumers.getConsumer(lcConfig.objectTasksTopic);
```
miniscruff, line 38: Not sure the best way to change this but using the `_consumers` directly feels off, but maybe it shouldn't be public so nothing we can do

miniscruff (changes requested):

lib/BackbeatConsumerManager.js
```diff
@@ -0,0 +1,81 @@
...
+        async.each(
+            Object.keys(this._consumers),
+            (id, done) => this._consumers[id].close(done),
+            err => {
+                if (err) {
+                    this._log.error('unable to stop consumers');
+                    return cb(err);
+                }
+                return cb();
+            });
```
miniscruff:
```suggestion
                if (err) {
                    this._log.error('unable to stop consumers');
                }
                cb(err);
            });
```

miniscruff (approved):

lib/models/ColdStorageActionQueueEntry.js
```diff
@@ -0,0 +1,68 @@
...
+    op: joi.string().valid('restore', 'delete', 'archive').required(),
+    bucketName: joi.string().required(),
+    objectKey: joi.string().required(),
+    objectVersion: joi.string().required(),
+    accountId: joi.string().required(),
+    archiveInfo: joi.object().required(),
+    requestId: joi.string().required(),
+});
+
+class ColdStorageActionQueueEntry {
```
rachedbenmustapha: This seems to describe the result of an action rather than an action, which was confusing me.

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * delete data in set in ObjectMD location
+     * @param {ColdStorageActionQueueEntry} entry - cold storage expire task
+     * @param {ObjectMD} objMD - object metadata
+     * @param {object} log - logger
+     * @param {function} done - callback
+     * @return {undefined}
+     */
+    _executeDeleteData(entry, objMD, log, done) {
```
rachedbenmustapha, line 39: I'm not sure it's a good idea to duplicate this non-trivial code with gc-specific business logic (which makes the gc extension not have any unique logic anymore).

rachedbenmustapha (changes requested):

lib/models/ColdStorageActionQueueEntry.js
alexanderchan-scality: will rename to be more specific to cold storage status as opposed to trying to be a general cold storage name

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
alexanderchan-scality: I wanted to localize the deletion to the transition processor; it didn't seem like a benefit to have the deletion be asynchronous  from the current expire task.

the alternative would to add additional information to the GC task messages such that a different action would be taken on completion or failure; it would however require additional changes in GC to introduce task retries, which could introduce some issues, so I had opted for this at the moment

rachedbenmustapha: If it's the shortest path to supporting archive right now let's go this way for now, but with a `debt` ticket in epic ARTESCA-3398 so it doesn't stay like this for too long.

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -2,9 +2,13 @@
 
 const Logger = require('werelogs').Logger;
 
+
+const ColdStorageStatusQueueEntry = require('../../../lib/models/ColdStorageStatusQeueuEntry');
```
rachedbenmustapha:
```suggestion
const ColdStorageStatusQueueEntry = require('../../../lib/models/ColdStorageStatusQueueEntry');
```

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
alexanderchan-scality: had taken a step back and looked over the current implementation and completely missed a benefit to having the gc handle the deletion task.
there is a cap on the retry, so if that is hit the garbage collection info will be lost and orphaned data created.

I would be more confident if the changes to GC are done now.

alexanderchan-scality: ticket tracking the needed changes to be made in GC: https://scality.atlassian.net/browse/BB-218

rachedbenmustapha (approved):

PR #2248, opened 2022-06-22
BB-193: add queue populator credentials

alexanderchan-scality:


extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -24,16 +20,15 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
-        } else {
-            this.stsAgent = new http.Agent({ keepAlive: true });
-        }
-        this.credentialsManager = new CredentialsManager('lifecycle', this.logger);
-        this._storeAWSCredentialsPromise();
+        this.vaultClientWrapper = new VaultClientWrapper(
+            LIFEYCLE_POPULATOR_CLIENT_ID,
+            params.vaultAdmin.host,
+            params.vaultAdmin.port,
+            this._autConfig,
```
miniscruff:
```suggestion
            this._authConfig,
```

extensions/utils/VaultClientWrapper.js
```diff
@@ -0,0 +1,145 @@
...
+        this._clientId = id;
+        this._vaultHost = host;
+        this._vaultPort = port;
+        this.logger = logger;
+
+        if (this._transport === 'https') {
+            this.stsAgent = new https.Agent({ keepAlive: true });
+        } else {
+            this.stsAgent = new http.Agent({ keepAlive: true });
+        }
```
miniscruff: This is quite nit picky and since we only have keep alive as an option may not be much better but we can do.
```suggestion
        const Agent = this._transport === 'https' ? https.Agent : http.Agent;
        this.stsAgent = new Agent({ keepAlive: true });
```

extensions/utils/VaultClientWrapper.js
```diff
@@ -0,0 +1,145 @@
...
+
+        this._tempCredsPromiseResolved = false;
+        this._storeAWSCredentialsPromise();
```
miniscruff: I think we should move this method call into the init method as it is part of the init process and is sort of like a pre-init as is

miniscruff (changes requested):

miniscruff (approved):

naren-rajendran (approved):

PR #2250, opened 2022-06-27
Feature: BB-213 fix oplog populator race condition

welansari:
A race condition occurs when multiple events are processed at the same time
and that affect the same connector, in which case the connector is not configured
correctly.

For example when two buckets are being added to the same connector at the same time,
only one of them gets added, this is caused by the two operations getting and updating
the list of buckets assigned to the connector at the same time, which results in one of
them overwriting the other.

The solution is using a scheduler that periodically updates all of the connectors.
Connectors are no longer updated on an event basis

### DIFF
--- a/extensions/oplogPopulator/OplogPopulatorConfigValidator.js
+++ b/extensions/oplogPopulator/OplogPopulatorConfigValidator.js
@@ -8,6 +8,7 @@ const joiSchema = joi.object({
     numberOfConnectors: joi.number().required().min(1),
     prefix: joi.string().optional(),
     probeServer: probeServerJoi.default(),
+    connectorsUpdateCronRule: joi.string().default('*/1 * * * * *'),
 });
 
 function configValidator(backbeatConfig, extConfig) {
--- a/extensions/oplogPopulator/modules/Connector.js
+++ b/extensions/oplogPopulator/modules/Connector.js
@@ -37,6 +37,18 @@ class Connector {
         this._name = params.name;
         this._config = params.config;
         this._buckets = new Set(params.buckets);
+        this._state = {
+            // Used to check if buckets assigned to this connector
+            // got modified from the last connector update
+            bucketsGotModified: true,
+            // Used to avoid concurrency issues when updating the state.
+            // the state value gets updated to false only when the connector
+            // update is successful. And because updating the connector is an
+            // asynchronous operation, multiple updates could of happened in the
+            // mean time, so we only set to false when no other update happened.
+            lastUpdated: Date.now(),
+            isUpdating: false,
+        };
         this._logger = params.logger;
         this._kafkaConnect = new KafkaConnectWrapper({
             kafkaConnectHost: params.kafkaConnectHost,
@@ -109,13 +121,14 @@ class Connector {
      * Add bucket to this connector
      * Connector is updated with the new bucket list
      * @param {string} bucket bucket to add
-     * @param {boolean} [doUpdate=true] updates connector if true
+     * @param {boolean} [doUpdate=false] updates connector if true
      * @returns {Promise|undefined} undefined
      * @throws {InternalError}
      */
-    async addBucket(bucket, doUpdate = true) {
+    async addBucket(bucket, doUpdate = false) {
+        this._buckets.add(bucket);
+        this._updateConnectorState(true);
         try {
-            this._buckets.add(bucket);
             await this.updatePipeline(doUpdate);
         } catch (err) {
             this._logger.error('Error while adding bucket to connector', {
@@ -132,13 +145,14 @@ class Connector {
      * Remove bucket from this connector
      * Connector is updated with new bucket list
      * @param {string} bucket bucket to add
-     * @param {boolean} [doUpdate=true] updates connector if true
+     * @param {boolean} [doUpdate=false] updates connector if true
      * @returns {Promise|undefined} undefined
      * @throws {InternalError}
      */
-    async removeBucket(bucket, doUpdate = true) {
+    async removeBucket(bucket, doUpdate = false) {
+        this._buckets.delete(bucket);
+        this._updateConnectorState(true);
         try {
-            this._buckets.delete(bucket);
             await this.updatePipeline(doUpdate);
         } catch (err) {
             this._logger.error('Error while removing bucket from connector', {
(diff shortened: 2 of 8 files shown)


extensions/oplogPopulator/OplogPopulatorConfigValidator.js
```diff
@@ -8,6 +8,7 @@ const joiSchema = joi.object({
     numberOfConnectors: joi.number().required().min(1),
     prefix: joi.string().optional(),
     probeServer: probeServerJoi.default(),
+    connectorsUpdateCronRule: joi.string().default('*/5 * * * * *'),
```
francoisferrand: every 5 minutes is a long time : we may miss lots of events  :-/

From what i understand, the problem is that we get the list of connectors every time : just getting the list once on startup (before listening to mongo events) should fix the problem as well, without introducing 5 minute delay?

Then, we may introduce a _short_ delay, to avoid too many reconfig : like once we get an update, wait a few seconds before applying it (so that we apply everything that happened since)...

welansari: I actually used the same value used for all backbeat recurring tasks 5 seconds.

The issue is not in getting the list of connectors but the list of buckets assigned to a connector, this can not be done once as the values change with each update.

So with each new bucket to add (or remove) from a connector we have to get that list of buckets and update it.

When having multiple updates at the same time, the update tasks might get the wrong list (not updated), add the task's bucket and then configure the connector. This can lead to not taking in consideration the other updates.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -245,6 +249,35 @@ class ConnectorsManager {
...
+    /**
+     * Schedules connector updates
+     * @returns {undefined}
+     */
+    scheduleConnectorUpdates() {
+        schedule.scheduleJob(this._cronRule, async () => {
+            const connectorsStatus = {};
+            await async.eachLimit(this._connectors, 10, async connector => {
+                try {
+                    connectorsStatus[connector.name] = {
```
jbertran: This does not have to be inside the `try` block.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -245,6 +249,35 @@ class ConnectorsManager {
...
+                } catch (err) {
+                    this._logger.error('Failed to updated connector', {
+                        method: 'ConnectorsManager.scheduleConnectorUpdates',
+                        connector: connector.name,
+                        bucketCount: connector.bucketCount,
+                        error: err.description || err.message,
+                    });
+                }
+            });
+            this._logger.info('Successfully updated connectors', {
```
jbertran: This output is a little bit misleading despite the dissimilar log levels.

Can we change the message according to whether all connector updates succeeded, or connectors partially succeeded?

jbertran (changes requested):

jbertran (approved):

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -245,6 +273,42 @@ class ConnectorsManager {
...
+                        method: 'ConnectorsManager.scheduleConnectorUpdates',
+                        connector: connector.name,
+                        bucketCount: connector.bucketCount,
+                        error: err.description || err.message,
+                    });
+                }
+            });
+            const logMessage = connectorUpdateFailed ? 'Partially updated the connectors' :
+                'Successfully updated all the connectors';
+            this._logger.info(logMessage, {
```
ghost: Nit: I think it would make sense to log as `error` if all updates failed (allowing to dump the debug logs as well if an error occurred), a `warn` if some of them failed and an `info` log it it went well.

ghost (commented):
If I understand correctly, while I understand that removing the event-based update of the connector bucket list will help us dealing with concurrent updates/race conditions, I am not 100% sure if it is acceptable to "ignore" changes every 5s for 5s as it may lead to data loss or unwanted writes, if we're unlucky (and we will, in production!).

So I was wondering if a _stream_ approach would help us keeping the events in the right order, as they arrive, and avoid concurrent requests overriding each other? Basically, anything that could converge towards a (100%) safe solution, is this possible?

welansari:
@williamlardier At a second thought i do agree that even 5 seconds, is a long time for the kind of updates we are doing.
A stream based system won't solve the concurrency issue as we are still dealing with async event handlers, i believe we need controlled connector updates.

What i did for now is reduce the delay between updates to one second, but this time we only update the connectors when there are actual changes made to them (buckets added or removed), this should help reduce network calls.

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -109,13 +121,14 @@ class Connector {
...
      * @param {string} bucket bucket to add
-     * @param {boolean} [doUpdate=true] updates connector if true
+     * @param {boolean} [doUpdate=false] updates connector if true
      * @returns {Promise|undefined} undefined
      * @throws {InternalError}
      */
-    async addBucket(bucket, doUpdate = true) {
+    async addBucket(bucket, doUpdate = false) {
+        this._buckets.add(bucket);
+        this._updateConnectorState(true);
```
ghost, line 130: What happen if the `updatePipeline` fails? do we still add the bucket to the list, as it is done before the try/catch?
(I'm curious to know if the approach will be stable in case of errors, so we don't end up in an inconsistent state, but I may lack a bit of context on this one)

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -170,19 +184,54 @@ class Connector {
...
 
+    /**
+     * Handles updating the values of _bucketsGotModified
+     * @param {boolean} bucketsGotModified value of _state.bucketsGotModified
+     * to set
+     * @param {number} timeBeforeUpdate time just before async
+     * connector update operation (only needed when updated connector)
+     * @returns {undefined}
+     */
+    _updateConnectorState(bucketsGotModified, timeBeforeUpdate = null) {
```
ghost: Nit:
```suggestion
    _updateConnectorState(bucketsGotModified, timeBeforeUpdate = 0) {
```

to be consistent with the jsdoc type

ghost (commented):
Thanks for the update, I think it's a better first step until controlled connector updates, maybe. Please find one question and one suggested change before I approve

extensions/oplogPopulator/modules/Connector.js
welansari: Yes the bucket is still added to the list but `_state.bucketsGotModified` stays as true, which will trigger the next update of the connector, so we will basically be retrying until it succeeds. We also get an error log each time a connector fails

welansari: i added this https://github.com/scality/backbeat/compare/17128feaebde3bc87cd6fe8afe642ba5063f0702..a0941c5e9287ea76513da7b7e7104476a3c38bc9 just to be sure the retry gets triggered

ghost (approved):
LGTM, but might be better to wait for senior reviews on this one :slightly_smiling_face:

PR #2251, opened 2022-06-28
BB-216: add logs to improve error understanding

ghost:
‘MongoQueueProcessor is not ready' log is not specific enough to know where the issue comes from.

This log comes from a ProbeServer [liveness check](https://github.com/scality/backbeat/blob/development/8.4/extensions/mongoProcessor/mongoProcessorTask.js#L39) which looks if the MongoQueueProcessor isReady(), which itself comes from a [boolean check](https://github.com/scality/backbeat/blob/development/8.4/lib/BackbeatConsumer.js#L238) of if a BackbeatConsumer isReady() after it is created on start of the MQP. 
This checks if kafkaBacklogMetrics.isReady() ([a check that a zookeeper client is in a connected state](https://github.com/scality/backbeat/blob/development/8.4/lib/KafkaBacklogMetrics.js#L74)) and if the _consumer of the BackbeatConsumer (a KafkaConsumer [object](https://github.com/scality/backbeat/blob/development/8.4/lib/BackbeatConsumer.js#L175)) isConnected() [here](https://github.com/scality/backbeat/blob/development/8.4/lib/BackbeatConsumer.js#L236) .

Proposed solution after exploration: add logs at these other levels to know specifically where the readiness check fails.

### DIFF
--- a/lib/KafkaBacklogMetrics.js
+++ b/lib/KafkaBacklogMetrics.js
@@ -73,9 +73,10 @@ class KafkaBacklogMetrics extends EventEmitter {
             this._zookeeper.getState().code ===
             zookeeper.State.SYNC_CONNECTED.code)) {
             this._log.error('KafkaBacklogMetrics is not ready, zookeeper not connected',
-                { error: this._zookeeper.getState().code });
+                { error: this._zookeeper.getState() });
             return false;
         }
+        this._log.info('KafkaBacklogMetrics is ready, zookeeper connected');
         return true;
     }
 
--- a/lib/KafkaBacklogMetrics.js
+++ b/lib/KafkaBacklogMetrics.js
@@ -69,9 +69,14 @@ class KafkaBacklogMetrics extends EventEmitter {
     }
 
     isReady() {
-        return this._zookeeper &&
+        if (!(this._zookeeper &&
             this._zookeeper.getState().code ===
-            zookeeper.State.SYNC_CONNECTED.code;
+            zookeeper.State.SYNC_CONNECTED.code)) {
+            this._log.error('KafkaBacklogMetrics is not ready, zookeeper not connected',
+                { error: this._zookeeper.getState().code });
+            return false;
+        }
+        return true;
     }
 
     /**
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -234,6 +234,9 @@ class BackbeatConsumer extends EventEmitter {
             cs = true;
         } else {
             cs = this._consumer && this._consumer.isConnected();
+            if (!cs) {
+                this._log.error('KafkaConsumer is not connected');
+            }
         }
         return kMetrics && cs;
     }


lib/KafkaBacklogMetrics.js
```diff
@@ -69,9 +69,13 @@ class KafkaBacklogMetrics extends EventEmitter {
     }
 
     isReady() {
-        return this._zookeeper &&
+        if (!(this._zookeeper &&
             this._zookeeper.getState().code ===
-            zookeeper.State.SYNC_CONNECTED.code;
+            zookeeper.State.SYNC_CONNECTED.code)) {
+            this._log.error('KafkaBacklogMetrics is not ready: zookeeper not connected');
```
ghost: Can we add the actual current code in the log, so that we know the current state of zookeeper?

lib/KafkaBacklogMetrics.js
```diff
@@ -131,7 +135,7 @@ class KafkaBacklogMetrics extends EventEmitter {
 
     _getOffsetZkPath(topic, partition, offsetType, label) {
         const basePath = `${this._getPartitionsOffsetsZkPath(topic)}/` +
-              `${partition}/${offsetType}`;
+            `${partition}/${offsetType}`;
```
francoisferrand: please don't reformat unrelated code, it makes reviewing (and `git blame`, afterwards) more complicated
(not just this chunk, but most of the following chunks as well)

lib/KafkaBacklogMetrics.js
```diff
@@ -171,15 +175,15 @@ class KafkaBacklogMetrics extends EventEmitter {
                 if (err.getCode() === zookeeper.Exception.NO_NODE) {
                     this._log.debug(
                         'requested kafka offset node does not exist', {
-                            topic, partition, offsetType, label,
-                        });
+                        topic, partition, offsetType, label,
```
francoisferrand: this re-indent is actually not correct, the brace introduces an extra "level" on top of the args, and its content should thus indeed be indented

lib/KafkaBacklogMetrics.js
ghost: apologies, didn't see that, I just used an automatic formatter, will fix

ghost: 13ca620341c8dbd1847a882af78bb711ebde126a

lib/KafkaBacklogMetrics.js
ghost: 13ca620341c8dbd1847a882af78bb711ebde126a

lib/KafkaBacklogMetrics.js
ghost: 13ca620341c8dbd1847a882af78bb711ebde126a

lib/KafkaBacklogMetrics.js
```diff
@@ -69,9 +69,14 @@ class KafkaBacklogMetrics extends EventEmitter {
...
-        return this._zookeeper &&
+        if (!(this._zookeeper &&
             this._zookeeper.getState().code ===
-            zookeeper.State.SYNC_CONNECTED.code;
+            zookeeper.State.SYNC_CONNECTED.code)) {
+            this._log.error('KafkaBacklogMetrics is not ready, zookeeper not connected',
+                { error: this._zookeeper.getState().code });
+            return false;
+        }
+        return true;
```
XinLi0207, line 80: do you think we can log `KafkaBacklogMetrics is ready, zookeeper connected` if it's ready?

lib/KafkaBacklogMetrics.js
```diff
@@ -69,9 +69,14 @@ class KafkaBacklogMetrics extends EventEmitter {
     }
 
     isReady() {
-        return this._zookeeper &&
+        if (!(this._zookeeper &&
             this._zookeeper.getState().code ===
-            zookeeper.State.SYNC_CONNECTED.code;
+            zookeeper.State.SYNC_CONNECTED.code)) {
+            this._log.error('KafkaBacklogMetrics is not ready, zookeeper not connected',
+                { error: this._zookeeper.getState().code });
```
miniscruff: I am not sure what else is part of the zookeeper state but it might be helpful to actually see why its failing and worth logging

lib/KafkaBacklogMetrics.js
ghost: this ok? 
786501173242d922bd7bc8375d92b1c731b7495b

lib/KafkaBacklogMetrics.js
ghost: done 786501173242d922bd7bc8375d92b1c731b7495b

lib/KafkaBacklogMetrics.js
ghost: maybe `log.info` instead? Don't know what would be best

XinLi0207: I feel like `log.info` would be better

XinLi0207 (approved):

lib/KafkaBacklogMetrics.js
```diff
@@ -69,9 +69,15 @@ class KafkaBacklogMetrics extends EventEmitter {
     }
 
     isReady() {
-        return this._zookeeper &&
+        if (!(this._zookeeper &&
             this._zookeeper.getState().code ===
-            zookeeper.State.SYNC_CONNECTED.code;
+            zookeeper.State.SYNC_CONNECTED.code)) {
+            this._log.error('KafkaBacklogMetrics is not ready, zookeeper not connected',
+                { error: this._zookeeper.getState().code, state: zookeeper.State });
```
ghost:
```suggestion
                { state: this._zookeeper.getState() });
```

No need to log the State variable I guess, it will contain everything, instead, you can log the current full state and remove the `code` as it will be part of the payload.

ghost: 476b6a5c5e0437a6d8354436b3ecc94167282bf4

ghost (approved):

miniscruff (approved):

ghost:
After a final check to see if these new logs will flood the terminal in case of disconnection, I saw no issues. Will merge.

ghost:
ping

PR #2252, opened 2022-06-29
Feature: BB-219 update mongo connector config

welansari:
Issue: [BB-219](https://scality.atlassian.net/browse/BB-219)
- Configure Kafka-connect MongoDB connector to only include bucket name in the Kafka message key. This will make events of the same bucket always land in the same partition as they will have the same key
- Configure Kafka-connect MongoDB connector to use an output string converter to avoid getting an over stringified JSON

### DIFF
--- a/extensions/oplogPopulator/constants.js
+++ b/extensions/oplogPopulator/constants.js
@@ -6,6 +6,43 @@ const constants = {
         'change.stream.full.document': 'updateLookup',
         'pipeline': '[]',
         'collection': '',
+        // Kafka message key config
+        // The message key is set to only contain the bucket where the event happend.
+        // This will make events of the same bucket always land in the same partition
+        // as they will have the same key
+        'output.format.key': 'schema',
+        'output.schema.key': JSON.stringify({
+            type: 'record',
+            name: 'keySchema',
+            fields: [{
+                name: 'ns',
+                type: [{
+                        name: 'ns',
+                        type: 'record',
+                        fields: [{
+                            name: 'coll',
+                            type: ['string', 'null'],
+                        }],
+                    }, 'null'],
+            }, {
+                name: 'fullDocument',
+                type: [{
+                   type: 'record',
+                   name: 'fullDocumentRecord',
+                   fields: [{
+                        name: 'value',
+                        type: [{
+                            type: 'record',
+                            name: 'valueRecord',
+                            fields: [{
+                                name: 'key',
+                                type: ['string', 'null'],
+                            }],
+                        }, 'null'],
+                   }],
+                }, 'null'],
+            }],
+        }),
     },
     extensionConfigField: {
         notification: 'notificationConfiguration',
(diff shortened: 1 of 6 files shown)


extensions/oplogPopulator/constants.js
```diff
@@ -6,6 +6,35 @@ const constants = {
...
+                        type: 'record',
+                        fields: [{
+                            name: 'coll',
+                            type: ['string', 'null']
+                        }]
+                    },
+                    'null'
+                ],
+            }]
+        }),
```
ghost: Can we have more consistent commas here? :slightly_smiling_face:

francoisferrand (approved):

extensions/oplogPopulator/constants.js
```diff
@@ -6,6 +6,35 @@ const constants = {
...
+            name: 'keySchema',
+            fields: [{
+                name: 'ns',
+                type: [
+                    {
+                        name: 'ns',
+                        type: 'record',
+                        fields: [{
+                            name: 'coll',
+                            type: ['string', 'null'],
```
francoisferrand, line 30: using only the bucket would prevent having multiple partitions for a single bucket : which could be an issue if the use-case is many objects in a few buckets.

But we should also keep the ordering of all events touching the same object.

→ the key should be based on the bucket (collection) and object key ?

@RachedBenMustapha What do you think about this? Do you think we should indeed drop ordering of events related to separate objects in order to support more scalable processing; or do you expect there would be enough buckets to support our horizontal scaling needs?

welansari: The object key we'll use will have to be the object name that is situated in the `value` field and not the `_id` as it can contain the version of the object which differ from version to version. 

The issue with this is that delete events don't have the `value` field and so the object key will be `null` which does not guarantee it'll land in the same partition as other messages.

Should we care about delete events ? i know they are not used in any backbeat extension except notification, but this will soon change with the implementation of the "object flagging before deletion"

ghost (approved):
LGTM, but I would suggest waiting for Rached's answer on the object keys

extensions/oplogPopulator/constants.js
francoisferrand: delete events should be handled soon, so hopefully that won't be a problem soon.

now the (functional) difference between `value` and `_id` will be that we may get multiple events related to the same object in different partitions (and thus possibly processed by different instances).

Are there cases where this relaxing of the ordering may be an issue (i.e. where the order of events related to different versions [or master] of the same object matters)?

welansari: In bucket notification we don't guarantee that events will come in the correct order so it shouldn't be an issue.

For other extensions like in replication, the versions must be stored in the same order as in the source site, and since we don't process master's in replication (master keys that have versions associated) we might end up with a different "latest" version in the replication site which is a problem.

PR #2254, opened 2022-07-05
BB-222: fix ColdStorageStatusQueueEntry parsing + improve error logging

alexanderchan-scality:


lib/models/ColdStorageStatusQueueEntry.js
```diff
@@ -4,15 +4,15 @@ const entrySchema = joi.object({
...
     requestId: joi.string().required(),
 });
 
 class ColdStorageStatusQueueEntry {
     static createFromKafkaEntry(kafkaEntry) {
-        const res = entrySchema.validate(kafkaEntry.value);
+        const res = entrySchema.validate(JSON.parse(kafkaEntry.value));
 
         if (res.error) {
             return {
```
rachedbenmustapha: The error returned is not really "malformed json" as that would make `JSON.parse` throw an exception (which needs to be caught btw). It's more "validation failed".

rachedbenmustapha (changes requested):

lib/models/ColdStorageStatusQueueEntry.js
alexanderchan-scality: I see.  `value` will be a  buffer/string type, so the parsing will be required. will add additional error for the parsing

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -185,7 +185,10 @@ class LifecycleObjectProcessor extends EventEmitter {
 
         const actionEntry = ActionQueueEntry.createFromKafkaEntry(kafkaEntry);
         if (actionEntry.error) {
-            this._log.error('malformed action entry', kafkaEntry.value);
+            this._log.error('malformed action entry', {
+                error: actionEntry.error,
+                entry: kafkaEntry.value,
+            ));
```
rachedbenmustapha: Syntax error

rachedbenmustapha (changes requested):

PR #2257, opened 2022-07-06
support replication of non versioned objects in OOB buckets

welansari:


extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -49,8 +49,16 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+
+        // Allow non versioned master keys as OOB buckets might contain objects
+        // that were created before versioning was enabled
+        const isNonVersionedMaster = !queueEntry.getVersionId();
+        // Allow objects created after version suspension, those have a null version master only
+        const isNullVersionMaster = queueEntry.getVersionId() && queueEntry.getIsNull();
+        // Allow a non-versioned object if being replicated from an NFS bucket.
+        const isNFS = queueEntry.getReplicationIsNFS();
+
+        if (isMasterKey(entry.key) && !isNFS && !isNullVersionMaster && !isNonVersionedMaster) {
```
francoisferrand, line 66: in ingestionQueuePopulator, there is this function (`IngestionQueuePopulator.js:134`)

```
    /**
     * Filter if the entry is considered a master key entry.
     * There is a case where a single null entry looks like a master key and
     * will not have a duplicate versioned key. They are created when you have a
     * non-versioned bucket with objects, and then convert bucket to versioned.
     * If no new versioned objects are added for given object(s), they look like
     * standalone master keys. The `isNull` case is undefined for these entries.
     * @param {ObjectQueueEntry} entry - object queue entry instance
     * @return {Boolean} true if we should filter entry
     */
    _isMasterKeyEntry(entry) {
        const isMaster = isMasterKey(entry.getObjectVersionedKey());
        // single null entries will have a version id as undefined or null.
        // do not filter single null entries
        if (isMaster && entry.getVersionId() !== undefined) {
            this.log.trace('skipping master key entry');
            return true;
        }
        return false;
    }
```

which is the appropriate way to check `isMasterKey` i guess : so maybe we should refactor somehow, to that this function is used in both cases...

PR #2265, opened 2022-07-12
Forward-port CRR replay

francoisferrand:
- Forward port CRR Replay implementation from 7.10 branch.
- Start all ReplayQueueProcessor instances in the same node.js process.
- Disable CRR replay from Data-Mover.

Issue: BB-234


extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -277,7 +371,7 @@ class UpdateReplicationStatus extends BackbeatTask {
             const params = { sourceEntry, refreshedEntry };
             const updatedSourceEntry = this._getUpdatedSourceEntry(params, log);
             if (!updatedSourceEntry) {
-                return done();
+                return process.nextTick(done);
```
francoisferrand, line 378: Original code had this 'nextTick()', but not sure it is still relevant in the updated code

@nicolas2bert Can you provide more context/feedback?

nicolas2bert: I will keep it to keep the callback function asynchronous in both cases.

francoisferrand: thanks

package.json
```diff
@@ -15,6 +15,7 @@
     "oplog_populator": "node extensions/oplogPopulator/OplogPopulatorTask.js",
     "mongo_queue_processor": "node extensions/mongoProcessor/mongoProcessorTask.js",
     "garbage_collector": "node extensions/gc/service.js",
+    "replay_processor": "node extensions/replication/replayProcessor/task.js",
```
nicolas2bert: Not sure it is needed anymore.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -628,6 +637,10 @@ class QueueProcessor extends EventEmitter {
     }
 
     _getZkSiteNode() {
+        if (this.isReplayTopic) {
+            return `${zookeeperNamespace}${zkReplayStatePath}/${this.topic}/${this.site}`;
```
nicolas2bert, line 641: Today, `${zookeeperNamespace}${zkReplayStatePath}/${this.topic}/${this.site}` and `${zookeeperNamespace}${zkStatePath}/${this.site}` zk paths will share the same state. 
Pausing and Resuming a site will pause/resume both the site replication and the site replay work. 
I think it is the expected behavior, just wanted to double check with you :)

nicolas2bert (commented):
Also, are we planning on having some CRR Replay monitoring? @miniscruff worked on it for S3C if you want to have a look :) : https://github.com/scality/backbeat/pull/2086

extensions/replication/queueProcessor/QueueProcessor.js
francoisferrand: We do keep separate states in zookeeper on purpose, to avoid race conditions (eg when updating the state);
but every instance will react to the same "commands" (eg pause/resume a site), so their states should be the same indeed.

francoisferrand:
> Also, are we planning on having some CRR Replay monitoring? @miniscruff worked on it for S3C if you want to have a look :) : https://github.com/scality/backbeat/pull/2086 

Added this code, though we will add metrics retrieval and dashboard separately: there is no replication monitoring at the moment in zenko, it will be addressed in separate epic.

jbertran (approved):

nicolas2bert (approved):

welansari (approved):
LGTM, tested locally

Kerkesni:
ping

erwan-b:
ping

tcarmet:
don't mind me just testing stuff.

PR #2267, opened 2022-07-12
Feature: BB-246 allow replication of non versioned objects

welansari:
Issue: [BB-246](https://scality.atlassian.net/browse/BB-246)

Source OOB buckets can have non versioned objects with only master keys that were created before versioning was enabled, we allow replication of those objects.

- [x] Tested in Zenko CI [Build Link](https://eve.devsca.com/github/scality/zenko/#/builders/12/builds/5627)
- [x] Tested in Local Artesca cluster the behaviour of replication with zero byte objects, normal non MPU objects, MPU objects. the objects where also tested as versioned, non versioned and versioning suspended

### DIFF
--- a/extensions/replication/ReplicationQueuePopulator.js
+++ b/extensions/replication/ReplicationQueuePopulator.js
@@ -51,7 +51,8 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
             return;
         }
         // Allow a non-versioned object if being replicated from an NFS bucket.
-        if (isMasterKey(entry.key) && !queueEntry.getReplicationIsNFS()) {
+        // Or if the master key is of a non versioned object
+        if (!this._entryCanBeReplicated(queueEntry)) {
             return;
         }
         if (queueEntry.getReplicationStatus() !== 'PENDING') {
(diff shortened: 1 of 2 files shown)


jbertran (approved):

anurag4DSB (approved):

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -95,6 +96,30 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+     * There is a case where a single null entry looks like a master key and
+     * will not have a duplicate versioned key. They are created when you have a
+     * non-versioned bucket with objects, and then convert bucket to versioned.
+     * If no new versioned objects are added for given object(s), they look like
+     * standalone master keys. The `isNull` case is undefined for these entries.
+     * Non-versioned objects if being replicated from an NFS bucket are also allowed
+     * @param {ObjectQueueEntry} entry - raw queue entry
+     * @return {Boolean} true if we should filter entry
+     */
+    _isValidMasterKeyEntry(entry) {
```
jonathan-gramain: The name of this helper function looks misleading, as it also returns true for version keys. I suggest a more generic name, e.g. `_entryCanBeReplicated`

PR #2270, opened 2022-07-13
BB-241 Forbid transitioning/replicating a cold object

nicolas2bert:


nicolas2bert:
ping

ghost (approved):
LGTM

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -895,6 +897,13 @@ class LifecycleTask extends BackbeatTask {
             next =>
                 this._getObjectMD(params, log, next),
             (objectMD, next) => {
+                const dataStoreName = objectMD.getDataStoreName();
```
naren-rajendran, line 903: nit: Turn this into an util fn maybe?

rachedbenmustapha (dismissed):

rachedbenmustapha:
@nicolas2bert I missed the case previously when an object is temporarily restored: its `dataStoreName` is not the cold location, so an extra check to take into account the presence of the `archive.restoreCompletedAt` attribute would be good when deciding not to transition the object.

alexanderchan-scality (approved):

PR #2271, opened 2022-07-14
BB-218: move archived source data deletion logic to GC task

alexanderchan-scality:


alexanderchan-scality:
ping

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -92,6 +53,7 @@ class LifecycleColdStatusArchiveTask extends LifecycleUpdateTransitionTask {
                 objectMD = res;
                 skipLocationDeletion = !locations ||
                     (Array.isArray(locations) && locations.length === 0);
```
nicolas2bert, line 55: If "data location" metadata value is empty, can we just return right away (no need to go any further since the transition for this entry would be already completed)?

alexanderchan-scality: 0 byte objects would have an empty location, so returning immediately may not be correct; the `putMetadata` call will still need to be done to update the `datastoreName`

nicolas2bert (approved):

miniscruff (approved):

PR #2280, opened 2022-07-18
BB-254: skip processing of master entries when objects are versioned

alexanderchan-scality:


naren-rajendran (approved):

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -183,6 +184,12 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
             !(entry.key && entry.key.startsWith(mpuBucketPrefix));
     }
 
+    _isVersionedObject(mdValue) {
+        return mdValue.versionId !== undefined ||
+            mdValue.isNull ||
+            mdValue.isDeleteMarker;
```
jonathan-gramain, line 190: I think `isDeleteMarker` implies a `versionId` exists, so may not be needed to check explicitly, but in doubt ok to keep it. (it may also be the case for `isNull` as well)

jonathan-gramain (approved):

PR #2281, opened 2022-07-20
bf: BB-249 lifecycle queue populator producer

naren-rajendran:
Lifecycle queue populator should create only one producer per location.

Tests have to be added.

### DIFF
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -1,3 +1,4 @@
+const async = require('async');
 const { constants } = require('arsenal');
 const { encode } = require('arsenal').versioning.VersionID;
 const { mpuBucketPrefix } = constants;
@@ -15,6 +16,7 @@ const VaultClientWrapper = require('../utils/VaultClientWrapper');
 const config = require('../../lib/Config');
 const { coldStorageRestoreTopicPrefix } = config.extensions.lifecycle;
 const BackbeatProducer = require('../../lib/BackbeatProducer');
+const locations = require('../../conf/locationConfig.json') || {};
 
 class LifecycleQueuePopulator extends QueuePopulatorExtension {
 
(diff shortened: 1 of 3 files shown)


naren-rajendran:
Tests have to be added.

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -197,6 +219,21 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
         const locationName = value.dataStoreName;
+        const locationConfig = locations[locationName];
+        if (!locationConfig) {
+            this.log.error('could not get location configuration', {
+                method: 'LifecycleQueuePopulator._handleRestoreOp',
+                location: locationName,
+            });
+            return;
+        }
+        if (!locationConfig.isCold) {
```
alexanderchan-scality: is there a need for the extra check for is cold?
wouldn't the `s3:ObjectRestore` operation type check handle this as the operation can only be executed for archived data

naren-rajendran: I wasn't sure, will remove it.

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -246,7 +279,8 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
                 accountId
             });
 
-            this._setupProducer(topic, producer => {
+            const producer = this._producers[topic];
+            if (producer) {
```
alexanderchan-scality, line 286: minor: could invert the check and return

alexanderchan-scality (approved):

lib/queuePopulator/QueuePopulator.js
```diff
@@ -407,21 +407,34 @@ class QueuePopulator {
...
+            }
+            if (extension.createZkPath) {
+                return extension.createZkPath(cb);
+            }
+            return cb();
+        });
+    }
+
     _setupExtensions(cb) {
         return async.each(this._extensions, (ext, next) => {
```
jonathan-gramain: Personally I prefer to name the callback `next` when the tasks are done one after another, and `done` for parallel execution or otherwise out-of-order. Here it is the opposite (`next` for `async.each` and `done` for `async.series`) which may be confusing to read. Maybe either swapping the names or renaming the callbacks to more specific names could help.

jonathan-gramain (approved):

nicolas2bert (commented):
Can we test the fix?

PR #2282, opened 2022-07-22
BB-257 Backbeat client requests never time out

nicolas2bert:
Today, “backbeat client” requests never timeout blocking the transition object processor flow when a response is never received.

This might happen due to a software bug (an error is thrown).

I set the timeout to 2 minutes (default AWS SDK value).

Also, the retry logic timeout after 5 minutes so it will retry a couple of times before giving up.

### DIFF
--- a/lib/clients/utils.js
+++ b/lib/clients/utils.js
@@ -1,6 +1,8 @@
 const S3 = require('aws-sdk/clients/s3');
 const BackbeatClient = require('./BackbeatClient');
 
+const TIMEOUT_MS = 1000 * 60 * 2; // 2 minutes in ms
+
 function attachReqUids(s3req, log) {
     s3req.on('build', () => {
         // eslint-disable-next-line no-param-reassign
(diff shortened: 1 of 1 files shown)


nicolas2bert:
ping

lib/clients/utils.js
```diff
@@ -1,6 +1,8 @@
 const S3 = require('aws-sdk/clients/s3');
 const BackbeatClient = require('./BackbeatClient');
 
+const TWO_MINUTES = 120000;
```
naren-rajendran: can we rename this to something appropriate? and may be the value be represented like `1000 * 60 * 2` if we want to know it is two minutes?

naren-rajendran:
nit: tests?

naren-rajendran (approved):

rachedbenmustapha (approved):

nicolas2bert:
@bert-e approve

PR #2284, opened 2022-07-25
bugfix: BB-260 fix S3C delete events not correctly deleting Zenko obj…

welansari:
…ect in OOB buckets

When receiving a delete event in RAFT log, OOB calls the generic deleteObject function of
the MongoClient with undefined options so that it calls the deleteObjectNoVer function to
directly delete the object without any further manipulation.

The behaviour of this function changed with the introduction of the v1 metadata in Zenko,
the function now adds a prefix to the object name before performing the deletion. Since the
function was intended for deletion of non versioned objects (objects that only have a master
key) it adds a master version prefix \x7fM.

This introduces an issue where deleted objects in S3C are not deleted in Zenko, as the function
only deletes the master keys and not the version. This creates an issue when listing objects either
in the UI or using the CLI, where the object is not visible in the normal listing function but is
when listing the versions.

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -8,6 +8,7 @@ const { replicationBackends, emptyFileMd5 } = require('arsenal').constants;
 const MongoClient = require('arsenal').storage
     .metadata.mongoclient.MongoClientInterface;
 const ObjectMD = require('arsenal').models.ObjectMD;
+const { extractVersionId } = require('../../lib/util/versioning');
 
 const Config = require('../../lib/Config');
 const BackbeatConsumer = require('../../lib/BackbeatConsumer');
@@ -385,14 +386,19 @@ class MongoQueueProcessor {
      */
     _processDeleteOpQueueEntry(log, sourceEntry, location, done) {
         const bucket = sourceEntry.getBucket();
-        const key = sourceEntry.getObjectVersionedKey();
+        const key = sourceEntry.getObjectKey();
+        const versionId = extractVersionId(sourceEntry.getObjectVersionedKey());
 
-        // Always call deleteObject with version params undefined so
-        // that mongoClient will use deleteObjectNoVer which just deletes
-        // the object without further manipulation/actions.
-        // S3 takes care of the versioning logic so consuming the queue
-        // is sufficient to replay the version logic in the consumer.
-        return this._mongoClient.deleteObject(bucket, key, undefined, log,
+        const options = versionId ? { versionId } : undefined;
+
+        // Calling deleteObject with undefined options to use deleteObjectNoVer which is used for
+        // deleting non versioned objects that only have master keys.
+        // When deleting a versioned object however we supply the version id in the options, which
+        // causes the function to call the deleteObjectVer function that is used to handle objects that
+        // have both a master and version keys. This handles the deletion of both the version and the master
+        // keys in the case where no other version is available, or deleting the version and updating the
+        // master key otherwise.
+        return this._mongoClient.deleteObject(bucket, key, options, log,
             err => {
                 if (err) {
                     this._normalizePendingMetric(location);
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -1,6 +1,17 @@
 const VID_SEP = require('arsenal').versioning.VersioningConstants
           .VersionId.Separator;
 
+/**
+ * @param {string} key full object key (_id in metadata)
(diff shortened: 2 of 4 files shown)


extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -385,14 +386,24 @@ class MongoQueueProcessor {
...
-        // Always call deleteObject with version params undefined so
-        // that mongoClient will use deleteObjectNoVer which just deletes
-        // the object without further manipulation/actions.
-        // S3 takes care of the versioning logic so consuming the queue
-        // is sufficient to replay the version logic in the consumer.
-        return this._mongoClient.deleteObject(bucket, key, undefined, log,
+        const key = sourceEntry.getObjectKey();
+        const versionId = extractVersionId(sourceEntry.getObjectVersionedKey());
+
+        let options = undefined;
```
jbertran: `const options = versionId ? { versionId } : undefined` ? No side-effect :)

lib/util/versioning.js
```diff
@@ -1,6 +1,26 @@
 const VID_SEP = require('arsenal').versioning.VersioningConstants
           .VersionId.Separator;
 
+/**
+ * @param {string} key full object key (_id in metadata)
+ * @returns {bool} true if key belongs to master object
+ */
+function isMasterKey(key) {
```
jbertran: This exists in Arsenal as `arsenal.versioning.isMasterKey` even in 8.4 . I can't find code that fits your `extractVersionId`, but let's import what we can from Arsenal in the meantime.

(Additionally, I can't tell why we reexport `isMasterKey`, I see no usage of it, but let's keep it just in case)

If we have time before RC2, it would be better to update Arsenal with this code, import both, and reexport isMasterKey directly from Arsenal.

tests/functional/ingestion/MongoQueueProcessor.js
```diff
@@ -127,7 +127,8 @@ class MongoClientMock {
     deleteObject(bucket, key, params, log, cb) {
         assert.strictEqual(bucket, BUCKET);
         assert([KEY, `${KEY}${VID_SEP}${VERSION_ID}`].includes(key));
-        this._deleted.push(key);
+        const versionId = params && params.versionId;
+        this._deleted.push({ key, versionId });
```
jbertran, line 131: This pushes the key-value pair `versionId: undefined` in all cases. Is this expected, or do we want to avoid pushing the key if we have no versionId?

jbertran (changes requested):

tests/functional/ingestion/MongoQueueProcessor.js
welansari: that's expected, we want to push both cases, to see if the params passed to the delete function are good

lib/util/versioning.js
welansari: i didn't add `isMasterKey` actually, it was already there i just changed the way it was written to be more consistent with the `extractVersionId` function. It was implemented directly in the `module.exports` object.
And i do agree, i don't see the use of it here as like you said there are no visible references in backbeat calling this function.

jbertran: I saw you didn't add it, but we can drive-by remove it and reexport it directly from Arsenal to reduce the amount of code duplication we have, since we're here anyway :wink:

jbertran (approved):

lib/util/versioning.js
```diff
@@ -1,6 +1,19 @@
+const { isMasterKey } = require('arsenal').versioning;
 const VID_SEP = require('arsenal').versioning.VersioningConstants
           .VersionId.Separator;
 
+/**
+ * @param {string} key full object key (_id in metadata)
+ * @returns {srting | null} version id
```
jonathan-gramain:
```suggestion
 * @returns {string | null} version id
```

lib/util/versioning.js
```diff
@@ -1,6 +1,19 @@
...
+function extractVersionId(key) {
+    if (key.includes(VID_SEP)) {
+        return key.split(VID_SEP)[1];
+    }
+    return null;
+}
+
 module.exports = {
-    isMasterKey: key => !key.includes(VID_SEP),
+    isMasterKey,
```
jonathan-gramain: is this export coming from Arsenal done on purpose?

jonathan-gramain (approved):

lib/util/versioning.js
welansari: That function is actually no longer used in backbeat, i ran e2e tests and everything seems to still work, i'm removing it

PR #2291, opened 2022-08-01
bugfix: BB-262 fix replication failures of objects with tags

welansari:
Issue: [BB-262](https://scality.atlassian.net/browse/BB-262)

when putting or deleting tags from an object that has a FAILED replication status, all future replication attempts fail due to the replication processor directly putting/deleting tags and assuming that the object was replicated before which is not the case.

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -1117,10 +1117,14 @@ class MultipleBackendTask extends ReplicateObject {
                     return next(errors.InvalidObjectState.customizeDescription(
                         errMessage));
                 }
-                if (content.includes('PUT_TAGGING')) {
+                // Only put/delete tags if object was previously
+                // replicated. Otherwise we replicate the whole object
+                // which should also replicate the updated tags.
+                const dataStoreVersionId = sourceEntry.getReplicationSiteDataStoreVersionId(this.site);
+                if (content.includes('PUT_TAGGING') && dataStoreVersionId) {
                     return this._putObjectTagging(sourceEntry, log, next);
                 }
-                if (content.includes('DELETE_TAGGING')) {
+                if (content.includes('DELETE_TAGGING') && dataStoreVersionId) {
                     return this._deleteObjectTagging(sourceEntry, log, next);
                 }
                 // Do a multipart upload when either the size is above
--- a/tests/unit/replication/MultipleBackendTask.js
+++ b/tests/unit/replication/MultipleBackendTask.js
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const jsutil = require('arsenal').jsutil;
+const sinon = require('sinon');
 
 const config = require('../../config.json');
 const MultipleBackendTask =
@@ -57,7 +58,8 @@ describe('MultipleBackendTask', function test() {
                 site: 'test-site-2',
                 notificationConfigManager: {
                     getConfig: () => null
-                }
+                },
+                logger: fakeLogger,
             }),
         });
     });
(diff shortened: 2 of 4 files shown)


jonathan-gramain (commented):
It this PR fixing an issue specifically for Artesca? For S3C CRR, this case is normally already handled in `ReplicateObject.js`, where the condition is marked by this log line:

```
            log.info('target object version does not exist, retrying ' +
                'a full replication',
```

So I'm wondering if this condition shouldn't also already be handled in the multiple backend code, or if not if we should consider reusing the same mechanism that is already implemented?

welansari:
> It this PR fixing an issue specifically for Artesca? For S3C CRR, this case is normally already handled in `ReplicateObject.js`, where the condition is marked by this log line:

Yes this PR is only for Artesca.

> So I'm wondering if this condition shouldn't also already be handled in the multiple backend code, or if not if we should consider reusing the same mechanism that is already implemented?

Unlike `ReplicateObject.js` here the condition is not handled, I tried using the same mechanism however the errors we get are different from RING CRR, and i'm assuming they are also backend specific.
Replication to AWS for example returns this:

````
ServiceUnavailable: Error returned from AWS: Version id cannot be the empty string
````
Another issue i found is that the errors returned by AWS are considered retryable, so we do `n` retries (default is 5 i believe) which take quiet a time for what it is before considering it a failure, the retry_processor also does the same. I think that hinders the performance sgnificantly.

**Review by jbertran — APPROVED** [review]

File: extensions/replication/tasks/MultipleBackendTask.js [hunk 1]
```diff
@@ -907,15 +907,26 @@ class MultipleBackendTask extends ReplicateObject {
...
-            shouldRetryFunc: err => err.retryable,
-            log,
-        }, cb);
+        const dataStoreVersionId = sourceEntry.getReplicationSiteDataStoreVersionId(this.site);
+        async.series([
+            next => {
+                // put object if it wasn't already replicated.
+                // A replicated object has a dataStoreVersionId
+                if (!dataStoreVersionId) {
+                    return this._putObject(sourceEntry, log, next);
```
nicolas2bert: If we replicate get/put the object at this stage, is there still a need to update the object's tag? 
I assume its metadata (tags...) is already up-to-date.

nicolas2bert:
Is this issue happening only when ingesting an "object's tag(s) update" and replicating the "object's tag(s) update" to a "failed replication" object? 
I don't see 'PUT_TAGGING' content being used anywhere else but in the MongoQueueProcessor.

extensions/replication/tasks/MultipleBackendTask.js
welansari: you're right, i wasn't aware that putting the object also puts in the updated tags, fixed it

welansari:
> Is this issue happening only when ingesting an "object's tag(s) update" and replicating the "object's tag(s) update" to a "failed replication" object?
> I don't see 'PUT_TAGGING' content being used anywhere else but in the MongoQueueProcessor.

The object doesn't have to be coming from an OOB bucket, the issue occurs whenever we put/delete tags on any object that initially failed replicating.

'PUT_TAGGING' is added to the object's metadata by Cloudserver [here](https://github.com/scality/cloudserver/blob/e70753446f3886b761344aec4fc77c55b7262199/lib/api/objectPutTagging.js#L81-L83) when putting tags, the same happens when we delete tags

tests/unit/replication/MultipleBackendTask.js
```diff
@@ -57,12 +58,13 @@ describe('MultipleBackendTask', function test() {
...
                     getConfig: () => null
-                }
+                },
+                logger: fakeLogger,
             }),
         });
     });
 
-    describe('::_setupClients', () => {
+    describe.skip('::_setupClients', () => {
```
nicolas2bert: Why is it being skipped? A comment with potentially mentioning a JIRA ticket would be welcome.

nicolas2bert (approved):

tests/unit/replication/MultipleBackendTask.js
welansari: 🤦 sorry about that, forgot to remove the skip when i finished testing

PR #2302, opened 2022-08-09
Feature: BB-244 move notification failed processing logic to replication status processor

welansari:
Issue: [BB-244](https://scality.atlassian.net/browse/BB-244)

This refactoring is needed for multiple reasons:
- Send “replication failed” notification only after CRR replay has failed
- Avoid duplicating “resources” (like NotificationConfigManager) in each instance of QueueProcessor
- Keep all status processor in a single place


KillianG (approved):

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -102,6 +115,9 @@ class UpdateReplicationStatus extends BackbeatTask {
             score,
         };
         this.failedCRRProducer.publishFailedCRREntry(JSON.stringify(message));
+        if (this.bucketNotificationConfig) {
```
francoisferrand: this function may not be called, e.g. if `repConfig.monitorReplicationFailures` is not set : but we should still post a bucket notification in that case.

tests/unit/replication/ReplicateObject.spec.js
```diff
@@ -1,166 +0,0 @@
-const assert = require('assert');
```
francoisferrand, line 1: we should port these tests to some `tests/unit/replication/UpdateReplicationStatus.js` ?

ghost (approved):
LGTM! Can we mention in the ticket that the `tests` workflow was updated with GHA caching? Just to keep track of why it was done in this PR

francoisferrand (approved):

PR #2303, opened 2022-08-10
BB-263 Blacklist lifecycle buckets and accounts

nicolas2bert:
More insight related to this PR: https://scality.atlassian.net/browse/S3C-6042

The Conductor's config will look like this:

```
"conductor": {
             ...
              "filter": {
                  "deny": {
                      "buckets": ["bucket1", "bucket2"],
                      "accounts": ["account1:be288756448dc58f61482903131e7ae533553d20b52b0e2ef80235599a1b9143"]
                  }
              }
          },
```

### DIFF
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -19,6 +19,7 @@ const { AccountIdCache } = require('../util/AccountIdCache');
 
 const DEFAULT_CRON_RULE = '* * * * *';
 const DEFAULT_CONCURRENCY = 10;
+const ACCOUNT_SPLITTER = ':';
 
 const LIFEYCLE_CONDUCTOR_CLIENT_ID = 'lifecycle:conductor';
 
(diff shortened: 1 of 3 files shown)


nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e reset

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -325,12 +367,15 @@ class LifecycleConductor {
...
                                 isTruncated = result.IsTruncated;
-                                nEnqueued += result.Contents.length;
 
                                 result.Contents.forEach(o => {
                                     marker = o.key;
                                     const [canonicalId, bucketName] = marker.split(constants.splitter);
-                                    queue.push({ canonicalId, bucketName });
+
+                                    if (!this._isBlacklisted(canonicalId, bucketName)) {
+                                        nEnqueued += 1;
```
rachedbenmustapha, line 377: I think keeping a count of the skipped ones as well (logging it alongside `nEnqueued`) would be helpful.

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -325,12 +367,15 @@ class LifecycleConductor {
...
+                                        queue.push({ canonicalId, bucketName });
+                                    }
```
rachedbenmustapha, line 387: Potential optimization here: at the end of the loop, if the last entry was a denied account id, there's a good chance that the next "page" also starts with the same account, and with bad luck, could go on for several thousands of pages.

We could cheat in this specific case by overriding the marker, to start listing at the next canid (would be trivial with an allowlist-style rule but here it needs a trick). For example:
- last entry of `Contents` is `abcdef:bucketfoo`
- (next list probably starts with `abcdef:bucketbar`)
- (could be several millions of entries to list and evaluate)
- after the `Contents.forEach` loop, if the last entry was skipped:
- overwrite the marker with `abcdef;` (because `;` is immediately after `:`)
- the next listing will start at the next available unskipped account

rachedbenmustapha (commented):
Overall the implementation looks good, just an optional suggestion to save a potential few million bucket listing entries. Maybe I'm overly optimistic, just something to consider.

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -95,9 +96,34 @@ class LifecycleConductor {
         // - some entries are expired for each listing
         this._accountIdCache = new AccountIdCache(this._concurrency);
 
+        const blacklist = (this.lcConfig.conductor.filter && this.lcConfig.conductor.filter.deny) || {};
```
miniscruff, line 99: There was a push a few years ago around switching from the terms whitelist and blacklist to alternatives ( see https://9to5google.com/2020/06/12/google-android-chrome-blacklist-blocklist-more-inclusive/ ). Just something to consider. Also you could do: `this.lcConfig.conductor.filter?.deny || {}`

nicolas2bert: Unfortunately, optional chaining is not supported yet on this version.

rachedbenmustapha (approved):

nicolas2bert:
ping

naren-rajendran (approved):

nicolas2bert:
@bert-e approve

PR #2306, opened 2022-08-12
BB-264 handle suspended versioning

jbertran:
Status:

- [x] manual tests for normal, suspended and non versioned replication with OOB
- [x] Release Arsenal & update version

Arsenal Changes: https://github.com/scality/Arsenal/pull/2033

Changes: 
- In OOB we avoid creating a version object by specifying the correct params for the MongoClient's `putObject` function
- In replication, allow replication of null versions (versioning suspended master keys)
- In replication, avoid creating a version object when updating the replication status of the versioning suspended object

### DIFF
--- a/extensions/ingestion/IngestionQueuePopulator.js
+++ b/extensions/ingestion/IngestionQueuePopulator.js
@@ -135,7 +135,10 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
         const isMaster = isMasterKey(entry.getObjectVersionedKey());
         // single null entries will have a version id as undefined or null.
         // do not filter single null entries
-        if (isMaster && entry.getVersionId() !== undefined) {
+        // Versioning suspended entries will have a version id but also a isNull tag.
+        // These master keys are considered a version and do not have a duplicate version,
+        // we also don't filter this type of entry.
+        if (isMaster && entry.getVersionId() !== undefined && !entry.getIsNull()) {
             this.log.trace('skipping master key entry');
             return true;
         }
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -200,7 +200,10 @@ class MongoQueueProcessor {
         const bucket = entry.getBucket();
         const key = entry.getObjectKey();
         const params = {};
-        if (entry.getVersionId()) {
+        // master keys with a 'null' version id comming from
+        // a versioning suspended bucket are considered a version
+        // we should not specify the version id in this case
+        if (entry.getVersionId() && !entry.getIsNull()) {
             params.versionId = entry.getVersionId();
         }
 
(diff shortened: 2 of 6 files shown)


extensions/replication/utils/utils.js
```diff
@@ -0,0 +1,17 @@
+const { encode } = require('arsenal').versioning.VersionID;
+const ObjectMD = require('arsenal').models.ObjectMD;
+
+function getVersionIdForReplication(sourceEntry, encoding = false) {
+    const nullCheckFn = sourceEntry instanceof ObjectMD ? entry => entry.getIsNull() : entry => entry.isNull;
+    const getVIDFn = sourceEntry instanceof ObjectMD ? entry => entry.getVersionId() : entry => entry.versionId;
```
francoisferrand: what other type can it be?
it would be better to just add the `getIsNull()` and `getVersionId()` methods in that type...

francoisferrand: Also, this replaces calls to `getEncodedVersionId()`, which is implemented in some (or all?) the types....

*To be confirmed*, but it looks to me like this logic should actually be in the getEncodedVersionId() function [from arsenal ObjectMD)?

(in backbeat at least, it is called only for replication and in a few BackbeatAPI calls : where i guess we probably want to pass `null` when needed as well...)

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -273,8 +273,13 @@ class MongoQueueProcessor {
...
         // as its dataStoreVersionId
-        const dataStoreVersionId = entry.getVersionId() ?
-            entry.getEncodedVersionId() : 'null';
+
+        let dataStoreVersionId;
+        if (entry.getIsNull() || (! entry.getVersionId())) {
+            dataStoreVersionId = 'null';
+        } else {
+            dataStoreVersionId = entry.getEncodedVersionId();
+        }
```
francoisferrand: is this not yet another call to `utils.getVersionIdForReplication()` ?

extensions/replication/utils/utils.js
welansari: Indeed the changes can be simplified a lot just by updating `getEncodedVersionId()` to handle the versioning suspended case.

This is done in the following PR: https://github.com/scality/Arsenal/pull/2033

extensions/mongoProcessor/MongoQueueProcessor.js
welansari: Not needed anymore

francoisferrand (approved):

extensions/ingestion/IngestionQueuePopulator.js
```diff
@@ -135,7 +135,10 @@ class IngestionQueuePopulator extends QueuePopulatorExtension {
         const isMaster = isMasterKey(entry.getObjectVersionedKey());
         // single null entries will have a version id as undefined or null.
         // do not filter single null entries
-        if (isMaster && entry.getVersionId() !== undefined) {
+        // Versioning suspended entries will have a version id but also a isNull tag.
+        // These master keys are considered a version an do not have a duplicate version,
```
ghost: Nit
```suggestion
        // These master keys are considered a version and do not have a duplicate version,
```

package.json
```diff
@@ -43,7 +43,7 @@
   "homepage": "https://github.com/scality/backbeat#readme",
   "dependencies": {
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/Arsenal#8.1.64",
+    "arsenal": "git+https://github.com/scality/Arsenal#bugfix/ARSN-278-handle-versioning-suspended-versionId",
```
ghost: Just a note so that we don't forget to use a proper tag before merging

ghost (approved):

PR #2307, opened 2022-08-17
BB-265 Set client timeout to free up consumer slots when connection idle

nicolas2bert:
Sets the socket to timeout after 120000 milliseconds of inactivity on the socket. It is the default AWS S3 JavaScript SDK.

The current value of 0 means infinity and for multiple reasons (server, network, internet connection) the server can stop sending data and leave the connection open and idle “forever“.

That will lead to:
- one consumer slot being taken “forever”
- “the consumer group offset” not committed and lag never decreased.

Note: we are only updating the `this.backbeatSource` client's timeout since it is the one responsible for getting and putting data. We also don't want to change too much the logic that has been working for years. 

TODO: Test that large files (> 1GB) are being replicated correctly and do not timeout. If it times out, we could increase the client timeout on replay processors.

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -457,6 +457,7 @@ class BackbeatConsumer extends EventEmitter {
             entry: { topic, partition, offset,
                      key: key && key.toString(), timestamp },
             committableOffset,
+            offsetLedger: this._offsetLedger.toString(),
         });
         // ensure consumer is active when calling offsetsStore() on
         // it, to avoid raising an exception (invalid state)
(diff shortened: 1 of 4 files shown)


miniscruff (approved):

francoisferrand (approved):

lib/BackbeatConsumer.js
```diff
@@ -457,6 +457,7 @@ class BackbeatConsumer extends EventEmitter {
             entry: { topic, partition, offset,
                      key: key && key.toString(), timestamp },
             committableOffset,
+            offsetLedger: this._offsetLedger.toString(),
```
francoisferrand, line 461: may also need logging on 'marked consumed entry'

francoisferrand (approved):

lib/BackbeatConsumer.js
nicolas2bert: You got it!

nicolas2bert:
@bert-e approve

PR #2310, opened 2022-08-24
Enable backlog metrics for QueueProcessor consumer

francoisferrand:
Issue: BB-268


francoisferrand:
this is not appropriate : the metrics are actually already generated (via calls to static functions of `KafkaBacklogMetrics`), and "enabling backlog metrics" is actually enabling some tracking for lifecycle through Zookeeper

PR #2312, opened 2022-09-08
Add replication metrics

francoisferrand:
- Add metrics computation to MultipleBackendTask
- Implement metrics route in QueueProcessor
- Enable replication metrics
- Remove containerName label from metrics
- Collect replicationStatusMetric
- feature: BB-133 add replay specific backlog metrics
- feature: BB-132 upload size metrics for replays
- Add target location to QueueProcessor metrics
- Remove unused kafka lag metrics
- Add location and replayCount to replay metrics

Issue: BB-62

### DIFF
--- a/bin/queuePopulator.js
+++ b/bin/queuePopulator.js
@@ -63,14 +63,9 @@ async.waterfall([
             probeServer.addHandler([DEFAULT_LIVE_ROUTE, DEFAULT_READY_ROUTE],
                 (res, log) => queuePopulator.handleLiveness(res, log)
             );
-            // TODO: set this variable during deployment
-            // enable metrics route only when it is enabled
-            if (process.env.ENABLE_METRICS_PROBE === 'true') {
-                probeServer.addHandler(
-                    DEFAULT_METRICS_ROUTE,
-                    (res, log) => queuePopulator.handleMetrics(res, log)
-                );
-            }
+            probeServer.addHandler(DEFAULT_METRICS_ROUTE,
+                (res, log) => queuePopulator.handleMetrics(res, log)
+            );
         }
         done();
     }),
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -6,7 +6,7 @@ const https = require('https');
 const { EventEmitter } = require('events');
 const Redis = require('ioredis');
 const schedule = require('node-schedule');
-const promClient = require('prom-client');
+const { ZenkoMetrics } = require('arsenal').metrics;
 
 const Logger = require('werelogs').Logger;
 
@@ -44,10 +44,6 @@ const {
     replicationBackends,
 } = require('../constants');
 
-promClient.register.setDefaultLabels({
-    origin: 'replication',
-});
-
 /**
  * Labels used for Prometheus metrics
  * @typedef {Object} MetricLabels
@@ -60,55 +56,58 @@ promClient.register.setDefaultLabels({
  * @property {string} [replicationStage] - Name of the replication stage
  */
 
-const dataReplicationStatusMetric = new promClient.Counter({
+const dataReplicationStatusMetric = ZenkoMetrics.createCounter({
     name: 'replication_data_status_changed_total',
     help: 'Number of status updates',
     labelNames: ['origin', 'location', 'replicationStatus'],
 });
 
-const metadataReplicationStatusMetric = new promClient.Counter({
+const metadataReplicationStatusMetric = ZenkoMetrics.createCounter({
     name: 'replication_metadata_status_changed_total',
     help: 'Number of status updates',
     labelNames: ['origin', 'location', 'replicationStatus'],
 });
 
-const dataReplicationBytesMetric = new promClient.Counter({
+const dataReplicationBytesMetric = ZenkoMetrics.createCounter({
     name: 'replication_data_bytes',
     help: 'Total number of bytes replicated for data operation',
     labelNames: ['origin', 'serviceName', 'location'],
 });
 
-const metadataReplicationBytesMetric = new promClient.Counter({
+const metadataReplicationBytesMetric = ZenkoMetrics.createCounter({
     name: 'replication_metadata_bytes',
     help: 'Total number of bytes replicated for metadata operation',
     labelNames: ['origin', 'serviceName', 'location'],
 });
 
-const sourceDataBytesMetric = new promClient.Counter({
+const sourceDataBytesMetric = ZenkoMetrics.createCounter({
     name: 'replication_source_data_bytes',
     help: 'Total number of data bytes read from replication source',
     labelNames: ['origin', 'serviceName', 'location'],
 });
 
-const readMetric = new promClient.Counter({
+const readMetric = ZenkoMetrics.createCounter({
     name: 'replication_data_read',
     help: 'Number of read operations',
     labelNames: ['origin', 'serviceName', 'location'],
 });
 
-const writeMetric = new promClient.Counter({
+const writeMetric = ZenkoMetrics.createCounter({
     name: 'replication_data_write',
     help: 'Number of write operations',
     labelNames: ['origin', 'serviceName', 'location', 'replicationContent'],
 });
 
-const timeElapsedMetric = new promClient.Histogram({
+const timeElapsedMetric = ZenkoMetrics.createHistogram({
     name: 'replication_stage_time_elapsed',
     help: 'Elapsed time of a specific stage in replication',
     labelNames: ['origin', 'serviceName', 'location', 'replicationStage'],
     buckets: [0.01, 0.1, 1, 10, 30, 60, 120, 300],
 });
 
+const defaultLabels = {
+    origin: 'replication',
+};
 /**
  * Contains methods to incrememt different metrics
  * @typedef {Object} MetricsHandler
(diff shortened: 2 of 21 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -1014,7 +1004,7 @@ class QueueProcessor extends EventEmitter {
      * @param {Logger} log - Logger
      * @returns {string} Error response string or undefined
      */
-    handleMetrics(res, log) {
+    static handleMetrics(res, log) {
```
miniscruff, line 1007: I don't think this needs to be static and am wondering why you changed it

tests/utils/mockEntries.js
```diff
@@ -7,6 +7,7 @@ const sourceEntry = {
     getLocation: () => ([]),
     getUserMetadata: () => {},
     getContentType: () => {},
+    getContentLength: () => { return 0; },
```
miniscruff: Can this just be?
```suggestion
    getContentLength: () => 0,
```

tests/unit/replication/UpdateReplicationStatus.js
```diff
@@ -31,6 +33,9 @@ function checkReplicationInfo(site, status, updatedSourceEntry) {
         updatedSourceEntry.getReplicationSiteStatus(site), status);
 }
 
+const metricsHandlers = ReplicationStatusProcessor.loadMetricHandlers(
```
miniscruff:
```suggestion
const metricHandlers = ReplicationStatusProcessor.loadMetricHandlers(
```
Kinda nit picky but I think "metric handlers" makes more sense then "metrics handlers"

miniscruff: This seems to have been a typo we made somewhere and I see both being used so maybe we can just swap them all to the same "metric handlers" now.

miniscruff (changes requested):

extensions/replication/queueProcessor/QueueProcessor.js
francoisferrand: since we create multiple instances of queue processor in the same process, it needs to be static indeed.

I could just move it to task.js, but I kept it here for consistency with the other "tasks".

tests/unit/replication/UpdateReplicationStatus.js
francoisferrand: I believe the "metrics" word is actually singular in this domain, and we should always speak of "metrics" even if there is a single one... so metricsHandler seems fine with me.

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -172,6 +218,7 @@ class ReplicationStatusProcessor {
         }
 
         this._setupVaultclientCache();
+        this.metricsHandlers = loadMetricHandlers(repConfig);
```
miniscruff:
```suggestion
        this.metricHandlers = loadMetricHandlers(repConfig);
```

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -479,7 +526,7 @@ class ReplicationStatusProcessor {
         }
         let task;
         if (sourceEntry instanceof ObjectQueueEntry) {
-            task = new UpdateReplicationStatus(this, metricsHandler);
+            task = new UpdateReplicationStatus(this, this.metricsHandlers);
```
miniscruff:
```suggestion
            task = new UpdateReplicationStatus(this, this.metricHandlers);
```

extensions/replication/queueProcessor/QueueProcessor.js
miniscruff: > since we create multiple instances of queue populator in the same process

we do? I thought we just create the one in `bin/queuePopulator.js` and pass it around everywhere. That was an assumption made when building the metric handlers.

francoisferrand: sorry i was speeaking of QueueProcessor, but mis-typed QueuePopulator... : and there is indeed one instance of QueueProcessor (and consumer group) for each replicaiton site.
In S3C we used to run multiple node processes for these, but in Zenko they all got moved into the same node process.

miniscruff: Wouldn't this mean we don't need this to be static then?

francoisferrand: It needs to be static since it called from task.js, in the /metrics route callback while there are multiple instances of QueueProcessor class (one per site)

miniscruff: oh right in 8x we have multiple processors. I guess making it a static function that just calls the prom stuff is fine. Personally I think a better solution would be to move this to an exported function or a function under the metric handler but that would break the single export we have.

francoisferrand: I agree it is not very nice, and quite redundant since we have this in a few places. But I will leave this cleanup for another time...

bin/queuePopulator.js
```diff
@@ -63,14 +63,11 @@ async.waterfall([
...
             );
-            // TODO: set this variable during deployment
             // enable metrics route only when it is enabled
-            if (process.env.ENABLE_METRICS_PROBE === 'true') {
-                probeServer.addHandler(
-                    DEFAULT_METRICS_ROUTE,
-                    (res, log) => queuePopulator.handleMetrics(res, log)
-                );
-            }
+            probeServer.addHandler(
```
ghost: We can remove the comment above now that we enable this by default

extensions/replication/queueProcessor/task.js
```diff
@@ -275,15 +275,10 @@ function initAndStart(zkClient) {
...
-                    // TODO: set this variable during deployment
                     // enable metrics route only when it is enabled
-                    if (process.env.ENABLE_METRICS_PROBE === 'true') {
-                        // TODO: implement metrics route for multi site setup, BB-23
-                        probeServer.addHandler(DEFAULT_METRICS_ROUTE, (res, log) => {
-                            log.info('queue processor metrics route not implemented');
-                            sendSuccess(res, log);
-                        });
-                    }
+                    probeServer.addHandler(DEFAULT_METRICS_ROUTE,
```
ghost, line 278: Same here

extensions/replication/replicationStatusProcessor/task.js
```diff
@@ -51,14 +51,11 @@ function initAndStart() {
...
                     );
-                    // TODO: set this variable during deployment
                     // enable metrics route only when it is enabled
-                    if (process.env.ENABLE_METRICS_PROBE === 'true') {
-                        probeServer.addHandler(
-                            DEFAULT_METRICS_ROUTE,
-                            (res, log) => replicationStatusProcessor.handleMetrics(res, log)
-                        );
-                    }
+                    probeServer.addHandler(
```
ghost: Same here

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -214,9 +214,21 @@ class UpdateReplicationStatus extends BackbeatTask {
...
             if (sourceEntry.getReplayCount() >= 0) {
-                this.metricsHandler.replaySuccess();
+                replayCount = this.replayTopics.length - sourceEntry.getReplayCount();
+                this.metricsHandler.replaySuccess({ location: site, replayCount });
             }
+
+            const labels = {
+                location: site,
+                replayCount,
+                replicationStatus: 'COMPLETED',
```
ghost, line 227: If we reuse such statuses, we could create constants, it will be easier to manage

ghost (approved):
LGTM, just some quick code-related remarks

extensions/replication/tasks/UpdateReplicationStatus.js
francoisferrand: i totally agree, `const` (or better yet a typescript `enum`) would be better... but these should actually be part of the metadata models (e.g. in arsenal), and are used much more widely than just here: so i will leave that for another day

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -6,7 +6,7 @@ const https = require('https');
 const { EventEmitter } = require('events');
 const Redis = require('ioredis');
 const schedule = require('node-schedule');
-const promClient = require('prom-client');
+const { ZenkoMetrics } = require('arsenal').metrics;
```
miniscruff, line 9: What is the reason behind going to zenko metrics from the prom client?

francoisferrand: Initial reason was consistency, since ZenkoMetrics is used everywhere. It also automatically inject the "default" node metrics.

Then i found out that not using it is an issue: other parts of the code use it, and actually end up using _another copy_ of prom-client via arsenal, and thus another singleton registry. Some metrics end up in ZenkoMetrics/arsenal copy (backbeatconsumer, backbeatproducer, default metrics) and replication ones go in directly imported copy: and it is ugly to merge both in the route callback.

Not sure why ZenkoMetrics was introduced in the first place, but I just followed through to avoid issues and collect all metrics as expected.

bin/queuePopulator.js
francoisferrand: good catch, will do

extensions/replication/queueProcessor/QueueProcessor.js
miniscruff: The default node metrics were intentionally not imported as they were very bloaty and did not seem to be helpful given the size increase. Backbeat consumer and producer not including there metrics could be a problem, I am not sure this effects 7x so maybe we just ignored it. I believe we were also concerned about trying to limit what metrics were being registered given backbeat has many processes in one project. I am not sure if this actually matters though.

Oh and by wrapping the prom client we lose all type hints which makes working with the metrics a little harder. Not a big deal as we wrap the metrics with the metric handler but that part is at least mostly type aware when using the prom client directly. 

Personally I think a static wrapper around a static registry is a bit odd and problematic, hence why we did not use zenko metrics and was trying to deprecate the use over time ( similar to the health check route and the probe server update ).

francoisferrand: Are there any measure as to the bloat of default metrics ?
Not having them was actually an issue on KB, to help us isolate issue related to node itself...

As for ZenkoMetrics itself, I fully agree we should not need a static wrapper: yet I don't think I have the time to remove it overall (and validate the change), so I'll prefer to stay consistent for now, and leave it for another day.

miniscruff: We did not measure them I don't think. We had them on at first and when I presented them there were some 40 or so metrics I think, compared to the like 3 we added. When I mentioned this to Rahul we chose to turn the defaults off to reduce prometheus memory usage.

miniscruff (approved):

PR #2313, opened 2022-09-12
Collect latency metric for status-processor

francoisferrand:
Issue: BB-278

### DIFF
--- a/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
+++ b/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
@@ -64,6 +64,13 @@ const loadMetricHandlers = jsutil.once(repConfig => {
         labelNames: ['origin', 'replicationStatus'],
     });
 
+    const replicationStatusDurationSeconds = ZenkoMetrics.createHistogram({
+        name: 'replication_status_process_duration_seconds',
+        help: 'Duration of replication status processing',
+        labelNames: ['origin', 'result', 'replicationStatus'],
+        buckets: [0.05, 0.1, 0.2, 0.3, 0.4, 0.5, 1],
+    });
+
     const replayAttempts = ZenkoMetrics.createCounter({
         name: 'replication_replay_attempts_total',
         help: 'Number of total attempts made to replay replication',
@@ -119,6 +126,8 @@ const loadMetricHandlers = jsutil.once(repConfig => {
     };
     return {
         status: wrapCounterInc(replicationStatusMetric, defaultLabels),
+        statusDuration: wrapHistogramObserve(replicationStatusDurationSeconds,
+            defaultLabels),
         replayAttempts: wrapCounterInc(replayAttempts, defaultLabels),
         replaySuccess: wrapCounterInc(replaySuccess, defaultLabels),
         replayQueuedObjects: wrapCounterInc(replayQueuedObjects, defaultLabels),
--- a/extensions/replication/tasks/UpdateReplicationStatus.js
+++ b/extensions/replication/tasks/UpdateReplicationStatus.js
@@ -400,30 +400,48 @@ class UpdateReplicationStatus extends BackbeatTask {
         return null;
     }
 
+    _updateStatusDurationMetric(sourceEntry, err, replicationStatus) {
+        let result = 'nochange';
+        if (err) {
+            result = 'error';
+        } else if (replicationStatus) {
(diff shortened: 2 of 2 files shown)


extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -64,6 +64,13 @@ const loadMetricHandlers = jsutil.once(repConfig => {
         labelNames: ['origin', 'replicationStatus'],
     });
 
+    const replicationStatusDurationSeconds = ZenkoMetrics.createHistogram({
+        name: 'replication_status_process_duration_seconds',
+        help: 'Duration of replication status processing',
```
ghost, line 69: Is this a duration or a latency? The ticket is about latency but here we use the word "duration".

francoisferrand: A latency is a duration, it basically depends on the point of view.
Here we are expected to measure 
* The average latency for operations → hence the "latency" in title
* Defined as "elapsed ms for the time when an entry is picked up from the Kafka to the time its status is successfully updated in MongoDB" → i.e. the duration of processing

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -400,30 +400,51 @@ class UpdateReplicationStatus extends BackbeatTask {
...
+    _updateStatusDurationMetric(sourceEntry, err, replicationStatus) {
+        const result = (() => {
+            if (err) {
+                return 'error';
+            }
+            if (replicationStatus) {
+                return 'success';
+            }
+            return 'nochange';
+        })();
```
miniscruff: I don't feel like a function here is any better then two if statements.
```suggestion
        let result = 'nochange';
        if (err) {
            result = 'error';
        }
        else if (replicationStatus) {
            result = 'success';
        }
```

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -400,30 +400,51 @@ class UpdateReplicationStatus extends BackbeatTask {
...
+        this.metricsHandler.statusDuration({ result, replicationStatus },
+            (Date.now() - sourceEntry.getStartProcessing()) / 1000);
```
miniscruff, line 411: This is more of a nit pick and is a general pattern I have noticed in our code base that I think we should stop doing is formatting around multilines. This style of continuing to the next and just letting it end wherever is quite hard to read. Especially when sandwiched together with many other function calls of similar style. Or when the second line is not tabbed over like I have seen a lot. 

My suggestion is:
```suggestion
        this.metricHandlers.statusDuration(
            { result, replicationStatus },
            (Date.now() - sourceEntry.getStartProcessing()) / 1000,
        );
```

Ending parenthesis on the same column as the opening line, and one parameter per line with a trailing comma.
If you had a few arguments that were not long you could use just one line eg:
```
functionCall(
    a, b, c, d,
);
```

This and your other PRs could have this change in a lot of places but I wasn't about to make 15 comments just to be resolved without changes if this is some standard we have I have not heard about. Like if we need to minimize our line count or something.

note: I believe it is `metricHandlers` now but could be mistaken

miniscruff (changes requested):

extensions/replication/tasks/UpdateReplicationStatus.js
francoisferrand: Sorry on this one, but it is indeed a general pattern, and I don't think it is the place to start enforcing such kind of rule.

This pattern is common in most languages, so I guess most developers out there would be accustomed to it. And personally I think having less lines makes the code easier to read, as you can see the whole flow of things.

But in any case, and whatever the style we would settle on, I believe the only way to have something is to first get some sort of agreement, and then most importantly have it automated : e.g. rely on a top to (re)format the code according to stable rules (clang-format for exemple, but there are probably other tools dedicated to Js)

francoisferrand: > I believe it is metricHandlers now but could be mistaken

maybe it should have been `metricHandlers` indeed, but it is `metricsHandlers` everywhere...

miniscruff: ok

miniscruff (approved):

ghost (approved):

PR #2319, opened 2022-09-16
bugfix: BB-289 set notification versionId to null if object is a null…

welansari:
… version

Issue [BB-289](https://scality.atlassian.net/browse/BB-289)

"null" versions have an internal versionId that should not be used externally, as it's not possible to access the object using that versionId


ghost (approved):
lgtm

jbertran (approved):

PR #2320, opened 2022-09-16
bugfix: BB-290 update all the connector configuration instead of only…

welansari:
… the pipeline

Issue: [BB-290](https://scality.atlassian.net/browse/BB-290)

OplogPopulator does not update the configuration of old connectors, it only updates the buckets assigned to the connector which might cause an issue when having a connector with an outdated config that has the wrong output topic for example


jbertran (approved):

ghost (approved):
LGTM

PR #2321, opened 2022-09-16
Bugfix/bb 286 fix kafka log consumer not letting log reader push to topics

welansari:
Issue: [BB-286](https://scality.atlassian.net/browse/BB-286)

**Original Issue:**
Previous implementation of async in the logReader produces flakiness where some oplog entries can get processed by the filter function but not pushed into their respective topic.
The issue is only present for backbeat extensions that use an async filter function (only Bucket Notification for now).

**Changes:**
- Implement proper support for async `filter` functions (same as in 7.x branch)
- Stringified the oplog event value (bug)
- Implement filterAsync in the NotificationQueuePopulator
- Avoid crashing when having an issue consuming messages from kafka. This was done to avoid getting crashes at startup before the oplog topic get created.

Zenko Build: https://github.com/scality/Zenko/actions/runs/3376337223

### DIFF
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -1,4 +1,5 @@
 const assert = require('assert');
+const util = require('util');
 
 const { isMasterKey } = require('arsenal').versioning;
 const { usersBucket, mpuBucketPrefix } = require('arsenal').constants;
@@ -24,6 +25,8 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
         this.bnConfigManager = params.bnConfigManager;
         assert(this.bnConfigManager, 'bucket notification configuration manager'
             + ' is not set');
+        // callbackify functions
+        this._processObjectEntryCb = util.callbackify(this._processObjectEntry).bind(this);
     }
 
     /**
@@ -136,29 +139,30 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
      * filter
      *
      * @param {Object} entry - log entry
-     * @return {Promise|undefined} Promise|undefined
+     * @param {Function} cb - callback
+     * @return {undefined} Promise|undefined
      */
-    async filter(entry) {
+    filterAsync(entry, cb) {
         const { bucket, key, type } = entry;
         const value = entry.value || '{}';
         const { error, result } = safeJsonParse(value);
         // ignore if entry's value is not valid
         if (error) {
             this.log.error('could not parse log entry', { value, error });
-            return undefined;
+            return cb();
         }
         // ignore bucket operations, mpu's or if the entry has no bucket
         const isUserBucketOp = !bucket || bucket === usersBucket;
         const isMpuOp = key && key.startsWith(mpuBucketPrefix);
         const isBucketOp = bucket && result && this._isBucketEntry(bucket, key);
         if ([isUserBucketOp, isMpuOp, isBucketOp].some(cond => cond)) {
-            return undefined;
+            return cb();
         }
         // object entry processing - filter and publish
         if (key && result) {
-            return this._processObjectEntry(bucket, key, result, type);
+            return this._processObjectEntryCb(bucket, key, result, type, cb);
         }
-        return undefined;
+        return cb();
     }
 }
 
--- a/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
+++ b/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
@@ -57,7 +57,8 @@ class ListRecordStream extends stream.Transform {
                 method: 'ListRecordStream._transform',
                 data,
             });
-            changeStreamDocument = {};
+            // skipping the event
+            return callback(null, null);
         }
         const streamObject = {
             timestamp: new Date(data.timestamp),
@@ -68,7 +69,7 @@ class ListRecordStream extends stream.Transform {
                 value: changeStreamDocument.fullDocument && JSON.stringify(changeStreamDocument.fullDocument.value),
             }],
         };
-        callback(null, streamObject);
+        return callback(null, streamObject);
     }
 }
 
--- a/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
+++ b/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
@@ -65,7 +65,7 @@ class ListRecordStream extends stream.Transform {
             entries: [{
                 key: changeStreamDocument.documentKey && changeStreamDocument.documentKey._id,
                 type: this._getType(changeStreamDocument.operationType),
-                value: changeStreamDocument.fullDocument && changeStreamDocument.fullDocument.value,
+                value: changeStreamDocument.fullDocument && JSON.stringify(changeStreamDocument.fullDocument.value),
             }],
         };
         callback(null, streamObject);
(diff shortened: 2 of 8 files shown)


lib/queuePopulator/LogReader.js
```diff
@@ -522,6 +484,94 @@ class LogReader {
         return undefined;
     }
 
+    _processLogEntry(batchState, record, entry, cb) {
+        function _executeFilter(ext, entry, cb) {
+            if (typeof ext.filterAsync === 'function') {
+                ext.filterAsync(entry, cb);
+            } else {
+                ext.filter(entry);
+                process.nextTick(cb);
```
francoisferrand, line 493: this should be just `callbackify(ext.filter)`

francoisferrand: also, wondering if this should not be done just once, e.g. on startup process the list of extenions to build the list of callbacks : then just call them for processing

welansari: IMO i think that adds unnecessary complexity to the code especially since we are not waiting on any return value

francoisferrand, line 493: can we make this more seamless and support async methods, by just checking the return?

e.g. maybe something like this:
```
res = ext.filter(entry)
if (res instanceof Promise) {
   res.then(cb)
} else {
   process.nextTick(cb);
}
```

And then maybe we do not even need to switch to callbacks in the extensions (or support filterAsync)....

francoisferrand (approved):

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -136,29 +139,30 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
-    async filter(entry) {
+    filterAsync(entry, cb) {
         const { bucket, key, type } = entry;
         const value = entry.value || '{}';
         const { error, result } = safeJsonParse(value);
         // ignore if entry's value is not valid
         if (error) {
             this.log.error('could not parse log entry', { value, error });
-            return undefined;
+            return cb(null);
```
ghost: I see a mix of `cb()` and `cb(null)`, maybe we can just keep the first one everywhere?

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -65,7 +65,7 @@ class ListRecordStream extends stream.Transform {
             entries: [{
                 key: changeStreamDocument.documentKey && changeStreamDocument.documentKey._id,
                 type: this._getType(changeStreamDocument.operationType),
-                value: changeStreamDocument.fullDocument && changeStreamDocument.fullDocument.value,
+                value: changeStreamDocument.fullDocument && JSON.stringify(changeStreamDocument.fullDocument.value),
```
ghost, line 69: Maybe having this in a try/catch will prevent `JSON.stringify` errors if the object is circular. I believe it's not supposed to be the case, but I would say it's better to handle this case, what do you think?

lib/queuePopulator/LogReader.js
```diff
@@ -522,6 +484,94 @@ class LogReader {
...
+             * Value field normally contains the timestamp
+             * of the event, which is needed for the notification
+             * extension.
+             * That gets added here, we use the oplog event timestamp
+             */
+            if (
+                entry.type === 'delete' &&
+                ext.constructor.name === 'NotificationQueuePopulator'
+            ) {
+                entry.value = JSON.stringify({
```
ghost, line 508: Same comment here for the JSON.stringify

lib/queuePopulator/LogReader.js
```diff
@@ -522,6 +484,94 @@ class LogReader {
...
+        if (!record || !record.entries) {
+            return done();
+        }
+        this._setEntryBatch(entriesToPublish);
+        return this._filterEntries(batchState, record, err => {
+            this._unsetEntryBatch();
+            if (err) {
+                this.log.error('error filtering entries from log', {
+                    method: 'LogReader._processFilterEntries',
+                    error: err,
```
ghost, line 547: Can it be better to include `batchState` in the error log to ease the debugging? (same for other logs)

lib/queuePopulator/LogReader.js
```diff
@@ -522,6 +484,94 @@ class LogReader {
...
+                });
+                return done(err);
+            }
+            return done();
+        });
+    }
+
+    _processFilterEntries(batchState, done) {
+        const { currentRecords } = batchState;
+        if (!currentRecords || currentRecords.length === 0) {
```
ghost: Maybe safer like that, in case length is not defined?

```suggestion
        if (!currentRecords || !currentRecords.length) {
```

lib/queuePopulator/LogReader.js
```diff
@@ -522,6 +484,94 @@ class LogReader {
...
+            logStats.nbLogEntriesRead += 1;
+            return this._processLogEntry(batchState, record, entry, next);
+        }, cb);
+    }
+
+    _processFilterEntry(batchState, record, done) {
+        const { entriesToPublish } = batchState;
+        if (!record || !record.entries) {
+            return done();
+        }
```
ghost, line 539: Can be done before line 537

ghost (commented):
Some code-related comments.

The logic LGTM as well, I did not see anything wrong according to your ticket objectives.

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: We already parsed the string before, so i don't see a reason why it can't be re-parsed again. If the object happens to be circular the first try/catch should handle the error

The thing to change here i think is the way we handle those errors, i think returning an empty event is irrelevant as it will be ignored anyway, i changed it to skip the event completely

lib/queuePopulator/LogReader.js
welansari: A batch can contain up to 10 000 events, i'm afraid logging the `batchState` would just give too much information that might not even help when debugging.

One thing to note here also is that at the end of each batch we log the `batchState.logStat` which contains information about the events we read
https://github.com/scality/backbeat/blob/db9e3815983b046b77bc41847dbf201006b20d0a/lib/queuePopulator/LogReader.js#L311-L317

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
ghost: I see. Now my question is: do we really need to parse the same string several time?

welansari: The thing here is that we can not parse a part of the string. we only need the fullDocument field as a string (this is where the object metadata lies)

welansari: Oh i just noticed, i said re-parsed when i should have said stringified 😅.
The idea is if it got parsed with no issue then it should stringify with no errors

lib/queuePopulator/LogReader.js
welansari: No need i think inside are values we set

https://github.com/scality/Arsenal/blob/c343820cae2c7a533fb73046b2ad4c13cb77fc8b/lib/storage/metadata/mongoclient/ListRecordStream.js#L133-L137

ghost (approved):
LGTM

PR #2322, opened 2022-09-22
Fix retry on connection error

francoisferrand:
- Fix name of _getAndPutMultipartUpload retry
- Make ECONNRESET errors retryable
- Remove timeout from _getAndPutMultipartUpload retry

Issue: BB-292

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -98,7 +98,7 @@ class MultipleBackendTask extends ReplicateObject {
     _getAndPutMPUPart(sourceEntry, destEntry, part, uploadId, log, cb) {
         this.retry({
             actionDesc: 'stream part data',
-            logFields: { entry: sourceEntry.getLogInfo() },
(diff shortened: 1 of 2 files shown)


anurag4DSB (approved):

welansari (approved):

jonathan-gramain (approved):

nicolas2bert (approved):

nicolas2bert:
Could we backport those 3 commits into `development/7.10` branch? It has the same logic and will be helpful for S3C as well. Thanks :)

francoisferrand:
> Could we backport those 3 commits into development/7.10 branch? It has the same logic and will be helpful for S3C as well.

For the first 2, no problem.

For the last one, i am not sure we should:
* there is no `maxRetries` in 7.10, so removing the timeout means we would retry forever;
* but then this noTimeout option is only used in `MultipleBackendTask`, which is not used in S3C : so it may be clearer to keep the commit anyway, so that the PR is "identical" between all branches

PR #2325, opened 2022-09-30
BB-276: process messages to requeue objects

rachedbenmustapha:
Adds action to clear transition-in-progress-flag from objects, with sorbet and sorbet cli as primary users.

This code was a lot more elegant with async/await but I changed back to old async after spending way too much time debugging why one of the nested promises was never `await`-ed.


nicolas2bert (approved):

ghost (approved):

nicolas2bert (approved):

rachedbenmustapha:
ping

PR #2332, opened 2022-10-05
bugfix: BB-301 fix issue with notification processor not sending noti…

welansari:
…fications

Issue: [BB-301](https://scality.atlassian.net/browse/BB-301)

After recent changes to the Kafka Producer it now uses the Backbeat Producer's send method that needs the message to be stringified before it is called.

### DIFF
--- a/extensions/notification/queueProcessor/QueueProcessor.js
+++ b/extensions/notification/queueProcessor/QueueProcessor.js
@@ -193,7 +193,7 @@ class QueueProcessor extends EventEmitter {
         try {
             return this._getConfig(bucket, (err, notifConfig) => {
                 if (err) {
-                    this.logger.err('Error while getting notification configuration', {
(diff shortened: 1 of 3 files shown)


francoisferrand (approved):

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -232,7 +232,7 @@ class QueueProcessor extends EventEmitter {
                             // for Kafka keyed partitioning, to map a
                             // particular bucket and key to a partition
                             key: `${bucket}/${key}`,
-                            message,
+                            message: JSON.stringify(message),
```
ghost, line 235: Could it be helpful to add a new test here, to avoid such bug in the future?

ghost (approved):

PR #2336, opened 2022-10-10
feature: BB-306 add bucket notification metrics

welansari:
Issue: [BB-306](https://scality.atlassian.net/browse/BB-306)

### DIFF
--- a/extensions/notification/NotificationConfigManager.js
+++ b/extensions/notification/NotificationConfigManager.js
@@ -1,9 +1,9 @@
 const joi = require('joi');
 
+const { ZenkoMetrics } = require('arsenal').metrics;
 const LRUCache = require('arsenal').algorithms
     .cache.LRUCache;
 const MongoClient = require('mongodb').MongoClient;
-
 const constants = require('./constants');
 
 const paramsJoi = joi.object({
@@ -13,6 +13,50 @@ const paramsJoi = joi.object({
 
 const MAX_CACHED_ENTRIES = Number(process.env.MAX_CACHED_BUCKET_NOTIFICATION_CONFIGS)
     || 1000;
+
+// should equal true if config manager's cache was hit during a get operation
+const CONFIG_MANAGER_CACHE_HIT = 'cache_hit';
+// Type of operation performed on the cache
+const CONFIG_MANAGER_OPERATION_TYPE = 'op';
+
+const cacheUpdates = ZenkoMetrics.createCounter({
+    name: 'notification_config_manager_cache_updates',
+    help: 'Total number of cache updates',
+    labelNames: [
+        CONFIG_MANAGER_OPERATION_TYPE,
+    ],
+});
+
+const configGetLag = ZenkoMetrics.createHistogram({
+    name: 'notification_config_manager_config_get_sec',
+    help: 'Time it takes in seconds to get a bucket notification config from MongoDB',
+    labelNames: [
+        CONFIG_MANAGER_CACHE_HIT,
+    ],
+    buckets: [0.001, 0.01, 1, 10, 100, 1000],
+});
+
+const cachedBuckets = ZenkoMetrics.createGauge({
+    name: 'notification_config_manager_cached_buckets_count',
+    help: 'Total number of cached buckets in the notification config manager',
+});
+
+function onConfigManagerCacheUpdate(op) {
+    cacheUpdates.inc({
+        [CONFIG_MANAGER_OPERATION_TYPE]: op,
+    });
+    if (op === 'add') {
+        cachedBuckets.inc({});
+    } else if (op === 'delete') {
+        cachedBuckets.dec({});
+    }
+}
+
+function onConfigManagerConfigGet(cacheHit, delay) {
+    configGetLag.observe({
+        [CONFIG_MANAGER_CACHE_HIT]: cacheHit,
+    }, delay);
+}
 
 /**
  * @class NotificationConfigManager
@@ -102,14 +146,17 @@ class NotificationConfigManager {
             notificationConfiguration : null;
         switch (change.operationType) {
             case 'delete':
-                // if no bucket config was cached, nothing is done
-                this._cachedConfigs.remove(change.documentKey._id);
+                if (cachedConfig) {
+                    this._cachedConfigs.remove(change.documentKey._id);
+                    onConfigManagerCacheUpdate('delete');
+                }
                 break;
             case 'replace':
             case 'update':
                 if (cachedConfig) {
                     // add() replaces the value of an entry if it exists in cache
                     this._cachedConfigs.add(change.documentKey._id, bucketNotificationConfiguration);
+                    onConfigManagerCacheUpdate('update');
                 }
                 break;
             default:
@@ -226,9 +273,12 @@ class NotificationConfigManager {
      * @return {Object|undefined} - configuration if available or undefined
      */
     async getConfig(bucket) {
+        const startTime = Date.now();
         // return cached config for bucket if it exists
         const cachedConfig = this._cachedConfigs.get(bucket);
         if (cachedConfig) {
+            const delay = (Date.now() - startTime) / 1000;
+            onConfigManagerConfigGet(true, delay);
             return cachedConfig;
         }
         try {
@@ -238,6 +288,9 @@ class NotificationConfigManager {
                 bucketMetadata.value.notificationConfiguration) || undefined;
             // caching the bucket configuration
             this._cachedConfigs.add(bucket, bucketNotificationConfiguration);
+            const delay = (Date.now() - startTime) / 1000;
+            onConfigManagerConfigGet(false, delay);
+            onConfigManagerCacheUpdate('add');
             return bucketNotificationConfiguration;
         } catch (err) {
             this._logger.error('An error occured when getting notification ' +
--- a/extensions/notification/destination/KafkaNotificationDestination.js
+++ b/extensions/notification/destination/KafkaNotificationDestination.js
@@ -1,5 +1,36 @@
 const NotificationDestination = require('./NotificationDestination');
 const KafkaProducer = require('./KafkaProducer');
+const { ZenkoMetrics } = require('arsenal').metrics;
+
+const notificationSize = ZenkoMetrics.createCounter({
+    name: 'notification_queue_processor_notification_size',
+    help: 'Total size of the notifications that were sent',
+    labelNames: ['target'],
+});
+
+const deliveryLag = ZenkoMetrics.createHistogram({
+    name: 'notification_queue_processor_notification_delivery_delay_sec',
+    help: 'Difference between the time a notification is sent and when it gets delivered',
+    labelNames: ['target', 'status'],
+    buckets: [0.001, 0.01, 1, 10, 100, 1000, 10000],
+});
+
+function onNotificationSent(target, status, messages, delay) {
+    if (status === 'success') {
+        // Update total size of notifications sent
+        const messageSize = new TextEncoder().encode(JSON.stringify(messages)).length;
+        notificationSize.inc({
+            target,
+        }, messageSize);
+    }
+    if (delay) {
+        // update notification delivery delay
+        deliveryLag.observe({
+            target,
+            status,
+        }, delay);
+    }
+}
 
 class KafkaNotificationDestination extends NotificationDestination {
     /**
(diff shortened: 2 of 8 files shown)


extensions/notification/NotificationConfigManager.js
```diff
@@ -1,19 +1,73 @@
...
+    origin: joi.string().required(),
     logger: joi.object().required(),
 }).required();
 
 const MAX_CACHED_ENTRIES = Number(process.env.MAX_CACHED_BUCKET_NOTIFICATION_CONFIGS)
     || 1000;
 
+// Config manager is used in both the QueuePopulator and QueueProcessor
+// we use this label to distinguish between the two
+const CONFIG_MANAGER_LABEL_ORIGIN =  'origin';
```
francoisferrand: do we actually need this?
We will already be able to differentiate by looking at the job/service/container labels (which are added by prometheus)

extensions/notification/NotificationConfigManager.js
```diff
@@ -102,14 +157,17 @@ class NotificationConfigManager {
             notificationConfiguration : null;
         switch (change.operationType) {
             case 'delete':
-                // if no bucket config was cached, nothing is done
-                this._cachedConfigs.remove(change.documentKey._id);
+                if (cachedConfig) {
+                    this._cachedConfigs.remove(change.documentKey._id);
+                    onConfigManagerCacheUpdate(this._origin, 'delete');
```
francoisferrand: 'delete' is used here, but 'remove' is expected in `onConfigManagerCacheUpdate`

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,19 +1,73 @@
...
+const cachedBuckets = ZenkoMetrics.createGauge({
+    name: 'notification_config_manager_cached_buckets_count',
+    help: 'Total number of cached buckets in the notification config manager',
+    labelNames: [
+        CONFIG_MANAGER_LABEL_ORIGIN,
+    ],
+});
+
+function onConfigManagerCacheUpdate(origin, op) {
+    cacheUpdates.inc({
```
francoisferrand, line 45: the op shoud probably be added as a label, to better see the 'repartitions' of updates

extensions/notification/destination/KafkaNotificationDestination.js
```diff
@@ -1,5 +1,42 @@
 const NotificationDestination = require('./NotificationDestination');
 const KafkaProducer = require('./KafkaProducer');
+const { ZenkoMetrics } = require('arsenal').metrics;
+
+const notificationsSent = ZenkoMetrics.createCounter({
+    name: 'notification_queue_processor_notifications_sent',
```
francoisferrand: "sent" is kind of redundant : bucket notificaiton is only about sending notifications.

maybe it would be clearer with 'count' in the name, to more explicitely describe what this is about:
`notification_queue_processor_notifications_sent`

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -5,13 +5,38 @@ const Logger = require('werelogs').Logger;
...
 const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
 const NotificationDestination = require('../destination');
 const configUtil = require('../utils/config');
 const messageUtil = require('../utils/message');
 const NotificationConfigManager = require('../NotificationConfigManager');
 
+const processedEvents = ZenkoMetrics.createCounter({
+    name: 'notification_queue_processor_processed_events',
+    help: 'Total number of successfully processed events',
+    labelNames: ['target', 'eventType'],
```
francoisferrand, line 19: just 'event' would be clearer i think

lib/queuePopulator/LogReader.js
```diff
@@ -290,14 +290,22 @@ class LogReader {
             timeoutMs: params.timeoutMs,
             logger: this.log.newRequestLogger(),
         };
-
+        const startTime = Date.now();
```
francoisferrand: startTime is already in the `batchState`

lib/queuePopulator/LogReader.js
```diff
@@ -290,14 +290,22 @@ class LogReader {
...
         async.waterfall([
             next => this._processReadRecords(params, batchState, next),
             next => this._processPrepareEntries(batchState, next),
             next => this._processPublishEntries(batchState, next),
             next => this._processSaveLogOffset(batchState, next),
         ],
             err => {
+                if (this._metricsHandler && this._metricsHandler.processingLag) {
```
francoisferrand: not sure if/how we should handle errors here : not sure if/when errors are expected, but it may very well be that either we don't care about duration in case of error OR we want to at least differentiate the duration of errors...

the easy way out is to just add some sort of 'result' label, so that we can get either the aggregated, succesful or failed metrics.

lib/queuePopulator/LogReader.js
```diff
@@ -290,14 +290,22 @@ class LogReader {
...
+                    const processingTime = (Date.now() - startTime) / 1000;
+                    const origin = process.env.BACKBEAT_QUEUEPOPULATOR_EXTENSIONS === 'notification' ?
+                    'notification' : 'replication';
+                    this._metricsHandler.processingLag.observe({
```
francoisferrand: i am wondering if this metric is really specific, and/or if we do not already have something similar/related...

* If it is in LogReader, it should be generic for each QueuePopulator. To make it speciifc to NotificationPopulator, it should be in the NotificationPopulatorExtension...
* Since it is (theorically) possible to have multiple extensions, as is the case with the "usual" QueuePopulator, I am wondering if this "batch" request duration is useful (at QueuePopulator level: e.g. just check how fast we go through the batch?) or needed for bucket notification (what specifically do we want to monitor for bucket notification, e.g. what could go wrong?)

In any case, it seems `this._metricsHandler && this._metricsHandler.processingLag` should be true, even if some tests need to be updated?

welansari: I decided to remove this metric and to instead watch how much time it takes to get a bucket's config from MongoDB, this was what i originally intended to meter with that metric

extensions/notification/NotificationConfigManager.js
```diff
@@ -14,6 +14,49 @@ const paramsJoi = joi.object({
...
+        [CONFIG_MANAGER_OPERATION_TYPE]: op,
+    });
+    if (op === 'add') {
+        cachedBuckets.inc({});
+    } else if (op === 'delete') {
+        cachedBuckets.dec({});
+    }
+}
+
+function onConfigManagerConfigGet(hitMiss, delay) {
```
francoisferrand: not sure about the semantics here: the label is "cache_hit" (e.g. true if it was in the cache) while the parameter is "hitMiss" (e.g. true if it was not in the cache)...

Looking at call sites, the label is correct, and param should be renamed

extensions/notification/destination/KafkaNotificationDestination.js
francoisferrand: The metric was _removed_ ?
Counting the notifications (to show the number of notifications per seconds) seems quite useful though, my earlier comment was really just about the _name_ of the metric.

extensions/notification/destination/KafkaNotificationDestination.js
```diff
@@ -1,5 +1,33 @@
...
+});
+
+function onNotificationSent(target, status, messages, delay) {
+    // Update total size of notifications sent
+    const messageSize = new TextEncoder().encode(JSON.stringify(messages)).length;
+    notificationSize.inc({
+        target,
+    }, messageSize);
+    // update notification delivery delay
+    if (delay) {
```
francoisferrand, line 26: in the code, this is passed as "-1" : which is a weird value for a delay, and will break the "total lag" counter (must be monotically increasing!) and average computations...

so the condition should be `delay>0`

extensions/notification/destination/KafkaNotificationDestination.js
```diff
@@ -1,5 +1,33 @@
...
+const deliveryLag = ZenkoMetrics.createSummary({
+    name: 'notification_queue_processor_notification_delivery_delay_sec',
+    help: 'Difference between the time a notification is sent and when it gets delivered',
+    labelNames: ['target', 'status'],
+});
+
+function onNotificationSent(target, status, messages, delay) {
+    // Update total size of notifications sent
+    const messageSize = new TextEncoder().encode(JSON.stringify(messages)).length;
+    notificationSize.inc({
```
francoisferrand: not sure if/how this metric should increase in case of failure : since we are not succeeding to publish the notif, the size may not really matter...
so i suggest either adding the 'status' label here as well, or increasing only on success

lib/queuePopulator/QueuePopulator.js
```diff
@@ -87,6 +92,7 @@ const metricsHandler = {
     logReadOffset: wrapGaugeSet(logReadOffsetMetric, defaultLabels),
     logSize: wrapGaugeSet(logSizeMetric, defaultLabels),
     logTimestamp: wrapGaugeSet(logTimestamp, defaultLabels),
+    notificationProcessedEvents,
```
francoisferrand: the "usage" here is to wrap the metric (using one of the wrapXXXX functions), so that the metricsHandler contains only functions which can be called directly : and instead of `this._metricsStore.notificationProcessedEvents.inc()` you will simply do `this._metricsStore.notificationProcessedEvents()`

we should stick to this approch, for consistency.

extensions/notification/destination/KafkaNotificationDestination.js
welansari: We can use the `deliveryLag` metric to do the same, It's a summary so we can just get the count

extensions/notification/NotificationConfigManager.js
```diff
@@ -14,6 +14,50 @@ const paramsJoi = joi.object({
...
+    ],
+});
+
+const configGetLag = ZenkoMetrics.createHistogram({
+    name: 'notification_config_manager_config_get_sec',
+    help: 'Time it takes in seconds to get a bucket notification config from MongoDB',
+    labelNames: [
+        CONFIG_MANAGER_CACHE_HIT,
+    ],
+    buckets: [0.0000001, 0.000001, 0.00001, 0.0001, 0.001, 0.01, 1, 10, 100, 1000, 10000],
```
francoisferrand: there is probably some cleanup to be done in this list of buckets : i doubt we would have GET operations take 10'000 seconds or 0.1 nano-second...

extensions/notification/NotificationConfigManager.js
```diff
@@ -14,6 +14,50 @@ const paramsJoi = joi.object({
...
+        [CONFIG_MANAGER_OPERATION_TYPE]: op,
+    });
+    if (op === 'add') {
+        cachedBuckets.inc({});
+    } else if (op === 'delete') {
+        cachedBuckets.dec({});
+    }
+}
+
+function onConfigManagerConfigGet(cacheHit, delay) {
```
francoisferrand, line 55: should be `onConfigManagerCacheGet` ?

extensions/notification/destination/KafkaNotificationDestination.js
```diff
@@ -1,5 +1,36 @@
...
+    name: 'notification_queue_processor_notification_size',
+    help: 'Total size of the notifications that were sent',
+    labelNames: ['target'],
+});
+
+const deliveryLag = ZenkoMetrics.createHistogram({
+    name: 'notification_queue_processor_notification_delivery_delay_sec',
+    help: 'Difference between the time a notification is sent and when it gets delivered',
+    labelNames: ['target', 'status'],
+    buckets: [0.0000001, 0.000001, 0.00001, 0.0001, 0.001, 0.01, 1, 10, 100, 1000, 10000],
```
francoisferrand: there is probably some cleanup to be done in this list of buckets : i doubt we would have GET operations take 10'000 seconds or 0.1 nano-second...

francoisferrand (approved):

extensions/notification/NotificationConfigManager.js
```diff
@@ -14,6 +14,50 @@ const paramsJoi = joi.object({
...
+    ],
+});
+
+const configGetLag = ZenkoMetrics.createHistogram({
+    name: 'notification_config_manager_config_get_sec',
+    help: 'Time it takes in seconds to get a bucket notification config from MongoDB',
+    labelNames: [
+        CONFIG_MANAGER_CACHE_HIT,
+    ],
+    buckets: [0.00001, 0.0001, 0.001, 0.01, 1, 10, 100, 1000],
```
ghost: I am not sure that we will get something below 0.001 using `Date.now()` as the level of details is "milliseconds"

ghost (approved):
LGTM, just one comment on the resolution for the lag

extensions/notification/NotificationConfigManager.js
welansari: makes sense, fixed it

PR #2337, opened 2022-10-14
BB-295 - bootstrap ts compilation

jbertran:
Outcome so far:

* TypeScript is integrated in the build environment, and compiles to the dist folder
* Jest is configured to use the mixed sources with on-the-fly compilation, and runs all tests that were run with mocha
* Some test scripts pass:
  * unit
  * routes
  * retry

Remaining work is simply ensuring all other test suites pass (bonus: make them separate parallel steps in the build process, such that the work doesn’t need to be sequential).

Zenko Build: https://github.com/scality/Zenko/actions/runs/3822509132


Dockerfile
```diff
@@ -28,11 +28,12 @@ RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSI
     && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
     && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
 
-COPY package.json yarn.lock /usr/src/app/
+COPY . /usr/src/app
```
francoisferrand, line 35: the goal here is to perform yarn install first, without depending on changing to the code to allow better layer caching : hence it is required to copy only package.json and yarn.lock

To deal with the build, you should thus do it in 2 steps here:

```
COPY package.json yarn.lock /usr/src/app/
RUN yarn install ...
COPY . /usr/src/app
RUN yarn build
```

Dockerfile
```diff
@@ -45,7 +46,8 @@ RUN apt-get update && \
 WORKDIR /usr/src/app
 
 # Keep the .git directory in order to properly report version
-COPY . /usr/src/app
+COPY . ./
```
francoisferrand: not needed? since data is copied from `dist/` as well..

tsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
```
tmacro, line 1: One of the things I found while standing up Scuba's TS config was https://www.npmjs.com/package/@tsconfig/node16
It basically just a base you can extend, tuned for a particular use case.
I used it here https://github.com/scality/scuba/blob/development/1.0/tsconfig.json
The options can be set manually of course, but I think it's nice to have a maintained base to build off of.

tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+  "compilerOptions": {
+      "target": "es6",
```
tmacro: I'd suggest setting this to `es2021`. Among other things it uses native node promises instead of a polyfill.
```suggestion
      "target": "es2021",
```

tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+      "checkJs": false,
+      "outDir": "dist",
+      "esModuleInterop": true,
+      "forceConsistentCasingInFileNames": true,
+      "strict": true,
+      "declaration": true,
+      "noImplicitAny": false,
+      "noEmitOnError": false,
+      "sourceMap": true,
+      "skipLibCheck": true
```
tmacro: Need to add the es2021 lib to use the target
```suggestion
      "skipLibCheck": true,
      "lib": ["es2021"]
```

package.json
```diff
@@ -61,21 +64,40 @@
     "node-rdkafka": "^2.12.0",
     "node-schedule": "^1.2.0",
     "node-zookeeper-client": "^1.1.3",
+    "typescript": "^4.8.4",
```
tmacro: This should be listed as a dev dependency

package.json
```diff
@@ -44,6 +46,7 @@
   },
   "homepage": "https://github.com/scality/backbeat#readme",
   "dependencies": {
+    "@types/node": "^18.11.9",
```
tmacro: This should be listed as a dev dependency

Dockerfile
```diff
@@ -28,11 +28,12 @@ RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSI
     && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
     && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
 
-COPY package.json yarn.lock /usr/src/app/
+COPY . /usr/src/app
 RUN yarn install --ignore-engines --frozen-lockfile --production --network-concurrency 1 \
     && rm -rf /var/lib/apt/lists/* \
     && rm -rf ~/.node-gyp \
     && rm -rf /tmp/yarn-*
+RUN yarn build
```
tmacro: When typescript is moved to a dev dependency this needs to be split if you want to continue copying the node_modules dir from the builder.
Also we can get rid of the `rm`s since the builder image will be trashed at the end anyway.

```suggestion
RUN yarn install --ignore-engines --frozen-lockfile --network-concurrency 1

RUN yarn build

RUN rm -rf node_modules \
    && yarn install --production --frozen-lockfile --ignore-optional
```

Dockerfile
tmacro: Only thing that needs to be copied is package.json if we intend to use yarn to launch things
```suggestion
COPY ./package.json ./
```

tmacro (changes requested):

tests/functional/replication/queueProcessor.spec.js
```diff
@@ -877,9 +877,9 @@ describe('queue processor functional tests with mocking', () => {
         httpServer.listen(7777);
     });
 
-    after(done => {
+    afterAll(done => {
         httpServer.close();
-        async.parallel([
+        async.series([
```
rachedbenmustapha, line 881: This doesn't seem to be strictly related to typescript/jest, is there a reason for this change?

tests/functional/notification/NotificationConfigManager.spec.js
```diff
@@ -42,7 +57,7 @@ describe('NotificationConfigManager ::', () => {
         manager = new NotificationConfigManager(params);
         const getCollectionStub = sinon.stub().returns({
             // mock change stream
-            watch: () => new events.EventEmitter(),
+            watch: () => new MockChangeStream(),
```
rachedbenmustapha, line 60: This doesn't seem to be strictly related to typescript/jest, is there a reason for this change?

tests/unit/OffsetLedger.spec.js
```diff
@@ -1,6 +1,6 @@
 const assert = require('assert');
 
-const OffsetLedger = require('../../lib/OffsetLedger');
+const OffsetLedger = require('../../lib/OffsetLedger').default;
```
rachedbenmustapha, line 3: Naming `default` explicitly feels off; isn't there some more typescript-native way of importing the default export without explicitly naming it?

tests/utils/syntheticbucketd/Dockerfile
```diff
@@ -1,10 +1,7 @@
 FROM node:16
 
 RUN mkdir /app
-COPY package.json /app
-COPY yarn.lock /app
+COPY * /app
```
rachedbenmustapha: Why this change? It invalidates the layer if anything changes even if unrelated to dependencies, and triggers an avoidable `yarn install`.

tests/functional/lib/BackbeatConsumer.spec.js
```diff
@@ -122,7 +121,7 @@ describe('BackbeatConsumer main tests', () => {
                 setTimeout(() => {
                     _checkZkMetrics(() => {
                         consumeCb();
-                        consumer.unsubscribe();
+                        consumer.pause();
```
rachedbenmustapha, line 126: Why this change (and the `subscribe`/`resume` one)? Was the test wrong before?

tests/unit/RoleCredentials.spec.js
```diff
@@ -55,100 +55,133 @@ function _assertCredentials(err, roleCredentials, cb) {
...
-    });
+    }, 10000);
 
-    it('should refresh credentials a bit before expiration', function test(done) {
-        this.timeout(10000);
+    it('should refresh credentials a bit before expiration', done => {
         roleCredentials.get(err => {
             if (err) {
-                return done(err);
+                done(err);
```
rachedbenmustapha, line 158: I think the missing return changes the test's flow.

tests/unit/RoleCredentials.spec.js
```diff
@@ -55,100 +55,133 @@ function _assertCredentials(err, roleCredentials, cb) {
...
         });
-    });
+    }, 10000);
 
-    it('should refresh credentials upon expiration', function test(done) {
-        this.timeout(10000);
+    it('should refresh credentials upon expiration', done => {
         roleCredentials.get(err => {
             if (err) {
-                return done(err);
```
rachedbenmustapha, line 108: I think the missing return changes the test's flow.

tests/unit/RoleCredentials.spec.js
```diff
@@ -55,100 +55,133 @@ function _assertCredentials(err, roleCredentials, cb) {
...
+            _assertCredentials(err, roleCredentials, done)
+        );
     });
 
-    it('should use same credentials if not expired or about to expire', function test(done) {
-        this.timeout(10000);
+    it('should use same credentials if not expired or about to expire', done => {
         roleCredentials.get(err => {
             if (err) {
-                return done(err);
```
rachedbenmustapha, line 86: I think the missing return changes the test's flow.

lib/api/BackbeatAPI.js
```diff
@@ -101,31 +101,31 @@ class BackbeatAPI {
             ingestionSites: this._ingestionSites,
             internalStart: this._internalStart,
         };
-        const metrics = new Metrics(metricsConfig, logger);
+        this._metrics = new Metrics(metricsConfig, logger);
```
rachedbenmustapha, line 113: Why is this needed?

francoisferrand: the objects needs to be 'stored' so the resources can be closed (in order not to leak) at the end of the tests:

```
bbapi._metrics.disconnect();
```

tests/unit/OffsetLedger.spec.js
francoisferrand: I think this is the way, the same is done in arsenal.

Maybe this is due to the dynamic transpiling of library's TS code, in the tests...
And the proper solution would be to migrate the test to typescript (or possibly use ES6 `import` instead of `require`), but this is not in scope of this ticket: which only deals with "bootstraping", i.e. having the toolchain in place so that it is _possible_ to start writing components in TS)

tmacro (approved):

francoisferrand:
Replaced with https://github.com/scality/backbeat/pull/2849

PR #2338, opened 2022-10-14
Feature: BB-307 add bucket notification dashboard

welansari:
Issue: [BB-307](https://scality.atlassian.net/browse/BB-307)

Screenshot 2022-10-21 at 17 28 17 (2)

### DIFF
--- a/monitoring/notification/dashboard.py
+++ b/monitoring/notification/dashboard.py
@@ -1,0 +1,518 @@
+from grafanalib.core import (
+    ConstantInput,
+    DataSourceInput,
+    RowPanel,
+    Threshold,
+    StateTimeline,
+    Templating,
+    Template,
+    Heatmap,
+    HeatmapColor,
+    YAxis,
+)
+from grafanalib import formatunits as UNITS
+from scalgrafanalib import(
+    layout,Target,
+    TimeSeries,
+    Dashboard,
+    Stat,
+    GaugePanel,
+    metrics,
+    Tooltip,
+)
+
+LABEL_EXPR = {
+    'notification-processors': '.*-(.*)-headless',
+    'topics': '.*backbeat-notification-(.*).*'
+}
+
+DEFAULT_JOB_PREFIX = 'artesca-data-'
+
+class Metrics:
+    DELIVERY_LAG = metrics.BucketMetric(
+        'notification_queue_processor_notification_delivery_delay_sec',
+        'target', 'job', 'status', namespace='${namespace}',
+    ).with_defaults(
+        'job=~"${notification_processor}"',
+        'target=~"${target}"',
+    )
+
+    CONSUMED_EVENTS = metrics.CounterMetric(
+        'notification_queue_populator_event',
+        job='${job_notification_producer}', namespace='${namespace}'
+    )
+
+    NOTIF_SIZE, PROCESSOR_EVENTS = [
+        metrics.CounterMetric(
+            name, 'job', 'target', *extraLabels, namespace='${namespace}'
+        ).with_defaults(
+            'job=~"${notification_processor}"',
+            'target=~"${target}"',
+        ) for name, extraLabels in {
+            'notification_queue_processor_notification_size': [],
+            'notification_queue_processor_event': ['eventType'],
+        }.items()
+    ]
+
+    CACHE_LAG = metrics.BucketMetric(
+        'notification_config_manager_config_get_sec',
+        'cache_hit', 'job', namespace='${namespace}',
+    ).with_defaults(
+        'job=~"${notification_processor}|${job_notification_producer}"'
+    ) 
+
+    CACHE_UPDATES, CACHED_BUCKETS = [
+        metrics.CounterMetric(
+            name, 'job', *extraLabels, namespace='${namespace}'
+        ).with_defaults(
+            'job=~"${notification_processor}|${job_notification_producer}"'
+        ) for name, extraLabels in {
+            'notification_config_manager_cache_updates': ['op'],
+            'notification_config_manager_cached_buckets_count': [],
+        }.items()
+    ]
+
+producer_up_current = Stat(
+    title='Notification producer',
+    description='Number of active notification producer instances',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='background',
+    noValue='0',
+    reduceCalc='last',
+    targets=[Target(
+        expr='\n'.join([
+            'sum(up{',
+            '    namespace="${namespace}",',
+            '    job="${job_notification_producer}"',
+            '})',
+        ]),
+    )],
+    thresholds=[
+        Threshold('red', 0, 0.),
+        Threshold('green', 1, 1.),
+    ])
+
+processors_up_current = StateTimeline(
+    title='Notification Processors',
+    description='Active notification processor instances',
+    dataSource='${DS_PROMETHEUS}',
+    legendDisplayMode='hidden',
+    showValue='never',
+    targets=[Target(
+        expr='\n'.join([
+            'label_replace(',
+            'sum by(job) (up{',
+            '    namespace="${namespace}",',
+            '    job=~"${notification_processor}"',
+            '}), "job", "$1", "job", "' + LABEL_EXPR['notification-processors'] +'")',
+        ]),
+        legendFormat='{{job}}',
+    )],
+    thresholds=[
+        Threshold('red', 0, 0.),
+        Threshold('green', 1, 1.),
+    ])
+
+zookeeper_quorum = Stat(
+    title='Zookeeper Quorum',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='background',
+    noValue='0',
+    reduceCalc='last',
+    targets=[Target(
+        expr='\n'.join([
+            'max(',
+            'quorum_size{',
+            '  namespace="${namespace}",',
+            '  job="${job_zookeeper}"',
+            '})',
+        ]),
+    )],
+    thresholds=[
+        Threshold('green', 1, 0.),
+    ])
+
+kafka_brokers = Stat(
+    title='Kafka Brokers',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='background',
+    noValue='0',
+    reduceCalc='last',
+    targets=[Target(
+        expr='\n'.join([
+            'count(',
+            'kafka_server_replicamanager_leadercount{',
+            ' namespace="${namespace}",',
+            ' job="${job_kafka}"',
+            '})',
+        ]),
+    )],
+    thresholds=[
+        Threshold('green', 1, 0.),
+    ])
+
+kafka_lag = TimeSeries(
+    title='Notification Topics Lag',
+    description='Consumer lag of each of the notification topics',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    legendDisplayMode='table',
+    legendPlacement='right',
+    unit=UNITS.SECONDS,
+    targets=[Target(
+        expr='\n'.join([
+            'label_replace(',
+            'max by(topic) (',
+            '    kafka_consumergroup_group_lag_seconds{'
+            '       topic=~"${notification_topics}"',
+            '}), "topic", "$1", "topic", "' + LABEL_EXPR['topics'] +'")',
+        ]),
+        legendFormat='{{topic}}',
+    )])
+
+success_rate = GaugePanel(
+    title='Success Rate',
+    dataSource='${DS_PROMETHEUS}',
+    min=0,
+    max=1,
+    noValue='-',
+    format=UNITS.PERCENT_UNIT,
+    targets=[Target(
+        expr='\n'.join([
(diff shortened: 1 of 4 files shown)


monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,521 @@
...
+    lineInterpolation='smooth',
+    lineWidth=3,
+    fillOpacity=10,
+    spanNulls=True,
+    legendDisplayMode='table',
+    legendPlacement='right',
+    targets=[Target(
+        expr='\n'.join([
+            'sum by(target) (',
+            '    notification_queue_processor_notifications_sent{status="success"}',
```
francoisferrand: this is missing selectors (e.g. job)

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,521 @@
...
+    lineWidth=3,
+    fillOpacity=10,
+    spanNulls=True,
+    legendDisplayMode='table',
+    legendPlacement='right',
+    targets=[Target(
+        expr='\n'.join([
+            'sum by(target) (',
+            'rate(',
+            '    notification_queue_processor_notifications_sent{status="success"}[$__rate_interval]',
```
francoisferrand: same (and many afterwards)

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,521 @@
...
+    title='Notification Delivery Rate',
+    description='Number of delivered notifications in a given time',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    lineWidth=3,
+    fillOpacity=10,
+    spanNulls=True,
+    legendDisplayMode='table',
+    legendPlacement='right',
```
francoisferrand, line 233: missing unit : op/s ?

monitoring/notification/dashboard.py
francoisferrand: this is a counter, it should not be used directly : if you want to display the number of notiifcation, it must pass through the `increase` function :

```
sum by(target) (
  increase(
    notification_queue_processor_notifications_sent{status="success"}[$__rate_interval]
  )
)
```

This information will not be fully accurate and present some rounding issues; you may thus want to limit decimals (to 0) to avoid having a fractional number of events.

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,521 @@
...
+    unit=UNITS.SECONDS,
+    targets=[Target(
+        expr='\n'.join([
+            'avg(',
+            'rate(',
+            '    queue_populator_processing_delay_sec{origin="notification"}[$__rate_interval]',
```
francoisferrand: I don't think this formula gives the expected value.

Durations metrics should use histogram/summary type, so that:
* we get easily the cumulated duration and number of events (allowing to compute the average duration: total/count)
* we get metrics by buckets, allowing to see how the distribution of processing time and compute percentiles

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,521 @@
...
+    title='Cached Notification Configurations',
+    description='Total number of notification configurations cached',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='value',
+    noValue='0',
+    reduceCalc='last',
+    targets=[Target(
+        expr='\n'.join([
+            'sum by (origin) (',
+            '    notification_config_manager_cached_buckets_count',
```
francoisferrand: is this one a `counter` or a `gauge` ?
if it is a counter, the expr must use rate() or increase()

Also, i am wondering if we should use a sum() here : we don't really care about the "total", but about the size in each pod. Maybe not aggregating pods would thus be preferable.

francoisferrand (changes requested):
You should use the `Metric`(s) classes from scalgrafanalib, as it makes the expression easier to write (less string manipulation, syntax highlighting, shorter names, default parameters) and safer (validation of labels...)

Also, please provide a screenshot(s) of the dashboard : this makes it easier to go through the code.

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+}
+
+DEFAULT_JOB_PREFIX = 'artesca-data-'
+
+class Metrics:
+    DELIVERY_LAG = metrics.BucketMetric(
+        'notification_queue_processor_notification_delivery_delay_sec',
+        'target', 'job', 'status', namespace='${namespace}',
+    ).with_defaults(
+        'status="success"',
```
francoisferrand: since this is an equality, it can be written more simply inline with the label defintion (line 34), like what you did for namespace

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+        'job=~"${notification_processor}"',
+    )
+
+    CONSUMED_EVENTS = metrics.CounterMetric(
+        'notification_queue_populator_event',
+        'job', namespace='${namespace}'
+    ).with_defaults(
+        'job="${job_notification_producer}"'
```
francoisferrand: same

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+            Template(
+                dataSource='${DS_PROMETHEUS}',
+                label='Notification processor',
+                name='notification_processor',
+                query='label_values(' + Metrics.PROCESSOR_EVENTS.raw('job=~"${job_notification_processors}.*"') + ', job)',
+                regex='/^(?<value>(' + DEFAULT_JOB_PREFIX + ')?(backbeat-notification-processor-)?(?<text>.*?)(-headless)?)$/',
+                allValue='${job_notification_processors}.*',
+                includeAll=True,
+                multi=True,
+            )
```
francoisferrand: may be nice to have a "target" variable here?
(and would just need to add a default value for target label in each metric : `target=~"$target"`)

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+        'job="${job_notification_producer}"'
+    )
+
+    NOTIF_SIZE, PROCESSOR_EVENTS = [
+        metrics.CounterMetric(
+            name, 'job', *extraLabels, namespace='${namespace}'
+        ).with_defaults(
+            'job=~"${notification_processor}"'
+        ) for name, extraLabels in {
+            'notification_queue_processor_notification_size': ['target'],
```
francoisferrand: 'target' is in each metrics' `extraLabels` : maybe should be moved out of this list

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+success_rate = GaugePanel(
+    title='Success Rate',
+    dataSource='${DS_PROMETHEUS}',
+    min=0,
+    max=1,
+    noValue='-',
+    format=UNITS.PERCENT_UNIT,
+    targets=[Target(
+        expr='\n'.join([
+            'sum(rate(',
```
francoisferrand: it now fits on a line : best to make it readable here (in the source) than in grafana...

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+            ' ' + Metrics.DELIVERY_LAG.count(),
+            ')) /',
+            'sum(rate(',
+            ' ' + Metrics.DELIVERY_LAG.count('status=~".*"'),
```
francoisferrand: it may be faster (in grafana/prometheus) to do it the other way: e.g. instead of asking grafana/prometheus to filter "all values" with regex, specifying no 'status' label will pick up everything...

that would mean removing the default value for DELIVERY_LAG's status label, and give the following expression (which is also more explicit i think):

```python
expr = '\n'.join([
   'sum(rate(' + Metrics.DELIVERY_LAG.count(status='success') + '))',
   '/',
   'sum(rate(' + Metrics.DELIVERY_LAG.count() + '))',
])
```

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+    colorMode='background',
+    noValue='0',
+    reduceCalc='last',
+    format=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='\n'.join([
+            'sum(rate(',
+            '' + Metrics.DELIVERY_LAG.count(),
+            '))',
+        ]),
```
francoisferrand:
```suggestion
        expr='sum(rate(' + Metrics.DELIVERY_LAG.count() + '))',
```

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+    colorMode='background',
+    noValue='0',
+    reduceCalc='last',
+    format=UNITS.BYTES_SEC,
+    targets=[Target(
+        expr='\n'.join([
+            'sum(rate(',
+            '' + Metrics.NOTIF_SIZE(),
+            '))',
+        ]),
```
francoisferrand:
```suggestion
        expr='sum(rate(' + Metrics.NOTIF_SIZE() + '))',
```

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    legendDisplayMode='table',
+    legendPlacement='right',
+    targets=[Target(
+        expr='\n'.join([
+            'label_replace(',
+            'sum('
+            '  notification_config_manager_cached_buckets_count{',
```
francoisferrand: should use Metrics

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+            '  job=~"${notification_processor}|${job_notification_producer}"'
+            '}) by (job), "job", "$1", "job", "' + LABEL_EXPR['notification-processors'] +'")',
+        ]),
+        legendFormat='{{job}}',
+    )],
+    thresholdType='absolute',
```
francoisferrand: this is the default, no need to make it explicit

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+    targets=[Target(
+        expr='\n'.join([
+            'label_replace(',
+            'sum by (job) (rate (',
+            ' ' + Metrics.CACHE_UPDATES(),
+            ')), "job", "$1", "job", "' + LABEL_EXPR['notification-processors'] +'")',
+        ]),
+        legendFormat='{{job}}',
+    )],
+    thresholdType='absolute',
```
francoisferrand: same

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,567 @@
...
+            ' ' + Metrics.CACHE_LAG.count(cache_hit="true"),
+            '))',
+            '/',
+            'sum by (job) (rate (',
+            ' ' + Metrics.CACHE_LAG.count(),
+            ')), "job", "$1", "job", "' + LABEL_EXPR['notification-processors'] +'")',
+        ]),
+        legendFormat='{{job}}',
+    )],
+    thresholdType='absolute',
```
francoisferrand: and here

francoisferrand (approved):

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,516 @@
...
+            'sum by (job) (rate (' + Metrics.CACHE_LAG.count(cache_hit="true") + '))',
+            '/',
+            'sum by (job) (rate (' + Metrics.CACHE_LAG.count() +'))',
+            ', "job", "$1", "job", "' + LABEL_EXPR['notification-processors'] +'")',
+        ]),
+        legendFormat='{{job}}',
+    )],
+    thresholds=[
+        Threshold('green', 0, 0.),
+    ])
```
ghost: I am wondering if we could stick to common standards heres, i.e. a red color if the rate is below 80%, as this is considered a poor cache hit rate, something like
```
thresholds=[
        Threshold("red",     1, 0.00),
        Threshold("orange",  2, 0.85),
        Threshold("green",   3, 0.90),
    ],
```

Mixed with more info in the gauge to ease the reading, like 
image

monitoring/notification/dashboard.py
```diff
@@ -0,0 +1,516 @@
...
+    title='Bandwidth',
+    description='Bandwidth used to send notifications to targets in a given time',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    legendDisplayMode='table',
+    legendPlacement='right',
+    unit=UNITS.BYTES,
```
ghost: How about adding the time unit in the y axis here? `UNIT.BYTES_SEC`

I am not sure, when looking at the graph, if this is a sum of all data processed by a unit of time, or per second (without looking at the code, as a simple user). If you agree having a 1 second resolution might help reading.

ghost (commented):
Please ping me for approval at any time :slightly_smiling_face:

monitoring/notification/dashboard.py
welansari: good point

monitoring/notification/dashboard.py
welansari: agreed

ghost (approved):

PR #2340, opened 2022-10-18
feature: BB-308 add bucket notification alerts

welansari:
Issue: [BB-308](https://scality.atlassian.net/browse/BB-308)

### DIFF
--- a/monitoring/notification/alerts.yaml
+++ b/monitoring/notification/alerts.yaml
@@ -1,0 +1,100 @@
+x-inputs:
+- name: namespace
+  type: constant
+  value: zenko
+- name: job_notification_producer
+  type: config
+  value: artesca-data-backbeat-notification-producer-headless
+- name: job_notification_processors
+  type: config
+  value: artesca-data-backbeat-notification-processor
+- name: notificationFailuresWarningThreshold
+  type: config
+  value: 0.1
+- name: notificationFailuresCriticalThreshold
+  type: config
+  value: 1
+- name: notification_producer_replicas
+  type: config
+  value: 1
+- name: notification_processor_replicas
+  type: config
+  value: 1
+
+groups:
+- name: Bucket Notification
+  rules:
+
+  - alert: NotificationProducerDegraded
+    Expr: |
+      sum(up{namespace="${namespace}",job="${job_notification_producer}"})
+        < ${notification_producer_replicas}
+    For: "30s"
+    Labels:
+     severity: warning
+    Annotations:
+      zenko_service: backbeat-notification-producer
+      description: "Less than 100% of notification producers are up and healthy"
+      summary: "Notification producer service is degraded"
+
+  - alert: NotificationProducerCritical
+    Expr: |
+      sum(up{namespace="${namespace}",job="${job_notification_producer}"}) * 2
+        < ${notification_producer_replicas}
+    For: "30s"
+    Labels:
+      severity: critical
+    Annotations:
+      zenko_service: backbeat-notification-producer
+      description: "Less than 50% of notification producers are up and healthy"
+      summary: "Notification producer service is in critical state"
+
+  - alert: NotificationProcessorsDegraded
+    Expr: |
+      label_replace(
+        sum(up{namespace="${namespace}",job=~"${job_notification_processors}.*"}) by(job)
+      , "job", "$1", "job", ".*-(.*)-headless") < ${notification_processor_replicas}
+    For: "30s"
+    Labels:
+      severity: warning
+    Annotations:
+      zenko_service: backbeat-notification-processor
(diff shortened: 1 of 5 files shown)


monitoring/notification/alerts.yaml
```diff
@@ -0,0 +1,105 @@
...
+    Annotations:
+      zenko_service: backbeat-notification-producer
+      description: "Less than 50% of notification producers are up and healthy"
+      summary: "Notification producer service is in critical state"
+
+  - alert: NotificationProcessorsDegraded
+    Expr: |
+      ${bucket_notification_enabled} == bool 1 
+        * (sum(up{namespace="${namespace}",job=~"${job_notification_processors}.*"}) 
+        < (${configured_notification_targets} * ${notification_processor_replicas}))
```
francoisferrand: we do not really care about the aggregate, this alert should be generated for each `target` : e.g. use `sum() by(target)` so that there is one alert for each target (which just checks for notification_processor_replicas, and does not use configured_notification_targets)

the target will be a label of expression. and can then be used in the alert summary/name

monitoring/notification/alerts.yaml
```diff
@@ -0,0 +1,105 @@
...
+    Annotations:
+      zenko_service: backbeat-notification-processor
+      description: "Less than 100% of notification processors are up and healthy"
+      summary: "Notification processor service is degraded"
+
+  - alert: NotificationProcessorsCritical
+    Expr: |
+      ${bucket_notification_enabled} == bool 1
+        * (sum(up{namespace="${namespace}",job=~"${job_notification_processors}.*"}) * 2 
+        < (${configured_notification_targets} * ${notification_processor_replicas}))
```
francoisferrand: same

monitoring/notification/alerts.yaml
```diff
@@ -0,0 +1,105 @@
...
+    For: "30s"
+    Labels:
+      severity: critical
+    Annotations:
+      zenko_service: backbeat-notification-processor
+      description: "Less than 50% of notification processors are up and healthy"
+      summary: "Notification processor service is in critical state"
+
+  - alert: NotificationDeliveryDegraded
```
francoisferrand, line 78: for these it may make sense to have alerts for each target, what do you think?
since each target (or remote server, more precisely) may have different error rate... but then we cannot specify different warning thresholds, so...

welansari: Yes i do think alerts should be precise and indicate exactly where the issue is, especially here since the targets might no be linked to each other and can be different machines.

francoisferrand (approved):

monitoring/notification/alerts.yaml
```diff
@@ -0,0 +1,100 @@
...
+      severity: warning
+    Annotations:
+      zenko_service: backbeat-notification-processor
+      description: "Less than 100% of notification processors for target {{ $labels.job }} are up and healthy"
+      summary: "Notification processor service for target {{ $labels.job }} is degraded"
+
+  - alert: NotificationProcessorsCritical
+    Expr: |
+      label_replace(
+        sum (up{namespace="${namespace}",job=~"${job_notification_processors}.*"}) by(job) * 2
```
ghost: Very <sup><sub>smol</sub></sup> consistency nitpicking:
```suggestion
        sum(up{namespace="${namespace}",job=~"${job_notification_processors}.*"}) by(job) * 2 
```

monitoring/notification/alerts.yaml
```diff
@@ -0,0 +1,100 @@
...
+      zenko_service: backbeat-notification-processor
+      description: "Less than 50% of notification processors for target {{ $labels.job }} are up and healthy"
+      summary: "Notification processor service for target {{ $labels.job }} is in critical state"
+
+  - alert: NotificationDeliveryDegraded
+    Expr: |
+      (sum by(target) (rate(notification_queue_processor_notification_delivery_delay_sec_count{status="fail",job=~"${job_notification_processors}.*",namespace="${namespace}"}[1m]))
+        / sum by(target) (rate(notification_queue_processor_notification_delivery_delay_sec_count{job=~"${job_notification_processors}.*",namespace="${namespace}"}[1m])))
+        >= ${notificationFailuresWarningThreshold}
+    For: "30s"
```
ghost, line 83: As we are already checking the last minute in the query, I wonder if having the alert only after 30s makes sense? I mean, if we have a degraded state for less than 30s, and that repeats, we never alert, right? (same below)

welansari: Prometheus polls the metrics every 30s, so putting less than that won't have the desired effect.
The alert is fired after having a degraded state for the last two consecutive polls.

ghost (approved):

alexanderchan-scality (approved):

PR #2349, opened 2022-11-18
BB-337 Implement metrics for Cold Storage Archive and Restore

nicolas2bert:


nicolas2bert:
ping

nicolas2bert:
@bert-e reset

extensions/utils/metricsUtil.js
```diff
@@ -0,0 +1,28 @@
+function getStatusLabel(err) {
```
francoisferrand: do we really need this?
"aggregating" by status code 'range' can be done at presentation time (in prometheus queries), and there are not that many code anyway ; yet it may useful at times to be able to dig and find the metric for a specific code ?

monitoring/lifecycle/dashboard.py
```diff
@@ -271,7 +271,7 @@ def kafka_row(topic, op):
             layout.row([lifecycle_global_s3_requests], height=10),
             layout.row(lifecycle_global_s3_error_rates, height=4),
             RowPanel(title="Kafka"),
-            layout.row(kafka_row("Expiration Bucket Task", "BucketTopic"), height=10),
+            layout.row(kafka_row("Lifecycle Bucket Task", "BucketTopic"), height=10),
             layout.row(kafka_row("Expiration Object Task", "ObjectTopic"), height=10),
```
francoisferrand, line 275: That one should be renamed as well?

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -1,4 +1,5 @@
 const { ZenkoMetrics } = require('arsenal').metrics;
+const { getStatusLabel } = require('../utils/metricsUtil');
 
 const LIFECYCLE_LABEL_ORIGIN =  'origin';
 const LIFECYCLE_LABEL_OP = 'op';
```
francoisferrand: It would be useful to have "duration" metrics as well: e.g.
* the duration of individual tasks (processing of an object by GC...)
* duration of bucket scan
* latency/delay between an "event" (expiration....) and the time when we actually apply the transition

extensions/gc/GarbageCollectorMetrics.js
```diff
@@ -0,0 +1,40 @@
...
+
+const gcS3Operations = ZenkoMetrics.createCounter({
+    name: 'gc_s3_operations',
+    help: 'Total number of S3 operations by the garbage collector processes',
+    labelNames: [
+        GC_LABEL_ORIGIN,
+        GC_LABEL_OP,
+        GC_LABEL_STATUS,
+    ],
+});
```
francoisferrand, line 15: GC may be called for reason other than cold storage expiration ; so we may want to also have some specific metric for restore expiration.

monitoring/lifecycle/dashboard.py
```diff
@@ -271,7 +271,7 @@ def kafka_row(topic, op):
             layout.row([lifecycle_global_s3_requests], height=10),
             layout.row(lifecycle_global_s3_error_rates, height=4),
```
francoisferrand, line 272: dashboard will be updated in a separate PR ?

monitoring/lifecycle/dashboard.py
nicolas2bert: No, the metrics collected are specific to expiration.

monitoring/lifecycle/dashboard.py
nicolas2bert: yes

extensions/utils/metricsUtil.js
nicolas2bert: Indeed. I am ok removing this aggregation

extensions/gc/GarbageCollectorMetrics.js
nicolas2bert: Is leveraging the origin label not ideal? (e.g. `GarbageCollectorMetrics.onS3Request(log, 'getMetadata', 'archive', err);` )

extensions/utils/metricsUtil.js
nicolas2bert: Since lifecycle dashboards are already using the "aggregating by status code rage" logic, I see this more as a tech depth that will be tackled by https://scality.atlassian.net/browse/BB-342.

extensions/lifecycle/LifecycleMetrics.js
nicolas2bert: This deserves a ticket for itself: https://scality.atlassian.net/browse/BB-343

extensions/utils/metricsUtil.js
francoisferrand: TechDebt indeed, though very quick to fix : i.e. use the 'raw' status instead of calling `getStatusLabel` in the code, and replace each of these filters in dashboard (e.g. `="2xx"` with `=~"2.."` for each of the ranges 2/3/4/5)

extensions/gc/GarbageCollectorMetrics.js
francoisferrand: indeed, i missed that (maybe after seeing the origin being hardcoded in other use cases...)

monitoring/lifecycle/dashboard.py
francoisferrand: 😕
this breaks the kind of "atomicity" we get from having the dashboards in the same repo as the code (and thus the safety of having both the dashboard and matching metrics in the product)

extensions/lifecycle/LifecycleMetrics.js
francoisferrand: ok

monitoring/lifecycle/dashboard.json
```diff
@@ -2045,7 +2045,7 @@
       "targets": [
         {
           "datasource": null,
-          "expr": "sum(rate(lifecycle_s3_operations{status=\"2xx\",op=\"deleteObject\",namespace=\"${namespace}\",job=\"${job_lifecycle_object_processor}\"}[$__rate_interval]))",
+          "expr": "sum(rate(lifecycle_s3_operations{status=\"200\",op=\"deleteObject\",namespace=\"${namespace}\",job=\"${job_lifecycle_object_processor}\"}[$__rate_interval]))",
```
francoisferrand, line 2048: same here: maybe we should have `=~"2.."` ?

monitoring/lifecycle/dashboard.py
```diff
@@ -66,9 +66,9 @@ def s3_request_error_rate_expr(process, job, code):
     divsLabel = 'namespace="${namespace}"'
 
     if code is not None:
-        divdLabel += f',status="{code}"'
+        divdLabel += f',status=~"{code}"'
     else:
-        divdLabel += ',status!="2xx"'
+        divdLabel += ',status!="200"'
```
francoisferrand, line 71: do we only ever use 200 on success?
otherwise we should have !~"2.." in such case...

monitoring/lifecycle/dashboard.py
```diff
@@ -110,8 +110,8 @@ def s3_request_error_rates(process=None, job=None):
 
 
 def s3_deletion_request_time_series(op):
-    successLabel = f'status="2xx",op="{op}",namespace="{"${namespace}"}",job="{"${job_lifecycle_object_processor}"}"'
-    errorLabel = f'status!="2xx",op="{op}",namespace="{"${namespace}"}",job="{"${job_lifecycle_object_processor}"}"'
+    successLabel = f'status="200",op="{op}",namespace="{"${namespace}"}",job="{"${job_lifecycle_object_processor}"}"'
+    errorLabel = f'status!="200",op="{op}",namespace="{"${namespace}"}",job="{"${job_lifecycle_object_processor}"}"'
```
francoisferrand, line 114: same here?

francoisferrand (approved):

monitoring/lifecycle/dashboard.py
nicolas2bert: Yes, it is either an error status code (3xx, 4xx, 5xx) or 200.

welansari (approved):

nicolas2bert:
@bert-e approve

PR #2350, opened 2022-11-23
BB-269 avoid clearing dataStoreVersionId when retrying a replication

welansari:
Issue: [BB-269](https://scality.atlassian.net/browse/BB-269)

Replication has special code to handle replication of tags: in that case, it will not replicate the object again, but simply update the tags on the target object.

However, in the (unlikely) case where that tag update fails, the datastore version id will be reset when the entry is put in the replay queue:

```
    toRetryEntry(site) {
        return this.clone()
            .setReplicationBackends([{
                site,
                status: 'PENDING',
                dataStoreVersionId: '',
            }])
            .setReplicationStorageClass(site)
            .setReplicationStatus('PENDING');
    }
```
Consequently, the replay-processor would trigger a full replication of the object, and the data will be copied again.

[BB-269]: https://scality.atlassian.net/browse/BB-269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/models/ObjectQueueEntry.js
+++ b/lib/models/ObjectQueueEntry.js
@@ -198,11 +198,8 @@ class ObjectQueueEntry extends ObjectMD {
 
     toRetryEntry(site) {
         return this.clone()
-            .setReplicationBackends([{
-                site,
-                status: 'PENDING',
-                dataStoreVersionId: '',
-            }])
+            .setReplicationBackends(this.getReplicationBackends().filter(o => o.site === site))
+            .setReplicationSiteStatus(site, 'PENDING')
             .setReplicationStorageClass(site)
             .setReplicationStatus('PENDING');
     }
--- a/tests/unit/lib/models/ObjectQueueEntry.spec.js
+++ b/tests/unit/lib/models/ObjectQueueEntry.spec.js
@@ -1,0 +1,22 @@
+const assert = require('assert');
+
+const QueueEntry = require('../../../../lib/models/QueueEntry');
+const { replicationEntry } = require('../../../utils/kafkaEntries');
+
+describe('ObjectQueueEntry', () => {
+    describe('toRetryEntry', () => {
+        it('Should not clear dataStoreVersionId when retrying', () => {
+            const entry = QueueEntry.createFromKafkaEntry(replicationEntry);
+            const dataStoreVersionId = entry.getReplicationSiteDataStoreVersionId('sf');
+            const retryEntry = entry.toRetryEntry('sf');
+            assert.strictEqual(retryEntry.getReplicationSiteDataStoreVersionId('sf'),
+                dataStoreVersionId);
+        });
+        it('Should only include failed site details', () => {
+            const entry = QueueEntry.createFromKafkaEntry(replicationEntry);
+            const retryEntry = entry.toRetryEntry('sf');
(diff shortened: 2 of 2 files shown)


lib/models/ObjectQueueEntry.js
```diff
@@ -198,11 +198,7 @@ class ObjectQueueEntry extends ObjectMD {
 
     toRetryEntry(site) {
         return this.clone()
-            .setReplicationBackends([{
-                site,
-                status: 'PENDING',
-                dataStoreVersionId: '',
-            }])
+            .setReplicationSiteStatus(site, 'PENDING')
```
francoisferrand, line 202: i think this would introduce a significant change : this may trigger a re-replication of other sites as well....
this may be OK (since either it is already replicated/pending, and would be ignored? - or it would be FAILED and should be picked up, which may not be bad?), but i fear it may introduce some side-effect or race condition.

Resetting the whole field (in the entry only, not the metadata!) avoids this kind of effect, by targeting only the specified site.

Not sure if this is an issue, but we need to think well about this and understand what is the correct/expected solution.

welansari: `setReplicationSiteStatus` actually only changes the replication status of the specified site
https://github.com/scality/Arsenal/blob/8f63687ef3d7dfe47de747a88f1673ae29a9b6ca/lib/models/ObjectMD.ts#L1048-L1056

In the original implementation we already set the global replication status
https://github.com/scality/backbeat/blob/2ff70385a6a59a23fdeb7186474a5c2b8feb6c72/lib/models/ObjectQueueEntry.js#L207

But it is not an issue as sites with a 'COMPLETED' state are ignored
https://github.com/scality/backbeat/blob/2ff70385a6a59a23fdeb7186474a5c2b8feb6c72/extensions/replication/tasks/MultipleBackendTask.js#L1169-L1185

francoisferrand: I don't think this is all. Setting a single site in the list (although doing it correctly!) will make sure the entry is skipped by all other replication processors: 
* so they will not try to replicate at all, even if there is a 'gap' in the filtering. You mentioned the filter on 'COMPLETED', but it still creates a warning which not approriate; and there are other states: typically already PENDING, or in ERROR...
* so we avoid the load of refreshing the metatadata in each data processor : the faster we drop the entry, the better

There is a significant difference with the "initial" update of the object: in that case, it is added to the kafka topic so that every processor will look at it and decide if replication is needed. In this specific case, we only want the processor for the failed destination to retry, so it would be better to stick with the minimal information. And remember this is only the kafka message (e.g. designed only the target processor) and not a real copy of the real objectMd.

welansari: yes i understand your point now, fixed

francoisferrand (approved):
LGTM, though it may be safer (as in "future proof") to rely more on the API provided by ObjectMD to perform the update we want:

```
this.clone()
            .setReplicationBackends(this.getReplicationBackends().filter(o => o.site === site))
            .setReplicationSiteStatus(site, PENDING)
            ...
```

ghost (approved):

PR #2351, opened 2022-11-23
Improvement: BB-335 ignore objects with pending replication in lifecycle

welansari:
Issue: [BB-335](https://scality.atlassian.net/browse/BB-335)

Skipping objects with a pending replication from getting expired or transitioned

[BB-335]: https://scality.atlassian.net/browse/BB-335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
+++ b/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
@@ -7,6 +7,7 @@ const { attachReqUids } = require('../../../lib/clients/utils');
 const { LifecycleMetrics } = require('../LifecycleMetrics');
 
 class ObjectLockedError extends Error {}
+class PendingReplicationError extends Error {}
 
 class LifecycleDeleteObjectTask extends BackbeatTask {
     /**
@@ -19,9 +20,14 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
         const procState = proc.getStateVars();
         super();
         Object.assign(this, procState);
+        this.objectMD = null;
     }
 
     _getMetadata(entry, log, done) {
+        // only retreiving object metadata once
+        if (this.objectMD) {
+            return done(null, this.objectMD);
+        }
         const { owner: canonicalId, accountId } = entry.getAttribute('target');
         const backbeatClient = this.getBackbeatMetadataProxy(accountId);
         if (!backbeatClient) {
@@ -56,7 +62,8 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
                     errors.InternalError.
                         customizeDescription('error parsing metadata blob'));
             }
-            return done(null, res.result);
+            this.objectMD = res.result;
+            return done(null, this.objectMD);
         });
     }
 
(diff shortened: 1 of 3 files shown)


extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -142,32 +145,39 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
...
 
-        return this._getMetadata(entry, log, (err, objMD) => {
-            if (err) {
-                return done(err);
-            }
-
-            if (objMD.getLegalHold()) {
-                return done(new ObjectLockedError('object locked'));
-            }
+        if (this.objectMD.getLegalHold()) {
```
francoisferrand: there is a small difference here (vs before): we _always_ retrieve the metadata, even if the object is not versioned.

This used to be skipped at the beginning of _checkObjectLockState, when the entry has no target version.

What is that specific case: old non-versioned objects from before versioning was enabled?
May this case have a significant impact?

welansari: The case is: we ignore object lock check when the object is not versioned
Always getting the metadata is something we need to do anyway, as the replication check needs to be done regardless of the versioning state of the object.

francoisferrand (approved):

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -170,6 +177,27 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
...
     }
 
+    /**
+     * Throws error if object has a 'PENDING' replication status
+     * @param {ActionQueueEntry} entry entry object
+     * @param {Logger} log logger instance
+     * @param {Function} done callback
+     * @returns {undefined}
+     */
+    _checkReplicationStatus(entry, log, done) {
```
nicolas2bert, line 187: So we decided to evaluate the replication status "late" to prevent an extra metadata call. Did anything else prevent us to do it one step earlier in task/LifecycleTask.js?

welansari: Actually in all cases we only do one metadata call (the value is kept from the first call and reused later), and so here i don't think it really makes a difference wether we check the replication before or after object lock, what do you think ?

nicolas2bert: I meant "late" by one async communication later. Basically, why don't we check it in the Lifecycle Bucket Processor, after consuming the bucket task topic?

welansari: Yes as you said the only reason is to save the metadata call nothing more

nicolas2bert: Ok, thanks, I added a note for it on my side.

Context: we are implementing new object listings (specific to lifecycle) that, inter alia, return extra object metadata fields to avoid those extra MD calls. So, we will add Replication Status as one of the objects MD fields returned from the list.

nicolas2bert (approved):

alexanderchan-scality (approved):

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -170,6 +177,27 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
...
+        return this._getMetadata(entry, log, (err, objMD) => {
+            if (err) {
+                return done(err);
+            }
+            const replicationStatus = objMD.getReplicationStatus();
+            if (replicationStatus === 'PENDING') {
```
nicolas2bert: What about 'PROCESSING' and 'FAILED' status?

Replication status:
- PROCESSING: 
      - (multiple sites) one site is replication COMPLETED while other sites are still PENDING
      - entry had a FAILED replication and was picked up by "FailedKafkaEntries" cronjob that updated the global replication status to PROCESSING along with sending an entry to the replication topic so that the queue processor re-attempts replication.
- PENDING: 
      - Fresh new object waiting to be replicated. PENDING is set by Cloudserver for objects created in a bucket with replication enabled. Meaning that no replication has been attempted before on the specific object.
- FAILED:
      - at least one replication site has FAILED

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
welansari: Got it, thanks for the context

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
welansari: Good catch, i think we should also include these other statuses

PR #2352, opened 2022-11-30
BB-341: OplogPopulator support lifecycle and replication

alexanderchan-scality:


extensions/oplogPopulator/OplogPopulatorTask.js
```diff
@@ -19,8 +19,17 @@ werelogs.configure({ level: config.log.logLevel,
...
+
+[
+    'notification',
+    'lifecycle',
+    'replication',
+].keys(config.extensions).forEach(ext => {
+    if (config.exntensions[ext]) {
+        activeExtensions.push(ext);
+    }
+});
```
francoisferrand: does this really work? You are trying to push into a `const` variable, this seems fishy...

It seems however it may be a simple use-case for `filter` :
```suggestion
const activeExtensions = [
    'notification',
    'lifecycle',
    'replication',
].filter(ext => config.extensions[ext]);
```

extensions/oplogPopulator/OplogPopulatorTask.js
```diff
@@ -19,8 +19,17 @@ werelogs.configure({ level: config.log.logLevel,
...
-// Notification extension is the only one using OplogPopulator for now
-const activeExtensions = config.extensions.notification ? ['notification'] : [];
+const activeExtensions = [];
+
+[
+    'notification',
+    'lifecycle',
+    'replication',
+].keys(config.extensions).forEach(ext => {
+    if (config.exntensions[ext]) {
```
francoisferrand:
```suggestion
    if (config.extensions[ext]) {
```

francoisferrand (dismissed):
please test the code

alexanderchan-scality:
@francoisferrand tested  `replication` and `lifecycle` as the active extensions; fixed the typos with your suggestion; it looks to be from a buffer issue, resulting in the mangled code

extensions/oplogPopulator/OplogPopulatorTask.js
alexanderchan-scality: > does this really work? You are trying to push into a const variable, this seems fishy...

yes. javascript quirk, the `const` is to the reference, the underlying array can still be populated
will use your filter suggestions

welansari:
To detect if replication & lifecycle is enabled for a bucket we use these two files :
https://github.com/scality/backbeat/blob/development/8.4/extensions/replication/ReplicationOplogPopulatorUtils.js
https://github.com/scality/backbeat/blob/development/8.4/extensions/lifecycle/LifecycleOplogPopulatorUtils.js

Might be worth checking if we cover all possible cases.

naren-rajendran (approved):

welansari (approved):

alexanderchan-scality:
ping

PR #2355, opened 2022-12-19
BB-353 - support s3:ObjectRestore:Post operation types in lifecycle

welansari:
Issue: [BB-353](https://scality.atlassian.net/browse/BB-353)

### DIFF
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -225,7 +225,10 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         const value = JSON.parse(entry.value);
 
         const operation = value.originOp;
-        if (operation !== 's3:ObjectRestore') {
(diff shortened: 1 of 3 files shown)


ghost (approved):
LGTM, aligned with what I found, thanks for adding a test as well!

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -225,7 +225,7 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         const value = JSON.parse(entry.value);
 
         const operation = value.originOp;
-        if (operation !== 's3:ObjectRestore') {
+        if (!['s3:ObjectRestore', 's3:ObjectRestore:Post'].includes(operation)) {
```
francoisferrand, line 231: a comment explaining that this s3:ObjectRestore value is used for backwards compatibility with older cloudserver would be nice

francoisferrand (approved):
Should target 8.5, since cloudserver 8.6 with the breaking change is used in Zenko 2.5 along with backbeat 8.5

PR #2357, opened 2022-12-21
Feature: BB-309 add oplog populator metrics

welansari:
Issue: [BB-309](https://scality.atlassian.net/browse/BB-309)

[BB-309]: https://scality.atlassian.net/browse/BB-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BB-309]: https://scality.atlassian.net/browse/BB-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -248,7 +248,14 @@ class OplogPopulator {
             // establish change stream
             this._setMetastoreChangeStream();
             // remove no longer valid buckets from old connectors
-            await this._connectorsManager.removeInvalidBuckets(validBuckets);
+            const oldConnectorBuckets = this._connectorsManager.oldConnectors
+                .map(connector => connector.buckets)
+                .flat();
+            const invalidBuckets = oldConnectorBuckets.filter(bucket => !validBuckets.includes(bucket));
+            await Promise.all(invalidBuckets.map(bucket => this._allocator.stopListeningToBucket(bucket)));
+            this._logger.info('Successfully removed invalid buckets from old connectors', {
+                method: 'ConnectorsManager.removeInvalidBuckets',
+            });
             // start scheduler for updating connectors
             this._connectorsManager.scheduleConnectorUpdates();
             this._logger.info('OplogPopulator setup complete', {
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -67,6 +67,7 @@ class OplogPopulator {
             const client = await MongoClient.connect(this._mongoUrl, {
                  replicaSet: this._replicaSet,
                  useNewUrlParser: true,
+                 useUnifiedTopology: true,
             });
             // connect to metadata DB
             this._mongoClient = client.db(this._database, {
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -6,6 +6,7 @@ const { constructConnectionString } = require('../utils/MongoUtils');
 const ChangeStream = require('../../lib/wrappers/ChangeStream');
 const Allocator = require('./modules/Allocator');
 const ConnectorsManager = require('./modules/ConnectorsManager');
+const { ZenkoMetrics } = require('arsenal').metrics;
 
 const paramsJoi = joi.object({
     config: joi.object().required(),
@@ -289,6 +290,21 @@ class OplogPopulator {
             this._logger.error('ready state', components);
         }
         return allReady;
+    }
+
+    /**
+     * Handle ProbeServer metrics
+     *
+     * @param {http.HTTPServerResponse} res - HTTP Response to respond with
+     * @param {Logger} log - Logger
+     * @returns {undefined}
+     */
+    handleMetrics(res, log) {
+        log.debug('metrics requested');
+        res.writeHead(200, {
+            'Content-Type': ZenkoMetrics.asPrometheusContentType(),
+        });
+        res.end(ZenkoMetrics.asPrometheus());
     }
 }
 
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -7,12 +7,14 @@ const ChangeStream = require('../../lib/wrappers/ChangeStream');
 const Allocator = require('./modules/Allocator');
 const ConnectorsManager = require('./modules/ConnectorsManager');
 const { ZenkoMetrics } = require('arsenal').metrics;
+const OplogPopulatorMetrics = require('./OplogPopulatorMetrics');
 
 const paramsJoi = joi.object({
     config: joi.object().required(),
     mongoConfig: joi.object().required(),
     activeExtensions: joi.array().required(),
     logger: joi.object().required(),
+    enableMetrics: joi.boolean().default(true),
 }).required();
 
 /**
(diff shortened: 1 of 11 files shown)


extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -67,6 +76,7 @@ class OplogPopulator {
             const client = await MongoClient.connect(this._mongoUrl, {
                  replicaSet: this._replicaSet,
                  useNewUrlParser: true,
+                 useUnifiedTopology: true,
```
francoisferrand, line 79: seems unrelated...

welansari: yes the two first commits are drive by fixes of some issues i found

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -6,12 +6,15 @@ const { constructConnectionString } = require('../utils/MongoUtils');
...
 const ConnectorsManager = require('./modules/ConnectorsManager');
+const { ZenkoMetrics } = require('arsenal').metrics;
+const OplogPopulatorMetrics = require('./OplogPopulatorMetrics');
 
 const paramsJoi = joi.object({
     config: joi.object().required(),
     mongoConfig: joi.object().required(),
     activeExtensions: joi.array().required(),
     logger: joi.object().required(),
+    enableMetrics: joi.boolean().default(true),
```
francoisferrand, line 17: do we need to make metrics optional?
i don't really see a use-case here or "precedent" in other modules, shouldn't metrics always be collected?

extensions/oplogPopulator/OplogPopulator.js
francoisferrand: we may want to validate these drive-by fixes in zenko CI before merging though...

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -200,7 +212,16 @@ class OplogPopulator {
                     '_id': 1,
                     'operationType': 1,
                     'documentKey._id': 1,
-                    'fullDocument.value': 1
+                    'fullDocument.value': 1,
+                    // transforming the BSON timestamp
+                    // into a usable date
```
francoisferrand, line 217: clusterTime should already be included, with a value like this, right?
```js
"clusterTime": Timestamp({ "t": 1666193824, "i": 1 }),
```

can't just read the timestamp (`clusterTime.t`) in oplogpopulator instead of converting to date string and parsing the date?

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -0,0 +1,164 @@
...
+        this.requestSize = null;
+        this.connectors = null;
+        this.reconfigurationLag = null;
+        this._logger = logger;
+    }
+
+    registerMetrics() {
+        this.acknowledgementLag = ZenkoMetrics.createHistogram({
+            name: 'oplog_populator_acknowledgement_lag_sec',
+            help: 'Delta between a config change in mongo and it getting processed by the oplogPopulator in seconds',
```
francoisferrand: _**Delay**_ between a config change in mongo and _**start of processing**_ by the oplogPopulator in seconds

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -0,0 +1,164 @@
...
+            labelNames: ['opType'],
+            buckets: [0.001, 0.01, 1, 10, 100, 1000, 10000],
+        });
+        this.connectorConfiguration = ZenkoMetrics.createCounter({
```
francoisferrand, line 26: is it the number of "reconfiguration" (e.g. running the whole loop)?
or the number of time we update the configuration of a single connector?

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -0,0 +1,164 @@
...
+            labelNames: ['connector'],
+        });
+        this.connectors = ZenkoMetrics.createGauge({
+            name: 'oplog_populator_connectors',
+            help: 'Total number of configured connectors',
+            labelNames: ['isOld'],
+        });
+        this.reconfigurationLag = ZenkoMetrics.createHistogram({
+            name: 'oplog_populator_reconfiguration_lag_sec',
+            help: 'Time it takes to reconfigure a connector',
```
francoisferrand: is it the time it takes to configure a _single_ connector? (e.g. issuing the command to kafka connect)
or does it also include the time it takes for a _single_ connector to "be ready" (e.g. tasks re-created/updated in kafka connect)?
or the overall "reconfigure time" (that it takes to reconfigure all connectors)

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -0,0 +1,164 @@
...
+            labelNames: ['connector'],
+            buckets: [0.001, 0.01, 1, 10, 100, 1000, 10000],
+        });
+        this.reconfiguration = ZenkoMetrics.createCounter({
```
francoisferrand: not clear what this measures & what "unit"

extensions/oplogPopulator/OplogPopulator.js
welansari: In normal cases we always have the metrics enabled, that's why it's set to true by default.
The option was put there to be able to run unit tests with no issues, as otherwise we would get errors about the metrics being registered multiple times.

extensions/oplogPopulator/OplogPopulator.js
welansari: I did this because i find it cleaner than the way we do in Arsenal which is not precise, i'd rather let mongo transform it into a usable timestamp.
```
 timestamp: new Date((itemObj.ts ? itemObj.ts.getHighBits() * 1000 : 0))
```

welansari: And the value `clusterTime` is equal to:
```
Timestamp { _bsontype: 'Timestamp', low_: 1, high_: 1672913394 }
```

extensions/oplogPopulator/OplogPopulator.js
welansari: No need i think since the code only impacts the oplog populator (no tests for now), and i tested everything locally and on an Artesca cluster

extensions/oplogPopulator/OplogPopulator.js
francoisferrand: we may indeed have something for the tests, but it should not propagate to the config...

francoisferrand (approved):

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -204,12 +208,14 @@ class ConnectorsManager {
                 const oldConnectors = await this._getOldConnectors(oldConnectorNames);
                 this._connectors.push(...oldConnectors);
                 this._oldConnectors.push(...oldConnectors);
+                this._metricsHandler.onConnectorsInstantiated(true, oldConnectors.length);
```
XinLi0207, line 211: onConnectorsInstantiated() catches the error inside the function if `this.connectors.inc` failes, but doesn't it matter for the following logic of  `this._metricsHandler.onConnectorsInstantiated(true, oldConnectors.length);` even if `this.connectors.inc` fails?

welansari: It does affect it, the value of the metric will be false.
We don't however want to crash the oplog populator here as that will lead to some events being missed in case one or more bucket configurations change.

XinLi0207 (approved):

PR #2358, opened 2022-12-23
Feature: BB-310 add oplog populator dashboard

welansari:
Issue: [BB-310](https://scality.atlassian.net/browse/BB-310)

Screenshot 2023-01-04 at 14 09 30
Screenshot 2023-01-04 at 14 09 38
Screenshot 2023-01-04 at 14 09 48

[BB-310]: https://scality.atlassian.net/browse/BB-310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BB-310]: https://scality.atlassian.net/browse/BB-310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/monitoring/oplog-populator/dashboard.py
+++ b/monitoring/oplog-populator/dashboard.py
@@ -1,0 +1,362 @@
+from grafanalib.core import (
+    ConstantInput,
+    DataSourceInput,
+    Threshold,
+    Templating,
+    Template,
+    Heatmap,
+    YAxis,
+    HeatmapColor,
+    RowPanel,
+)
+from grafanalib import formatunits as UNITS
+from scalgrafanalib import(
+    metrics,
+    layout,Target,
+    TimeSeries,
+    Dashboard,
+    Stat,
+    PieChart,
+    Tooltip,
+)
+
+class Metrics:
+    RECONFIG_LAG = metrics.BucketMetric(
+        'oplog_populator_reconfiguration_lag_sec', 'connector',
+        job="${oplog_populator_job}", namespace="${namespace}"
+    ).with_defaults(
+        'connector=~"${connector}"',
+    )
+
+    ACK_LAG = metrics.BucketMetric(
+        'oplog_populator_acknowledgement_lag_sec',
+        'opType', 'job', namespace="${namespace}"
+    ).with_defaults(
+        'job="${oplog_populator_job}"',
+    )
+
+    CONNECTOR_CONFIG, CONNECTOR_CONFIG_APPLIED, BUCKETS, REQUEST_SIZE, PIPELINE_SIZE = [
+        metrics.CounterMetric(
+            name, 'connector', *extraLabels,
+            job="${oplog_populator_job}", namespace="${namespace}"
+        ).with_defaults(
+            'connector=~"${connector}"',
+        ) for name, extraLabels in {
+            'oplog_populator_connector_configuration': ['opType'],
+            'oplog_populator_connector_configuration_applied': ['success'],
+            'oplog_populator_connector_buckets': [],
+            'oplog_populator_connector_request_bytes': [],
+            'oplog_populator_connector_pipeline_bytes': [],
+        }.items()
+    ]
+
+    CONNECTORS = metrics.CounterMetric(
+        'oplog_populator_connectors', 'isOld',
+        job="${oplog_populator_job}", namespace="${namespace}"
+    )
+
+oplog_populator_up_current = Stat(
+    title='Oplog Populator Instances',
+    description='Number of active oplog populator instances',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='background',
+    noValue='-',
+    reduceCalc='last',
+    targets=[Target(
+        expr='\n'.join([
+            'sum(up{',
+            '    namespace="${namespace}",',
+            '    job="${oplog_populator_job}"',
+            '})',
+        ]),
+    )],
+    thresholds=[
+        Threshold('red', 0, 0.),
+        Threshold('green', 1, 1.),
+    ])
+
+kafka_connect_up_current = Stat(
+    title='Kafka Connect Instances',
+    description='Number of active kafka-connect instances',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='background',
+    noValue='-',
+    reduceCalc='last',
+    targets=[Target(
+        expr='\n'.join([
+            'sum(up{',
+            '    namespace="${namespace}",',
+            '    job="${kafka_connect_job}"',
+            '})',
+        ]),
+    )],
+    thresholds=[
+        Threshold('red', 0, 0.),
+        Threshold('green', 1, 1.),
+    ])
+
+connector_instances = Stat(
+    title='Active Connectors',
+    description='Number of active kafka-connect connector instances',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='value',
+    noValue='-',
+    reduceCalc='last',
+    targets=[Target(
+        expr='sum(' + Metrics.CONNECTORS.raw() + ')',
+    )],
+    thresholds=[
+        Threshold('red', 0, 0.),
+        Threshold('green', 1, 1.),
+    ])
+
+total_number_buckets = Stat(
+    title='Monitored Buckets',
+    description='Number of buckets configured on the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='value',
+    noValue='-',
+    reduceCalc='last',
+    targets=[Target(
+        expr='sum(' + Metrics.BUCKETS.raw() + ')',
+    )],
+    thresholds=[
+        Threshold('green', 0, 0.),
+    ])
+
+max_pipeline_size = Stat(
+    title='Max Pipeline Size',
+    description='Maximum size of the mongo pipeline',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='value',
+    noValue='-',
+    reduceCalc='last',
+    format=UNITS.BYTES,
+    orientation='horizontal',
+    targets=[Target(
+        expr='max (' + Metrics.PIPELINE_SIZE.raw() + ')',
+    )],
+    thresholds=[
+        Threshold('green', 0, 0.),
+    ])
+
+bandwidth = Stat(
+    title='Bandwidth',
+    description='Bandwidth used for reconfiguring the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='value',
+    noValue='-',
+    reduceCalc='last',
+    format=UNITS.BYTES_SEC,
+    orientation='horizontal',
+    targets=[Target(
+        expr='sum (rate(' + Metrics.REQUEST_SIZE() + '))',
+    )],
+    thresholds=[
+        Threshold('green', 0, 0.),
+    ])
+
+config_applied_success = TimeSeries(
+    title='Connector Reconfiguration Rate',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    legendDisplayMode='hidden',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum (rate(' + Metrics.CONNECTOR_CONFIG_APPLIED('success="true"') + '))',
+    )])
+
+config_applied_failure = TimeSeries(
+    title='Failed Connector Reconfiguration Rate',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    legendDisplayMode='hidden',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum (rate(' + Metrics.CONNECTOR_CONFIG_APPLIED('success="false"') + '))',
+    )])
+
+
+bucket_distribution = PieChart(
+    title='Bucket Distribution',
+    description='How buckets are distributed among the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    legendValues=['percent', 'value'],
+    pieType='donut',
+    legendDisplayMode='table',
+    legendPlacement='right',
+    targets=[Target(
+        expr='sum by(connector) (' + Metrics.BUCKETS.raw() + ')',
+        legendFormat='{{connector}}',
+    )])
+
+bucket_per_connector = TimeSeries(
+    title='Buckets Per Connector',
+    description='Evolution of buckets per connector',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    targets=[Target(
+        expr='sum by(connector) (' + Metrics.BUCKETS.raw() + ')',
+        legendFormat='{{connector}}',
+    )])
+
+bucket_additions = TimeSeries(
+    title='Bucket Addition Rate',
+    description='Rate at which buckets are getting added to the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    legendDisplayMode='hidden',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum (rate(' + Metrics.CONNECTOR_CONFIG(opType="add") + '))',
+    )])
+
+bucket_removals = TimeSeries(
+    title='Bucket Removal Rate',
+    description='Rate at which buckets are getting removed from the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    legendDisplayMode='hidden',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum (rate(' + Metrics.CONNECTOR_CONFIG(opType="delete") + '))',
+    )])
+
+
+ack_lag = Heatmap(
+    title='Acknowledgement Lag',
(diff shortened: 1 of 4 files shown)


francoisferrand (commented):
monitoring/oplog-populator/__pycache__/dashboard.cpython-310.pyc should not be commited

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    Stat,
+    PieChart,
+)
+
+class Metrics:
+    RECONFIG_LAG = metrics.BucketMetric(
+        'oplog_populator_reconfiguration_lag_sec',
+        'job', 'connector', 'namespace'
+    ).with_defaults(
+        'job="${oplog_populator_job}"',
```
francoisferrand: this can be added as named parameter in the constructor parameters:

```python
metrics.BucketMetric(
        'oplog_populator_reconfiguration_lag_sec', 'connector',
        job="${oplog_populator_job}", namespace="${namespace}"
    ).with_defaults('connector=~"${connector}"')
```

(unfortunately this does not yet work for other operators...)

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+        'oplog_populator_acknowledgement_lag_sec',
+        'opType', 'job', 'namespace'
+    ).with_defaults(
+        'job="${oplog_populator_job}"',
+        'namespace="${namespace}"',
+    )
+
+    CONNECTOR_CONFIG, BUCKETS, REQUEST_SIZE, PIPELINE_SIZE = [
+        metrics.CounterMetric(
+            name, 'job', 'namespace', *extraLabels
```
francoisferrand: `connector` is always present, it should be move out of `extraLabels` and put here : to avoid redundancy and make it consistent with the default value set for connector [which will fail if connector label is not declared in constructor]

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    )
+
+oplog_populator_up_current = Stat(
+    title='Oplog Populator Instances',
+    description='Number of active oplog populator instances',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='background',
+    noValue='0',
+    reduceCalc='last',
+    graphMode='none',
```
francoisferrand: why remove the graph? We have it in most other dashboards

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+        expr='\n'.join([
+            'sum(up{',
+            '    namespace="${namespace}",',
+            '    job="${kafka_connect_job}"',
+            '})',
+        ]),
+    )],
+    thresholds=[
+        Threshold('red', 0, 0.),
+        Threshold('green', 1, 1.),
```
francoisferrand, line 95: may be yellow if less than expected number of replicas, and green only when we reach that number?

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='value',
+    noValue='0',
+    reduceCalc='last',
+    graphMode='none',
+    targets=[Target(
+        expr='sum(' + Metrics.CONNECTORS.raw(isOld='true') + ')',
+    )],
+    thresholds=[
+        Threshold('red', 0, 0.),
```
francoisferrand: not sure 'red' is correct here: on first startup, we would not have reused ay connector?

francoisferrand: not sure we actually need to display this one actually : not clear to me in which case it would actually be meaningful...

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+        Threshold('green', 1, 1.),
+    ])
+
+total_number_buckets = Stat(
+    title='Total Buckets',
```
francoisferrand: "Bucket count" ? or just "Buckets" ?
Should use the same label which is used in Cloudserver dashboard i guess.

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    description='Number of buckets configured on the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='value',
+    noValue='0',
```
francoisferrand: no 'noValue' here : if we do not have any metric, best display the default ("-") instead of 0, since we don't know the acutal number...

monitoring/oplog-populator/dashboard.py
francoisferrand: may be worth making it clear in the title this is the number of "monitored" buckets

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+bucket_distribution = PieChart(
+    title='Bucket Distribution',
+    description='How buckets are distributed among the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    legendValues=['percent', 'value'],
+    pieType='donut',
+    targets=[Target(
+        expr='sum by(connector) (' + Metrics.BUCKETS.raw() + ')',
+        legendFormat='{{connector}}',
+    )])
```
francoisferrand, line 197: maybe we should also display a graph showing the deviation/variation of distribution over time : e.g. how far from the average (and optimal distribution) each connector is...
→ e.g. displaying `stddev(sum by(connector)(Metrics.BUCKETS))` or `stdvar(sum by(connector)(Metrics.BUCKETS))`

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    description='Maximum size of the mongo pipeline per connector',
+    dataSource='${DS_PROMETHEUS}',
+    colorMode='value',
+    noValue='0',
+    reduceCalc='last',
+    format=UNITS.BYTES,
+    orientation='horizontal',
+    graphMode='none',
+    targets=[Target(
+        expr='sum by(connector) (' + Metrics.PIPELINE_SIZE.raw() + ')',
```
francoisferrand, line 171: i don't think we need to display the per-connector graph: just the overall value is enough (and specific connector can be filtered if needed)

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    title='Bucket Addition Rate',
+    description='Rate at which buckets are getting added to the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum by(connector) (rate(' + Metrics.CONNECTOR_CONFIG(opType="add") + '))',
```
francoisferrand: i don't think we need to display the per-connector graph: just the overall value is enough (and specific connector can be filtered if needed)

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    title='Bucket Removal Rate',
+    description='Rate at which buckets are getting removed from the connectors',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum by(connector) (rate(' + Metrics.CONNECTOR_CONFIG(opType="delete") + '))',
```
francoisferrand: same

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    title='Reconfig Lag By Connector',
+    description='Time it takes to reconfigure a connector in seconds',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.SECONDS,
+    targets=[Target(
+        expr='sum by(connector) (rate(' + Metrics.RECONFIG_LAG.sum() + '))',
```
francoisferrand: i think we are mostly interested in the 'max' of this value : e.g. how long does the system take to reconfigure

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    title='Acknowledgement Lag By Operation Type',
+    description='Delta between a config change in mongo and it getting processed by the oplogPopulator in seconds',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.SECONDS,
+    targets=[Target(
+        expr='sum by(opType) (rate(' + Metrics.ACK_LAG.sum() + '))',
```
francoisferrand: I am not sure that value is correct : the _average_ lag is `Metrics.ACK_LAG.sum()/Metrics.ACK_LAG.count()`, but this format is the rate at which the cumulated lag increases...

(so the by type average would be `sum by(opType)(Metrics.ACK_LAG.sum()) / sum by(opType) (Metrics.ACK_LAG.count())`)

We can display that average value, either as a panel and/or as a graph.
And it may be useful to use the "full" histogram metric to display the repartition of the lag as a heatmap

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+reconfig_lag = TimeSeries(
+    title='Reconfig Lag By Connector',
+    description='Time it takes to reconfigure a connector in seconds',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.SECONDS,
+    targets=[Target(
```
francoisferrand, line 207: same here for the formula

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    targets=[Target(
+        expr='sum by(connector) (' + Metrics.BUCKETS.raw() + ')',
+        legendFormat='{{connector}}',
+    )])
+
+request_body_size_per_connector = TimeSeries(
+    title='Request Body Size Per Connector',
```
francoisferrand: this does not match the snapshot in the PR message ;)
Please update the snapshot, and remember it will actually be useful when you need to provide screenshots for documentation!

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+    description='Size of kafka connect request in bytes',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.BYTES_SEC,
+    targets=[Target(
+        expr='sum by(connector) (rate(' + Metrics.REQUEST_SIZE() + '))',
```
francoisferrand: that metric should be a GAUGE (since the request size _will_ go down, when we stop monitoring some buckets) : so we must not use `rate()`

on top of that, I guess just displaying the value is fine in this case : we basically just want to see the value, which should generally not evolve much, and possibly compare it with the mongodb limit...

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,335 @@
...
+bucket_per_connector = TimeSeries(
+    title='Buckets Per Connector',
+    description='Evolution of buckets per connector',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    targets=[Target(
+        expr='sum by(connector) (' + Metrics.BUCKETS.raw() + ')',
```
francoisferrand: we may want to use `rate()` in here?
and not sure if/how to aggregate....

alternatively, we already have the rate of add and delete, so tracking the number directly here may not make such sense.... but the deviation/variance may (c.f. ⬆️)

monitoring/oplog-populator/dashboard.py
welansari: This metric contains the accumulation of all request body sizes hence making it a counter metric, we can also use it to get the overall bandwidth used by the oplog populator.

monitoring/oplog-populator/dashboard.py
welansari: We don't really need the rate in this case, as we want to see the number of buckets assigned to each connector. The restart of the pod won't create any issues as the number of assigned buckets will be reset to where it was before.

monitoring/oplog-populator/dashboard.py
welansari: removed it

monitoring/oplog-populator/dashboard.py
welansari: i think this one might be a bit confusing for the person using the dashboard.
Also we already have that information in the `Bucket Distribution` panel as we can clearly see when the connectors are unbalanced

monitoring/oplog-populator/dashboard.py
welansari: I don't think this is needed for this dashboard as we are not monitoring Kafka connect

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,362 @@
...
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    legendDisplayMode='hidden',
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum (rate(' + Metrics.CONNECTOR_CONFIG_APPLIED('success="false"') + '))',
+    )])
```
dvasilas, line 183: I am wondering if it would be more clear for the user to see the actual number of errors instead of the rate.

Also, in the screenshot this graph show "No data". Is this the default behavior when there have been no failures ? (if yes, it's probably better to show the graph with 0)

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,362 @@
...
+            'oplog_populator_connector_configuration': ['opType'],
+            'oplog_populator_connector_configuration_applied': ['success'],
+            'oplog_populator_connector_buckets': [],
+            'oplog_populator_connector_request_bytes': [],
+            'oplog_populator_connector_pipeline_bytes': [],
+        }.items()
+    ]
+
+    CONNECTORS = metrics.CounterMetric(
+        'oplog_populator_connectors', 'isOld',
```
dvasilas, line 54: Is `isOld` used somewhere? :thinking:

francoisferrand (approved):

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,362 @@
...
+bucket_per_connector = TimeSeries(
+    title='Buckets Per Connector',
+    description='Evolution of buckets per connector',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    fillOpacity=10,
+    spanNulls=True,
+    targets=[Target(
+        expr='sum by(connector) (' + Metrics.BUCKETS.raw() + ')',
```
dvasilas, line 208: For my understanding: what is an "expected" number (order of magnitude) of connectors ?
(what I am thinking is that this graph may become hard to read if there are 10s of connectors)

monitoring/oplog-populator/dashboard.py
francoisferrand: this is actually a _developer_ bias : we usually run with small number of objects, and we want to be precise : so the precise number is what we expect... but in production, precise values are much higher and don't matter so much: what matters is how frequent changes happen...

dvasilas: Ok, this makes sense.
What I had in mind was that this should ideally stay to 0 and we will want to pay attention to even a low number of failures.

monitoring/oplog-populator/dashboard.py
```diff
@@ -0,0 +1,362 @@
...
+    fillOpacity=10,
+    spanNulls=True,
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum (rate(' + Metrics.CONNECTOR_CONFIG_APPLIED('success="false"') + '))',
+    )])
+
+
+bucket_distribution = PieChart(
+    title='Bucket Distribution',
```
ghost, line 187: I feel like this is misleading, the Bucket from prometheus and the Bucket from S3... Maybe we can have something easier to understand for a not-so-technical reader?

monitoring/oplog-populator/dashboard.py
welansari: We don't know yet actually, the number will depend on the performance of the connectors, meaning that if a single connector can handle tens of thousands of buckets at a time we'll probably have few connectors, and it's true the other way around.

There's a ticket for doing that performance test which will also help determine how we manage the connectors, for now the number of configured connectors is a set value that can be configured at Zenko CR level. This might change later on to make it dynamic in case of sudden degradation of performance with the addition of buckets to connectors.

ghost (approved):
LGTM beyond the nit comment

monitoring/oplog-populator/dashboard.py
welansari: Not anymore, but the metric does have that property so i think maybe it makes sense to keep it in the definition of the metric ?

monitoring/oplog-populator/dashboard.py
dvasilas: Thanks for the explanation !

dvasilas (approved):

PR #2359, opened 2022-12-23
Feature: BB-311 add oplog populator alerts

welansari:
Issue: [BB-311](https://scality.atlassian.net/browse/BB-311)

[BB-311]: https://scality.atlassian.net/browse/BB-311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/monitoring/oplog-populator/alerts.yaml
+++ b/monitoring/oplog-populator/alerts.yaml
@@ -1,0 +1,53 @@
+x-inputs:
+- name: namespace
+  type: constant
+  value: zenko
+- name: oplog_populator_job
+  type: config
+  value: artesca-data-backbeat-oplog-populator-headless
+- name: oplogPopulatorChangeStreamLagThreshold
+  type: config
+  value: 10
+
+groups:
+- name: Oplog Populator
+  rules:
+
+  - alert: OplogPopulatorUnavailable
+    Expr: |
+      sum(up{namespace="${namespace}",job="${oplog_populator_job}"}) < 1
+    For: "30s"
+    Labels:
+     severity: critical
+    Annotations:
+      zenko_service: backbeat-oplog-populator
+      description: "Oplog populator pod is down"
+      summary: "Oplog populator service is in critical state"
+
+  - alert: KafkaConnectFailedConnectorConfiguration
+    Expr: |
+        sum by(connector) (increase(oplog_populator_reconfiguration{success="false",job="${oplog_populator_job}",namespace="${namespace}"}[1m]))
+        > 0
+    For: "5m"
+    Labels:
+     severity: critical
+    Annotations:
+      zenko_service: backbeat-oplog-populator
+      description: "Oplog populator failed to configure connector"
+      summary: "Oplog populator couldn't update kafka connect connector"
+
+  - alert: OplogPopulatorMetastoreChangeStreamLagThreshold
+    Expr: |
+        histogram_quantile(
+          0.99,
+          sum by(le) (rate(oplog_populator_acknowledgement_lag_sec_bucket{job="${oplog_populator_job}",namespace="${namespace}"}[1m]))
+        )
+        >= ${oplogPopulatorChangeStreamLagThreshold}
+    For: "5m"
+    Labels:
+     severity: critical
+    Annotations:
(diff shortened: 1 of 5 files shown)


monitoring/oplog-populator/alerts.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - alert: OplogPopulatorUnavailable
+    Expr: |
+      sum(up{namespace="${namespace}",job="${oplog_populator_job}"}) < 1
+    For: "30s"
+    Labels:
+     severity: critical
+    Annotations:
+      zenko_service: backbeat-oplog-populator
+      description: "Oplog populator pod is down"
+      summary: "Oplog populator service is critical"
```
ghost: I am not sure "service is critical" is very meaningful, why not "service state is in critical state" or something like that?

francoisferrand (commented):
should we also have alerts:
- if the we fail to update kafka-connect config
- if updating kafka-connect config takes too long (since metastore update)

francoisferrand:
Also, practically this PR may be stacked in https://github.com/scality/backbeat/pull/2358 : to avoid merge conflicts eventually...

monitoring/oplog-populator/alerts.yaml
```diff
@@ -0,0 +1,58 @@
...
+    Labels:
+     severity: critical
+    Annotations:
+      zenko_service: backbeat-oplog-populator
+      description: "Oplog populator pod is down"
+      summary: "Oplog populator service is in critical state"
+
+  - alert: KafkaConnectFailedConnectorConfiguration
+    Expr: |
+        (sum (rate(oplog_populator_reconfiguration{success="false",job="${oplog_populator_job}",namespace="${namespace}"}[1m]))
```
francoisferrand: not sure the "ratio" is significant : we want the system to be "fully" configured properly, i.e. for each run of the oplog populator we expect that it configured every connector properly. If 1 connector is not configured properly, we will not get all events, which is an issue → just raise an alert.

here, this would simply mean raise an alert when `increase(oplog_populator_reconfiguration{success=false}[1m]) > 0` i guess.

(and we may check this over 5 minutes maybe, via the `For:` directive, to filter out temporary failures)

monitoring/oplog-populator/alerts.yaml
```diff
@@ -0,0 +1,58 @@
...
+    Labels:
+     severity: critical
+    Annotations:
+      zenko_service: backbeat-oplog-populator
+      description: "Oplog populator failed to configure connectors"
+      summary: "Oplog populator couldn't update kafka connect connectors"
+
+  - alert: OplogPopulatorMetastoreChangeStreamLagThreshold
+    Expr: |
+        sum(rate(
```
francoisferrand: instead of checking the average, we should prbably compare the 95th or 99th quantile

monitoring/oplog-populator/alerts.yaml
```diff
@@ -48,7 +48,7 @@ groups:
           oplog_populator_acknowledgement_lag_sec_count{job="${oplog_populator_job}",namespace="${namespace}"}[1m]
         ))
         >= ${oplogPopulatorChangeStreamLagThreshold}
-    For: "30s"
+    For: "5m"
```
XinLi0207, line 46: if I understand correctly, `For` clause means prom waits for 5m to trigger the alert after it first encounters the issue in Expr, right?
just out of curiosity, how we choose this duration? some of them are 30s, some are 1m some are 5m

welansari: Yes that's what it means, but only if the condition persists for that amount of time.
There's no real way of choosing it in this case, as we don't know yet how the feature behaves in prod

XinLi0207: ok I see, so may adjust these numbers accordingly by how they behave in the prod later right?

welansari: yes

XinLi0207 (approved):

francoisferrand (approved):

PR #2360, opened 2023-01-03
BB-355 use updateDescription field to get object metadata in kafka lo…

welansari:
…g consumer

Issue: [BB-355](https://scality.atlassian.net/browse/BB-355)

**Problem:**
Some times oplog update events we get when using the Kafka log consumer don't have any metadata. This mostly happens when deleting the object right after the update.

The issue happens because we use the `fullDocument` field to retrieve the metadata, when reading the mongodb docs we see that the field represents a copy of the entire updated document at a point in time after the update, which means that in some cases this value can be unavailable like when deleting right after the update.

**Fix:**
Use the `updateDescription` field to get object metadata instead of fullDocument.
Values from `updateDescription` represent the exact changes that occurred to the object.
It is possible to have non complete metadata in this field but this shouldn't be an issue as we update the full object each time in the mongoClient.
One key updates like we do in the update we perform before each deletion will be ignored as we don't even have the `value` field that's supposed to contain the metadata.

[BB-355]: https://scality.atlassian.net/browse/BB-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BB-355]: https://scality.atlassian.net/browse/BB-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
+++ b/lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
@@ -51,6 +51,29 @@ class ListRecordStream extends stream.Transform {
     }
 
     /**
+     * Get object metadata from oplog event
+     * @param {ChangeStreamDocument} changeStreamDocument changeStreamDocument
+     * @returns {string|undefined} stringified objectMd
+     */
+    _getObjectMd(changeStreamDocument) {
+        // Metadata values from updateDescription are used instead of
+        // fullDocument as they represent the exact changes that occured
+        // to the object. We only need this for the update operation.
+
+        // The fullDocument value contains a copy of the entire updated
+        // document at a point in time after the update, this can cause issues
+        // when the object gets deleted just after the update (only for updates).
+        const updateDescriptionMd = changeStreamDocument.updateDescription
+            && changeStreamDocument.updateDescription.updatedFields
+            && changeStreamDocument.updateDescription.updatedFields.value;
+        if (changeStreamDocument.operationType === 'update') {
+            return JSON.stringify(updateDescriptionMd);
+        }
+        const fullDocumentMd = changeStreamDocument.fullDocument && changeStreamDocument.fullDocument.value;
+        return JSON.stringify(fullDocumentMd);
+    }
+
+    /**
      * Formats change stream entries
      * @param {Object} data chunk of data
      * @param {Buffer} data.value message contents as a Buffer
@@ -77,13 +100,15 @@ class ListRecordStream extends stream.Transform {
             // skipping the event
             return callback(null, null);
         }
+        const objectMd = this._getObjectMd(changeStreamDocument);
+        const opType = this._getType(changeStreamDocument.operationType, objectMd);
         const streamObject = {
             timestamp: new Date(data.timestamp),
             db: changeStreamDocument.ns && changeStreamDocument.ns.coll,
             entries: [{
                 key: changeStreamDocument.documentKey && changeStreamDocument.documentKey._id,
-                type: this._getType(changeStreamDocument.operationType, changeStreamDocument.fullDocument),
-                value: changeStreamDocument.fullDocument && JSON.stringify(changeStreamDocument.fullDocument.value),
+                type: opType,
+                value: objectMd,
             }],
         };
         return callback(null, streamObject);
--- a/extensions/oplogPopulator/constants.js
+++ b/extensions/oplogPopulator/constants.js
@@ -3,7 +3,6 @@ const constants = {
     defaultConnectorName: 'source-connector',
     defaultConnectorConfig: {
         'connector.class': 'com.mongodb.kafka.connect.MongoSourceConnector',
-        'change.stream.full.document': 'updateLookup',
         'pipeline': '[]',
         'collection': '',
         // JSON output converter config
--- a/tests/unit/lib/queuePopulator/kafkaLogConsumer/ListRecordStream.js
+++ b/tests/unit/lib/queuePopulator/kafkaLogConsumer/ListRecordStream.js
@@ -19,25 +19,34 @@ const changeStreamDocument = {
         }
     }
 };
-const kafkaMessage = {
-    value: Buffer.from(JSON.stringify(changeStreamDocument)),
-    timestamp: Date.now(),
-    size: 2,
-    topic: 'oplog-topic',
-    offset: 1337,
-    partition: 0,
-    key: Buffer.from('key'),
+const changeStreamDocumentUpdate = {
+    ns: {
+        db: 'metadata',
+        coll: 'example-bucket',
+    },
+    documentKey: {
+        _id: 'example-key',
+    },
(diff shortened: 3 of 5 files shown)


lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -50,6 +50,29 @@ class ListRecordStream extends stream.Transform {
...
+    _getObjectMd(changeStreamDocument) {
+        // Metadata values from updateDescription are used instead of
+        // fullDocument as they represent the exact changes that occured
+        // to the object. We only need this for the update operation.
+
+        // The fullDocument value contains a copy of the entire updated
+        // document at a point in time after the update, this can cause issues
+        // when the object gets deleted just after the update (only for updates).
+        const updateDescriptionMd = changeStreamDocument.updateDescription
+            && changeStreamDocument.updateDescription.updatedFields &&
```
francoisferrand: inconsistent wrapping: either put the operator at the end of the line (before line break) or at the beginning of next line (after line break), but do not mix...
(not sure what rule to apply, but at least it should be consistent)

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -83,7 +106,7 @@ class ListRecordStream extends stream.Transform {
             entries: [{
                 key: changeStreamDocument.documentKey && changeStreamDocument.documentKey._id,
                 type: this._getType(changeStreamDocument.operationType, changeStreamDocument.fullDocument),
```
francoisferrand: we still use fullDocument here : should be changed, e.g. pass `this._getObjectMd(changeStreamDocument)` instead to `getType()` ?

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -50,6 +50,29 @@ class ListRecordStream extends stream.Transform {
...
+        // The fullDocument value contains a copy of the entire updated
+        // document at a point in time after the update, this can cause issues
+        // when the object gets deleted just after the update (only for updates).
+        const updateDescriptionMd = changeStreamDocument.updateDescription
+            && changeStreamDocument.updateDescription.updatedFields &&
+            changeStreamDocument.updateDescription.updatedFields.value;
+        if (changeStreamDocument.operationType === 'update') {
+            return updateDescriptionMd && JSON.stringify(updateDescriptionMd);
+        }
+        const fullDocumentMd = changeStreamDocument.fullDocument && changeStreamDocument.fullDocument.value;
```
francoisferrand, line 72: `change.stream.full.document` is disabled, is it still relevant (e.g. in case 'set' operationType i guess) ?

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -50,6 +50,29 @@ class ListRecordStream extends stream.Transform {
...
+        // to the object. We only need this for the update operation.
+
+        // The fullDocument value contains a copy of the entire updated
+        // document at a point in time after the update, this can cause issues
+        // when the object gets deleted just after the update (only for updates).
+        const updateDescriptionMd = changeStreamDocument.updateDescription
+            && changeStreamDocument.updateDescription.updatedFields &&
+            changeStreamDocument.updateDescription.updatedFields.value;
+        if (changeStreamDocument.operationType === 'update') {
+            return updateDescriptionMd && JSON.stringify(updateDescriptionMd);
```
francoisferrand: are there not 2 cases, e.g. when we set the whole value vs when we update a single 'deep' field ?

From yesterday, i understand we update single field only when setting the 'delete' flag, so we may actually just skip that oplog entry?

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: yes `insert` operations which we do when first creating an object only have the fullDocument field

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
welansari: it will be skipped at the logReader level

welansari: https://github.com/scality/backbeat/blob/d9aa80bfcc4335f65379223e4226c504da16ce52/lib/queuePopulator/LogReader.js#L496-L498

francoisferrand: may be useful to add a comment in `getObjectMD()` explaining what we expect here (and how it will be fixed otherwise)

francoisferrand (approved):
LGTM

lib/queuePopulator/KafkaLogConsumer/ListRecordStream.js
```diff
@@ -50,6 +50,29 @@ class ListRecordStream extends stream.Transform {
...
+        // document at a point in time after the update, this can cause issues
+        // when the object gets deleted just after the update (only for updates).
+        const updateDescriptionMd = changeStreamDocument.updateDescription
+            && changeStreamDocument.updateDescription.updatedFields
+            && changeStreamDocument.updateDescription.updatedFields.value;
+        if (changeStreamDocument.operationType === 'update') {
+            return updateDescriptionMd && JSON.stringify(updateDescriptionMd);
+        }
+        const fullDocumentMd = changeStreamDocument.fullDocument && changeStreamDocument.fullDocument.value;
+        return fullDocumentMd && JSON.stringify(fullDocumentMd);
```
XinLi0207: just a question out of my curiosity

the entries.value before it was `changeStreamDocument.fullDocument && JSON.stringify(changeStreamDocument.fullDocument.value),`

now we kind of changed it to sth like 
```
changeStreamDocument.fullDocument && 
changeStreamDocument.fullDocument.value &&
JSON.stringify(changeStreamDocument.fullDocument && 
changeStreamDocument.fullDocument.value)
```

what do we want to return here as value? why we don't just let the last one as return value

welansari: Looking at it, there is a redundant check indeed, `JSON.stringify(undefined)` should equal `undefined`.
So this `return fullDocumentMd && JSON.stringify(fullDocumentMd);` should become `return JSON.stringify(fullDocumentMd);`

XinLi0207: ok cool! you already updated that 👍

XinLi0207 (approved):

PR #2361, opened 2023-01-17
BB-338 Suspend Lifecycle transition to paused location

welansari:
Issue: [BB-338](https://scality.atlassian.net/browse/BB-338)

*Changes :*

- Adapt the same Pause/Resume API routes used for replication and ingestion to also support lifecycle routes
- Lifecycle routes will use mongo to store the location status
- Backbeat API will be the only process modifying the state of locations in MongoDB, hence scheduling a resume job will be handled in Backbeat API contrary to how it works in other extensions.
- The lifecycle bucket processor will be listening to mongo events using change streams and will be pausing and resuming locations based on the modifications to the state applied to MongoDB

**Note for reviewers:** 
Please review per commit as it will make understanding the changes easier.

[BB-338]: https://scality.atlassian.net/browse/BB-338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
+++ b/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
@@ -13,6 +13,7 @@ const LifecycleTask = require('../tasks/LifecycleTask');
 const safeJsonParse = require('../util/safeJsonParse');
 const ClientManager = require('../../../lib/clients/ClientManager');
 const { authTypeAssumeRole } = require('../../../lib/constants');
+const LocationStatusStream = require('../../utils/LocationStatusStream');
 
 const PROCESS_OBJECTS_ACTION = 'processObjects';
 
@@ -55,14 +56,16 @@ class LifecycleBucketProcessor {
      * @param {Object} s3Config - s3 config
      * @param {String} s3Config.host - host ip
      * @param {String} s3Config.port - port
+     * @param {Object} mongoConfig - MongoDb Connection Config
      * @param {String} transport - http or https
      */
-    constructor(zkConfig, kafkaConfig, lcConfig, repConfig, s3Config, transport = 'http') {
+    constructor(zkConfig, kafkaConfig, lcConfig, repConfig, s3Config, mongoConfig, transport = 'http') {
         this._log = new Logger('Backbeat:Lifecycle:BucketProcessor');
         this._zkConfig = zkConfig;
         this._kafkaConfig = kafkaConfig;
         this._lcConfig = lcConfig;
         this._repConfig = repConfig;
+        this._mongoConfig = mongoConfig;
         this._producer = null;
         this._kafkaBacklogMetrics = null;
 
@@ -80,6 +83,9 @@ class LifecycleBucketProcessor {
 
         this._producerReady = false;
         this._consumerReady = false;
+
+        this._pausedLocations = new Set();
+        this._locationStatusStream = null;
 
         this.clientManager = new ClientManager({
             id: 'lifecycle',
@@ -151,8 +157,42 @@ class LifecycleBucketProcessor {
             objectTasksTopic: this._lcConfig.objectTasksTopic,
             kafkaBacklogMetrics: this._kafkaBacklogMetrics,
             ncvHeap: this.ncvHeap,
+            pausedLocations: this._pausedLocations,
             log: this._log,
         };
+    }
+
+    /**
+     * Pause bucket(s) based on given location
+     * @param {String} location - location
+     * @return {undefined}
+     */
+    _pauseServiceForLocation(location) {
+        const paused = this._pausedLocations.has(location);
+        if (!paused) {
+            this._pausedLocations.add(location);
+            this._log.info('location paused', {
+                method: 'LifecycleBucketProcessor._pauseServiceForLocation',
+                location,
+            });
+        }
+    }
+
+    /**
+     * Resume lifecycle operations for a location
+     * @param {String} location - location
+     * @param {Date} [date] - optional date object for scheduling resume
+     * @return {undefined}
+     */
+    _resumeServiceForLocation(location) {
+        const paused = this._pausedLocations.has(location);
+        if (paused) {
+            this._pausedLocations.delete(location);
+            this._log.info('location resumed', {
+                method: 'LifecycleBucketProcessor._resumeServiceForLocation',
+                location,
+            });
+        }
     }
 
     /**
@@ -366,6 +406,14 @@ class LifecycleBucketProcessor {
             done => this._setupProducer(done),
             done => this._initKafkaBacklogMetrics(done),
             done => this._setupConsumer(done),
+            done => {
+                this._locationStatusStream = new LocationStatusStream('lifecycle',
+                    this._mongoConfig,
+                    this._pauseServiceForLocation.bind(this),
+                    this._resumeServiceForLocation.bind(this),
+                    this._log);
+                this._locationStatusStream.start(done);
+            }
         ], done);
     }
 
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -26,6 +26,8 @@ const errorObjectTemporarilyRestored = errors.InternalError.
     customizeDescription('object temporarily restored');
 const errorReplicationInProgress = errors.InternalError.
     customizeDescription('replication of the object is currently in progress');
+const errorLocationPaused = errors.InternalError.
+    customizeDescription('lifecycle events to location have been paused');
 
 
 // Default max AWS limit is 1000 for both list objects and list object versions
@@ -1065,12 +1067,20 @@ class LifecycleTask extends BackbeatTask {
      */
     _applyTransitionRule(params, log) {
         async.waterfall([
+            next => {
+                // No transition if lifecycle events to location were paused
+                if (this.pausedLocations.has(params.site)) {
+                    return next(errorLocationPaused);
+                }
+                return next();
+            },
             next =>
                 this._getObjectMD(params, log, (err, objectMD) => {
                     LifecycleMetrics.onS3Request(log, 'getMetadata', 'bucket', err);
                     return next(err, objectMD);
                 }),
             (objectMD, next) => {
+                // No transition when replication is ongoing or failed
                 const replicationStatus = objectMD.getReplicationStatus();
                 if (['PENDING', 'PROCESSING', 'FAILED'].includes(replicationStatus)) {
                     return next(errorReplicationInProgress);
--- a/extensions/utils/LocationStatusStream.js
+++ b/extensions/utils/LocationStatusStream.js
@@ -1,0 +1,167 @@
+const { MongoClient } = require('mongodb');
+const async = require('async');
+
+const { locationStatusCollection } = require('../../lib/constants');
+const { constructConnectionString } = require('./MongoUtils');
+const ChangeStream = require('../../lib/wrappers/ChangeStream');
+
+/**
+ * Pause a location
+ * @typedef { (locationName: string,) => void } PauseLocationFn
+ */
+/**
+ * Resume a location
+ * @typedef { (locationName: string) => void } ResumeLocationFn
+ */
+
+/**
+ * @class LocationStatusStream
+ *
+ * @classdesc uses change streams to listen to a location statuss
+ * for a service (crr, ingestion, lifecycle)
+ */
+class LocationStatusStream {
+
+    /**
+     * @constructor
+     * @param {string} serviceName service name (i.e crr, ingestion, lifecycle)
+     * @param {Object} mongoConfig mongodb configuration object
+     * @param {PauseLocationFn} pauseFn pause location function
+     * @param {ResumeLocationFn} resumeFn resume location function
+     * @param {Logger} logger logger instance
+     */
+    constructor(serviceName, mongoConfig, pauseFn, resumeFn, logger) {
+        this._serviceName = serviceName;
+        this._mongoConfig = mongoConfig;
+        this._mongoClient = null;
+        this._locationStatusColl = null;
+        this._pauseServiceForLocation = pauseFn;
+        this._resumeServiceForLocation = resumeFn;
+        this._log = logger;
+    }
+
+    /**
+     * Set up the mongo client and initiate the change stream
+     * @param {function} done - callback
+     * @return {undefined}
+     */
+    start(done) {
+        async.series([
+            done => this._setupMongoClient(done),
+            done => this._setChangeStream(done),
+            done => this._initializeLocationStatuses(done),
+        ], done);
+    }
+
+    /**
+     * Connects to MongoDB using and retreives
+     * the location status collection
+     * @param {function} cb callback
+     * @returns {undefined} undefined
+     */
+    _setupMongoClient(cb) {
+        const mongoUrl = constructConnectionString(this._mongoConfig);
+        MongoClient.connect(mongoUrl, {
+                replicaSet: this._mongoConfig.replicaSet,
+                useNewUrlParser: true,
+                useUnifiedTopology: true,
+        }, (err, client) => {
+            if (err) {
+                this._log.error('Could not connect to MongoDB', {
+                    method: 'ServiceStatusManager._setupMongoClient',
+                    error: err.message,
+                });
+                return cb(err);
+            }
+            // connect to metadata DB
+            this._mongoClient = client.db(this._mongoConfig.database, {
+                ignoreUndefined: true,
+            });
+            this._locationStatusColl = this._mongoClient.collection(locationStatusCollection);
+            this._log.info('Connected to MongoDB', {
+                method: 'ServiceStatusManager._setupMongoClient',
+            });
+            return cb();
+        });
+    }
+
+    /**
+     * Initialized the list of paused locations
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _initializeLocationStatuses(cb) {
+        this._locationStatusColl.find({})
+            .toArray((err, locations) => {
+                if (err) {
+                    this._log.error('Could not fetch location statuses from mongo', {
+                        method: 'ServiceStatusManager._initializeLocationStatuses',
+                        error: err.message,
+                    });
+                    return cb(err);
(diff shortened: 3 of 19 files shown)


extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -328,6 +373,129 @@ class LifecycleBucketProcessor {
...
+                return cb();
+            });
+    }
+
+    /**
+     * Initializes a change stream on the metastore collection
+     * @param {function} cb callback
+     * @returns {undefined}
+     */
+    _setMetastoreChangeStream(cb) {
```
francoisferrand: this is not the "metastore" collection (i.e. the collection with the list of buckets), but the "location status" collection

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -340,6 +508,9 @@ class LifecycleBucketProcessor {
             done => this._setupProducer(done),
             done => this._initKafkaBacklogMetrics(done),
             done => this._setupConsumer(done),
+            done => this._setupMongoClient(done),
+            done => this._initializeLocationStatuses(done),
```
francoisferrand: there is a race condition here:
* Read initial state → paused
* User triggers a resume, which is added to mongo & all...
* Create the change stream → no event received
* Processor is still paused...

we should first setup the change stream, then look for the initial state ; and in the change stream callback we need to be resilient to both:
* being the first event, e.g. happening before we could do the initial state management
* being already in the expected state
(should not be hard, but just to keep in ming)

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -328,6 +373,129 @@ class LifecycleBucketProcessor {
         });
     }
 
+    /**
+     * Connects to MongoDB using and retreives
+     * the location status collection
+     * @param {function} cb callback
+     * @returns {undefined} undefined
+     */
+    _setupMongoClient(cb) {
```
francoisferrand: it seems to me the whole mongodb change stream / status management could easily be extracted and made reusable : just need to eventually be able to call "resume" and "suspend" functions on the processor (possibly via callbacks)

it will then be easier to port this change to CRR and Ingestion....

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -328,6 +373,129 @@ class LifecycleBucketProcessor {
...
+                return cb(err);
+            }
+            // connect to metadata DB
+            this._mongoClient = client.db(this._mongoConfig.database, {
+                ignoreUndefined: true,
+            });
+            // creating a collection in case it doesn't exist, this is done as a safety measure
+            // but normally it will be created by backbeat API that gets deployed before the
+            // lifecycle bucket processor.
+            return this._mongoClient.createCollection(locationStatusCollection, err => {
```
francoisferrand: > locationStatusCollection

don't we need to add "filters" in other parts of Backbeat, to ensure this collection is not unexpectedly processed by replication/ingestion/... ?

lib/api/constants.js
```diff
@@ -0,0 +1,5 @@
+const constants = {
+    locationStatusCollection: '__locationStatusStore',
```
francoisferrand: does not seem related to "api", i would expect to find this closer to database : is there some "central" place where we already define the name of the metastore or other system collections?

lib/api/BackbeatAPI.js
```diff
@@ -1088,22 +1124,27 @@ class BackbeatAPI {
...
-        let sites;
-        if (details.site === 'all') {
-            sites = details.extensions[service].filter(s => s !== 'all');
-        } else {
-            sites = [details.site];
-        }
-        sites.forEach(site => {
+        const sites = this.getRequestedSites(details);
+        async.eachLimit(sites, 10, (site, next) => {
+            if (service === 'lifecycle') {
```
francoisferrand: this does not look really nice, with this state management intermingled everywhere (which was indeed already the case)

I am wondering if it would not be better to introduce some kind of interface "LocationStatusManager", called by this api : so we could have 2 implementations (redis/zookeeper and mongo), and easily switch one or the other...

That would be much cleaner IMO, but i am wondering if this is worth it (and how long that would take), esp. since we should eventually change this for CRR and ingestion as well... On the positiive side, it would make it easier to keep both codes around, progressively migrate (e.g. just ingestion) and possibly keep the code compatible with RING (where we will not have mongo, in case we want to upgrade RING to backbeat 8.x)

lib/api/BackbeatAPI.js
```diff
@@ -1171,7 +1261,18 @@ class BackbeatAPI {
             schedule.setHours(schedule.getHours() + hours);
         }
 
-        sites.forEach(site => {
+        return async.eachLimit(sites, 10, (site, next) => {
+            if (service === 'lifecycle') {
```
XinLi0207: we have three services here and doing different kinds of things for each, just my personal thinking, maybe a `switch(service) case` logic is more clear?

lib/api/constants.js
welansari: We don't have anything like that actually, the `metastore` collection name for example is hardcoded into the mongo client in arsenal

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
welansari: I think there's no need for that, in change stream implementation we only listen to buckets defined on the metastore, and in the oplog implementation, this would still be ignored as we only replicate proper objects that have a replication status.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -916,6 +918,11 @@ class LifecycleTask extends BackbeatTask {
                     return next(err, objectMD);
                 }),
             (objectMD, next) => {
+                // No transition if lifecycle events to location were paused
+                if (this.pausedLocations.has(params.site)) {
```
nicolas2bert: *perf* What about moving it up as the "first task" to avoid extra GET requests (from _getObjectMD)?

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -916,6 +918,11 @@ class LifecycleTask extends BackbeatTask {
...
+                    return next(errorLocationPaused);
```
nicolas2bert: *debug* Should we add a log with the paused location name?

extensions/utils/LocationStatusStream.js
```diff
@@ -0,0 +1,180 @@
...
+ */
+/**
+ * Resume a location
+ * @typedef { (locationName: string) => void } ResumeLocationFn
+ */
+
+/**
+ * @class LocationStatusStream
+ *
+ * @classdesc uses change streams to listen to a location statuss
```
nicolas2bert, line 20: *nit* status

extensions/utils/LocationStatusStream.js
```diff
@@ -0,0 +1,180 @@
...
+                return cb(err);
+            }
+            // connect to metadata DB
+            this._mongoClient = client.db(this._mongoConfig.database, {
+                ignoreUndefined: true,
+            });
+            // creating a collection in case it doesn't exist, this is done as a safety measure
+            // but normally it will be created by backbeat API that gets deployed before the
+            // lifecycle bucket processor.
+            return this._mongoClient.createCollection(locationStatusCollection, err => {
```
nicolas2bert: *nit* No need to explicitly create the MongoDB collection unless specific options/config are needed.

extensions/utils/LocationStatusStream.js
```diff
@@ -0,0 +1,180 @@
...
+                // in case the collection already exists, we ignore the error
+                if (err && err.codeName !== 'NamespaceExists') {
+                    this._log.error('Could not create mongo collection', {
+                        collection: locationStatusCollection,
```
nicolas2bert: *debug* You could add  the method to this log.

extensions/utils/LocationStatusStream.js
```diff
@@ -0,0 +1,180 @@
...
+                return cb();
+            });
+    }
+
+    /**
+     * Initializes a change stream on the location status collection
+     * @param {function} cb callback
+     * @returns {undefined}
+     */
+    _setChangeStream(cb) {
```
nicolas2bert, line 118: *nit* This function is synchronous.

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,552 @@
...
+    _listCollectionDocuments(cb) {
+        this._locationStatusColl.find({}, (err, cursor) => {
+            if (err) {
+                this._logger.error('Could not list documents', {
+                    method: 'LocationStatusManager._listCollectionDocuments',
+                    collection: locationStatusCollection,
+                    error: err.message,
+                });
+            }
+            return cursor.toArray(cb);
```
nicolas2bert, line 100: same

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,552 @@
...
+     * @param {Object[]} locations list of locations data from mongo
+     * @param {function} cb callback
+     * @returns {undefined}
+     */
+    _getPreviousLocationStates(locations, cb) {
+        const validLocationNames = Object.keys(validLocations);
+        locations
+            .filter(loc => validLocationNames.includes(loc._id))
+            .forEach(loc => {
+                this._locationStatusStore[loc._id] = new LocationStatus(supportedServices, loc.value);
```
nicolas2bert: *archi* Why is `LocationStatus` in Arsenal? It seems to be "highly" correlated with Backbeat.

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,552 @@
...
+        });
+    }
+
+    /**
+     * Add valid mongo locations to local store
+     * @param {Object[]} locations list of locations data from mongo
+     * @param {function} cb callback
+     * @returns {undefined}
+     */
+    _getPreviousLocationStates(locations, cb) {
```
nicolas2bert, line 110: *nit* This function looks synchronous.

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,552 @@
...
+        const invalidLocations = locations
+            .filter(loc => !validLocationNames.includes(loc._id))
+            .map(loc => loc._id);
+        return this._locationStatusColl.deleteMany({
+            _id: {
+                $in: invalidLocations
+            }
+        }, err => {
+            if (err) {
+                this._logger.error('Couldn not delete invalid locations', {
```
nicolas2bert: *nit* Could

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,552 @@
...
+            }
+            if (reqBody.hours === '' || isNaN(reqBody.hours) ||
+                parseInt(reqBody.hours, 10) <= 0) {
+                return {
+                    error: errors.MalformedPOSTrequest.customizeDescription(
+                        `${msg}: hours must be an integer greater than 0`),
+                };
+            }
+            return { hours: parseInt(reqBody.hours, 10) };
+        } catch (e) {
```
nicolas2bert, line 401: *debug* should we log the error?

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,552 @@
...
+                return next();
+            }
+            const paused = this._locationStatusStore[location].getServicePauseStatus(service);
+            if (paused) {
+                this._deleteResumeJob(location, service);
+                this._locationStatusStore[location].setServiceResumeSchedule(service, null);
+                return this._updateServiceStatusForLocation(location, next);
+            }
+            return next();
+        }, () => {
```
nicolas2bert: What if `_updateServiceStatusForLocation` fails, should we return an error to our user?
Also, might want to make sure it returns an "arsenal type of error".

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,552 @@
...
+                this._pushActionToRedis(service, location, actions.pause);
+                return next();
+            }
+            const paused = this._locationStatusStore[location].getServicePauseStatus(service);
+                if (!paused) {
+                    this._locationStatusStore[location].pauseLocation(service);
+                    return this._updateServiceStatusForLocation(location, next);
+                }
+            return next();
+        }, () => {
```
nicolas2bert: same for error.

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,552 @@
...
+            const paused = this._locationStatusStore[location].getServicePauseStatus(service);
+            if (paused) {
+                this._locationStatusStore[location].resumeLocation(service, schedule);
+                if (schedule) {
+                    this._scheduleResumeJob(location, service, schedule);
+                }
+                return this._updateServiceStatusForLocation(location, next);
+            }
+            return next();
+        }, () => {
```
nicolas2bert: same for error

extensions/lifecycle/tasks/LifecycleTask.js
welansari: we already have that in https://github.com/scality/backbeat/blob/375383c1b4651629349b550784cdcb3bdbd8369d/extensions/lifecycle/tasks/LifecycleTask.js#L976-L983

lib/util/LocationStatusManager.js
welansari: The initial reason was cause it's a MongoDB document model, which means that it can get accessed by everyone.

But i guess currently the information is only accessed by backbeat and the logic of pause/resume is also very specific to backbeat. I moved the model to backbeat

extensions/utils/LocationStatusStream.js
welansari: The reason i used callbacks for this and all the other synchronous functions was for cleanness of the code, so that they can fit neatly into the async.series/waterfalls

tests/unit/api/BackbeatAPI.spec.js
```diff
@@ -5,15 +5,18 @@ const BackbeatRequest = require('../../../lib/api/BackbeatRequest');
...
     let bbapi;
 
     const destconfig = config.extensions.replication.destination;
     const { site } = destconfig.bootstrapList[0];
     const ingestSite = `${destconfig.bootstrapList[1].site}-ingestion`;
+    const coldSite = Object.keys(locationConfig)
+        .filter(site => locationConfig[site].isCold)[0];
 
-    before(() => {
+    beforeEach(() => {
```
KillianG: Do we really need to create a new `BackbeatAPI` before every test ? Why wasn't it like that before ?

lib/api/BackbeatAPI.js
```diff
@@ -1383,14 +1253,47 @@ class BackbeatAPI {
...
+                {
+                    crr: {
+                        namespace: zookeeperReplicationNamespace,
+                        statePath: zkReplicationStatePath,
+                    },
+                    ingestion: {
+                        namespace: zookeeperIngestionNamespace,
+                        statePath: zkIngestionStatePath,
+                    },
+                },
```
francoisferrand: nit: maybe we should merge the "topics" and "zookeeper" params in a single map:

```suggestion
                {
                    crr: {
                        topic: this._crrTopic,
                        namespace: zookeeperReplicationNamespace,
                        statePath: zkReplicationStatePath,
                    },
                    ingestion: {
                        topic: this._ingestionTopic,
                        namespace: zookeeperIngestionNamespace,
                        statePath: zkIngestionStatePath,
                    },
                },
```

francoisferrand: and also merge the information from `mongoLocationStatusServices`, somehow: e.g. should have an entry to the service even if it is mongo, though in that case it would NOT have the redis topic or zookeeper state.

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,583 @@
...
+        this._zkClient = zkClient;
+        this._redis = redis;
+        this._topics = topics;
+        this._zkDetails = zkDetails;
+        this._logger = logger;
+
+        this._locationStatusColl = null;
+
+        this._locationStatusStore = {};
+        this._scheduledResumeJobs = {
```
francoisferrand, line 57: can this/should this be an attribute of `LocationStatus` ?
(and thus avaialble through the `locationStatusStore` map)

lib/util/LocationStatusManager.js
```diff
@@ -0,0 +1,583 @@
...
+     * @param {string[]} locations location names to pause
+     * @param {Function} cb callback
+     * @returns {undefined}
+     */
+    _getStateDetails(service, locations, cb) {
+        if (this._mongoServices.includes(service)) {
+            const selectedLocations = {};
+            locations.forEach(loc => {
+                selectedLocations[loc] = this._locationStatusStore[loc];
+            });
```
francoisferrand: this should do the mapping done in the later functions (`.paused = .getServicePauseStatus(service)` etc, c.f. lines 315, 342...), to complete the abstraction :

```suggestion
            locations.forEach(loc => {
                selectedLocations[loc] = this._locationStatusStore[loc].getValue()[service]
            });
```

(or better introduce a `LocationStatus.getService()` function 😉)

tests/unit/lib/util/LocationStatusManager.spec.js
```diff
@@ -0,0 +1,366 @@
...
+            },
+            fakeLogger,
+        );
+    });
+
+    afterEach(() => {
+        sinon.restore();
+    });
+
+    it('pauseService:: should pause selected service', done => {
```
francoisferrand: nit: `pauseService::` should be in a describe block instead, to keep the "flow" of describing the tests

francoisferrand (approved):

lib/util/LocationStatusManager.js
welansari: `_scheduledResumeJobs` contains the actual scheduled jobs for all services (that use both mongo and zookeeper), this can't be put in the `LocationStatus` as it'll only work for mongo services

KillianG (approved):

PR #2364, opened 2023-02-10
feature/BB 358/supportNewerNoncurrentVersion

alexanderchan-scality:
- BB-358: add support for NewerNonccurrentVersions
- BB-358: add tests for NewerNoncurrentVersion helpers
- bump aws-sdk to v2.1035.0

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -8,6 +8,7 @@ const {
     LifecycleDateTime,
     LifecycleUtils,
 } = require('arsenal').s3middleware.lifecycleHelpers;
+const { CompareResult, MinHeap } = require('arsenal').algorithms.Heap;
 
 const config = require('../../../conf/Config');
 const { attachReqUids } = require('../../../lib/clients/utils');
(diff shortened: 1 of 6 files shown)


extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -241,7 +374,23 @@ class LifecycleTask extends BackbeatTask {
...
+                    if (err) {
+                        return done(err);
+                    }
+
+                    if (data.IsTruncated) {
+                        this._ncvHeapBucketClear(bucketData.target.bucket);
+                        return done();
+                    }
+
+                    this._ncvHeapObjectsClear(
```
nicolas2bert: Should we not keep NextMarker versions only if the listing is truncated?
Also, you might want to add a comment to ease the code review.

Is this pagination logic tested somewhere?

alexanderchan-scality: with the uniqueKey set, NextMarkers key versions are kept when IsTruncated is true. 

you're correct. the logic is flipped here ; the  IsTruncated check should be for falsy as the clearing of the buckets should only be done at the end of the listing.

nicolas2bert (approved):

naren-rajendran (approved):

PR #2365, opened 2023-02-10
BB-362 remove kafka connect connector config from logs

welansari:
kafka connect connector config contains the mongodb connection string which should not be logged

also fixing a bug with the oplog populator where if a connector update fails cause the connector can't be found, it is not considered an error

### DIFF
--- a/lib/wrappers/KafkaConnectWrapper.js
+++ b/lib/wrappers/KafkaConnectWrapper.js
@@ -141,7 +141,6 @@ class KafkaConnectWrapper {
         } catch (err) {
             this._logger.error('Error when creating connector', {
                 method: 'KafkaConnectManager.createConnector',
-                config,
                 error: err.description,
             });
             throw err;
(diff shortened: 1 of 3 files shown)


lib/wrappers/KafkaConnectWrapper.js
```diff
@@ -242,18 +241,16 @@ class KafkaConnectWrapper {
...
                 throw err;
             }
         }
-        // do nothing if connector not found
         this._logger.error('Can\'t update connector config, connector not found', {
             method: 'KafkaConnectManager.updateConnectorConfig',
             connector: connectorName,
         });
-        return undefined;
+        throw errors.InternalError.customizeDescription('Could not find connector to update');
```
francoisferrand, line 253: not the same commit : this is unrelated to the logs...

francoisferrand (commented):
.

KillianG (approved):

francoisferrand (approved):

PR #2366, opened 2023-02-14
Add azure archive location type to the config

KillianG:
The purpose of this PR is to add azure archive location type to the config
Issue: BB-359

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -21,6 +21,7 @@ const locationTypeMatch = {
     'location-scality-artesca-s3-v1': 'aws_s3',
     'location-ceph-radosgw-s3-v1': 'aws_s3',
     'location-dmf-v1': 'dmf',
+    'location-azure-archive-v1': 'azure_archive'
 };
 // NOTE: currently only s3 connector (s3c 7.4.3 and above) are supported
 const ingestionTypeMatch = {


welansari (approved):

lib/Config.js
```diff
@@ -21,6 +21,7 @@ const locationTypeMatch = {
     'location-scality-artesca-s3-v1': 'aws_s3',
     'location-ceph-radosgw-s3-v1': 'aws_s3',
     'location-dmf-v1': 'dmf',
+    'location-azure-archive-v1': 'azure_archive'
```
francoisferrand, line 24: seeing the extent of this change, i wonder what/how this value is used...

KillianG: It looks like it's only used [here](https://github.com/scality/backbeat/blob/375383c1b4651629349b550784cdcb3bdbd8369d/lib/Config.js#L200)

Which is used [here](https://github.com/scality/backbeat/blob/375383c1b4651629349b550784cdcb3bdbd8369d/lib/management/patchConfiguration.js#L38)

Which is used [here](https://github.com/scality/backbeat/blob/375383c1b4651629349b550784cdcb3bdbd8369d/lib/management/patchConfiguration.js#L70)

Appearently its used to patch configuration file which comes from [this ticket ](https://scality.atlassian.net/browse/BB-66)

Location type match comes from this PR : https://github.com/scality/backbeat/commit/a7035c1e5d04123f8fa125cf327dcaeff95a6758

And appearently there is a ticket BACKBEAT-3 which i don't have access to...

KillianG: It's only usefull for Zenko-1 because we are now using a config file for that so I guess I can remove this change

francoisferrand: that is not the end of the story I guess.

* We still have the code [to handle dynamic config changes, like with Orbit in Zenko 1], so the code must be updated still. It is a "feature" of the code base, even if not used currently by the product, and it **must** thus be maintained. (or we decide we won't need it, and remove it from the code base; but that is not the scope of this ticket).
* That said the question is not "where is the `locationTypeMatch` used", but where is the value (dmf, aws_s3...) defined/used. I.e. where/how is `config.bootstrapList[*].type` used

francoisferrand (approved):

PR #2370, opened 2023-03-03
BB-364: reduce image size 

alexanderchan-scality:


ghost (approved):

francoisferrand (dismissed):
This will not reduce the real size of the image : the size of "resulting" file system is indeed smaller, but there is still the layer with all caches.

The proper way to do this is to use multi-layer build, similar to how it is done in 8.x

images/nodesvc-base/Dockerfile
```diff
@@ -1,15 +1,20 @@
...
 
-# Remove when gitcache is sorted out
-RUN rm /root/.gitconfig
+RUN yarn cache clean \
+    && yarn install --production --frozen-lockfile --ignore-optional --ignore-engines --network-concurrency 1 \
+    && yarn cache clean
 
-RUN yarn install --production --frozen-lockfile --network-concurrency 1
+################################################################################
+FROM registry.scality.com/federation/nodesvc-base:7.10.5.0
```
jonathan-gramain: maybe a good time to bump the base image to `7.10.7` as well?

jonathan-gramain (approved):

images/nodesvc-base/Dockerfile
alexanderchan-scality: my mistake. had only bumped the builder base

PR #2374, opened 2023-03-20
Change handleMetrics function to async to be able to await the call to asPrometheus in ZenkoMetrics which now returns an async after upgrade of prom-client

KillianG:
Issue: BB-370


naren-rajendran (commented):
Looks good to me. Just needs linter corrections.

PR #2377, opened 2023-03-24
BB-352 Implement lifecycle rule aggregator

nicolas2bert:
**Backbeat lifecycle rule aggregator**

The Backbeat “lifecycle rule aggregator” determines the list of API calls needed to list all eligible objects based on the bucket lifecycle rules.

The lifecycle rule aggregator will group rules by "list type".

For each "list type" corresponds a listing call:
- **current**: {Expiration, Transition} → ListLifecycleObjects or ListLifecycleCurrentVersions depending on the bucket versioning state.
- **noncurrent**: {NoncurrentVersionExpiration, NoncurrentVersionTransition} → ListLifecycleNonCurrentVersions
- **orphan**: {Expiration, Expiration.ExpiredObjectDeleteMarker} → ListLifecycleDeleteMarkers

For a given "list type", the rules will be aggregated by the wildest prefix to not list the same key multiple times. _E.g. rules: [{ prefix: 's', days: 10 }, { prefix: 's/c', days: 5 }, { prefix: 's/c/a', days: 1 }]  should return [{ prefix: 's', days: 1 }] after aggregating by prefix._

### DIFF
--- a/extensions/lifecycle/util/RulesReducer.js
+++ b/extensions/lifecycle/util/RulesReducer.js
@@ -1,0 +1,282 @@
+const lowest = (acc, cur) => (acc < cur ? acc : cur);
+
+class RulesReducer {
+    /**
+     * Constructor of RulesReducer
+     *
+     * @constructor
+     * @param {string} versioningStatus - bucket's version status: 'Enabled', 'Disabled' or 'Suspended'
+     * @param {Date} currentDate - current date
+     * @param {Object} bucketLCRules - lifecycle rules
+     * @param {Object} options - rule lifecycle options
+     * @param {boolean} options.expireOneDayEarlier - moves lifecycle expiration deadlines 1 day earlier
+     * @param {boolean} options.transitionOneDayEarlier - moves lifecycle transition deadlines 1 day earlier
+    */
+    constructor(versioningStatus, currentDate, bucketLCRules, options) {
+        this._isVersioning = versioningStatus === 'Enabled' || versioningStatus === 'Suspended';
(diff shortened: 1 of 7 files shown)


extensions/lifecycle/util/RulesReducer.js
```diff
@@ -0,0 +1,282 @@
...
+                days = this._decrementExpirationDay(r.Expiration.Days);
+            } else if (r.Expiration.Date) {
+                if (r.Expiration.Date <= this._currentDate) {
+                    days = 0;
+                }
+            }
+        }
+
+        if (isTransitions) {
+            // NOTE: Transitions Days cannot be 0.
```
welansari: Not sure i understood this comment, seems it still works when having a rule set to `0 days`.
Also i believe pensieve accepts a value of 0 for current transitions
https://github.com/scality/pensieve-api/blob/b86dfec12c2d1a3e3d20d5a23dc46625f5ef7dee/gen/models/bucket_workflow_transition_v2_all_of1.go#L35-L37

nicolas2bert: yes, good catch, my comment is wrong but the code logic should be right. (I am checking for `if (r.Transitions[0].Days !== undefined) {` )
My comment should have been `NOTE: Transitions Days can be 0`.

welansari (approved):

alexanderchan-scality (approved):

nicolas2bert:
@bert-e approve

nicolas2bert:
ping

PR #2381, opened 2023-03-30
add s3 lifecycle expiration event

KillianG:
Add new aws route to internal cloudserver to delete an object from lifecycle expiration

This route is used in cloudserver to raise the correct event notification

- [x] when listing is fixed with version suspended, test that this is raising the right event and not multiple ones as it does actually

ISSUE: BB-375

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
+++ b/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
@@ -99,12 +99,6 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
 
     _executeDelete(entry, log, done) {
         const { accountId } = entry.getAttribute('target');
-        const s3Client = this.getS3Client(accountId);
-        if (!s3Client) {
-            log.error('failed to get S3 client', { accountId });
-            return done(errors.InternalError
-                .customizeDescription('Unable to obtain client'));
-        }
 
         const reqParams = {
             Bucket: entry.getAttribute('target.bucket'),
@@ -117,13 +111,27 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
         let reqMethod;
 
         const actionType = entry.getActionType();
+        let req = null;
         if (actionType === 'deleteObject') {
             reqMethod = 'deleteObject';
+            const backbeatClient = this.getBackbeatClient(accountId);
+            if (!backbeatClient) {
+                log.error('failed to get backbeat client', { accountId });
+                return done(errors.InternalError
+                    .customizeDescription('Unable to obtain client'));
+            }
+            req = backbeatClient.deleteObjectFromExpiration(reqParams);
         } else if (actionType === 'deleteMPU') {
             reqParams.UploadId = entry.getAttribute('details.UploadId');
             reqMethod = 'abortMultipartUpload';
+            const s3Client = this.getS3Client(accountId);
+            if (!s3Client) {
+                log.error('failed to get S3 client', { accountId });
+                return done(errors.InternalError
+                    .customizeDescription('Unable to obtain client'));
+            }
+            req = s3Client[reqMethod](reqParams);
         }
-        const req = s3Client[reqMethod](reqParams);
         attachReqUids(req, log);
         return req.send(err => {
             LifecycleMetrics.onS3Request(log, reqMethod, 'expiration', err);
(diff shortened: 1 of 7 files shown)


extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -99,6 +99,12 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
 
     _executeDelete(entry, log, done) {
         const { accountId } = entry.getAttribute('target');
+        const backbeatClient = this.getBackbeatClient(accountId);
+        if (!backbeatClient) {
+            log.error('failed to get backbeat client', { accountId });
+            return done(errors.InternalError
+                .customizeDescription('Unable to obtain client'));
+        }
```
welansari: We might want to avoid initializing both clients when only one will be used ?

tests/functional/lifecycle/LifecycleObjectProcessor.spec.js
```diff
@@ -19,15 +21,18 @@ const {
 
 werelogs.configure({ level: 'warn', dump: 'error' });
 
-describe('Lifecycle Object Processor', function lifecycleObjectProcessor() {
+const backbeatRoutes = ['deleteObject'];
+
+describe.only('Lifecycle Object Processor', function lifecycleObjectProcessor() {
```
KazToozs: only here

KazToozs (commented):
looks good currently otherwise, waiting for the other tests

tests/functional/lifecycle/LifecycleObjectProcessor.spec.js
KillianG: Its for testing purpose, need to be remove.. Should have set it in the PR desc mb

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -117,13 +111,27 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
...
         } else if (actionType === 'deleteMPU') {
             reqParams.UploadId = entry.getAttribute('details.UploadId');
             reqMethod = 'abortMultipartUpload';
+            const s3Client = this.getS3Client(accountId);
+            if (!s3Client) {
+                log.error('failed to get S3 client', { accountId });
+                return done(errors.InternalError
+                    .customizeDescription('Unable to obtain client'));
+            }
+            req = s3Client[reqMethod](reqParams);
```
jonathan-gramain, line 133: I believe that eventually the "abortMPU" triggered by Lifecycle would also need its own origin op? Maybe comment here if that's indeed the case.

tests/utils/BackbeatClientMock.js
```diff
@@ -0,0 +1,67 @@
...
+        this.stubMethod('deleteObjectFromExpiration', {});
+    }
+
+    makeRetryableError() {
+        const err = errors.ServiceUnavailable.customizeDescription('failing on purpose');
+        err.retryable = true;
+        return err;
+    }
+
+    stubMethod(methodName, successResult, stubError) {
```
jonathan-gramain, line 20: I'm fine with it, but I feel the kind of processing done here could be taken care of by using [Sinon stubs](https://sinonjs.org/releases/latest/stubs/)

jonathan-gramain, line 20: I'm fine with it, but I feel the kind of processing done here could be taken care of by using [Sinon stubs](https://sinonjs.org/releases/latest/stubs/)

jonathan-gramain (approved):

KazToozs (approved):
lgtm

tests/utils/BackbeatClientMock.js
KillianG: I did it the way S3Client is mocked as well to keep consistancy

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
KillianG: There is no lifecycle specific event for abortMPU according to AWS documentation https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html

The s3 call here already sets the originOP accordingly so we don't need to do anything specific for this case

KillianG: There is no lifecycle specific event for abortMPU according to AWS documentation https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html

The s3 call here already sets the originOP accordingly so we don't need to do anything specific for this case

PR #2387, opened 2023-04-13
BB-377: use LifecycleTaskV2 for eligible buckets

alexanderchan-scality:
* add helper method for creating LifecycleTasks
** enable V2 listing for bucket containing V2 indexes

### DIFF
--- a/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
+++ b/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
@@ -18,6 +18,7 @@ const LocationStatusStream = require('../../utils/LocationStatusStream');
 const {
     updateCircuitBreakerConfigForImplicitOutputQueue
 } = require('../../../lib/CircuitBreaker');
+const { lifecycleTaskVersions } = require('../../../lib/constants');
 
 const PROCESS_OBJECTS_ACTION = 'processObjects';
 
@@ -249,7 +250,7 @@ class LifecycleBucketProcessor {
             });
             return process.nextTick(() => cb(errors.InternalError));
         }
-        const { bucket, owner, accountId } = result.target;
+        const { bucket, owner, accountId, taskVersion } = result.target;
         if (!bucket || !owner || (!accountId && this._authConfig.type === authTypeAssumeRole)) {
             this._log.error('kafka bucket entry missing required fields', {
                 method: 'LifecycleBucketProcessor._processBucketEntry',
@@ -305,8 +306,10 @@ class LifecycleBucketProcessor {
             }
 
             let task;
-            // TODO: ZENKO-4501 Check if bucket is indexed before using new lifecycle improvement solution
-            if (this._lcConfig.bucketProcessor.forceLegacyListing) {
+
+            if (this._lcConfig.bucketProcessor.forceLegacyListing ||
+                taskVersion === lifecycleTaskVersions.v1 ||
+                !taskVersion) {
                 task = new LifecycleTask(this);
             } else {
                 task = new LifecycleTaskV2(this);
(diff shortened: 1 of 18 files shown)


extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -304,28 +305,92 @@ class LifecycleBucketProcessor {
...
+        let isV2 = false;
+        let task;
+
+        async.series([
+            next => backbeatMetadataProxy.getBucketIndexes(
+                params.bucket,
+                this._log,
+                (err, bucketIndexes) => {
+                    if (err) {
+                        this._log.warn('unable tparamso retrieve indexing info; using LifecycleTaskV1', {
```
rachedbenmustapha:
```suggestion
                        this._log.warn('unable to retrieve indexing info; using LifecycleTaskV1', {
```

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -304,28 +305,92 @@ class LifecycleBucketProcessor {
...
+                            bucket: params.Bucket,
+                            error: err,
+                        });
+
+                        task = new LifecycleTask(this);
+                        return next(null);
+                    }
+
+                    isV2 = indexesForFeature.lifecycle.v2.every(
```
rachedbenmustapha: Does this ensure that the index is fully "built" (making listing efficient) or is this true if indexes are still building (which can make listings + building very expensive).

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -304,28 +305,92 @@ class LifecycleBucketProcessor {
...
-                rules: config.Rules,
-                value: result,
-                s3target: s3,
-                backbeatMetadataProxy,
-            }, cb);
+        });
+
+    }
+
+    _createLifecycleTask(backbeatMetadataProxy, params, cb) {
```
rachedbenmustapha: I don't know how expensive it is to list indexes before each listing call. Is there a way to do this check only once before the first listing of the chain?

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -304,28 +305,92 @@ class LifecycleBucketProcessor {
...
+                        return next(null);
+                    }
+
+                    task = new LifecycleTaskV2(this);
+                    return next(null);
+
+                }),
+            next => {
+                if (!isV2) {
+                    return backbeatMetadataProxy.putBucketIndexes(
```
rachedbenmustapha: Is this serialized (through the kafka message key/partition or otherwise)? Not sure what happens if a bunch of these requests are sent at once.

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
alexanderchan-scality: the `getIndexes` method only return indexes that have been built

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
alexanderchan-scality: indexing should not be expensive to query since most of it would be in memory, but there could be some reads from disk; this could also done as part of the conductor instead, so I  will check if it possible to perform the query there as opposed to checking per task in the bucket processor

rachedbenmustapha (approved):

rahulreddy (approved):

PR #2390, opened 2023-04-18
Use resumeAfter to resume changestream on 4.0

francoisferrand:
On 4.2, we should use `startAfter` parameter to resume the change stream.
However, this is not available on 4.0, so we must use `resumeAfter`
instead.

Changes:
- Use `resumeAfter` or `startAfter` depending on mongo version
- Use Change Steam wrapper in notification config manager to avoid duplicate logic

Issue: BB-386

### DIFF
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -1,8 +1,9 @@
 const joi = require('joi');
+const semver = require('semver');
 const { errors } = require('arsenal');
 const { MongoClient } = require('mongodb');
 const constants = require('./constants');
-const { constructConnectionString } = require('../utils/MongoUtils');
+const { constructConnectionString, getMongoVersion } = require('../utils/MongoUtils');
 const ChangeStream = require('../../lib/wrappers/ChangeStream');
 const Allocator = require('./modules/Allocator');
 const ConnectorsManager = require('./modules/ConnectorsManager');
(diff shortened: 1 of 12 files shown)


extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -231,6 +233,7 @@ class OplogPopulator {
             pipeline: changeStreamPipeline,
             handler: this._handleChangeStreamChangeEvent.bind(this),
             throwOnError: false,
+            useStartAfter: !this._mongoVersion.startsWith("4.0.")
```
ghost: I am not sure this will work with the current BB CI, can we do this a bit more "semver"-like?

https://github.com/scality/backbeat/blob/development/8.6/.github/workflows/tests.yaml#L79

ghost: Also, I am not sure the version _alone_ is good enough. In case of upgrades, unless we explicitely set the featureVersion of mongodb, we might be in 4.2 but without `startAfter` support. Instead, you can do something like this:

```
db.adminCommand( 
    { 
        getParameter: 1, 
        featureCompatibilityVersion: 1 
    } 
)
```

it will return something like `{ "featureCompatibilityVersion" : { "version" : "4.2" }, "ok" : 1 }`, which will be better to test the compatibility

ghost (changes requested):
Some quick comments, plus some linter warnings that can be easily solved

ghost (approved):

anurag4DSB (approved):

tests/functional/notification/NotificationConfigManager.js
```diff
@@ -100,13 +104,11 @@ describe('NotificationConfigManager ::', () => {
                 _id: 'example-bucket-1',
             },
         };
-        manager._metastoreChangeStream.emit('change', changeStreamEvent);
+        manager._changeStreamWrapper._changeStream.emit('change', changeStreamEvent);
```
francoisferrand: it would have been clearer to keep the existing name (_metastoreChangeStream), even if the type changes : it better conveys the intention that we listen to metadata (vs just knowing this is a change stream, where we need "outside" knowledge to know what we use it)

```suggestion
        manager._metastoreChangeStream._changeStream.emit('change', changeStreamEvent);
```

francoisferrand:
LGTM (but I can't approve since I was the one creating the PR initially)

ghost (approved):

welansari (approved):

PR #2399, opened 2023-05-03
BB-392 : Run transition when expiration rules are also defined for the bucket

welansari:
Fixed transition getting skipped when an expiration rule was defined for the same bucket even when not applied.

Issue: BB-392

### DIFF
--- a/tests/functional/lifecycle/LifecycleTask.js
+++ b/tests/functional/lifecycle/LifecycleTask.js
@@ -508,9 +508,9 @@ describe('lifecycle task functional tests', function dF() {
 
     // Perform the bucket setup as defined in the parameters, and call the
     // lifecycle task bucket processor method.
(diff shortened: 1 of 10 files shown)


tests/functional/lifecycle/LifecycleTask.js
```diff
@@ -596,9 +592,15 @@ describe('lifecycle task functional tests', function dF() {
                         }])
                         .build(),
                 ],
-                expectedKeys: [],
             };
-            testTransition(params, done);
+
+            const expectedKeys = [];
+
+            setuptRunLifecycleTask(params, (err, data) => {
```
KillianG: nit: left a t after setup `setupRunLifecycleTask`

KillianG (approved):

KazToozs (approved):
lgtm

PR #2400, opened 2023-05-03
BB-397: integrate circuit breaker in ingestion producer

welansari:
- Integrate circuit breaker in ingestion producer
- Add control flow panel in ingestion producer and processor panels
- Add CircuitBreaker in Queue Populator config

**Ingestion Producer Dashboard**
ingestion_producer_dashboard

**Ingestion Processor Dashboard**
ingestion_processor_dashboard

Issue: BB-397

### DIFF
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -2,6 +2,7 @@ const async = require('async');
 const url = require('url');
 const Redis = require('ioredis');
 const schedule = require('node-schedule');
+const { CircuitBreaker, BreakerState } = require('breakbeat').CircuitBreaker;
 
 const Logger = require('werelogs').Logger;
 
@@ -12,6 +13,9 @@ const MetricsConsumer = require('../MetricsConsumer');
 const MetricsProducer = require('../MetricsProducer');
 const { metricsExtension } = require('../../extensions/ingestion/constants');
 const IngestionPopulatorMetrics = require('./IngestionPopulatorMetrics');
+const {
+    startCircuitBreakerMetricsExport,
+} = require('../CircuitBreaker');
 
 const {
     zookeeperNamespace,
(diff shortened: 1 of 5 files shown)


KillianG (approved):

tmacro (approved):

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -749,29 +766,55 @@ class IngestionPopulator {
...
+            return cb();
+        }
+        // skipping if location is paused
+        if (this._pausedLocations[logReader.getLocationConstraint()] !== undefined) {
+            this.log.debug('location paused, skipping', {
+                zenkoBucket: logReader.getTargetZenkoBucketName(),
+                location: logReader.getLocationConstraint(),
+            });
+            return cb();
+        }
```
francoisferrand: does not seem related : did we not already stop ingestion when location was paused?
or is it juste an "optimization", to skip the last few entries which we may still get immediately when ingestion is paused

lib/queuePopulator/IngestionReader.js
```diff
@@ -190,11 +190,7 @@ class IngestionReader extends LogReader {
             if (err) {
                 return done(err);
             }
-            if (hasMoreLog) {
-                // keep ingesting new log without waiting
-                return this.processLogEntries(params, done);
-            }
-            return done();
+            return done(null, hasMoreLog);
```
francoisferrand: could this have an effect on performance of ingestion?

someone decided to immediately process log entries here, and according to the comment this avoids introducing a delay : has this delay been removed? or is there/was there another issue?

lib/queuePopulator/IngestionPopulator.js
francoisferrand: actually, this one is associated with another ticket (BB-399) and is the latest one : should we move it a separate, stacked PR maybe?
(or should it actually not have been pushed in this PR, and you simply need to reset `improvement/BB-397` to `bugfix/BB-399^` ?)

welansari: yes i forgot to target the BB-399 PR, fixed that

lib/queuePopulator/IngestionReader.js
welansari: The behaviour should not change, there is still no delay in the initial scan, i only added pre checks before every log processing (or listing), where we now check the circuit breaker state as well as the location status.

This change belongs to BB-399 also btw

francoisferrand (approved):

PR #2402, opened 2023-05-05
BB-389: Port DMF's hack for 0-byte objects to LifeCycleTaskV2

welansari:
Issue: BB-389

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTaskV2.js
+++ b/extensions/lifecycle/tasks/LifecycleTaskV2.js
@@ -292,7 +292,8 @@ class LifecycleTaskV2 extends LifecycleTask {
             this._checkAndApplyExpirationRule(bucketData, obj, rules, log)) {
             return process.nextTick(cb);
         }
-        if (rules.Transition) {
+        if (rules.Transition &&
+            !this._bb383SkipDMFTransition(obj, rules.Transition.StorageClass, log)) {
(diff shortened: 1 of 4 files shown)


francoisferrand (approved):

extensions/lifecycle/tasks/LifecycleTaskV2.js
```diff
@@ -292,7 +292,8 @@ class LifecycleTaskV2 extends LifecycleTask {
             this._checkAndApplyExpirationRule(bucketData, obj, rules, log)) {
             return process.nextTick(cb);
         }
-        if (rules.Transition) {
+        if (rules.Transition &&
+            !this._bb383SkipDMFTransition(obj, rules.Transition.StorageClass, log)) {
```
KillianG, line 296: why this function's name starts with  `_bb383` ?

KillianG (approved):

alexanderchan-scality (approved):

extensions/lifecycle/tasks/LifecycleTaskV2.js
welansari: The changes introduced by the ticket BB-383 are temporary and will need to be reverted once DMF supports 0 byte objects.

The prefix `_bb383` is there to show that the function is temporary and that it shouldn't be used for something else

PR #2403, opened 2023-05-09
BB-399: make initial ingestion scan take into consideration the location status

welansari:
The initial bucket scan doesn't respect the cron rule as we continuously list the objects to converge faster, the listing is done in a single `_processLogEntries` call which makes it ignore the location status (paused, resumed).

Issue: BB-399

### DIFF
--- a/lib/queuePopulator/IngestionReader.js
+++ b/lib/queuePopulator/IngestionReader.js
@@ -190,11 +190,7 @@ class IngestionReader extends LogReader {
             if (err) {
                 return done(err);
             }
-            if (hasMoreLog) {
-                // keep ingesting new log without waiting
-                return this.processLogEntries(params, done);
-            }
-            return done();
+            return done(null, hasMoreLog);
         });
     }
 
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -749,29 +749,47 @@ class IngestionPopulator {
         return false;
     }
 
+    _processLogReaderEntries(logReader, params, cb) {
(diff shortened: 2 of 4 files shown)


lib/queuePopulator/IngestionReader.js
```diff
@@ -190,11 +190,7 @@ class IngestionReader extends LogReader {
             if (err) {
                 return done(err);
             }
-            if (hasMoreLog) {
-                // keep ingesting new log without waiting
-                return this.processLogEntries(params, done);
-            }
-            return done();
+            return done(null, hasMoreLog);
```
francoisferrand, line 193: > The initial bucket scan doesn't respect the cron rule as we continuously list the objects to converge faster, the listing is done in a single _processLogEntries call which makes it ignore the location status (paused, resumed).

I think this is by design, the initial scan should indeed not wait the next cron job... (but it must look for pause & circuit breakers indeed)

and since you moved the checks inside `_processLogEntries()`, it should be fine:
* here we will try to process immediately
* if paused/throttled, it will skip until the "next" cron call

francoisferrand: or is the "caller" of this code already using the `hasMoreLog` callback parameter to check if it see if it can wait for the next cron 'tick' or should proceed immediately ?

francoisferrand: ok, actually you moved this "keep ingestion without waiting" from IngestionReader to IngestionPopulator

francoisferrand (approved):

KillianG (approved):

PR #2407, opened 2023-05-25
Discard restore expiry request if date not reached

francoisferrand:
There is a sort of race condition if the `s3:restore` api is used to
update the expiration date of the restore object at the last moment,
just while it is in-flight: e.g. after it is queued for expiration, but
before this queue message gets processed by
backbeat.

To avoid this, we do an extra check and simply discard the message in
that case, relying on cold storage to queue the object for
expiration again once it will have reached the "new" expiration date.

Issue: BB-402

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
+++ b/extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
@@ -152,8 +152,15 @@ class LifecycleUpdateExpirationTask extends BackbeatTask {
                     Object.keys(locations).find(name => locations[name].isCold);
 
                 const archive = objMD.getArchive();
-                // Reset archive flags to no longer
-                // show it as restored
+
+                // Confirm the object has indeed expired: it can happen that the
+                // expiration date is updated while the expiry was "in-flight" (e.g.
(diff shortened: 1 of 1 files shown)


extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
```diff
@@ -152,8 +152,15 @@ class LifecycleUpdateExpirationTask extends BackbeatTask {
                     Object.keys(locations).find(name => locations[name].isCold);
 
                 const archive = objMD.getArchive();
-                // Reset archive flags to no longer
-                // show it as restored
+
+                // Confirm the object has indeed expired: it can happen that the
+                // expiration date is updated while the expiry was "in-flight" (e.g.
+                // queued for expiry but not yet expired)
+                if (new Date(archive.getRestoreWillExpireAt()) > new Date(Date.now())) {
```
welansari:
```suggestion
                if (new Date(archive.getRestoreWillExpireAt()) > new Date()) {
```
`new Date()` already returns the current date no need to add the `Date.now()`

nicolas2bert (approved):

alexanderchan-scality (approved):

PR #2408, opened 2023-05-25
Set 'null' as version id when replicating non versioned objects

welansari:
Non versioned object are also considered a null version when in a versioned bucket, we should specify 'null' as query version id for Cloudserver to properly replicate these objects

Issue: BB-403

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -200,7 +200,7 @@ class MultipleBackendTask extends ReplicateObject {
             }
         }
 
-        this.retry({
+        return this.retry({
             actionDesc: 'stream part data',
             logFields: { entry: sourceEntry.getLogInfo(), range },
             actionFunc: done => this._getRangeAndPutMPUPartOnce(sourceEntry,
@@ -949,7 +949,7 @@ class MultipleBackendTask extends ReplicateObject {
             }
         }
 
-        this.retry({
+        return this.retry({
             actionDesc: 'send object tagging XML data',
             entry: sourceEntry,
             logFields: { entry: sourceEntry.getLogInfo() },
@@ -1004,7 +1004,7 @@ class MultipleBackendTask extends ReplicateObject {
             }
         }
 
-        this.retry({
+        return this.retry({
             actionDesc: 'delete object tagging',
             logFields: { entry: sourceEntry.getLogInfo() },
             actionFunc: done => this._deleteObjectTaggingOnce(sourceEntry,
@@ -1057,7 +1057,7 @@ class MultipleBackendTask extends ReplicateObject {
             }
         }
 
-        this.retry({
+        return this.retry({
             actionDesc: 'put delete marker',
             logFields: { entry: sourceEntry.getLogInfo() },
             actionFunc: done => this._putDeleteMarkerOnce(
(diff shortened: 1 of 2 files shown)


extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -126,7 +126,7 @@ class MultipleBackendTask extends ReplicateObject {
         const params = {
             bucket: sourceEntry.getBucket(),
             objectKey: sourceEntry.getObjectKey(),
-            versionId: sourceEntry.getEncodedVersionId(),
+            versionId: sourceEntry.getEncodedVersionId() || 'null',
```
francoisferrand, line 129: should we instead modify `getEncodedVersionId()` to return 'null' ?
(i.e. always return an actual version id...)

francoisferrand: ....but that function is in arsenal :-(

welansari: no cause the function can also be used for non versioned buckets, in which case it should return undefined

welansari: in replication we are sure the bucket is versioned that's why we do this

extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -273,15 +273,13 @@ class UpdateReplicationStatus extends BackbeatTask {
...
-        const versionId = updatedSourceEntry.getIsNull() ?
-            undefined : updatedSourceEntry.getEncodedVersionId();
         return client.putMetadata({
             bucket: updatedSourceEntry.getBucket(),
             objectKey: updatedSourceEntry.getObjectKey(),
-            versionId,
+            // non versioned object are also considered a null version
+            // in a versioned bucket, we should set the version id to 'null'
+            // for it to be handled correctly in cloudserver
+            versionId: updatedSourceEntry.getEncodedVersionId() || 'null',
```
francoisferrand, line 282: should we do the same for `getMetadata()` call, line 73 ?

extensions/replication/tasks/MultipleBackendTask.js
francoisferrand: ok, let's leave it for now.

but it seems it should be the other way around : getEncodedVersionId() should alwats return the version id (which may be "null") and it would be up to the caller to check if the bucket is versionned → call getEncodedVersionId() or non-versionned → completely ignore the version id.

What do you think? Should we create a TechDebt (cleanup) ticket ?

francoisferrand (approved):

extensions/replication/tasks/MultipleBackendTask.js
nicolas2bert: For non-versioned buckets, putMetadata("null") should update the object master, so we should be good and
 return "null" if bucket is non-versioned.

extensions/replication/tasks/UpdateReplicationStatus.js
welansari: we should yes, i missed that one

jonathan-gramain (approved):

nicolas2bert:
Just to clarify, there seems to be 2 different cases:

**Lifecycle Bucket Processor**: In this case, the entry MD is sourced from either the listObject or listObjectVersions, and thus the "null" version is already appropriately set.

**All the other processes**: Here, we receive the pure object MD, and if the version id is undefined, a "null" version id is used for both getMetadata and putMetadata operations. CloudServer.putMetadata() will then determine how to handle the null version without a version id.

It would be interesting to consolidate/abstract the logic, so that when we utilize a versionId, we don't need to consider whether it has already been "processed" or not. 

This task can be addressed in a subsequent ticket.

PR #2412, opened 2023-06-02
When receive a retryRestore action in object task topic, update object MD with origin op to trigger a restore from oplog reader

KillianG:
After the restoration process is completed, it will be necessary to remove the "x-amz-meta-scal-s3-restore-attempt" in cloudserver. ( cf. CLDSRV-403 )

Issue: BB-408

### DIFF
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
@@ -108,17 +108,16 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
                 return new LifecycleResetTransitionInProgressTask(this);
             case 'requeueRestore':
                 return new LifecycleRetriggerRestoreTask(this);
+            case 'copyLocation':
+                if (actionEntry.getContextAttribute('ruleType') === 'transition') {
+                    return new LifecycleUpdateTransitionTask(this);
+                }
+                // fall through
             default:
+                this._log.warn(`skipped unsupported  action ${actionType}`,
+                            actionEntry.getLogInfo());
+                return null;
         }
-
-        if (actionType !== 'copyLocation' ||
-            actionEntry.getContextAttribute('ruleType') !== 'transition') {
-            this._log.warn(`skipped unsupported  action ${actionType}`,
-                             actionEntry.getLogInfo());
-            return null;
-        }
-
-        return new LifecycleUpdateTransitionTask(this);
     }
 
     processColdStorageStatusEntry(kafkaEntry, done) {
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
@@ -103,10 +103,12 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
     getTask(actionEntry) {
         const actionType = actionEntry.getActionType();
 
-        if (actionType === 'requeueTransition') {
-            return new LifecycleResetTransitionInProgressTask(this);
-        } else if (actionType === 'requeueRestore') {
-            return new LifecycleRetriggerRestoreTask(this);
+        switch (actionType) {
+            case 'requeueTransition':
+                return new LifecycleResetTransitionInProgressTask(this);
+            case 'requeueRestore':
+                return new LifecycleRetriggerRestoreTask(this);
+            default:
         }
 
         if (actionType !== 'copyLocation' ||
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
@@ -10,6 +10,7 @@ const LifecycleColdStatusArchiveTask =
 const { LifecycleResetTransitionInProgressTask } =
       require('../tasks/LifecycleResetTransitionInProgressTask');
 const { updateCircuitBreakerConfigForImplicitOutputQueue } = require('../../../lib/CircuitBreaker');
+const { LifecycleRetriggerRestoreTask } = require('../tasks/LifecycleRetriggerRestoreTask');
 
 class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
 
@@ -104,6 +105,8 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
 
         if (actionType === 'requeueTransition') {
             return new LifecycleResetTransitionInProgressTask(this);
+        } else if (actionType === 'requeueRestore') {
+            return new LifecycleRetriggerRestoreTask(this);
         }
 
         if (actionType !== 'copyLocation' ||
--- a/extensions/lifecycle/tasks/LifecycleRequeueTask.js
+++ b/extensions/lifecycle/tasks/LifecycleRequeueTask.js
@@ -1,0 +1,102 @@
+'use strict'; // eslint-disable-line
+
+const async = require('async');
+
+const { errors } = require('arsenal');
+const ObjectMD = require('arsenal').models.ObjectMD;
+
+const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
+const { LifecycleMetrics } = require('../LifecycleMetrics');
+
+class LifecycleRequeueTask extends BackbeatTask {
+
+    /**
+     * Execute the action specified in action entry
+     *
+     * @param {ActionQueueEntry} entry - action entry to execute
+     * @param {Function} done - callback funtion
+     * @return {undefined}
+     */
+    processActionEntry(entry, done) {
+        const log = this.logger.newRequestLogger(entry.actionId);
+        const { byAccount } = entry.getAttribute('target') || {};
+
+        async.reduce(
+            Object.keys(byAccount),
+            0,
+            (objsPerBatch, accountId, done) => {
+                const buckets = byAccount[accountId] || {};
+                const bucketNames = Object.keys(buckets);
+                this.handleBatch(
+                    accountId,
+                    buckets,
+                    bucketNames,
+                    log,
+                    (err, res) => {
+                        if (err) {
+                            return done(err);
+                        }
+
+                        const sum = res.reduce((acc, v) => acc + v, 0);
+                        return done(null, sum + objsPerBatch);
+                    }
+                );
+            },
+            (err, objectCount) => {
+                if (err) {
+                    log.error('could not process message', { error: err, objectCount });
+                } else {
+                    log.info('processed requeue message', { objectCount });
+                }
+
+                done(err);
+            }
+        );
+    }
+
+    handleBatch(accountId, buckets, bucketNames, log, cb) {
+        async.map(
+            bucketNames,
+            (bucketName, next) =>
+                async.reduce(
+                    buckets[bucketName] || [],
+                    0,
+                    (objsPerBucket, { objectKey, objectVersion, eTag, ...rest }, nextObject) =>
+                        this.requeueObjectVersion(
+                            accountId,
+                            bucketName,
+                            objectKey,
+                            objectVersion,
+                            eTag,
+                            rest.try,
+                            log,
+                            (err, res) => {
+                                if (err) {
+                                    return nextObject(err);
+                                }
+
+                                return nextObject(null, res + objsPerBucket);
+                            }
+                        ),
+                    (err, res) => {
+                        if (err) {
+                            return next(err);
+                        }
+
+                        return next(null, res);
+                    }
+                ),
+            (err, res) => {
+                if (err) {
+                    return cb(err);
+                }
+
+                return cb(null, res);
+            }
+        );
+    }
+}
+
+module.exports = {
+    LifecycleRequeueTask
+};
--- a/extensions/lifecycle/tasks/LifecycleRequeueTask.js
+++ b/extensions/lifecycle/tasks/LifecycleRequeueTask.js
@@ -11,6 +11,72 @@ const { LifecycleMetrics } = require('../LifecycleMetrics');
 class LifecycleRequeueTask extends BackbeatTask {
 
     /**
+     * Process a lifecycle object entry
+     *
+     * @constructor
+     * @param {LifecycleObjectProcessor} proc - object processor instance
+     */
+    constructor(proc) {
+        const procState = proc.getStateVars();
+        super();
+        Object.assign(this, procState);
+    }
+
+    requeueObjectVersion(accountId, bucketName, objectKey, objectVersion, etag, try_, bucketLogger, cb) {
+        const client = this.getBackbeatMetadataProxy(accountId);
+        if (!client) {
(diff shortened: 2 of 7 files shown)


extensions/lifecycle/tasks/LifecycleRequeueRestoreArchiveTask.js
```diff
@@ -0,0 +1,134 @@
+const async = require('async');
+
+const { errors } = require('arsenal');
+const ObjectMD = require('arsenal').models.ObjectMD;
+const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
+
+class LifecycleRequeueRestoreArchiveTask extends BackbeatTask {
```
welansari: Could we reuse `LifecycleResetTransitionInProgressTask` by renaming it and making it handle both transitions and restores ? as the logic is almost entirely the same and we will also need to duplicate the batch reading logic if we keep them separate.

KillianG: I implemented a similar solution by utilizing inheritance. As a result, we now have a new class called `LifecycleRetriggerRestoreTask` that shares the same base class as the original one. Both classes inherit from `LifecycleRequeueTask`, which handles the common code between retrying archive and restore (primarily batching handling, which is done in `LifecycleRequeueTask`).

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            if (this.shouldSkipObject(md, log)) {
+                return cb(null, 0);
+            }
+
+            md.setUserMetadata({
+                'x-amz-meta-scal-s3-restore-attempt': try_,
+            });
+            md.setOriginOp('s3:ObjectRestore:Retry')
```
francoisferrand: all the getMetadata/metrics/putMetadata/... logic is common : should be moved to the base class, just calling a new function (`updateObjectMD(md, log) → bool`) whose implementations differ between both subclasses

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -0,0 +1,95 @@
...
+     * Process a lifecycle object entry
+     *
+     * @constructor
+     * @param {LifecycleObjectProcessor} proc - object processor instance
+     */
+     constructor(proc) {
+        const procState = proc.getStateVars();
+        super();
+        Object.assign(this, procState);
+    }
```
francoisferrand: can't this be moved to LifecycleRequeueTask as well ?
(`the getStateVars()` and `assign` at least)

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -104,6 +105,8 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
 
         if (actionType === 'requeueTransition') {
             return new LifecycleResetTransitionInProgressTask(this);
+        } else if (actionType === 'requeueRestore') {
```
francoisferrand: prefer a switch case?

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -0,0 +1,40 @@
...
+        }
+        md.setUserMetadata({
+            'x-amz-meta-scal-s3-restore-attempt': try_,
+        });
+        md.setOriginOp('s3:ObjectRestore:Retry');
+        return true;
+    }
+
+    shouldSkipObject(md, log) {
+        if (md.getArchive()?.archiveInfo?.archiveId === undefined) {
```
francoisferrand: should check other conditions:
* restore must be in progress
* must not be already restored

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -0,0 +1,40 @@
...
+class LifecycleRetriggerRestoreTask extends LifecycleRequeueTask {
+    /**
+     * Process a lifecycle object entry
+     *
+     * @constructor
+     * @param {LifecycleObjectProcessor} proc - object processor instance
+     */
+     constructor(proc) {
+        super(proc);
+        this.processName = 'retriggerRestore';
```
francoisferrand: should be just 'restore'

(btw this is not a "process" as in Linux process, but the "workflow" for use in lifecycle metrics: other values are 'transition', 'archive', 'expiration'...)

francoisferrand: or maybe we should keep 'transition' as well in this case: @nicolas2bert what do you think?

KillianG: > (btw this is not a "process" as in Linux process, but the "workflow" for use in lifecycle metrics: other values are 'transition', 'archive', 'expiration'...)

I took this name from the function where we use this variable https://github.com/scality/backbeat/blob/b33d280015eecdf697ba4f292474236d4b44399b/extensions/lifecycle/LifecycleMetrics.js#L91
Didn't know about the other values though :)

nicolas2bert: Lifecycle Retrigger Restore Task is a component of the "restore workflow".

extensions/lifecycle/tasks/LifecycleResetTransitionInProgressTask.js
```diff
@@ -1,169 +1,28 @@
...
-            (err, res) => {
-                if (err) {
-                    return cb(err);
-                }
-
-                return cb(null, res);
-            }
-        );
+        super(proc);
+        this.processName = 'transition';
```
francoisferrand: `processName` should be an additional parameter of `LifecycleRequeueTask` constructor : since it is required by that class

extensions/lifecycle/tasks/LifecycleResetTransitionInProgressTask.js
```diff
@@ -1,169 +1,28 @@
...
-                err => {
-                    LifecycleMetrics.onS3Request(log, 'putMetadata', 'transition', err);
-                    if (err) {
-                        return cb(err);
-                    }
-
-                    return cb(null, 1);
-                }
-            );
         });
```
francoisferrand, line 23: I think there was an issue in the original bug : the "originOp" is not set when writing the metadata, which may cause extra notification... To avoid any issue, we should probably set it now (`setOriginOp("s3:LifecycleTransition:Retry")`) to avoid issues...

(it can be done in a separate dedicated ticket)

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    shouldSkipObject(md, log) {
+        const isObjectAlreadyRestored = md.getArchive()
+            && md.getArchive().restoreCompletedAt
+            && new Date(md.getArchive().restoreWillExpireAt) >= new Date();
+
+        if (!md.getArchive()?.restoreRequestedAt ||
+            !md.getArchive()?.restoreRequestedDays || isObjectAlreadyRestored) {
+                log.error('object is already restored, skipping');
```
francoisferrand, line 36: message is not correct:

* `isObjectAlreadyRestored` → object is already restored indeed
* `!restoreRequestedAt || !restoreRequestedDays` → restore was not requested

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -0,0 +1,50 @@
...
+            'x-amz-meta-scal-s3-restore-attempt': try_,
+        });
+        md.setOriginOp('s3:ObjectRestore:Retry');
+        return true;
+    }
+
+    shouldSkipObject(md, log) {
+        const isObjectAlreadyRestored = md.getArchive()
+            && md.getArchive().restoreCompletedAt
+            && new Date(md.getArchive().restoreWillExpireAt) >= new Date();
```
francoisferrand: if object is restored (getArchive && md.getArchive().restoreCompletedAt), but restoreWillExpireAt < date : we should probably not retrigger the restore...

```suggestion
        const isObjectAlreadyRestored = md.getArchive()
            && md.getArchive().restoreCompletedAt;
```

tests/unit/lifecycle/LifecycleRetriggerRestoreTask.spec.js
```diff
@@ -0,0 +1,90 @@
...
+            null,
+            null,
+            backbeatMetadataProxyClient,
+            null,
+            new werelogs.Logger('test:LifecycleRetriggerRestoreTask'));
+
+        task = new LifecycleRetriggerRestoreTask(objectProcessor);
+    });
+
+    it('should skip object not archived', done => {
```
francoisferrand, line 67: missing test cases for the various conditions to skip/!skip : already restored, restored but expired, restore not requested, ...

extensions/lifecycle/tasks/LifecycleResetTransitionInProgressTask.js
KillianG: Not sure I understand your point, maybe creating a ticket for that would be best ?

KillianG: Created a ticket for that: https://scality.atlassian.net/browse/BB-421

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -102,8 +103,12 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
         const actionType = actionEntry.getActionType();
 
-        if (actionType === 'requeueTransition') {
-            return new LifecycleResetTransitionInProgressTask(this);
+        switch (actionType) {
+            case 'requeueTransition':
+                return new LifecycleResetTransitionInProgressTask(this);
+            case 'requeueRestore':
+                return new LifecycleRetriggerRestoreTask(this);
+            default:
```
francoisferrand, line 116: what is the purpose of keeping the empty 'default' ?

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -0,0 +1,57 @@
...
+        md.setUserMetadata({
+            'x-amz-meta-scal-s3-restore-attempt': try_,
+        });
+        md.setOriginOp('s3:ObjectRestore:Retry');
+        return true;
+    }
+
+    shouldSkipObject(md, log) {
+        const isObjectAlreadyRestored = md.getArchive()
+            && md.getArchive().restoreCompletedAt;
```
francoisferrand:
```suggestion
        const isObjectAlreadyRestored = md.getArchive()?.restoreCompletedAt;
```

francoisferrand: actually, should be moved to line 36 :  no reason to declare it on top

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -0,0 +1,57 @@
...
+                log.error('object is already restored, skipping');
+                return true;
+        }
+
+        if (new Date(md.getArchive().restoreWillExpireAt) < new Date()) {
+            log.error('object restore has expired, skipping');
+            return true;
+        }
+
+        if (!md.getArchive()?.restoreRequestedAt || !md.getArchive()?.restoreRequestedDays) {
```
francoisferrand: the `?` after getArchive() is irrelevant here : you did not check on line 41 already, and actually indeed the first check on line 31 already ensures it is set in followup code.

it would also simply the code to use a variable for `md.getArchive()` in this function

francoisferrand (approved):

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
KillianG: It is to be "exhaustive".. in fact it does nothing but I like it to be exhaustive in a switch case, anyway I don't mind removing it if you feel this is really useless tho

francoisferrand: Being "exhaustive" would be to cover each individual case explicitly ; and would make sense only with something done or at the very an explanation why we don't do anything.

Adding an empty `default:` is just telling the compiler/linter that you handle all cases, without actually doing anything : i.e. just blindsiding the tools (and reviewers)

extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask.js
```diff
@@ -25,25 +25,24 @@ class LifecycleRetriggerRestoreTask extends LifecycleRequeueTask {
...
-            && md.getArchive().restoreCompletedAt;
-
-        if (!md.getArchive()?.archiveInfo?.archiveId) {
+        const archive = md.getArchive();
+        if (!archive?.archiveInfo?.archiveId) {
             log.error('object is not archived, skipping');
             return true;
         }
 
+        const isObjectAlreadyRestored = archive?.restoreCompletedAt;
```
francoisferrand: `?` not needed here either...

tests/unit/lifecycle/LifecycleRetriggerRestoreTask.spec.js
```diff
@@ -0,0 +1,132 @@
...
+const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
+const { LifecycleRetriggerRestoreTask } = require(
+    '../../../extensions/lifecycle/tasks/LifecycleRetriggerRestoreTask');
+
+const {
+    BackbeatMetadataProxyMock,
+    ProcessorMock,
+} = require('../mocks');
+
+describe.only('LifecycleResetTransitionInProgressTask', () => {
```
welansari:
```suggestion
describe('LifecycleResetTransitionInProgressTask', () => {
```

extensions/lifecycle/tasks/LifecycleRequeueTask.js
```diff
@@ -0,0 +1,170 @@
...
+                                return nextObject(null, res + objsPerBucket);
+                            }
+                        ),
+                    (err, res) => {
+                        if (err) {
+                            return next(err);
+                        }
+
+                        return next(null, res);
+                    }
```
welansari:
```suggestion
                    (err, res) => next(err, res)
```
i feel like this can be simplified

extensions/lifecycle/tasks/LifecycleRequeueTask.js
```diff
@@ -0,0 +1,170 @@
...
+                ),
+            (err, res) => {
+                if (err) {
+                    return cb(err);
+                }
+
+                return cb(null, res);
+            }
```
welansari:
```suggestion
            (err, res) => cb(err, res)
```
this too

welansari (approved):

PR #2415, opened 2023-06-06
Configure heartbeat in the kafka connect connectors

welansari:
Hearbeat prevents having an outdated resume token in the connectors by constantly updating the offset to the last object in the oplog

Outdated offsets can happen in two cases:
- Not enough updates in MongoDB, so the offset stays the same until the oplog event holding the offset gets expired from the oplog.
- Too many updates in the DB, which causes the event holding the offset to be overwritten as the oplog is a capped collection that overwrites the earliest events when it fills up.

Issue: BB-407

### DIFF
--- a/extensions/oplogPopulator/modules/ConnectorsManager.js
+++ b/extensions/oplogPopulator/modules/ConnectorsManager.js
@@ -17,6 +17,7 @@ const paramsJoi = joi.object({
     oplogTopic: joi.string().required(),
     cronRule: joi.string().required(),
     prefix: joi.string(),
+    heartbeatIntervalMs: joi.number().required(),
     kafkaConnectHost: joi.string().required(),
     kafkaConnectPort: joi.number().required(),
     metricsHandler: joi.object()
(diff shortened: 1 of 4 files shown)


KillianG (approved):

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -82,7 +82,11 @@ class ConnectorsManager {
             'connection.uri': this._mongoUrl,
             'topic.namespace.map': JSON.stringify({
                 '*': this._oplogTopic,
-            })
+            }),
+            // hearbeat prevents having an outdated resume token in the connectors
+            // by constantly updating the offset to the last object in the oplog
+            'heartbeat.interval.ms': process.env.OPLOG_HEARTBEAT_INTERVAL_MS ?
```
francoisferrand: should this not be configured via backbeat config file instead?

francoisferrand (approved):
LGTM

should we add the param to the `https://github.com/scality/backbeat/blob/HEAD/conf/config.json`, and possibly add tests?

PR #2422, opened 2023-06-14
bf: BB-419 increase notification default concurrency

jonathan-gramain:
**Replaces https://github.com/scality/backbeat/pull/2421**

- increase default notification BackbeatConsumer concurrency

  Notification QueueProcessor was limited by a concurrency of 10, which meant it could only wait for 10 delivery reports from the producer before consuming more entries.

  Increased the default concurrency to 1000 to improve throughput and add a test to show that it behaves well with this number of pending reports (the processing does not involve anything else asynchronous than waiting for delivery reports), as well as a few other extra tests.

  Also fix and improve logging related to processing errors and error reporting from delivery reports.

- [cleanup] BackbeatProducer.pollIntervalMs default as constant

  Use a code constant for `pollIntervalMs`, and add a JSDoc about it (and `messageMaxBytes`)

- [impr] configurable notif Kafka poll interval

  Make the external Kafka producer poll interval configurable: while the default is 2000ms, it could be helpful to reduce its value, e.g. to reduce latency of delivery in case the external Kafka cluster cannot receive all notifications in time.

  It's not clear at this point what would be a better value due to lack of testing to the limits, and 2000ms doesn't look unreasonable as a trade-off between latency and polling overhead, so leaving the default value as it is. It is however important to have a concurrency value high enough to not limit the possible throughput per notification processor, which is limited by "concurrency / poll_interval_seconds" messages per second. The concurrency has been raised to 1000 by default which therefore allows 500 messages per second to go through per processor with a 2000ms poll interval.

### DIFF
--- a/tests/unit/notification/QueueProcessor.spec.js
+++ b/tests/unit/notification/QueueProcessor.spec.js
@@ -1,0 +1,166 @@
+const assert = require('assert');
+const async = require('async');
+const sinon = require('sinon');
+
+const QueueProcessor = require(
+    '../../../extensions/notification/queueProcessor/QueueProcessor');
+
+describe('notification QueueProcessor', () => {
+    let qp;
+
+    before(done => {
+        qp = new QueueProcessor(null, null, null, {
+            host: 'external-kafka-host',
+        }, 'destId');
+        qp.bnConfigManager = {
+            getConfig: () => ({
+                notificationConfiguration: {
+                    queueConfig: [
+                        {
+                            queueArn: 'arn:scality:bucketnotif:::destId',
+                            events: [
+                                's3:ObjectCreated:*',
+                            ],
+                        },
+                    ],
+                },
+            }),
+            setConfig: () => {},
+        };
+        qp.bnConfigManager.setConfig('mybucket', {
+            host: 'foo',
+        });
+        qp._setupDestination('kafka', done);
+    });
+
+    it('should send notification to external destination and invoke callback on delivery report', done => {
+        const sendCb = sinon.stub().callsFake(cb => cb());
+        const send = sinon.stub().callsFake((messages, cb) => setTimeout(() => sendCb(cb), 100));
+        qp._destination._notificationProducer = {
+            send,
+        };
+        qp.processKafkaEntry({
+            value: JSON.stringify({
+                bucket: 'mybucket',
+                key: 'key',
+                eventType: 's3:ObjectCreated:Put',
+                value: '{}',
+            }),
+        }, err => {
+            assert.ifError(err);
+
+            assert(send.calledOnce);
+            assert(sendCb.calledOnce);
+            assert(Array.isArray(send.args[0][0]));
+            assert(send.args[0][0][0].key === 'mybucket/key');
+            assert(typeof send.args[0][0][0].message === 'object');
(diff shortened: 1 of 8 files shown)


tests/unit/notification/QueueProcessor.spec.js
```diff
@@ -0,0 +1,151 @@
...
+            value: JSON.stringify({
+                bucket: 'mybucket',
+                key: 'key',
+                // no "eventType"
+                value: '{}',
+            }),
+        }, err => {
+            assert.ifError(err);
+            // should not have been sent
+            assert(!send.calledOnce);
```
watsaqat: could this be true if `send.calledTwice == true`? Is `.notCalled` what is wanted instead?

tests/unit/notification/QueueProcessor.spec.js
```diff
@@ -0,0 +1,151 @@
...
+            done();
+        });
+    });
+});
```
watsaqat, line 166: is it possible that JSON parsing might fail in qp.processKafkaEntry?

tests/unit/notification/QueueProcessor.spec.js
```diff
@@ -0,0 +1,151 @@
...
+            send,
+        };
+        qp.processKafkaEntry({
+            value: JSON.stringify({
+                bucket: 'mybucket',
+                key: 'key',
+                // no "eventType"
+                value: '{}',
+            }),
+        }, err => {
```
watsaqat, line 144: if there's no eventType, is there an error/log message or it just silently doesn't send?

watsaqat (approved):

tests/unit/notification/QueueProcessor.spec.js
jonathan-gramain: yes, good point

tests/unit/notification/QueueProcessor.spec.js
jonathan-gramain: I think this is where the check happens: https://github.com/scality/backbeat/blob/development/8.6/extensions/notification/utils/config.js#L85

I believe there is no particular log.

For the context, I closed a previous iteration of this PR (https://github.com/scality/backbeat/pull/2421) that was not waiting for the callback of `send` but we decided to rework it differently, the point of this test was then mostly to trigger a case where a message is not sent for regression testing, not particularly to check all details of the particular case where `eventType` is not passed. I kept it because I think it has some value to test this case but it's not anymore directly related to the change.

tests/unit/notification/QueueProcessor.spec.js
jonathan-gramain: It actually throws an exception if the input is not valid JSON.

It's not directly related to the scope of this PR, but it can be a good idea to return a proper error instead and add a test case for it indeed.

jonathan-gramain: I fixed JSON parsing exception and added a test

tests/unit/notification/QueueProcessor.spec.js
jonathan-gramain: fixed, using `send.notCalled`

naren-rajendran (approved):

PR #2423, opened 2023-06-15
BB-420: fix conductor race condition

alexanderchan-scality:
issue: race condition occurs when queued bucket completion checks
resolving to true while async operations are still in progress, resulting
in the improper reset of state variables and preventing the creation of
indexes for queued buckets; this issue is caused by the combination of
the mBucketsQueued state variable being incremented before the
completion of the async operations and the completion interval check

fix: moved the incrementation of nBucketsQueued to after the completion
of the async operations

### DIFF
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -332,8 +332,6 @@ class LifecycleConductor {
                 return done();
             }
 
-            nBucketsQueued += tasks.length;
-
             const unknownCanonicalIds = new Set(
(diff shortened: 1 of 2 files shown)


rachedbenmustapha (approved):

naren-rajendran (approved):

jonathan-gramain:
@alexanderchan-scality wouldn't this fix also apply to the 7.x branches?

alexanderchan-scality:
@jonathan-gramain I believe that would also be the case; will open a separate ticket handling the 7.x branches

PR #2427, opened 2023-06-27
BB-425 Short version IDs break ordering - use lastModified

nicolas2bert:
**Summary:**

- Short version IDs no longer preserve ordering.
- We will use the "last-modified" date to sort version and delete markers.

**Description:**

- Prior to the introduction of short version IDs, a lower version ID indicated a more recent version. However, with short version IDs, the ordering is no longer preserved.
- To address this issue, we will use the "last-modified" date to sort version and delete markers from the object versions listing. This will ensure that the versions are always sorted in chronological order, regardless of the length of the version ID.
- We considered decoding the version ID as a solution, but this would be quite expensive in terms of CPU time and memory usage.

**Encoding algorithms history:**

The "short version id" encoding uses lossy base62 encoding algorithms (Base62Integer, Base62String) that do not preserve the order of the timestamps in the version ID. Previously, encoding was done using the hex encoding algorithm, which is a lossless encoding algorithm because it does not lose any information during the encoding process.

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -483,35 +483,13 @@ class LifecycleTask extends BackbeatTask {
                     sortedList.push(deleteMarkers[dmIdx++]);
                 }
             } else {
-                //    NOTE: VersionId may be null
-                const nullVersion = (versions[vIdx].VersionId === 'null'
-                    || deleteMarkers[dmIdx].VersionId === 'null');
-                const isVersionVidNewer = (versions[vIdx].VersionId
-                    < deleteMarkers[dmIdx].VersionId);
-
-                // if there is a null version, handle accordingly
-                if (nullVersion) {
-                    const isVersionLastModifiedNewer = (new Date(versions[vIdx].LastModified)
-                         > new Date(deleteMarkers[dmIdx].LastModified));
-                    const isDMLastModifiedNewer = (new Date(deleteMarkers[dmIdx].LastModified)
-                         > new Date(versions[vIdx].LastModified));
-                    // 2. by LastModified, find newer
-                    if (isVersionLastModifiedNewer) {
-                        sortedList.push(versions[vIdx++]);
-                    } else if (isDMLastModifiedNewer) {
-                        sortedList.push(deleteMarkers[dmIdx++]);
-                    } else {
-                        // 3. choose one randomly since all conditions match
-                        // TODO: to be fixed
-                        sortedList.push(versions[vIdx++]);
-                    }
+                const isVersionLastModifiedNewer = (new Date(versions[vIdx].LastModified)
+                    >= new Date(deleteMarkers[dmIdx].LastModified));
+                // 2. by LastModified, find newer
+                if (isVersionLastModifiedNewer) {
+                    sortedList.push(versions[vIdx++]);
                 } else {
-                    // 4. by VersionId, lower number means newer
-                    if (isVersionVidNewer) {
-                        sortedList.push(versions[vIdx++]);
-                    } else {
-                        sortedList.push(deleteMarkers[dmIdx++]);
-                    }
+                    sortedList.push(deleteMarkers[dmIdx++]);
                 }
             }
         }
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -1,7 +1,7 @@
 'use strict'; // eslint-disable-line
 
 const async = require('async');
-const { errors } = require('arsenal');
+const { errors, versioning } = require('arsenal');
 const { ObjectMD } = require('arsenal').models;
 const { supportedLifecycleRules } = require('arsenal').constants;
 const {
@@ -15,6 +15,7 @@ const { attachReqUids } = require('../../../lib/clients/utils');
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
 const ReplicationAPI = require('../../replication/ReplicationAPI');
+const { decode } = versioning.VersionID;
 
 // Default max AWS limit is 1000 for both list objects and list object versions
 const MAX_KEYS = process.env.CI === 'true' ? 3 : 1000;
@@ -319,9 +320,13 @@ class LifecycleTask extends BackbeatTask {
                 return done(err);
             }
             // all versions including delete markers
-            const allVersions = this._mergeSortedVersionsAndDeleteMarkers(
-                data.Versions, data.DeleteMarkers
+            const { error, sortedList: allVersions } = this._mergeSortedVersionsAndDeleteMarkers(
+                data.Versions, data.DeleteMarkers, log,
             );
+            if (error) {
+                return done(error);
+            }
+
             // for all versions and delete markers, add stale date property
             const allVersionsWithStaleDate = this._addStaleDateToVersions(
                 bucketData.details, allVersions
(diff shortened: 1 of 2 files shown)


rahulreddy (commented):
@nicolas2bert Why not decode the version id as they are guaranteed to be ordered?

nicolas2bert:
The "base62Decode" function ([Arsenal/lib/versioning/VersionID.ts at development/7.10 · scality/Arsenal](https://github.com/scality/Arsenal/blob/development/7.10/lib/versioning/VersionID.ts#L215) ) could be quite expensive in terms of CPU time and memory usage. There are a few reasons for this:

- The function makes three calls to the base62Integer.decode() and base62String.decode() functions. These functions are themselves relatively expensive, so making three calls to them adds up.

- The function has to iterate through the string to find the start of each encoded value. This is also an expensive operation.

- The function has to slice the string three times. This is another relatively expensive operation.

All of these factors add up to make this function relatively expensive.

nicolas2bert:
@rahulreddy  Is there any advantage to using version ID over creation date?

The only disadvantage, I see, of using the creation date is that two versions could be created at the same time.
In that case, we could decode the version ID to decide.

rahulreddy:
> @rahulreddy Is there any advantage to using version ID over creation date?
> 
> The only disadvantage, I see, of using the creation date is that two versions could be created at the same time. In that case, we could decode the version ID to decide.

Yup, my main concern is the collision. You will have two versions created at the same ms, we already see at at a lot of deployments today. I would add another layer of entropy to be unique, may be you decode version ids and compare only for this case.

nicolas2bert:
Yes, I agree. This condition is handled in the PR.

francoisferrand:
AFAICT, we also rely on ordering of keys in MongoClientInterface (and possibly metadata ?) to compute the last "version" to use as master... Does it mean we will need to do the same kind of fix in Arsenal : https://github.com/scality/Arsenal/blob/bdb59a0e63c20578078bd169e70c78a0c49f6e60/lib/storage/metadata/mongoclient/MongoClientInterface.js#L1055 ?

alexanderchan-scality:
do correct me if I am wrong, short version ID is only used at the encoding/decoding step in cloudserver, and the `generateVersionID` (https://github.com/scality/Arsenal/blob/bdb59a0e63c20578078bd169e70c78a0c49f6e60/lib/versioning/VersionID.ts#L100) is still timestamp based, so list order of keys should still be in some chronological order; the description may be a bit misleading attributing the issue to shortVersionID.

this issue would then look like an edge-case (race condition?)  scenario that will also affect the long version ID and would likely cause issue to the `getLatestVersion` logic pointed out by  Francois

nicolas2bert:
Internal medatada methods (server side) use decoded version id. For decoded version ids, the order is preserved. 

However, the lifecycle "merging and sorting" (client side) method used base62 encoded version id to sort the listing (with merging versions and delete markers) that does not seem to preserve the order (based on creation date). Cf PR description. 


I am not sure to understand your concerns @francoisferrand @alexanderchan-scality .

alexanderchan-scality:
@nicolas2bert i see. I had confused the versionID comparison to be with the internal version id; then I do agree with you that the base62 encode does lose the lexigraphic order property needed for the comparison. 

In that case, the find latest version would not be affected since that uses the internal VID for comparison.

francoisferrand:
> Internal medatada methods (server side) use decoded version id. For decoded version ids, the order is preserved.

thanks @nicolas2bert , that is the part I missed, and got me worried of similar issue with arsenal.

s3 listObjectVersions is supposed to "_request returns objects in the order that they were stored, returning the most recently stored object first_". I do not see any mention (in the doc) that this order covers both versions & delete markers, but it seems to be consistent with the above statement and the exemples there...

So I wonder why we do this sorting in lifecycle, and if we should not implement this in cloudserver/arsenal instead? (This may avoid issues in the UI as well)

nicolas2bert:
@francoisferrand The listObjectVersions API returns versions and delete markers in separate arrays. To determine the object's "stale date," the arrays must be merged and sorted.
With the introduction of lifecycle optimization listings, this logic will be deprecated.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -453,13 +458,42 @@ class LifecycleTask extends BackbeatTask {
...
     }
 
+    /** _decodeVID - decode the version id
+     * @param {string} versionId - version ID
+     * @param {Logger.newRequestLogger} log - logger object
+     * @return {Object} result - { error, versionId }
+     * @return {Error} result.error - if decoding failed
+     * @return {String} result.versionId - decoded version id
+     */
+    _decodeVID(versionId, log) {
```
francoisferrand, line 468: is this the same as arsenal's `versioning/decode` ?

alexanderchan-scality (approved):

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -483,40 +517,57 @@ class LifecycleTask extends BackbeatTask {
...
-                        // 3. choose one randomly since all conditions match
-                        // TODO: to be fixed
-                        sortedList.push(versions[vIdx++]);
-                    }
+                // The `Key` names of the versions and delete markers are the same.
+                // Compare the `LastModified` timestamps of the versions and delete markers.
+                const versionLastModified = new Date(versions[vIdx].LastModified);
+                const deleteMarkerLastModified = new Date(deleteMarkers[dmIdx].LastModified);
+                const isVersionLastModifiedNewer = versionLastModified > deleteMarkerLastModified;
+                const isDMLastModifiedNewer = deleteMarkerLastModified > versionLastModified;
```
francoisferrand, line 525: nit: this makes the code harder to read IMO, as it "hides" the real intent here... i.e. just comparing the date, and handling the 3 possible outcomes of the comparison....

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -483,40 +517,57 @@ class LifecycleTask extends BackbeatTask {
...
+                if (isVersionLastModifiedNewer) {
+                    sortedList.push(versions[vIdx++]);
+                } else if (isDMLastModifiedNewer) {
+                    sortedList.push(deleteMarkers[dmIdx++]);
                 } else {
-                    // 4. by VersionId, lower number means newer
-                    if (isVersionVidNewer) {
+                    // If the version and the delete marker have the same last modified date
+                    const nullVersion = (versions[vIdx].VersionId === 'null'
+                    || deleteMarkers[dmIdx].VersionId === 'null');
```
francoisferrand, line 534: nit: missing indent

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -483,40 +517,57 @@ class LifecycleTask extends BackbeatTask {
...
+                        }
+
+                        const { error: decodedDMError, decodedVersionId: decodedDMId } =
+                            this._decodeVID(deleteMarkers[dmIdx].VersionId, log);
+                        if (decodedDMError) {
+                            return { error: decodedDMError, sortedList: null };
+                        }
+
+                        const isVersionVidNewer = decodedVersionId < decodedDMId;
+                        if (isVersionVidNewer) {
```
francoisferrand, line 560: these push to `sortedList` and incrementation are repeated in multiple places, while they are actually the critical step of the algorithm.

to make core safer & easier to test, would be better to introduce a `bool compare(version, deleteMarker)` function, which would only perform the comparison (and be easy to test all corner cases), and keep the code simple & readable in the algorithm (handling only the progression and update of the list, ie. all `sortedList.push(XXXXX[YYYYY++])`)

francoisferrand (approved):

nicolas2bert:
@bert-e approve

nicolas2bert:
ping

PR #2429, opened 2023-07-05
Short version IDs break ordering - use lastModified 

nicolas2bert:
Backporting change to `hotfix/7.10.8` branch.

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -489,35 +489,13 @@ class LifecycleTask extends BackbeatTask {
                     sortedList.push(deleteMarkers[dmIdx++]);
                 }
             } else {
-                //    NOTE: VersionId may be null
-                const nullVersion = (versions[vIdx].VersionId === 'null'
-                    || deleteMarkers[dmIdx].VersionId === 'null');
-                const isVersionVidNewer = (versions[vIdx].VersionId
-                    < deleteMarkers[dmIdx].VersionId);
-
-                // if there is a null version, handle accordingly
-                if (nullVersion) {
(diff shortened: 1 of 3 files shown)


nicolas2bert:
@bert-e help

nicolas2bert:
@bert-e bypass_jira_check

nicolas2bert:
ping

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -459,13 +463,42 @@ class LifecycleTask extends BackbeatTask {
...
+            const invalidErr = errors.InternalError.customizeDescription('Invalid version id');
+            log.error('error decoding version id', {
+                method: 'LifecycleTask._decodeVID',
+                error: invalidErr,
+                versionId,
+            });
+            return { error: invalidErr, versionId: null };
+        }
+
+        return { error: null, decodedVersionId: decoded };
```
ghost: To match with the function jsdoc, should both returns return `versionId` instead of `decodedVersionId`? Also there is no unit test for this function

ghost (commented):
LGTM otherwise

extensions/lifecycle/tasks/LifecycleTask.js
nicolas2bert: updated

alexanderchan-scality (approved):

extensions/lifecycle/tasks/LifecycleTask.js
ghost: Thanks! But I still see a line with a different return name:
```js
            return { error: invalidErr, versionId: null };
```

is that expected? Should we document it in the jsdoc?

ghost (approved):
LGTM other than my last comment

nicolas2bert:
@bert-e approve

PR #2432, opened 2023-07-20
BB-189 support cold object deletion for DMF

welansari:
Issue: BB-189

### DIFF
--- a/extensions/lifecycle/LifecycleConfigValidator.js
+++ b/extensions/lifecycle/LifecycleConfigValidator.js
@@ -63,6 +63,7 @@ const joiSchema = joi.object({
     },
     coldStorageArchiveTopicPrefix: joi.string().default('cold-archive-req-'),
     coldStorageRestoreTopicPrefix: joi.string().default('cold-restore-req-'),
+    coldStorageGCTopicPrefix: joi.string().default('cold-gc-req-'),
     coldStorageStatusTopicPrefix: joi.string().default('cold-status-'),
 });
 
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -15,7 +15,7 @@ const METASTORE = '__metastore';
 const VaultClientWrapper = require('../utils/VaultClientWrapper');
 
 const config = require('../../lib/Config');
-const { coldStorageRestoreTopicPrefix } = config.extensions.lifecycle;
+const { coldStorageRestoreTopicPrefix, coldStorageGCTopicPrefix } = config.extensions.lifecycle;
 const BackbeatProducer = require('../../lib/BackbeatProducer');
 const locations = require('../../conf/locationConfig.json') || {};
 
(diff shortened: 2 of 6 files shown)


extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -63,6 +63,7 @@ const joiSchema = joi.object({
     },
     coldStorageArchiveTopicPrefix: joi.string().default('cold-archive-req-'),
     coldStorageRestoreTopicPrefix: joi.string().default('cold-restore-req-'),
+    coldStorageGCTopicPrefix: joi.string().default('c19d1d32-7c40-44c8-8fdd-babd5ab960ad.cold-gc-req-'),
```
KillianG: I think you left UUID here

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -320,13 +322,92 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         });
     }
 
+    _handleDeleteOp(entry) {
+        if (entry.type !== 'delete' ||
```
KillianG: As this is tested before calling the function, [here](https://github.com/scality/backbeat/pull/2432/files#diff-37c9ecda10c884ef79419b6744d2491598248b6511813899f07af1909cbf2f43R406) do we really need to check it again ?

KillianG: I see that you made the change but you also removed this test, is it intended ?
```            
entry.key.startsWith(mpuBucketPrefix)) {
```

KillianG (approved):

extensions/lifecycle/LifecycleQueuePopulator.js
welansari: With the switch to the oplog populator and change streams, this check is no longer needed as we don't listen to changes on the shadow buckets anymore

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -320,13 +322,87 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+                    this.log.error('error publishing object delete request entry', {
+                        error: err,
+                        method: 'LifecycleQueuePopulator._handleDeleteOp',
+                    });
+                }
+            });
+        } else {
+            this.log.error(`producer not available for location ${locationName}`, {
+                method: 'LifecycleQueuePopulator._handleDeleteOp',
+            });
```
ghost, line 390: shall we log the topic here as well, to ease debugging if anything happens?

tests/unit/lifecycle/LifecycleQueuePopulator.spec.js
```diff
@@ -172,4 +173,162 @@ describe('LifecycleQueuePopulator', () => {
...
+                md: {
+                    ...objMd,
+                    'archive': {},
+                    'x-amz-storage-class': 'STANDARD',
+                    'dataStoreName': 'us-east-1',
+                },
+                called: false,
+            },
+            {
+                it: 'should skip versined masters',
```
ghost: Nit
```suggestion
                it: 'should skip versioned masters',
```

tests/unit/lifecycle/LifecycleQueuePopulator.spec.js
```diff
@@ -172,4 +173,162 @@ describe('LifecycleQueuePopulator', () => {
...
+                send: kafkaSendStub,
+            };
+            lcqp._producers[`${coldStorageGCTopicPrefix}dmf-v1`] = {
+                send: kafkaSendStub,
+            };
+        });
+        afterEach(() => {
+            kafkaSendStub.reset();
+        });
+        [
```
ghost, line 233: Should we add unit tests for when the location config is not found, and we process a delete?

ghost (approved):
LGTM, just some minor comments

extensions/lifecycle/LifecycleQueuePopulator.js
welansari: I'm thinking it's not necessary, as there is only one deletion topic per cold location, so having only the location name should be good for debugging.

welansari: The topic name is composed of a fixed prefix + the location name, finding it will be an easy tasks

PR #2438, opened 2023-08-03
disable oplog when only having expiration rules

welansari:
Issue: BB-430


tests/unit/lifecycle/LifecycleOplogPopulatorUtils.spec.js
```diff
@@ -0,0 +1,140 @@
...
+                                abortMpuRule,
+                                removeDeleteMarkersRule,
+                            ]),
+                        ],
+                    },
+                },
+                expectedReturn: false,
+            },
+            {
+                it: 'should return true when there is at least one transition rules',
```
francoisferrand: maybe duplicate this test, with only `nonCurrentTransitionRule` enabled: at the moment i think your test set would work as well if we forgot the NonCurrentTransition in the test...

francoisferrand (approved):

rachedbenmustapha (approved):

PR #2439, opened 2023-08-04
Ensure _getBackbeatClient Always Returns BackbeatMetadataProxy

nicolas2bert:
The `_getBackbeatClient` function from the `BucketProcessor` should consistently return an instance of `BackbeatMetadataProxy`. However, there is an issue in the code where the function returns an instance of `BackbeatClient` when caching. 

**Impact:**

- **Ring S3C**: no direct impact today, since `backbeatClient` is only used for "lifecycle Transition workflow" that has been backported but is not supported yet. However, the inconstant result of `_getBackbeatClient` blocks additional development such as “lifecycle listing optimization “.

- **Artesca**: no impact since the logic has been fixed already with the ClientManager: https://github.com/scality/backbeat/blob/development/8.7/lib/clients/ClientManager.js 

_Note:_ the getBackbeatClient from the ObjectProcessor is correctly implemented.

### DIFF
--- a/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
+++ b/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
@@ -210,10 +210,10 @@ class LifecycleBucketProcessor {
     }
 
     /**
-     * Return an backbeat client instance
+     * Return metadata proxy with backbeat client
      * @param {String} canonicalId - The canonical ID of the bucket owner.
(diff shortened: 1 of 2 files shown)


nicolas2bert:
ping

jonathan-gramain (approved):

tmacro (approved):

nicolas2bert:
@bert-e approve

PR #2441, opened 2023-08-11
BB-432: remove useless kafka connectors

welansari:
Issue: BB-432

- Avoid spawning the connectors at startup
- Delete connectors when no bucket is assigned to them

### DIFF
--- a/extensions/oplogPopulator/OplogPopulatorMetrics.js
+++ b/extensions/oplogPopulator/OplogPopulatorMetrics.js
@@ -133,6 +133,22 @@ class OplogPopulatorMetrics {
     }
 
     /**
+     * updates s3_oplog_populator_connectors metric
+     * when a connector is destroyed
+     * @returns {undefined}
+     */
+    onConnectorDestroyed() {
+        try {
+            this.connectors.dec({ isOld: false }, 1);
+        } catch (error) {
+            this._logger.error('An error occured while pushing metrics', {
+                method: 'OplogPopulatorMetrics.onConnectorDestroyed',
+                error: error.message,
+            });
+        }
+    }
+
+    /**
      * updates s3_oplog_populator_reconfiguration_lag_seconds metric
      * @param {Connector} connector connector instance
      * @param {Boolean} success true if reconfiguration was successful
--- a/extensions/oplogPopulator/OplogPopulatorMetrics.js
+++ b/extensions/oplogPopulator/OplogPopulatorMetrics.js
@@ -46,7 +46,6 @@ class OplogPopulatorMetrics {
         this.connectors = ZenkoMetrics.createGauge({
             name: 's3_oplog_populator_connectors',
             help: 'Total number of configured connectors',
-            labelNames: ['isOld'],
         });
         this.reconfigurationLag = ZenkoMetrics.createHistogram({
             name: 's3_oplog_populator_reconfiguration_lag_seconds',
@@ -121,9 +120,7 @@ class OplogPopulatorMetrics {
      */
     onConnectorsInstantiated(isOld, count = 1) {
         try {
-            this.connectors.inc({
-                isOld,
-            }, count);
+            this.connectors.inc(count);
         } catch (error) {
             this._logger.error('An error occured while pushing metrics', {
                 method: 'OplogPopulatorMetrics.onConnectorsInstantiated',
@@ -139,9 +136,9 @@ class OplogPopulatorMetrics {
      */
     onConnectorDestroyed() {
         try {
-            this.connectors.dec({ isOld: false }, 1);
+            this.connectors.dec(1);
         } catch (error) {
-            this._logger.error('An error occured while pushing metrics', {
+            this._logger.error('An error occurred while pushing metrics', {
                 method: 'OplogPopulatorMetrics.onConnectorDestroyed',
                 error: error.message,
             });
--- a/extensions/oplogPopulator/modules/Connector.js
+++ b/extensions/oplogPopulator/modules/Connector.js
@@ -6,6 +6,7 @@ const connectorParams = joi.object({
     name: joi.string().required(),
     config: joi.object().required(),
     buckets: joi.array().required(),
+    isRunning: joi.boolean().required(),
     logger: joi.object().required(),
     kafkaConnectHost: joi.string().required(),
     kafkaConnectPort: joi.number().required(),
@@ -25,6 +26,7 @@ class Connector {
      * @constructor
      * @param {Object} params connector config
      * @param {string} params.name connector name
+     * @param {Boolean} params.isRunning true if connector is running
      * @param {Object} params.config Kafka-connect MongoDB source
      * connector config
      * @param {string[]} params.buckets buckets assigned to this connector
@@ -37,6 +39,7 @@ class Connector {
         this._name = params.name;
         this._config = params.config;
         this._buckets = new Set(params.buckets);
+        this._isRunning = params.isRunning;
         this._state = {
             // Used to check if buckets assigned to this connector
             // got modified from the last connector update
@@ -83,6 +86,12 @@ class Connector {
     get config() { return this._config; }
 
     /**
+     * Getter for connector running state
+     * @returns {Boolean} connector running state
+     */
+    get isRunning() { return this._isRunning; }
+
+    /**
      * Calculate config size in bytes
      * @returns {number} config size
      */
@@ -111,6 +120,7 @@ class Connector {
                 name: this._name,
                 config: this._config,
             });
+            this._isRunning = true;
         } catch (err) {
             this._logger.error('Error while spawning connector', {
                 method: 'Connector.spawn',
@@ -129,6 +139,7 @@ class Connector {
     async destroy() {
         try {
             await this._kafkaConnect.deleteConnector(this._name);
+            this._isRunning = false;
         } catch (err) {
             this._logger.error('Error while destroying connector', {
                 method: 'Connector.destroy',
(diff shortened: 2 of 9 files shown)


extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -238,47 +223,97 @@ class ConnectorsManager {
...
+                });
+                return true;
+            } else if (!connector.isRunning && connector.bucketCount > 0) {
+                await connector.spawn();
+                this._logger.info('Successfully destroyed a connector', {
+                    method: 'ConnectorsManager._spawnOrDestroyConnector',
+                    connector: connector.name
+                });
+                return true;
+            }
```
francoisferrand, line 251: would it not simplify the code [below] if there is an additional `else` here?

```
else if (connector.isRunning) {
   return await connector.updatePipeline(true);
}
```

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -105,40 +105,24 @@ class ConnectorsManager {
...
+        const connector = new Connector({
+            name: connectorName,
+            config,
+            buckets: [],
+            isRunning: false,
+            logger: this._logger,
+            kafkaConnectHost: this._kafkaConnectHost,
+            kafkaConnectPort: this._kafkaConnectPort,
+        });
+        return connector;
```
francoisferrand, line 124: not sure how to handle `onConnectorsInstantiated` metrics, now that the connectors are not created on startup :-/
it seems to me we are interested in the number of connectors created & destroyed (as 2 counters, when we actually creaste/destroy one connector), but not sure how it relates to this one.

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -6,6 +6,7 @@ const connectorParams = joi.object({
     name: joi.string().required(),
     config: joi.object().required(),
     buckets: joi.array().required(),
+    isRunning: joi.boolean().required(),
```
francoisferrand, line 9: `isRunning` is good enough for now, as long as the allocator is always give a _static_ list of connectors.

however, we should (eventually, not here) implement a mecanism to create more connectors when we reach a limit (max mongo query size, pre-configured number of collections...) : and at that point the allocator should be able to allocate more connectors if needed. But this is not for today.

tests/unit/oplogPopulator/Connector.js
```diff
@@ -40,20 +41,25 @@ describe('Connector', () => {
...
         });
     });
 
     describe('destroy', () => {
         it('Should destroy connector', async () => {
             const deleteStub = sinon.stub(connector._kafkaConnect, 'deleteConnector')
                 .resolves();
+            connector._isRunning = true;
+            assert.strictEqual(connector.isRunning, true);
             await connector.destroy();
```
francoisferrand, line 60: what is the point of this assert?

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -105,40 +105,24 @@ class ConnectorsManager {
...
-            });
-            return connector;
-        } catch (err) {
-            this._logger.error('An error occurred while creating connector', {
-                method: 'ConnectorsManager.addConnector',
-                error: err.description || err.message,
-            });
-            throw errors.InternalError.customizeDescription(err.description);
-        }
+    addConnector() {
```
francoisferrand, line 109: function is not async anymore, so call site should be updated:

```js
const newConnector = await this.addConnector();
                     ^^^^^
```

francoisferrand (commented):
looks mostly good, though a deeper refactoring should be done (eventually) to make the list really dynamic instead of virtually pre-creating it to a specified length.

tests/unit/oplogPopulator/Connector.js
welansari: makes sure the getter works properly

francoisferrand: oh yes, i missed the missing underscore and getter...

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -132,6 +132,22 @@ class OplogPopulatorMetrics {
...
     }
 
+    /**
+     * updates s3_oplog_populator_connectors metric
+     * when a connector is destroyed
+     * @returns {undefined}
+     */
+    onConnectorDestroyed() {
+        try {
+            this.connectors.dec({ isOld: false }, 1);
```
francoisferrand: this is not correct: we may remove either old or new connector : the "old" param is based on how the connector was created  (retrieved from kafka-connect or explcitely recreated), so must be tracked throughout...

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -238,47 +222,98 @@ class ConnectorsManager {
...
+     * or destroys connector with no buckets configured
+     * @param {Connector} connector connector instance
+     * @returns {Promise<Boolean>} true if connector state changed
+     * @throws {InternalError}
+     */
+    async _spawnOrDestroyConnector(connector) {
+        try {
+            if (connector.isRunning && connector.bucketCount === 0) {
+                await connector.destroy();
+                this._metricsHandler.onConnectorDestroyed(false);
```
francoisferrand: there is no param in this function....
and if it were, it could be either true or false depending on how that specific connector was created...

francoisferrand (approved):

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -220,9 +205,8 @@ class ConnectorsManager {
             // Add connectors if required number of connectors not reached
             const nbConnectorsToAdd = this._nbConnectors - this._connectors.length;
             await timesLimit(nbConnectorsToAdd, 10, async () => {
-                const newConnector = await this.addConnector();
+                const newConnector = this.addConnector();
```
nicolas2bert, line 207: Nothing is asynchronous here. What is the need for `timesLimit`? It could be misleading.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -238,47 +222,98 @@ class ConnectorsManager {
...
+     * @param {Connector} connector connector instance
+     * @returns {Promise<Boolean>} true if connector state changed
+     * @throws {InternalError}
+     */
+    async _spawnOrDestroyConnector(connector) {
+        try {
+            if (connector.isRunning && connector.bucketCount === 0) {
+                await connector.destroy();
+                this._metricsHandler.onConnectorDestroyed();
+                this._logger.info('Successfully spawned a connector', {
```
nicolas2bert: Log should be `Successfully destroyed a connector`? 
Next condition to be updated as well.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -238,47 +222,98 @@ class ConnectorsManager {
...
+                    connector: connector.name
+                });
+                return true;
+            } else if (!connector.isRunning && connector.bucketCount > 0) {
+                await connector.spawn();
+                this._metricsHandler.onConnectorsInstantiated(false);
+                this._logger.info('Successfully destroyed a connector', {
+                    method: 'ConnectorsManager._spawnOrDestroyConnector',
+                    connector: connector.name
+                });
```
nicolas2bert, line 247: Should we not also update the connector pipeline after spawning?

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -238,47 +222,98 @@ class ConnectorsManager {
...
     }
 
+    /**
+     * Spawns a connector when buckets are configured for it and is not running,
+     * or destroys connector with no buckets configured
+     * @param {Connector} connector connector instance
+     * @returns {Promise<Boolean>} true if connector state changed
+     * @throws {InternalError}
+     */
+    async _spawnOrDestroyConnector(connector) {
```
nicolas2bert, line 231: Now that bucket can be allocated to non-running connector(s) (before connector gets spawn), could this have an impact on the design? (e.g. potentially missing some MongoDB changes until the connector gets spawned...?)

francoisferrand: not really I think, we anyway have a delay between the "activation of new feature which requires oplog" and the moment when a connector is indeed running (i.e. the time for the oplog populator to setup everything) : it may only increase the delay ever so slightly...

but the immediate benefit is that we avoid running any change stream in Veeam/backup use case, and thus avoid risk of breaking the platform.

extensions/oplogPopulator/modules/ConnectorsManager.js
francoisferrand: updating after should not be needed (the `config` is passed to creation api), but indeed `config.pipeline` should be updated before making the call...

extensions/oplogPopulator/modules/ConnectorsManager.js
francoisferrand: i guess it is a remain from the origin (async) code, i'll check

extensions/oplogPopulator/modules/ConnectorsManager.js
francoisferrand: ok we are good: `updatePipeline()` is called automatically by `addBucket()` and `removeBucket()`, and will update the pipeline field automatically. However, it is called with the `doUpdate` param set to `false`, in order to only update in memory : then the config is either applied by the create all (which passes the struct) or by an explicit extra call to `updatePipeline(true)` when the connector was already there.

nicolas2bert (approved):

PR #2447, opened 2023-09-29
Bucket-processor : do not retry the whole listing range when their is an error on individual objects

francoisferrand:
Misc tweaks to lifecycle, which should avoid duplicated processing.

- Properly handle async calls to `_applyTransitionRule`
- Do not retry whole batch when a single entry fails

Issue: BB-437

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -894,18 +894,32 @@ class LifecycleTask extends BackbeatTask {
                 return process.nextTick(cb);
             }
 
-            return async.waterfall([
-                next => this._getRules(bucketData, lcRules, obj, log, next),
-                (applicableRules, next) => {
-                    if (versioningStatus === 'Enabled'
-                    || versioningStatus === 'Suspended') {
-                        return this._compareVersion(bucketData, obj, contents,
-                            applicableRules, versioningStatus, log, next);
-                    }
-                    return this._compareObject(bucketData, obj, applicableRules, log,
-                        next);
+            // We don't want to retry the _whole_ list if only a single entry fails,
+            // so we possibly retry each individual entry here, and ignore errors.
+            // Ignoring error is not too bad, the entry will be picked up again on
+            // next lifecycle run
+            return this.retry({
+                actionDesc: 'compare rules lifecycle entry',
+                logFields: {
+                    key: obj.Key,
+                    versionId: obj.VersionId,
+                    staleDate: obj.staleDate,
+                    versioningStatus,
                 },
-            ], cb);
+                actionFunc: done => async.waterfall([
+                    next => this._getRules(bucketData, lcRules, obj, log, next),
+                    (applicableRules, next) => {
+                        if (versioningStatus === 'Enabled' || versioningStatus === 'Suspended') {
+                            return this._compareVersion(bucketData, obj, contents,
+                                applicableRules, versioningStatus, log, next);
+                        }
+                        return this._compareObject(bucketData, obj, applicableRules, log,
+                            next);
+                    },
+                ], done),
+                shouldRetryFunc: err => err.retryable,
+                log,
+            }, () => cb());
         }, done);
     }
 
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -1284,7 +1284,7 @@ class LifecycleTask extends BackbeatTask {
                     (err, data) => {
                         if (err) {
                             // error already logged at source
-                            return done(err);
+                            return next(err);
                         }
                         const allVersions = [...data.Versions,
                             ...data.DeleteMarkers];
@@ -1469,7 +1469,7 @@ class LifecycleTask extends BackbeatTask {
                 return done();
             }
             if (rules.Transition) {
-                this._applyTransitionRule({
+                return this._applyTransitionRule({
                     owner: bucketData.target.owner,
                     accountId: bucketData.target.accountId,
                     bucket: bucketData.target.bucket,
@@ -1477,8 +1477,7 @@ class LifecycleTask extends BackbeatTask {
                     eTag: obj.ETag,
                     lastModified: obj.LastModified,
                     site: rules.Transition.StorageClass,
-                }, log);
-                return done();
+                }, log, done);
             }
 
             return done();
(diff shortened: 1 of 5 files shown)


nicolas2bert:
Thanks, could it target development/7.10 branch?

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -894,18 +894,32 @@ class LifecycleTask extends BackbeatTask {
...
-                next => this._getRules(bucketData, lcRules, obj, log, next),
-                (applicableRules, next) => {
-                    if (versioningStatus === 'Enabled'
-                    || versioningStatus === 'Suspended') {
-                        return this._compareVersion(bucketData, obj, contents,
-                            applicableRules, versioningStatus, log, next);
-                    }
-                    return this._compareObject(bucketData, obj, applicableRules, log,
-                        next);
+            // We don't want to retry the _whole_ list if only a single entry fails,
```
nicolas2bert, line 912: The same logic has to be applied to `LifecycleTaskV2._compareRulesToList()`.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -894,18 +894,32 @@ class LifecycleTask extends BackbeatTask {
...
+            // so we possibly retry each individual entry here, and ignore errors.
```
nicolas2bert, line 913: Should we even retry? Retry has exponential backoff and could take seconds to complete (to be checked).

francoisferrand: I think we should retry, at least until lifecycle optimization is used everywhere and we can run lifecycle more often : otherwise a minor retriable error could delay lifecycle for 12h.

That said, indeed retry has delay which we need to consider, esp. w/ respect to being a kafka consumer.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1469,16 +1483,15 @@ class LifecycleTask extends BackbeatTask {
                 return done();
             }
             if (rules.Transition) {
-                this._applyTransitionRule({
+                return this._applyTransitionRule({
```
nicolas2bert, line 1526: While I believe it's a beneficial concept to implement asynchronous logic for better flow control, there's a possibility that it could introduce a delay in the execution of the bucket processor task, as it would need to wait for the callback to complete, potentially affecting the Kafka consumer (as discussed earlier).

nicolas2bert: This could clarify why the processing of Lifecycle bucket tasks doesn't seem very asynchronous – we weren't waiting for the callback to be invoked. This might be due to concerns about its potential impact on the Kafka consumer.

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -747,12 +747,18 @@ class LifecycleConductor {
...
-            // check that object tasks topic consumer lag is 0
-            objectTasksCheck:
+            // check that object tasks topic consumer lag for expirations is 0
+            objectTasksExpirationCheck:
             next => this._kafkaBacklogMetrics.checkConsumerLag(
                 this.lcConfig.objectTasksTopic,
                 this.lcConfig.objectProcessor.groupId, 0, next),
 
+            // check that object tasks topic consumer lag for transitions is 0
+            objectTasksTransitionCheck:
```
rachedbenmustapha: There might be several topics to check for transition, as each cold location has its own.

rachedbenmustapha (approved):

extensions/lifecycle/tasks/LifecycleTask.js
francoisferrand: individual objects are processed in parallel (hard-coded to 10 at a time), so it should be fine : and if this introduces a delay, we will need to _manage_ it, not just run uncontrolled number of operations in parallel.
Another hint is that the callback is used in lifecycle v2 task.

extensions/lifecycle/tasks/LifecycleTask.js
francoisferrand: actually, looking at the code of `_applyTransitionRule()`, we already skip the errors from `s3:_getObjectMD`... though some errors can still go through, e.g. when  calling in `s3:listObjectVersions` ( `_checkAndApplyEODMRule()`) or `s3:headObject` in non-versionned bucket...

extensions/lifecycle/conductor/LifecycleConductor.js
francoisferrand: After discussion, it seems this may not be appropriate after all:
* the original goal of checking lag is to ensure we do not start a new listing until the previous one has been terminated, i.e. that each bucket identified by the conductor has been listed → so we should only check that the listing is complete, i.e. no lag on bucket topic
* checking too many topics may make unexpectedly skip the iteration : which is fine when the cron job runs every few minute, and kind of acceptable when the system is still busy processing the last batch... though it may become unacceptable if "other action" (restore expiration, restoration, end of archive...) were to delay the listing, possibly by 
* processing expiration is fast and transitions update the metadata, but it may be better to delay next listing anyway : to avoid filling the topic with objects which are already in it
* yet we certainly don't want the next listing to be delayed by operations not directly related to the listing : restore request, restore expiration, transition status update, ...

**NOTE**: If the cronjob is not running very often, skipping a run may have a significant impact. We should probably generate an alert if the scan lasts longer than the cronjob period, and/or if the scan period is more than 24h → https://scality.atlassian.net/browse/ZENKO-4665

francoisferrand: Topics/consumers groups to check identified, but there may be more work needed: moved to https://scality.atlassian.net/browse/BB-449

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -35,6 +35,20 @@ const MAX_KEYS = process.env.CI === 'true' ? 3 : 1000;
 // concurrency mainly used in async calls
 const CONCURRENCY_DEFAULT = 10;
 
+// Max number of retries (with exp. backoff) that can be be performed for a single
+// entry. We don't use the default, to avoid retrying for a long time (5 minutes)
+// which could make Kafka timeout. 4 retries (e.g. 5 attempts) should be around 8
+// seconds.
+const MAX_RETRIES = 4;
+
+// Maximum number of retries when processing entries in the rang
```
benzekrimaha:
```suggestion
// Maximum number of retries when processing entries in the range
```

benzekrimaha (approved):

nicolas2bert (approved):

PR #2449, opened 2023-10-10
Check if object is archived before trying to cold delete it and avoid…

KillianG:
… logging an error in that case

Issue: BB-448


welansari (approved):

benzekrimaha (approved):
LGTM

PR #2451, opened 2023-10-12
Improve bucket processor logs

francoisferrand:
Log rdkafka events, with a env-var flag allowing to enable all debug logs from the lib, for troubelshooting.

Also log bucket processor task `RequestId`, and pass the conductor's `RequestId` through kafka: so that it is possible to follow the chain of messages.

Issue: BB-439

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -179,6 +179,7 @@ class BackbeatConsumer extends EventEmitter {
             'offset_commit_cb': this._onOffsetCommit.bind(this),
             // automatically create topic
             'allow.auto.create.topics': true,
+            'rebalance_cb': true,
         };
         const topicParams = {};
         if (this._fromOffset !== undefined) {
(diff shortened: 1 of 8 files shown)


lib/BackbeatConsumer.js
```diff
@@ -187,6 +188,9 @@ class BackbeatConsumer extends EventEmitter {
         if (this._fetchMaxBytes !== undefined) {
             consumerParams['fetch.message.max.bytes'] = this._fetchMaxBytes;
         }
+        if (process.env.RDKAFKA_DEBUG_LOGS) {
+            consumerParams.debug = process.env.RDKAFKA_DEBUG_LOGS;
```
jonathan-gramain, line 192: Is this attribute used as a boolean? In which case, might be cleaner to do:

```
consumerParams.debug = process.env.RDKAFKA_DEBUG_LOGS ? true : false;
```

jonathan-gramain (approved):

lib/BackbeatConsumer.js
francoisferrand: no, it is a string containing the list of rdkafka "modules" to enable logs for: `consumer`, `producer`, ... (or `all` to get everyting)

KillianG (approved):

PR #2455, opened 2023-10-17
Call get metadata after error to discover new topic created by another consumer or producer

KillianG:
Issue: BB-453

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -424,6 +424,9 @@ class BackbeatConsumer extends EventEmitter {
                 this._log.debug(err ? `error, retry in 1s: ${err.message}` :
                     'no message is available yet, retry in 1s',
                     { topic: this._topic, groupId: this._groupId });
+                if (err && this._consumer.isConnected() && !this.isPaused()) {
+                    this._consumer.getMetadata({ topic: this._topic }, () => {});
+                }
                 this._scheduleNextTryConsume();
             }
         });
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -180,6 +180,9 @@ class BackbeatConsumer extends EventEmitter {
             // automatically create topic
             'allow.auto.create.topics': true,
             'rebalance_cb': true,
+            // decrease max metadata age to 5s to decrease potential lag between
+            // consumer and producer when a partition is added to a topic
+            'metadata.max.age.ms': 5000,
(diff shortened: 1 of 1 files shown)


lib/BackbeatConsumer.js
```diff
@@ -424,6 +426,18 @@ class BackbeatConsumer extends EventEmitter {
...
+                if (err && this._consumer.isConnected() && !this.isPaused()) {
+                    this._consumer.getMetadata({ topic: this._topic }, err => {
+                        if (err) {
+                            this._log.error('error getting kafka metadatas', {
+                                error: err,
+                                topic: this._topic,
+                                groupId: this._groupId,
+                            });
+                            this.emit('error', err);
+                        }
```
francoisferrand: do we really care ?
if there is another error, we should already catch it ; and if there is no error, we don't really care that `getMetadata` itself failed... (this is kind of a probabilistic workaround)

lib/BackbeatConsumer.js
```diff
@@ -180,6 +180,8 @@ class BackbeatConsumer extends EventEmitter {
             // automatically create topic
             'allow.auto.create.topics': true,
             'rebalance_cb': true,
+            // decrease max metadata age to 5s to avoid lag
```
francoisferrand: "lag" does not really help in this context : you need to explain what _lag_ it avoids...

francoisferrand:
this should go on 8.6, this is bugfix

jonathan-gramain (approved):
LGTM, but couldn't it be better as a solution to this issue to pre-create the archive topic with the wanted number of partitions, e.g. at deployment time, or when the process that produces to this topic starts?

KillianG:
> LGTM, but couldn't it be better as a solution to this issue to pre-create the archive topic with the wanted number of partitions, e.g. at deployment time, or when the process that produces to this topic starts?

I don't think so, because in the case the customer wants to add more partitions to a topic, we don't want the default 10mn lag..

benzekrimaha (approved):

francoisferrand:
> couldn't it be better as a solution to this issue to pre-create the archive topic with the wanted number of partitions, e.g. at deployment time, or when the process that produces to this topic starts?

yes, this is (kind of) planned. We already started managing topics from ZKOP, and will finish the effort in next quarter(s): e.g. covering the few topics (esp. cold) which are not yet managed there, then disabling the creating of the topics by consumers & producers.

PR #2458, opened 2023-10-19
Wait for current tasks to finish before shutdown in BackbeatConsumer

welansari:
To avoid processing Kafka messages multiple times by different consumers, the BackbeatConsumer needs to wait for the current jobs to finish before committing the updated offsets.

This has to be done twice, once when Kafka rebalances and the second on shutdown.

This PR addresses the shutdown part.

Issue: BB-455

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -585,15 +585,24 @@ class BackbeatConsumer extends EventEmitter {
                     logger.bind(this._log)('rdkafka.commit failed', { e: e.toString(), assignment });
                 }
 
-                try {
-                    this._consumer.unassign();
-                } catch (e) {
-                    // Ignore exceptions if we are not connected
-                    const logger = this._consumer.isConnected() ? this._log.error : this._log.info;
-                    logger.bind(this._log)('rdkafka.unassign failed', { e: e.toString(), assignment });
-                }
+                const doUnassign = () => {
+                    try {
+                        this._consumer.unassign();
+                    } catch (e) {
+                        // Ignore exceptions if we are not connected
+                        const logger = this._consumer.isConnected() ? this._log.error : this._log.info;
+                        logger.bind(this._log)('rdkafka.unassign failed', { e: e.toString(), assignment });
+                    }
 
-                this.emit('unassign', status);
+                    this.emit('unassign', status);
+                };
+
+                // publish offsets to zookeeper
+                if (this._kafkaBacklogMetricsConfig) {
+                    this._publishOffsetsCron(doUnassign);
+                } else {
+                    doUnassign();
+                }
             });
 
             if (!this._processingQueue || this._processingQueue.idle()) {
@@ -911,37 +920,30 @@ class BackbeatConsumer extends EventEmitter {
         this._circuitBreaker.stop();
         return async.waterfall([
             next => {
-                if (this._kafkaBacklogMetricsConfig) {
-                    // publish offsets to zookeeper
-                    return this._publishOffsetsCron(() => next());
+                if (this._consumer.isConnected()) {
+                    const subscription = this._consumer.subscription() || [];
+                    if (subscription.length > 0) {
+                        this._consumer.unsubscribe();
+                        // Wait for partition unassign to complete before
+                        // disconnecting, the rebalance callback will handle
+                        // waiting for current jobs to complete as well as commit
+                        // the latest offsets
+                        this.once('unassign', () => next());
+                        return;
+                    }
                 }
-                return process.nextTick(next);
+                process.nextTick(next);
             },
             next => {
                 if (this._zookeeper) {
                     this._zookeeper.close();
+                }
+                if (this._consumer.isConnected()) {
+                    this._consumer.disconnect();
+                    this._consumer.once('disconnected', () => next());
                 } else {
                     process.nextTick(next);
                 }
-            },
-            next => {
-                const subscription = this._consumer.subscription() || [];
-                if (subscription.length > 0) {
-                    this._consumer.unsubscribe();
-                    // Wait for partition unassign to complete before
-                    // disconnecting, the rebalance callback will handle
-                    // waiting for current jobs to complete as well as commit
-                    // the latest offsets
-                    this.on('unassign', next);
-                } else {
-                    next(UNASSIGN_STATUS.IDLE);
-                }
-            },
-            (status, next) => {
-                if (status !== UNASSIGN_STATUS.TIMEOUT) {
-                    this._consumer.disconnect();
-                }
-                this._consumer.once('disconnected', () => next());
             },
         ], () => cb());
     }
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -20,6 +20,12 @@ const CONCURRENCY_DEFAULT = 1;
 const CLIENT_ID = 'BackbeatConsumer';
 const { withTopicPrefix } = require('./util/topic');
 
+const UNASSIGN_STATUS = {
+    IDLE: 'idle',
+    DRAINED: 'drained',
+    TIMEOUT: 'timeout',
+};
+
 /**
  * Stats on how we are consuming Kafka
  * @typedef {Object} ConsumerStats
@@ -447,7 +453,6 @@ class BackbeatConsumer extends EventEmitter {
         });
     }
 
-
     /**
      * Track processing of a task.
      * @param {*} entry - kafka entry being processed
@@ -587,14 +592,16 @@ class BackbeatConsumer extends EventEmitter {
                     const logger = this._consumer.isConnected() ? this._log.error : this._log.info;
                     logger.bind(this._log)('rdkafka.unassign failed', { e: e.toString(), assignment });
                 }
+
+                this.emit('unassign', status);
             });
 
             if (!this._processingQueue || this._processingQueue.idle()) {
-                unassign('idle');
+                unassign(UNASSIGN_STATUS.IDLE);
                 return;
             }
 
-            this._processingQueue.drain = () => unassign('drained');
+            this._processingQueue.drain = () => unassign(UNASSIGN_STATUS.DRAINED);
 
             // Set timeout of `max.poll.interval.ms`), to ensure we complete the rebalance
             // eventually (even if a task is really stuck), and don't get kicked out of the consumer
@@ -602,7 +609,7 @@ class BackbeatConsumer extends EventEmitter {
             // disconnect the consumer, so that the healthcheck will fail and the process will get
             // restarted.
             this._drainProcessQueueTimeout = setTimeout(() => {
-                unassign('timeout');
+                unassign(UNASSIGN_STATUS.TIMEOUT);
 
                 this._log.error('rdkafka.rebalance timeout: consumer stuck, disconnecting');
                 this._consumer.disconnect();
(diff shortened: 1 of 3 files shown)


lib/BackbeatConsumer.js
```diff
@@ -778,6 +779,17 @@ class BackbeatConsumer extends EventEmitter {
         }
         this._circuitBreaker.stop();
         return async.series([
+            next => {
+                // stopping consumption of new messages while
+                // waiting for current ones to be processed
+                this._isShuttingDown = true;
+                // waiting for all current jobs to end before exiting
+                if (this._processingQueue && !this._processingQueue.idle()) {
+                    this._processingQueue.drain = next;
```
francoisferrand: Should we not _call_ drain?
```suggestion
                    this._processingQueue.drain(next);
```

welansari: Welp... it seems we have multiple dependencies of different versions of the same library `async`.
In version 2.x of `async`, the drain looks like [this](https://github.com/caolan/async/blob/c6bdaca4f9175c14fc655d3783c6af6a883e6514/lib/queue.js#L86-L89):
```
q.drain = function() {
    console.log('all items have been processed');
};
```
And in 3.x, like [this](https://github.com/caolan/async/blob/72de392b6dffc466419ae7cab419d3b120d2459d/lib/queue.js#L78-L81)
```
q.drain(() => {
    console.log('all done')
})
```
I'm not sure yet how the choice of version is made, but the 2.x works locally and on the CI, and the 3.x implementation works on tests in an actual Artesca lab :/

I pushed the code just to test if it passes in the CI, but i stil have to figure out the dependency issue before merging anything.

KillianG (approved):

lib/BackbeatConsumer.js
welansari: seems something was wrong with the image i was testing with. The code does indeed only use version 2.x

lib/BackbeatConsumer.js
```diff
@@ -783,9 +784,31 @@ class BackbeatConsumer extends EventEmitter {
...
+                this._isShuttingDown = true;
+                // waiting for all current jobs to end before exiting
+                if (this._processingQueue && !this._processingQueue.idle()) {
+                    this._processingQueue.drain = () => {
+                        this._log.info('drained processing queue', {
+                            method: 'BackbeatConsumer.close',
+                        });
+                        next();
+                    };
+                    return;
```
francoisferrand: nit: not needed (there is no other return here)

lib/BackbeatConsumer.js
```diff
@@ -783,9 +784,31 @@ class BackbeatConsumer extends EventEmitter {
         }
         this._circuitBreaker.stop();
         return async.series([
+            next => {
+                // stopping consumption of new messages while
+                // waiting for current ones to be processed
+                this._isShuttingDown = true;
+                // waiting for all current jobs to end before exiting
+                if (this._processingQueue && !this._processingQueue.idle()) {
+                    this._processingQueue.drain = () => {
```
francoisferrand: * to be on the safe side, we should reset the drain callback once it is called
* what happens if there is a rebalance during shutdown? we may have an issue, since we overwrite the callback there as well... :-/

lib/BackbeatConsumer.js
```diff
@@ -797,7 +820,7 @@ class BackbeatConsumer extends EventEmitter {
                 if (this._zookeeper) {
                     this._zookeeper.close();
                 }
-                if (this._consumer) {
+                if (this._consumer && this._consumer.isConnected()) {
```
francoisferrand: when is this happening?

lib/BackbeatConsumer.js
```diff
@@ -381,8 +382,8 @@ class BackbeatConsumer extends EventEmitter {
         // needs explicit polling to get new messages
 
         const nNewConsumeRequests = this._getAvailableSlotsInPipeline();
-        if (nNewConsumeRequests === 0) {
-            // processing pipeline is 100% busy, do not attempt to consume
+        if (nNewConsumeRequests === 0 || this._isShuttingDown) {
```
francoisferrand: should we use a flag (as you did), or `pause()` the consumer?
the risk would be that we fail to poll, and get kicked out of the group without waiting for us to commit...

lib/BackbeatConsumer.js
welansari: if i don't return here `process.nextTick(next);` will be called and we won't wait for the drain

lib/BackbeatConsumer.js
welansari: This is linked to your PR, when a rebalance times out and the consumer gets disconnected we don't want in that case to wait for the disconnect event.

lib/BackbeatConsumer.js
welansari: i don't think having the consumer getting kicked is an issue here, the only case where the `close()` function gets called is when the pods shutdown.

Using the pause function already implemented is probably a bad idea as it unsubscribes from the topic, so the consumer won't be able to commit.

Using the Kafka consumer's pause function is going to add more complexity that is not really needed, as we have to track the paused topic/partitions and update them in case of an assignment from a rebalance, ...

lib/BackbeatConsumer.js
welansari: Since we block the polls when shutting down, the rebalance callback will not be called.

If a rebalance was already in progress, the drain function will be overwritten, and as soon as the jobs finish the consumer will commit and disconnect.

The commit will work as the `unassign` of the rebalance will not be called because of the overwrite.

lib/BackbeatConsumer.js
francoisferrand: I was talking about consumer.pause()

francoisferrand: > as we have to track the paused topic/partitions and update them in case of an assignment from a rebalance

Do we? This is shutdown, so we should not take part in the assign...

Btw, maybe we should first unsubscribe: then wait for revoke to complete (eg drain, commi, unassign) : then disconnect and quit...

welansari: good idea, we should use the logic already present in the rebalance callback to wait for jobs and commit the offsets

lib/BackbeatConsumer.js
welansari: used the rebalance callback to handle all the logic, we shouldn't have these problems anymore

lib/BackbeatConsumer.js
```diff
@@ -923,10 +918,21 @@ class BackbeatConsumer extends EventEmitter {
...
+                if (this._consumer && this._consumer.isConnected()) {
                     this._consumer.unsubscribe();
-                    this._consumer.disconnect();
-                    this._consumer.on('disconnected', () => next());
+                    // Wait for partition unassign to complete before
+                    // disconnecting, the rebalance callback will handle
+                    // waiting for current jobs to complete as well as commit
+                    // the latest offsets
+                    this._waitForUnassign = true;
+                    const interval = setInterval(() => {
```
francoisferrand: why not "emit" a signal when unassigned, and simply waiting for that signal here?
(or other way to explicitely trigger the continuation instead of a timer)

lib/BackbeatConsumer.js
```diff
@@ -923,10 +918,21 @@ class BackbeatConsumer extends EventEmitter {
...
+                        if (this._waitForUnassign) {
+                            return;
+                        }
+                        clearInterval(interval);
+                        this._consumer.disconnect();
```
francoisferrand: if we unassign due to timeout, we may be disconnect already here?
(and that function may thus fail?)

lib/BackbeatConsumer.js
```diff
@@ -923,10 +921,20 @@ class BackbeatConsumer extends EventEmitter {
...
+                if (this._consumer && this._consumer.isConnected()) {
                     this._consumer.unsubscribe();
-                    this._consumer.disconnect();
-                    this._consumer.on('disconnected', () => next());
+                    // Wait for partition unassign to complete before
+                    // disconnecting, the rebalance callback will handle
+                    // waiting for current jobs to complete as well as commit
+                    // the latest offsets
+                    this.once('unassign', status => {
+                        if (status === UNASSIGN_STATUS.TIMEOUT) {
```
francoisferrand: when this happens, can we/should we wait for `disconnected` ?
disconnect() will have been called, not sure if we will get the event...

(can we write a test case for close() with a timeout on the rebalance ?)

welansari: I added a test case for this.
If the event already got emitted before adding the event handler then we will never receive the event yes.

lib/BackbeatConsumer.js
```diff
@@ -587,22 +592,24 @@ class BackbeatConsumer extends EventEmitter {
...
             });
 
             if (!this._processingQueue || this._processingQueue.idle()) {
-                unassign('idle');
+                unassign(UNASSIGN_STATUS.IDLE);
                 return;
             }
 
-            this._processingQueue.drain = () => unassign('drained');
+            this._processingQueue.drain = () => unassign(UNASSIGN_STATUS.DRAINED);
```
francoisferrand, line 604: shouldn't we stick to the 'empty' callback?
the queue should never drain (or fill) unless we go through this revoke-ation process anyway, so it seems safer to not emit the signal...

lib/BackbeatConsumer.js
```diff
@@ -902,17 +909,8 @@ class BackbeatConsumer extends EventEmitter {
...
-                        this._consumer.commit();
-                    } catch (e) {
-                        // Ignore exceptions if we are not connected
-                        const logger = this._consumer.isConnected() ? this._log.error : this._log.debug;
-                        logger.bind(this._log)('commit failed', { e: e.toString() });
-                    }
-                }
                 if (this._kafkaBacklogMetricsConfig) {
                     // publish offsets to zookeeper
                     return this._publishOffsetsCron(() => next());
```
francoisferrand: publishing offsets must be done _after_ the unsusbribe is complete : so that it comes after the `commit`...

(but possibly before we un-assign, if it changes anything...)
(and definitely before the timeout and associated disconnect...)

lib/BackbeatConsumer.js
```diff
@@ -922,14 +920,28 @@ class BackbeatConsumer extends EventEmitter {
...
                 } else {
-                    process.nextTick(next);
+                    next(UNASSIGN_STATUS.IDLE);
+                }
+            },
+            (status, next) => {
+                if (status !== UNASSIGN_STATUS.TIMEOUT) {
+                    this._consumer.disconnect();
                 }
+                this._consumer.once('disconnected', () => next());
```
francoisferrand: to be on the safe side, we should check if we are already disconnected.
(`consumer.isConnected()` only looks at the state of the "javascript" consumer, not the underlying librdkafka client ; and that state is updated on `disconnected` event)

and looking at the same code, it is safe to call disconnect() while disconnecting.

so we may simple do this:
```
if (this._consumer.isConnected()) {
  this._consumer.disconnect();
  this._consumer.once('disconnected', () => next());
} else {
  next();
}
```

lib/BackbeatConsumer.js
francoisferrand: maybe an easy wait to get around this is to introduce a flag to skip `disconnect` on timeout during shutdown: this way we can perform the `unsubscribe`, and when unassign'ed (and commited) we can do the offset publishing..

francoisferrand: actually, maybe publishing to zookeeper should always be done?

benzekrimaha (approved):

PR #2459, opened 2023-10-19
BB-452: add version-id to delete notifications

rachedbenmustapha:
Includes backported logic from 8.x branch, adapted to match available log entries.

### DIFF
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -120,21 +120,80 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
     }
 
     /**
+     * Returns the correct versionId
+     * to display according to the
+     * versioning state of the object
+     * @param {Object} value log entry object
+     * @param {Object} overheadFields - extra fields missing from the log entry
+     * @return {String} versionId
+     */
+    _getVersionId(value, overheadFields) {
+        const versionId = value.versionId || (overheadFields && overheadFields.versionId);
+        const isNullVersion = value.isNull;
+        const isVersioned = !!versionId;
+        // Versioning suspended objects have
+        // a versionId, however it is internal
+        // and should not be used to get the object
+        if (isNullVersion || !isVersioned) {
+            return null;
+        }
+        return versionId;
+    }
+
+    /**
+     * Decides if we should process the entry.
+     * Since we get both master and version events,
+     * we need to avoid pushing two notifications for
+     * the same event.
+     * - For non-versioned buckets, we process the master
+     * objects' events.
+     * - For versioned buckets, we process version events
+     * and ignore all master events.
+     * - For versioning suspended buckets, we need to process
+     * both master and version events, as the master is considered
+     * a separate version.
+     * @param {String} key object key
+     * @param {Object} value object metadata
+     * @return {boolean} - true if entry is valid
+     */
+    _shouldProcessEntry(key, value) {
+        const isMaster = isMasterKey(key);
+        const hasVersionId = !!value.versionId;
+
+        if (!isMaster) {
+            // versioned keys do generate a notifications. FIXME: in some cases
+            // they may duplicate the notification with a master update of the
+            // same null version.
+            return true;
+        }
+        // generate a notification for non-versioned or null-versioned master
+        // keys, but not for regular versions as then the versioned key triggers
+        // the notification.
+        return !hasVersionId || value.isNull;
+    }
+
+    /**
      * Process object entry from the log
      *
      * @param {String} bucket - bucket
      * @param {String} key - object key
      * @param {Object} value - log entry object
      * @param {String} type - entry type
-     * @param {String} commitTimestamp - when the entry was written, used as a fallback
+     * @param {Object} overheadFields - extra context fields missing from the log entry
+     * @param {String} overheadFields.commitTimestamp - when the entry was written, used as a fallback
      *   if no last-modified MD attribute available
+     * @param {String} overheadFields.versionId - version id involved in this operation, if the log entry
+     *   does not contain it
      * @return {undefined}
      */
-    _processObjectEntry(bucket, key, value, type, commitTimestamp) {
-        const versionId = value.versionId || null;
-        if (!isMasterKey(key)) {
+    _processObjectEntry(bucket, key, value, type, overheadFields) {
+        if (!this._shouldProcessEntry(key, value)) {
             return undefined;
         }
+
+        const versionId = this._getVersionId(value, overheadFields);
+        const baseKey = this._extractVersionedBaseKey(key);
+
         const config = this.bnConfigManager.getConfig(bucket);
         if (config && Object.keys(config).length > 0) {
             const { eventMessageProperty }
@@ -145,11 +204,13 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                 = value[eventMessageProperty.dateTime];
             if (eventType === undefined && type === 'del') {
                 eventType = notifConstants.deleteEvent;
-                dateTime = commitTimestamp;
+                if (!dateTime) {
+                    dateTime = (overheadFields && overheadFields.commitTimestamp) || null;
+                }
             }
             const ent = {
                 bucket,
-                key,
+                key: baseKey,
                 eventType,
                 versionId,
                 dateTime,
@@ -171,7 +232,7 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                     eventTime: message.dateTime,
                 });
                 this.publish(this.notificationConfig.topic,
-                    `${bucket}/${key}`,
+                    `${bucket}/${baseKey}`,
                     JSON.stringify(message));
             }
             return undefined;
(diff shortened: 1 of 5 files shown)


KillianG (approved):

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -119,22 +119,78 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+        }
+        return versionId;
+    }
+
+    /**
+     * Decides if we should process the entry.
+     * Since we get both master and version events,
+     * we need to avoid pushing two notifications for
+     * the same event.
+     * - For non versiond buckets, we process the master
```
jonathan-gramain:
```suggestion
     * - For non-versioned buckets, we process the master
```

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -119,22 +119,78 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+        const isMaster = isMasterKey(key);
+        const isVersionedMaster = isMaster && !!value.versionId;
+        const isNullVersion = isMaster && value.isNull;
+        if (!isMaster
+            || !isVersionedMaster
+            || isNullVersion
+            ) {
+            return true;
+        }
+        return false;
```
jonathan-gramain: Style suggestion, but should be reworked due to the above comment.

```suggestion
    return (!isMaster
        || !isVersionedMaster
        || isNullVersion
        );
```

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -119,22 +119,78 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+     * both master and version events, as the master is considered
+     * a separate version.
+     * @param {String} key object key
+     * @param {Object} value object metadata
+     * @return {boolean} - true if entry is valid
+     */
+    _shouldProcessEntry(key, value) {
+        const isMaster = isMasterKey(key);
+        const isVersionedMaster = isMaster && !!value.versionId;
+        const isNullVersion = isMaster && value.isNull;
```
jonathan-gramain: Beware that in V0 format, null versions may have both a master and a versioned key. This is due to a design flaw, it has been fixed in the rework of the null versions format for V1 (and for V0 when V1 is enabled). For this reason, I think we should also exclude versioned keys that have the `isNull` attribute.

But... in V1, null versions may **only** be in a versioned key, however this version key will then be a null key (i.e. its version ID is the empty string, or the raw key ends with `\0`).

So, I think to be correct we should, in addition to this logic, exclude versioned keys (with a non-empty version ID) that have the `isNull` attribute. Null keys are guaranteed not to be duplicated with a master key in the new logic.

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -119,22 +119,78 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
-    _processObjectEntry(bucket, key, value, type, commitTimestamp) {
-        const versionId = value.versionId || null;
-        if (!isMasterKey(key)) {
+    _processObjectEntry(bucket, key, value, type, overheadFields) {
+        if (!this._shouldProcessEntry(key, value)) {
             return undefined;
         }
+
+        const versionId = this._getVersionId(value, overheadFields);
+        const baseKey = this._extractVersionedBaseKey(key);
```
jonathan-gramain, line 195: By the way, is `key` the raw key returned from the oplog? If so, we need to strip it of its V1 prefix `\x7fM` or `\x7fV`. But it's not related to this PR, we should maybe have a separate ticket to handle this.

extensions/notification/NotificationQueuePopulator.js
francoisferrand: @Kerkesni how do we handle this in backbeat 8 ? (if we actually do handle it...)

jonathan-gramain: "null" **keys** (those ending with `\0`) have only been implemented in S3C, Artesca uses master or versioned keys for null versions.

extensions/notification/NotificationQueuePopulator.js
rachedbenmustapha: I was trying to keep the code and jsdoc the same as in backbeat 8.x where it comes from, to ease forward porting and eventual use of backbeat 8.x in RING, but given the discrepancy you pointed out between Artesca and RING it's not much use anymore.

extensions/notification/NotificationQueuePopulator.js
rachedbenmustapha: `LogReader._processLogEntry()` already strips the V1 prefixes before feeding entries to the filters.

extensions/notification/NotificationQueuePopulator.js
rachedbenmustapha: @jonathan-gramain I added a unit test to that effect.

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -119,22 +119,75 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+     * @param {String} key object key
+     * @param {Object} value object metadata
+     * @return {boolean} - true if entry is valid
+     */
+    _shouldProcessEntry(key, value) {
+        const isMaster = isMasterKey(key);
+        const hasVersionId = !!value.versionId;
+        const isVersionedMaster = isMaster && hasVersionId;
+        const isNullVersion = isMaster && value.isNull;
+        const isNullWithVersionId = hasVersionId && value.isNull;
```
jonathan-gramain: To exclude the null keys from this, we also need to check the key itself: if it ends with `\0`, it is a null key and should not be excluded (because then we want to process that entry as there is no master key associated). Null keys have a `versionId` field in their metadata because they correspond to null versions from a versioning-suspended bucket, or in V1 they may also be a current delete marker on a versioned bucket.

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -119,22 +119,76 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+     * @param {Object} value object metadata
+     * @return {boolean} - true if entry is valid
+     */
+    _shouldProcessEntry(key, value) {
+        const isMaster = isMasterKey(key);
+        const hasVersionId = !!value.versionId;
+        const keyHasNullVersion = key.endsWith('\0');
+        const isVersionedMaster = isMaster && hasVersionId;
+        const isNullVersion = (isMaster || keyHasNullVersion) && value.isNull;
+        const isNullWithVersionId = hasVersionId && value.isNull && !keyHasNullVersion;
```
jonathan-gramain: minor points:

- `hasVersionId` is redundant because versions with `isNull` attribute always have a version ID.

- on the `keyHasNullVersion` variable name: I've used "null key" to refer to those keys ending with `\0` in Metadata, not claiming that this is the best name, but it could be better to stick with it for consistency (so naming the variable `isNullKey` instead of `keyHasNullVersion` which I also find confusing as "null version" has another meaning)

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -119,22 +119,76 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+
+        return (!isMaster || !isVersionedMaster || isNullVersion) && !isNullWithVersionId;
```
jonathan-gramain: since `isVersionedMaster` implies `isMaster` I think we can get rid of the `!isMaster ||` to simplify.

I think the logic is right, but it looks to me a bit convoluted as is. Also some comments would help understanding. Here's a suggestion to change the test while maintaining the same truth condition, and get rid of some of the intermediate variables (it's arguably a bit longer to read but I think simplifies the logic overall):

```
if (keyHasNullVersion) { // could be named 'isNullKey'
  return true; // null keys always generate a notification because they are never duplicated with a versioned key or a master key
}
if (!isMaster) {
  return !version.isNull; // versioned keys (other than null keys) representing a null version never generate a notification because they always duplicate the reference null-versioned master key. Other, regular (non-null) versioned keys generate a notification.
}
return !hasVersionId || version.isNull; // generate a notification for non-versioned or null-versioned master keys, but not for regular versions as then the versioned key triggers the notification.

Comment (jonathan-gramain) [reply extensions/notification/NotificationQueuePopulator.js]: Actually I just realized that we also need to generate notifications for non-current null versions, which are normally stored in versioned keys in V0. If we don't have the context of the master key when we process a versioned key, we cannot know if it's a duplicate or not.

Is it possible to process versioned keys that have `isNull` knowing the context of the master? For example, knowing that it was updated atomically with the same value could allow to ignore the versioned update.

Without this context, the best we can do is always generate notifications for `isNull` versioned keys, leading to potential duplicates with the master. (it's better than skipping notifications).

Comment (jonathan-gramain) [reply extensions/notification/NotificationQueuePopulator.js]: So this would be the new logic (I think) without the context of master updates:

```
if (!isMaster) {
  return true; // versioned keys do generate a notifications. FIXME: in some cases they may duplicate the notification with a master update of the same null version.
}
return !hasVersionId || version.isNull; // generate a notification for non-versioned or null-versioned master keys, but not for regular versions as then the versioned key triggers the notification.
```

jonathan-gramain: A potential solution: what about keeping a short-lived cache of notifications we published? We could use it to deduplicate notifications on this pathological case.

rachedbenmustapha: @jonathan-gramain I made the cosmetic change, seeing duplicates when suspending/unsuspending versioning between put and delete (that I don't think occurred with the more verbose version), but that should be taken care of by `env_s3. enable_null_version_compat_mode`.

jonathan-gramain (approved):

PR #2461, opened 2023-10-20
Remove default behaviour of selecting the first cold location on expiration of cold objects

welansari:
We don’t need this logic anymore as sorbet always sets the location name and this might introduce unwanted behaviours when having multiple locations

Issue: BB-417

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
+++ b/extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
@@ -4,7 +4,6 @@ const { errors } = require('arsenal');
 const ObjectMD = require('arsenal').models.ObjectMD;
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
-const locations = require('../../../conf/locationConfig.json') || {};
 
 class LifecycleUpdateExpirationTask extends BackbeatTask {
     /**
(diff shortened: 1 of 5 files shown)


extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
```diff
@@ -146,10 +145,15 @@ class LifecycleUpdateExpirationTask extends BackbeatTask {
         async.waterfall([
             next => this._getMetadata(entry, log, next),
             (objMD, next) => {
-                const coldLocation = entry.getAttribute('target.location') ||
-                    // If location not specified, use the first (and only) location
-                    // This is a temporary fix, until Sorbet is fixed to provide the information
-                    Object.keys(locations).find(name => locations[name].isCold);
+                const coldLocation = entry.getAttribute('target.location');
```
francoisferrand, line 147: this should be done before retrieving metadata?

tests/unit/lifecycle/LifecycleUpdateExpirationTask.spec.js
```diff
@@ -124,4 +124,28 @@ describe('LifecycleUpdateExpirationTask', () => {
             done();
         });
     });
+
+    it.only('should throw an error when location is not specified', done => {
```
KillianG: remove `.only`

KillianG (changes requested):

KillianG (approved):

francoisferrand (approved):

PR #2463, opened 2023-10-25
Complete & commit current task(s) on rebalance

francoisferrand:
Implement rebalance callback, so that we will now properly handle the rebalance process at
application level, and thus avoid extra rebalances or processing the same entry multiple times:
* On `unassign`, wait until current tasks are complete, by draining the queue;
* Then explicitely commit, to ensure the persisted timely (since we are usually relying on a
  auto-commit timer, which may happen at any time in the process)
* Finally call the unassign() method of consumer to let the cluster sync and proceed with the
  rebalance.

We are expected (by kafka) to do this whole process within `max.poll.interval.ms` (default 5min),
similar to how we are expected to poll periodically. This should be fine in general, except in case
of a bug where a task is really stuck. In that case, the consumer will get kicked out of the
cluster, and commit may fail. If tasks eventually unblock, the consumer will reconnect to the
cluster on next call to `consume()`, which will trigger a rebalance. There is no guard yet against
this, and we keep the same behavior as before in those cases.

- Store offset in BackbeatConsumer bootstrap
- Add kafka consumer rebalance test
- Drain the queue on rebalance
- Commit before unassigning
- Fix BackbeatConsumer breakbeat tests

Issue: BB-441

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -601,6 +601,11 @@ class BackbeatConsumer extends EventEmitter {
                             { topic: self._topic, groupId: self._groupId });
                         clearInterval(producerTimer);
                         clearInterval(consumerTimer);
+                        self._consumer.offsetsStore([{
+                            topic: self._topic,
+                            partition: message.partition,
+                            offset: message.offset + 1,
+                        }]);
                         self._consumer.commit();
                         self._consumer.unsubscribe();
                         producer.close(() => {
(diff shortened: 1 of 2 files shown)


welansari (approved):

lib/BackbeatConsumer.js
```diff
@@ -512,6 +512,15 @@ class BackbeatConsumer extends EventEmitter {
             });
 
             if (!this._processingQueue || this._processingQueue.idle()) {
+                try {
```
fmonjalet, line 515: minor: this may be worth a sub-function to ensure we always commit before unassigning and also keep on logging consistently when the code changes.

KillianG (approved):

fmonjalet (approved):
No question to ask in the end, the code answered it itself!

lib/BackbeatConsumer.js
francoisferrand: in next PR :-)

PR #2464, opened 2023-10-25
BB-462 Expiring non-versioned objects fails

nicolas2bert:
**Note1: The change won't be forward-ported into Artesca branches.**
Note2: Only buckets with versioning enabled can be configured for replication.

To expire non-versioned objects correctly, we currently rely on the InvalidBucketState error returned when the S3C Backbeat API is used for a non-versioned bucket's object, until CLDSRV-461 is implemented to add support for non-versioned buckets to the Backbeat API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "7.10.13",
+  "version": "7.10.14",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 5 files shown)


nicolas2bert:
ping

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "7.10.13",
+  "version": "7.10.14",
```
jonathan-gramain, line 3: Can you do the version bump in a separate commit

nicolas2bert: Done.

nicolas2bert:
@bert-e reset

jonathan-gramain (approved):

ghost (approved):
LGTM

nicolas2bert:
@bert-e approve

PR #2465, opened 2023-10-25
Change connector partition name on restart

welansari:
The partition name is used as an indicator in the offsets topic to distinguish between resume tokens of the different connectors.
    
When a connector is respawned it should start processing event from the current moment and not use the previous resume token as it can be invalid if the connector stayed shut for a long time and the mongo oplog rotated.
  
Issue: BB-463

### DIFF
--- a/extensions/oplogPopulator/modules/Connector.js
+++ b/extensions/oplogPopulator/modules/Connector.js
@@ -115,6 +115,13 @@ class Connector {
      * @throws {InternalError}
      */
     async spawn() {
+        if (this._isRunning) {
+            this._logger.error('tried spawning an already created connector', {
+                method: 'Connector.spawn',
+                connector: this._name,
+            });
+            return;
+        }
         try {
             await this._kafkaConnect.createConnector({
                 name: this._name,
@@ -137,6 +144,13 @@ class Connector {
      * @throws {InternalError}
      */
     async destroy() {
+        if (!this._isRunning) {
+            this._logger.error('tried destroying an already destroyed connector', {
+                method: 'Connector.destroy',
+                connector: this._name,
+            });
+            return;
+        }
         try {
             await this._kafkaConnect.deleteConnector(this._name);
             this._isRunning = false;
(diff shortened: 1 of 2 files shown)


tests/unit/oplogPopulator/Connector.js
```diff
@@ -49,6 +49,22 @@ describe('Connector', () => {
...
             assert.strictEqual(connector.isRunning, true);
         });
+        it('Should change partition name on creation', async () => {
+            sinon.stub(connector._kafkaConnect, 'createConnector')
+                .resolves();
+            await connector.spawn();
+            const partition = connector.config['offset.partition.name'];
+            connector._isRunning = false;
+            await connector.spawn();
+            assert.notStrictEqual(partition, connector.config['offset.partition.name']);
```
KillianG: nit: 
```suggestion
            await connector.spawn();
            const partitionName = connector.config['offset.partition.name'];
            connector._isRunning = false;
            await connector.spawn();
            assert.notStrictEqual(partitionName, connector.config['offset.partition.name']);
```

KillianG (approved):

tests/unit/oplogPopulator/Connector.js
```diff
@@ -49,6 +49,22 @@ describe('Connector', () => {
             }));
             assert.strictEqual(connector.isRunning, true);
         });
+        it('Should change partition name on creation', async () => {
```
francoisferrand, line 52: do we need a test to explicitely verify that we _keep_ the current partition name if the connector is reused,
or is this already checked by the other tests?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -109,12 +110,25 @@ class Connector {
...
+        this._config['offset.partition.name'] = `partition-${uuid.v4()}`;
+    }
+
     /**
      * Creates the Kafka-connect mongo connector
      * @returns {Promise|undefined} undefined
      * @throws {InternalError}
      */
     async spawn() {
+        if (this._isRunning) {
```
francoisferrand, line 127: this actually never happens, it is just an extra safety check?
→ should log an error if it happens

francoisferrand (approved):

extensions/oplogPopulator/modules/Connector.js
welansari: yes just a safety check

tests/unit/oplogPopulator/Connector.js
welansari: this is already tested in the `ConnectorsManager` tests, we test if we keep reused connector's config

PR #2466, opened 2023-10-26
Add timeout on rebalance

francoisferrand:
Add timeout on rebalance' drain, and disconnect the kafka client from the cluster. We would get kicked out of the cluster anyway (causing the message to be picked'up by another consumer and preventing us from commiting, and in most cases it happens if a task is "stuck": doing it on our side however ensures the consumer is updated about this, and eventually causes the healthcheck to fail, so the process will get restarted.

To give better visibility on this, introduce a few metrics and alerts to detect and warn of "slow" task and rebalance.

Also fix a minor issue, ensuring we always try to fill the queue, even if consume returned a few entries to process.

Issue: BB-440

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -414,9 +414,13 @@ class BackbeatConsumer extends EventEmitter {
                     this._offsetLedger.onOffsetConsumed(
                         entry.topic, entry.partition, entry.offset);
                     this._messagesConsumed++;
+
+                    const finishProcessingTask = this._startProcessingTask(entry);
                     this._processingQueue.push(entry, (err, completionArgs) => {
                         this._onEntryProcessingDone(err, entry, completionArgs);
+                        finishProcessingTask(err);
                     });
+
                     // update Zenko metrics with the latest consumed
                     // message timestamp, to later allow computing
                     // backlog metrics on demand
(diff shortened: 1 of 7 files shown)


lib/BackbeatConsumer.js
```diff
@@ -417,7 +430,7 @@ class BackbeatConsumer extends EventEmitter {
                     return undefined;
                 });
             }
-            if (err || entries.length === 0) {
+            if (err || entries.length < nNewConsumeRequests) {
                 this._log.debug(err ? `error, retry in 1s: ${err.message}` :
                     'no message is available yet, retry in 1s',
```
welansari: maybe we should update this message

lib/BackbeatConsumer.js
```diff
@@ -82,6 +84,7 @@ class BackbeatConsumer extends EventEmitter {
                     intervalS: joi.number().default(60),
                 },
                 site: joi.string(),
+                maxPollIntervalMs: joi.number().min(45000).default(300000),
```
welansari, line 87: Should we expose this parameter in backbeat config ?

francoisferrand: I don't' think this is needed at the moment, we should keep things "simple" (as much as possible at least)

monitoring/lifecycle/alerts.yaml
```diff
@@ -181,3 +181,23 @@ groups:
...
+      summary: "Kafka message processing time too high"
+
+  - alert: KafkaConsumerRebalanceTimeout
+    Expr: |
+      sum(s3_zenko_queue_rebalance_total_count{namespace="${namespace}", status="timeout"}) by(pod) > 0
+    Labels:
+     severity: critical
+    Annotations:
+      description: >-
+        Kafka rebalance has timed out for pod `{{ $labels.pod }}`, which indicates that the consumer
```
ghost, line 206: `should` -> `must`? Same for others, I guess, as the level is critical...

monitoring/lifecycle/alerts.yaml
```diff
@@ -181,3 +181,23 @@ groups:
...
 
+  - alert: KafkaConsumerSlowTask
+    Expr: |
+      sum(s3_zenko_queue_slowTasks_count{namespace="${namespace}"}) by(job) > 0
+    For: "5m"
+    Labels:
+     severity: warning
+    Annotations:
+      description: Some tasks are taking too long to process in {{ $labels.job }}.
+      summary: "Kafka message processing time too high"
```
ghost: If that's displayed to customers shall we be a bit more explicit on the impacts of the alerts?

welansari (approved):

ghost (approved):
LGTM, for this one I may lack some context on the exact root causes but according to your comments it looks good as well

Just two non blocking comments on the alerts

monitoring/lifecycle/alerts.yaml
francoisferrand: i still have some doubt that there may be corner-cases where a delay is appropriate, so i will stick with `should` :-)

anyway users _should_ (😉) never this see, since the process is expected to become unhealthy when this happens...

PR #2468, opened 2023-11-07
Use separate topic for lifecycle transition tasks

welansari:
The lifecycle conductor uses the lag on the object tasks to throttle the lifecycle scan.
However, the object tasks topic handles events that are not in the direct path and that does not impact the lifecycle scan.
We use a different topic for those events to avoid unwanted throttle.

Issue: BB-467

### DIFF
--- a/extensions/lifecycle/LifecycleConfigValidator.js
+++ b/extensions/lifecycle/LifecycleConfigValidator.js
@@ -12,6 +12,7 @@ const joiSchema = joi.object({
     zookeeperPath: joi.string().required(),
     bucketTasksTopic: joi.string().required(),
     objectTasksTopic: joi.string().required(),
+    transitionTasksTopic: joi.string().default(parent => parent.objectTasksTopic),
     coldStorageTopics: joi.array().items(joi.string()).unique().default([]),
     auth: authJoi.optional(),
     forceLegacyListing: joi.boolean().default(true),
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
@@ -41,8 +41,7 @@ class LifecycleObjectProcessor extends EventEmitter {
      * @param {Object} lcConfig - lifecycle configuration object
      * @param {String} lcConfig.auth - authentication info
      * @param {String} lcConfig.objectTasksTopic - lifecycle object topic name
-     * consumer group id
-     *  of max allowed concurrent operations
+     * @param {String} lcConfig.transitionTasksTopic - lifecycle transition topic name
      * @param {Object} s3Config - S3 configuration
      * @param {Object} s3Config.host - s3 endpoint host
      * @param {Number} s3Config.port - s3 endpoint port
@@ -73,7 +72,7 @@ class LifecycleObjectProcessor extends EventEmitter {
         return 'object-processor';
     }
 
-    _getObjecTaskConsumerParams() {
+    _getTopicConsumerParams(topic) {
         return {
             zookeeper: {
                 connectionString: this._zkConfig.connectionString,
@@ -83,7 +82,7 @@ class LifecycleObjectProcessor extends EventEmitter {
                 site: this._kafkaConfig.site,
                 backlogMetrics: this._kafkaConfig.backlogMetrics,
             },
-            topic: this._lcConfig.objectTasksTopic,
+            topic,
             groupId: this._processConfig.groupId,
             concurrency: this._processConfig.concurrency,
             queueProcessor: this.processObjectTaskEntry.bind(this),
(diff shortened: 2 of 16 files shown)


extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -12,6 +12,7 @@ const joiSchema = joi.object({
     zookeeperPath: joi.string().required(),
     bucketTasksTopic: joi.string().required(),
     objectTasksTopic: joi.string().required(),
+    transitionTasksTopic: joi.string().required(),
```
francoisferrand: should it be required (i.e. require ZKOP to set the value on newer backbeat versions), or can we default to objectTasksTopic if this field is not specified (i.e. maintain backwards compatibility, degrading to the old behavior) ?

This may be done like this:

```suggest
transitionTasksTopic: joi.string().required().default(config => config.objectTaskTopic),
```

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -94,9 +93,11 @@ class LifecycleObjectProcessor extends EventEmitter {
         };
     }
 
-    getConsumerParams() {
+    getConsumerParams(isTransitionProcessor = false) {
```
francoisferrand: instead of adding the topic here, it should be added/overriden directly in the derived class: e.g. in `transitionProcessor.getConsumerParams()`

tests/functional/lifecycle/LifecycleTask.js
```diff
@@ -328,10 +334,10 @@ class ProducerMock {
 
     sendToTopic(topicName, entries, cb) {
         const entry = JSON.parse(entries[0].message);
-        if (topicName === 'bucket-tasks') {
+        if (topicName === bucketTasksTopic) {
             this.sendCount.bucket++;
             this.entries.bucket.push(entry);
-        } else if (topicName === 'object-tasks') {
+        } else if (topicName === objectTasksTopic) {
```
francoisferrand, line 340: shoudl we also count entries pushed to transition topic ?

welansari: No need as nothing is pushed into it here, only ones pushing to it are either sorbet or the replication process

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -48,8 +49,32 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
         return 'transition-processor';
     }
 
+    _getTransitionTaskConsumerParams() {
```
francoisferrand: should be refactored to avoid duplicating the actual "definition" of the params

maybe introduce a function `getConsumerParam(topic)` in ObjectProcessor, or just add an (optional) param to `getConsumerParams(topic = this._lcConfig.objecTasksTopic)`...

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -73,7 +72,7 @@ class LifecycleObjectProcessor extends EventEmitter {
         return 'object-processor';
     }
 
-    _getObjecTaskConsumerParams() {
+    _getObjecTaskConsumerParams(topic) {
```
francoisferrand: nit: `_getTopicConsumerParams` ?

francoisferrand (approved):

tests/unit/lifecycle/LifecycleObjectExpirationProcessor.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    beforeEach(() => {
+        objectProcessor = new LifecycleObjectExpirationProcessor(
+            config.zookeeper,
+            config.kafka,
+            config.extensions.lifecycle,
+            config.s3,
+        );
+    });
+
+    it('consumer params should contain object tasks topic', () => {
```
KillianG:
```suggestion
    it('should contain object tasks topic in consumer params', () => {
```

tests/unit/lifecycle/LifecycleObjectTransitionProcessor.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    beforeEach(() => {
+        objectProcessor = new LifecycleObjectTransitionProcessor(
+            config.zookeeper,
+            config.kafka,
+            config.extensions.lifecycle,
+            config.s3,
+        );
+    });
+
+    it('consumer params should contain transition tasks topic', () => {
```
KillianG:
```suggestion
    it('should contain transition tasks topic in consumer params', () => {
```

KillianG (approved):
2 little comment about phrasing the tests

PR #2471, opened 2023-11-15
BB-466 Dealing with expired zookeeper sessions

nicolas2bert:
Introducing the ZookeeperManager Class, which offers:

* Connection Management: This class efficiently manages connections to the ZooKeeper server. It ensures continuity by handling reconnections whenever the session expires. 

* Error and State Management: The class handles errors and meticulously logs state changes. It is essential for debugging. 

* Abstraction Layer: It creates an abstraction over the database-specific code, simplifying the integration of application logic. This layer not only facilitates an easier transition to different databases in the future but also aids in simplifying the mocking and testing processes.

_Why isn't this issue present in Artesca?_
Artesca uses Kubernetes as its orchestrator, which uses liveness probes to check if containers are operating correctly. Zookeeper state is part of the liveness probe of the Backbeat services. If a liveness probe fails, Kubernetes will understand that the application in the container is not functioning properly. In response, Kubernetes typically restarts the problematic container to try to resolve the issue.
This logic is not present in S3C/Federation.

However, to ensure consistency between the two projects, this change will be forward-ported to Artesca.

**TODO**: while forward-porting, make sure all the Zookeeper client uses the new ZookeeperManager class.

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -1,7 +1,6 @@
 'use strict'; // eslint-disable-line strict
 
 const async = require('async');
-const zookeeper = require('node-zookeeper-client');
 
 const { errors } = require('arsenal');
 const { RedisClient } = require('arsenal').metrics;
@@ -17,6 +16,7 @@ const Healthcheck = require('./Healthcheck');
 const routes = require('./routes');
 const { getSortedSetKey, getSortedSetMember } =
     require('../util/sortedSetHelper');
+const ZookeeperManager = require('../clients/ZookeeperManager');
 
 // StatsClient constant defaults
 // TODO: This should be moved to constants file
@@ -180,15 +180,6 @@ class BackbeatAPI {
     _checkProducersReady() {
         return this._crrProducer.isReady() && this._metricProducer.isReady()
             && this._crrStatusProducer.isReady();
-    }
-
-    /**
-     * Check if Zookeeper and Producer are connected
-     * @return {boolean} true/false
-     */
-    isConnected() {
-        return this._zkClient.getState().name === 'SYNC_CONNECTED'
-            && this._checkProducersReady();
     }
 
     /**
@@ -1079,13 +1070,12 @@ class BackbeatAPI {
         const zookeeperUrl =
             `${this._zkConfig.connectionString}${populatorZkPath}`;
 
-        const zkClient = zookeeper.createClient(zookeeperUrl, {
+        const zkClient = new ZookeeperManager(zookeeperUrl, {
             autoCreateNamespace: this._zkConfig.autoCreateNamespace,
-        });
-        zkClient.connect();
+        }, this._logger);
 
         zkClient.once('error', cb);
-        zkClient.once('connected', () => {
+        zkClient.once('ready', () => {
             zkClient.removeAllListeners('error');
             this._zkClient = zkClient;
             return cb();
--- a/lib/api/Healthcheck.js
+++ b/lib/api/Healthcheck.js
@@ -11,7 +11,7 @@ class Healthcheck {
     /**
      * @constructor
      * @param {object} repConfig - extensions.replication configs
-     * @param {node-zookeeper-client.Client} zkClient - zookeeper client
+     * @param {ZookeeperManager} zkClient - zookeeper client manager
      * @param {BackbeatProducer} crrProducer - producer for CRR topic
      * @param {BackbeatProducer} crrStatusProducer - CRR status producer
      * @param {BackbeatProducer} metricProducer - producer for metric
--- a/lib/clients/ZookeeperManager.js
+++ b/lib/clients/ZookeeperManager.js
@@ -1,0 +1,324 @@
+const zookeeper = require('node-zookeeper-client');
+const EventEmitter = require('events');
+const async = require('async');
+
+class ZookeeperManager extends EventEmitter {
+    /**
+     * Constructs an instance of the ZookeeperManager.
+     *
+     * @param {string} connectionString - The connection string for the ZooKeeper server.
+     * @param {Object} options - Configuration options for the ZooKeeper client.
+     * @param {Logger} log - An instance of a logging utility.
+     */
+    constructor(connectionString, options, log) {
+        super();
+
+        this.connectionString = connectionString;
+        this.log = log;
+        this.options = options;
+        this.client = null;
+
+        this._connect();
+    }
+
+    /**
+     * Establishes a connection to the ZooKeeper server.
+     *
+     * This method initializes a new ZooKeeper client using the provided connection string
+     * and options. It sets up event listeners for various client states such as 'connected',
+     * 'disconnected', and 'expired'. The method handles reconnection logic in case of
+     * client expiration and forwards ZooKeeper events through this class's EventEmitter interface.
+     *
+     * If 'autoCreateNamespace' is set in the options and the base path does not exist in ZooKeeper,
+     * this method tries to create the necessary path (namespace) in ZooKeeper.
+     *
+     * @method _connect
+     * @return {undefined}
+     */
+    _connect() {
+        // clean up exists client before reconnect
+        if (this.client) {
+            // close and clean up the existing ZooKeeper client connection.
+            this.close();
+            this.client.removeAllListeners();
+        }
+
+        this.client = zookeeper.createClient(this.connectionString, this.options);
+
+        let nsIndex;
+        let namespace;
+        let rootZkClient;
+        async.series([
+            // Check zookeeper client is connected
+            next => this.client.once('connected', next),
+            // Create namespace if it exists and options.autoCreateNamespace is true
+            next => {
+                // TODO: ARTESCA-10337 The 'autoCreateNamespace' functionality is currently specific to
+                // Artesca and may be removed in future versions once the Zenko Operator can handle base path creation.
+                // Once removed, we can simply rely on the 'connected' state instead of the 'ready' state and
+                // stop listening on 'error' event.
+                if (this.options && this.options.autoCreateNamespace) {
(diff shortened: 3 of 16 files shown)


nicolas2bert:
ping

lib/api/BackbeatAPI.js
```diff
@@ -173,7 +173,7 @@ class BackbeatAPI {
      * @return {boolean} true/false
      */
     isConnected() {
-        return this._zkClient.getState().name === 'SYNC_CONNECTED'
+        return this._zkClient && this._zkClient.getState() && this._zkClient.getState().name === 'SYNC_CONNECTED'
```
francoisferrand:
```suggestion
        return this._zkClient?.getState()?.name === 'SYNC_CONNECTED'
```

that said, what is a pre-existing bug?
Not sure what changed that we need these check if they were not needed before...

(we can always add more checks, but this is also extra noise, making the code harder to read and maintain...)

lib/api/Healthcheck.js
```diff
@@ -33,8 +33,9 @@ class Healthcheck {
     _getConnectionDetails() {
         return {
             zookeeper: {
-                status: this._zkClient.getState().name === 'SYNC_CONNECTED' ?
-                    'ok' : 'error',
+                status: this._zkClient &&
+                    this._zkClient.getState() &&
+                    this._zkClient.getState().name === 'SYNC_CONNECTED' ? 'ok' : 'error',
                 details: this._zkClient.getState(),
```
francoisferrand: if we need to check `this._zkClient` (like we do for previous line), we should check it here as well...

lib/api/BackbeatAPI.js
nicolas2bert: I agree, not needed.

lib/api/Healthcheck.js
nicolas2bert: I agree, not needed.

tests/functional/notification/NotificationConfigManager.js
```diff
@@ -2,16 +2,28 @@ const assert = require('assert');
...
 
+const ZookeeperManager = require('../../../lib/clients/ZookeeperManager');
+
 const logger = new werelogs.Logger('NotificationConfigManager:test');
 const zkConfigParentNode = 'config';
 const concurrency = 10;
 const bucketPrefix = 'bucket';
 const timeoutMs = 100;
 
+function mockZookeeperManager() {
```
rachedbenmustapha: Not a big deal, feel free to ignore, but this helper could be factored out as it's used in more than 1 place.

tests/functional/lib/ZookeeperManager.js
```diff
@@ -0,0 +1,281 @@
...
+        beforeEach(done => {
+            const connectionString = 'localhost:2181';
+            const options = {};
+            zkClient = new ZookeeperManager(connectionString, options, log);
+            zkClient.once('connected', () => done());
+        });
+
+        it('should trigger the disconnected event after closing the connection', done => {
+            zkClient.close();
+            zkClient.on('disconnected', () => done());
```
rachedbenmustapha: Not sure how often this would happen, but I think there's a risk of test flakiness if attaching to an event after triggering it - depending on how async `close` is.

rachedbenmustapha (approved):

lib/clients/ZookeeperManager.js
```diff
@@ -0,0 +1,313 @@
...
+     * If 'autoCreateNamespace' is set in the options and the base path does not exist in ZooKeeper,
+     * this method tries to create the necessary path (namespace) in ZooKeeper.
+     *
+     * @method _connect
+     * @return {undefined}
+     */
+    _connect() {
+        // clean up exists client before reconnect
+        if (this.client) {
+            this.client.removeAllListeners();
```
jonathan-gramain, line 43: Should we call `this.client.close()` before removing the listeners?

lib/clients/ZookeeperManager.js
```diff
@@ -0,0 +1,313 @@
...
+                // NO_NODE when base path does not exist, hence use
+                // getData() instead
+                this.getData('/', err => {
+                    if (err && err.name !== 'NO_NODE') {
+                        this.emit('error', err);
+                        return;
+                    }
+                    // NO_NODE error and autoCreateNamespace is enabled
+                    if (err) {
+                        const nsIndex = this.connectionString.indexOf('/');
```
jonathan-gramain: The connection string may not always contain a namespace. Maybe in our products we always have one, but for more genericity it could be better to handle cases where there's no namespace.

lib/clients/ZookeeperManager.js
```diff
@@ -0,0 +1,313 @@
...
+     * @method _connect
+     * @return {undefined}
+     */
+    _connect() {
+        // clean up exists client before reconnect
+        if (this.client) {
+            this.client.removeAllListeners();
+        }
+
+        this.client = zookeeper.createClient(this.connectionString, this.options);
```
jonathan-gramain, line 46: Structurally I think we should turn the whole construct into an async series or waterfall to avoid nesting too many levels. Also we would then be able to treat success/error in a unique place at the end of the series and emit the appropriate `ready` or `error` event.

lib/clients/ZookeeperManager.js
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * @method create
+     * @param {String} path The node path.
+     * @param {Buffer} [data=undefined] The data buffer.
+     * @param {Array} [acls=ACL.OPEN_ACL_UNSAFE] An array of ACL object.
+     * @param {CreateMode} [mode=CreateMode.PERSISTENT] The creation mode.
+     * @param {Function} callback The callback function.
+     * @return {undefined}
+     */
+    create(path, data, acls, mode, callback) {
```
jonathan-gramain, line 149: I'm not too fond of having to reproduce the original API of the zookeeper client, for the maintenance burden notably if we have to use new functions later and for the extra code.

I am thinking, would it be better to extend the zookeeper client itself in such case, as if we're manipulating the zookeeper client but with extra logic inside to deal with session expiration etc., possibly it would cause some issues on its own, so not sure. Not blocking for me but just mentioning it doesn't look like an ideal solution.

rachedbenmustapha: I'm not too sure about extending the zk client itself; the client we use is pretty antiquated and not super well maintained so we may want to migrate at some point to a more modern client (`async` based?) and having `ZookeeperManager` be an abstraction layer can help make the transition smoother.

lib/clients/ZookeeperManager.js
nicolas2bert: We already close it here already: https://github.com/scality/backbeat/pull/2471/files#diff-ab8f069d2d46a96dbf066330d2185a0e4705edacbc2cb222cfc40abf4511b9d3R91

but moving `this.client.close()` here might make more sense to make it more generic.

lib/clients/ZookeeperManager.js
```diff
@@ -0,0 +1,323 @@
...
+            // Create namespace if it exists and options.autoCreateNamespace is true
+            next => {
+                // TODO: ARTESCA-10337 The 'autoCreateNamespace' functionality is currently specific to
+                // Artesca and may be removed in future versions once the Zenko Operator can handle base path creation.
+                // Once removed, we can simply rely on the 'connected' state instead of the 'ready' state and
+                // stop listening on 'error' event.
+                if (this.options && this.options.autoCreateNamespace) {
+                    nsIndex = this.connectionString.indexOf('/');
+                    namespace = this.connectionString.slice(nsIndex);
+                    if (nsIndex > -1 && namespace !== '/') {
```
jonathan-gramain, line 63: The connection string could look like `1.2.3.4:2181`, I don't think we would even see a `/` in this case.

lib/clients/ZookeeperManager.js
```diff
@@ -0,0 +1,323 @@
...
+                rootZkClient = zookeeper.createClient(hostPort, this.options);
+                rootZkClient.connect();
+                rootZkClient.once('connected', next);
+            },
+            // Once connected, use the root zookeeper client to create the namespace
+            next => rootZkClient.mkdirp(namespace, err => {
+                if (err && err.name !== 'NODE_EXISTS') {
+                    return next({ event: 'error', err });
+                }
+                return next({ event: 'ready' });
```
jonathan-gramain, line 93: Do you need to disconnect/close the `rootZkClient` once the path exists?

lib/clients/ZookeeperManager.js
nicolas2bert: this has been tested already: https://github.com/scality/backbeat/pull/2471/files#diff-5fc9cf9601779e975b71eba0b498d453274fe5807e82c8ad97850b43eb3e2098R33
Not sure what you meant.

nicolas2bert: And this as well: https://github.com/scality/backbeat/pull/2471/files#diff-5d37e364a4ba55f3567e02ae1eaca3e7304140cf610d9c5da922751d5fc309f9R79

lib/clients/ZookeeperManager.js
nicolas2bert: Good catch. Thanks!

lib/clients/ZookeeperManager.js
jonathan-gramain: Ah yes I missed the `nsIndex > -1` check, we precompute the `namespace` though without checking if there's a namespace, but in this case calling `slice(-1)` should not cause issues except it will return "garbage" in `namespace` but we don't use it. Might be slightly cleaner to compute `namespace` only if `nsIndex > -1` but it's minor and not blocking for me.

jonathan-gramain (approved):

nicolas2bert:
@bert-e approve

PR #2473, opened 2023-11-23
Delete orphaned cold object

welansari:
When an object is deleted (from S3) while it was archived, a message is sent to the cold storage backend to notify the removal.

When this is happens during the transition (archival) process, the object is not yet archived, and we cannot sent such message to the backend. In most cases, this should be fine, as the archiving will fail with an error related to the absence of the (removed) object.

However, there is a race condition, and it is possible that the archival succeeds in backend if the object is removed at the end of this process or in between this result from the backend and the update of metadata in backbeat : the metadata update will fail (as expected), but the object will still be in cold storage and remain there forever (not referenced from anywhere, and thus orphaned)

Issue: BB-469

### DIFF
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
@@ -10,6 +10,7 @@ const { LifecycleResetTransitionInProgressTask } =
       require('../tasks/LifecycleResetTransitionInProgressTask');
 const { updateCircuitBreakerConfigForImplicitOutputQueue } = require('../../../lib/CircuitBreaker');
 const { LifecycleRetriggerRestoreTask } = require('../tasks/LifecycleRetriggerRestoreTask');
+const BackbeatProducer = require('../../../lib/BackbeatProducer');
 
 class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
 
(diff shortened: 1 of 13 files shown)


extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -45,6 +46,45 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
+     * @param {function} cb callback called when producer
+     * startup is complete
+     * @return {undefined}
+     */
+    setupProducer(cb) {
+        const producer = new BackbeatProducer({
+            kafka: { hosts: this._kafkaConfig.hosts },
+            maxRequestSize: this._kafkaConfig.maxRequestSize,
+        });
+        producer.once('error', () => {});
```
francoisferrand: why: are we expecting exactly one error before being ready?

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -40,6 +42,38 @@ class LifecycleColdStatusArchiveTask extends LifecycleUpdateTransitionTask {
...
+        this.coldProducer.sendToTopic(coldGcTopic, [{ message: gcMessage }], err => {
+            if (err) {
+                log.error('error sending cold object deletion entry', {
+                    error: err,
+                    entry: entry.getLogInfo(),
+                    method: 'LifecycleColdStatusArchiveTask._deleteColdObject',
+                });
+                return cb(err);
+            }
+            return cb(new SkipMdUpdateError('cold object deleted'));
```
francoisferrand, line 73: do we need to return an error here?
the "error" is actually handled, and it requires you to ignore that error at line 161 : so it may be best to just return `null` here instead...

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
welansari: Missing a call to the callback when getting an error before being ready

francoisferrand (approved):

KillianG (approved):

PR #2474, opened 2023-11-29
BB-468: propagate restore-expiration to cold backend

rachedbenmustapha:
When a `RestoreObject` call is performed on a restored object to update its lifetime, propagate to the backend cold storage.

Issue: BB-468

### DIFF
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -15,7 +15,11 @@ const METASTORE = '__metastore';
 const VaultClientWrapper = require('../utils/VaultClientWrapper');
 
 const config = require('../../lib/Config');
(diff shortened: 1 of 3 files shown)


extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -323,6 +330,42 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         });
     }
 
+    _adjustRestoreMaxAge(md) {
+        // We might be receiving this message with some delay, so ignore if already expired.
+        const objectExpired = new Date(md.archive.restoreWillExpireAt) < new Date();
```
rachedbenmustapha, line 335: This check might not actually be a good thing, if shortening a restore's lifetime and getting this message later. Open to removing it.

francoisferrand (approved):

welansari (approved):

PR #2476, opened 2023-11-29
Add metrics for lifecycle lateness & duration

francoisferrand:
- Add missing S3 ops tracking in GC & lifecycle
- Fix duplicated getMetadata call in metrics
- Add kafka publish metric for data-mover topic
- Add metric to track lifecycle triggering
- Add metric to track lifecycle completion
- Set transition-in-progress flag for Transition as well
- Add metrics for GC duration
- Add metrics for lifecycle lateness
- Add `transitionTime` to cold storage request messages
- Bump arsenal

Issue: BB-447


extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
```diff
@@ -72,7 +70,8 @@ class LifecycleUpdateTransitionTask extends BackbeatTask {
         objMD.setLocation(location)
             .setDataStoreName(newLocationName)
             .setAmzStorageClass(newLocationName)
-            .setOriginOp('s3:LifecycleTransition');
+            .setOriginOp('s3:LifecycleTransition')
+            .setTransitionInProgress(false);
```
welansari, line 74: Resetting this field might make the bucket processor continuously try to transition these objects, i don't think we have checks on `dataStoreName` or on `storageClass` to skip these objects ?

francoisferrand: This happens at the end of the transition, when the object has been moved to the new location.
It must be reset (similar to the end of the cold storage), to ensure the object becomes elligible to the "next" transition...

I think there should be a check already on the actual location of the object though (e.g. to avoid moving an object to the location where it is already), I will double check.

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -130,11 +137,18 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
...
+            LifecycleMetrics.onLifecycleStarted(log, 'expiration:mpu',
+                location, startTime - transitionTime);
+
             req = s3Client[reqMethod](reqParams);
         }
         attachReqUids(req, log);
         return req.send(err => {
             LifecycleMetrics.onS3Request(log, reqMethod, 'expiration', err);
+            LifecycleMetrics.onLifecycleCompleted(log,
+                actionType === 'deleteMPU' ? 'expiration:mpu' : 'expiration',
```
KillianG, line 150: Is there a reason not to follow the same case rule ? shouldn't it be `delete:mpu` or `expirationMPU` ?

KillianG (approved):

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
francoisferrand: this follows the same kind of convention as bucket notifications, and thus makes MPU expiration a kind of "subset" of expirations. Specifically, this allows having metrics for all expirations, by filtering on `type =~ '^expiration'`.

extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
francoisferrand: Ok, found the protection, in [Arsenal's `getApplicableRules`](https://github.com/scality/arsenal/blob/acd13ff31b92b3d9469e8625c9a9f671d06a1ec2/lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts):

```js
        // Do not transition to a location where the object is already stored.
        if (applicableRules.Transition
            && applicableRules.Transition.StorageClass === metadata.StorageClass) {
            applicableRules.Transition = undefined;
        }

        if (applicableRules.NoncurrentVersionTransition
            && applicableRules.NoncurrentVersionTransition.StorageClass === metadata.StorageClass) {
            applicableRules.NoncurrentVersionTransition = undefined;
        }
```

This expects the `StorageClass` to match the name of the current location, which is the case here: the storage class will indeed be "STANDARD" most of the time, except for cold objects (where it will indeed be [name of the location](https://github.com/scality/backbeat/blob/a9b8814a779eb277efbcd28ee278e51cfbb414d0/extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js)) or transistioned object (where it will be set to the [destination location after the copy](https://github.com/scality/backbeat/blob/9d9893209a1fe21bd2d16cbbfa66c185990c6590/extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js))

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1019,7 +1024,12 @@ class LifecycleTask extends BackbeatTask {
                 .setAttribute('target.bucket', bucketData.target.bucket)
                 .setAttribute('target.accountId', bucketData.target.accountId)
                 .setAttribute('target.key', obj.Key)
-                .setAttribute('details.lastModified', obj.LastModified);
+                .setAttribute('details.lastModified', obj.LastModified)
+                .setAttribute('details.dataStoreName', obj.StorageClass || '')
```
francoisferrand, line 1028: in case this is 'STANDARD', the dataStoreName should actually be the bucket's `region`... we don't currently have access to it, but it may be retrieved through a simple `HEAD` on the bucket.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1217,30 +1233,31 @@ class LifecycleTask extends BackbeatTask {
...
             },
-            (entry, objectMD, next) =>
-                ReplicationAPI.sendDataMoverAction(this.producer, entry, log, err =>
-                    next(err, entry, objectMD)),
+            (entry, objectMD, next) => {
+                const locationName = params.site;
+                LifecycleMetrics.onLifecycleTriggered(this.log, 'bucket',
+                    locationsConfig[locationName].isCold ? 'archive' : 'transition',
+                    locationName, Date.now() - params.transitionTime);
+                if (locationsConfig[locationName].isCold) {
```
nicolas2bert: Should we make sure that `locationsConfig[locationName]` still exists? What if the location gets deleted while lifecycle is ongoing?

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1217,30 +1233,31 @@ class LifecycleTask extends BackbeatTask {
...
-
-                return process.nextTick(next);
+                objectMD.setTransitionInProgress(true, params.transitionDate);
+                const putParams = {
+                    bucket: params.bucket,
+                    objectKey: params.objectKey,
+                    versionId: params.versionId,
+                    mdBlob: objectMD.getSerialized(),
+                };
+                return this._putObjectMD(putParams, log, err => {
```
nicolas2bert, line 1258: This change (removing the location.isCold check), results in a putMetadata being added for every eligible object, thereby increasing the number of metadata calls and consequently consuming more resources. 
Why is this necessary? 
Can't we rely on the conductor's throttling mechanism to prevent the same object from being transitioned again in a new batch?

francoisferrand: this is necessary to ensure we don't try to transition the object while it is being replicated : i.e. after it has been processed from the `data-mover` topic but before copy to the new location is complete (which will update the storage class eventually, and make the object not-elligible).

The conductor throttles itself based on the length of the data-mover queue indeed; and `Lifecycle.getApplicableRules` skips the rule if the storage class is already changed: but there is nothing preventing the object from being transitioned again in between.

That said, practically we have no "non cold" transition, and the performance impact is acceptable for transitions to cold : so I am wondering if the performance impact would really make a difference for these transitions ?

extensions/lifecycle/tasks/LifecycleTask.js
francoisferrand: that error is handled already in `sendDataMoverAction`, and indeed we should not "crash" before....

nicolas2bert (approved):

PR #2477, opened 2023-11-29
Fix missing originOp on lifecycle metadata updates

francoisferrand:
Incorrect or missing `originOp` causes incorrectly duplicated bucket notifications, missing notifications or wrong notifications being sent.

- Fix bucket notification at the end of archive
- Clear originOp when updating replication status
- Fix flaky BackbeatConsumer rebalance test

Issue: BB-472

### DIFF
--- a/extensions/replication/tasks/UpdateReplicationStatus.js
+++ b/extensions/replication/tasks/UpdateReplicationStatus.js
@@ -246,6 +246,7 @@ class UpdateReplicationStatus extends BackbeatTask {
         }
         const versionId =
             sourceEntry.getReplicationSiteDataStoreVersionId(site);
+        entry.setOriginOp('');
         entry.setReplicationSiteDataStoreVersionId(site, versionId);
         entry.setSite(site);
         return entry;
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -248,6 +248,7 @@ class GarbageCollectorTask extends BackbeatTask {
                 objMD.setLocation()
                     .setDataStoreName(newLocation)
                     .setAmzStorageClass(newLocation)
+                    .setOriginOp('s3:LifecycleTransition')
                     .setTransitionInProgress(false)
                     .setUserMetadata({
                         'x-amz-meta-scal-s3-transition-attempt': undefined,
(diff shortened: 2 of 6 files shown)


extensions/replication/tasks/UpdateReplicationStatus.js
```diff
@@ -246,6 +246,7 @@ class UpdateReplicationStatus extends BackbeatTask {
         }
         const versionId =
             sourceEntry.getReplicationSiteDataStoreVersionId(site);
+        entry.setOriginOp('');
```
KillianG, line 249: For my understanding can you explain why we set an empty origin op ?

jonathan-gramain: This is to avoid double notifications essentially, as keeping the original field value would trigger a new notification on update, but maybe we could/should set an actual value here instead like `s3:ReplicateObject` (if that's your original question).

jonathan-gramain (commented):
Left a comment replying to @KillianG , otherwise LGTM

extensions/replication/tasks/UpdateReplicationStatus.js
KillianG: `s3:ReplicateObject` is not [AWS standard](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html#:~:text=s3%3AReplication%3A*,s3%3AReplication%3AOperationNotTracked).. 
here are the only standard notifications for replication: 
```
s3:Replication:*
s3:Replication:OperationFailedReplication
s3:Replication:OperationMissedThreshold
s3:Replication:OperationReplicatedAfterThreshold
s3:Replication:OperationNotTracked
```

I don't see any that can fit our need here..

Can we introduce a non standard notification for that case (should we ?) ?

francoisferrand: This is not a standard notification indeed, but will not generate any notification anyway : since it will not be possible to register for this kind of notification.

The purpose of adding an `originOp` in that case would be to make it easier to see (in oplog or while troubleshooting) the "reason" for the metadata write operation; and it makes it easier to handle the case where we want to introduce a notification in that case (we can simply allow registering for such notification, or easily identification this case and set the proper id).

We could thus set `s3:ReplicateObject`, which would solve the issue just as well and not create any divergence form standard, since does not start with `s3:Repication:` and will thus not be matched by `s3:Replication:*`.

welansari (approved):

rachedbenmustapha (approved):

PR #2479, opened 2023-12-01
BB-473 Probe response logic should be handled in the handler 

nicolas2bert:
Currently, the probe response logic is distributed between Backbeat probe handlers and Arsenal's onRequest method.
This scattered approach causes confusion for developers and results in bugs.
The solution is to centralize the probe response logic exclusively within the Backbeat probe handlers.

### DIFF
--- a/lib/util/probe.js
+++ b/lib/util/probe.js
@@ -32,25 +32,58 @@ function startProbeServer(config, callback) {
     probeServer.start();
 }
 
-function sendError(res, log, error, optMessage) {
-    res.writeHead(error.code);
-    let message;
-    if (optMessage) {
-        message = optMessage;
-    } else {
-        message = error.description || '';
-    }
-    log.debug('sending back error response', { httpCode: error.code,
-        errorType: error.message,
-        error: message });
-    res.end(`${JSON.stringify({ errorType: error.message,
-        errorMessage: message })}\n`);
+/**
+ * Send an Error response with multiple errors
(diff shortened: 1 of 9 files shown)


rachedbenmustapha (approved):

lib/util/probe.js
```diff
@@ -32,30 +32,64 @@ function startProbeServer(config, callback) {
...
+ * @param {Logger} log - Werelogs instance for logging if you choose to
+ * @param {Object} errorMessages - error messages
+ * @return {undefined}
+ */
+function sendMultipleErrors(res, log, errorMessages) {
+    const messages = JSON.stringify(errorMessages);
+    const errorCode = 500;
+    log.error('sending back error response', {
+        httpCode: errorCode,
+        error: messages,
```
jonathan-gramain: It's not ideal to log JSON-serialized contents, as it will be serialized again so not very readable as is.

It's likely that for indexing purpose we want to keep the `error` field as a string.

Maybe we can separate this logic in two cases:
- if there is a single error, just log its message in `error`
- if there is more than one error, add `errors: errorMessages` to the log, and have `error` log something generic like `multiple errors (see "errors" field for details)`. Alternatively we may also concatenate error messages in a "human-readable" string but we may end up with a long error message, not necessarily very readable either.

That would be my recommendation but I leave it up to you.

jonathan-gramain (approved):

nicolas2bert:
@bert-e approve

PR #2485, opened 2023-12-08
Pause consumer on breakbeat trigger

welansari:
Pause consumer when breakbeat is triggered
    
The current implementation just skips the call to consume() when breakbeat is triggered, this will cause the consumer to be kicked out of the consumer group after the max poll delay is reached.

Polling is only done when calling consume(), hence we should keep calling that function and instead use the pause/resume methods that pause/resume the partition fetchers.

Pre-fetched messages are supposedly not purged from the internal queue, however they are not returned when calling consume() in a paused state.

The pause/resume state of the partitions is only kept locally, so this implementation is safe in case of forced shutdown or restart.

Issue: BB-456

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -150,6 +150,7 @@ class BackbeatConsumer extends EventEmitter {
 
         this._circuitBreaker = new CircuitBreaker(circuitBreaker);
         this._circuitBreakerMetrics = circuitBreakerMetrics;
+        this._circuitBreaker.on('state-changed', this._onCircuitBreakerStateChanged.bind(this));
 
         this._init();
         return this;
@@ -376,12 +377,6 @@ class BackbeatConsumer extends EventEmitter {
             this._tryConsumedTimeout = null;
         }
 
-        if (this._circuitBreaker.state !== BreakerState.Nominal) {
-            this._log.debug('circuitbreaker tripped, retry in 1s');
-            this._scheduleNextTryConsume();
-            return undefined;
-        }
-
         // use non-flowing mode of consumption to add some flow
         // control: explicit consumption of messages is required,
         // needs explicit polling to get new messages
@@ -536,6 +531,94 @@ class BackbeatConsumer extends EventEmitter {
     }
 
     /**
+     * Pauses consumption of assigned partitions
+     * @returns {undefined}
+     */
+    _pauseAssignments() {
+        try {
+            const assignments = this._consumer?.assignments();
+
+            if (!assignments || assignments.length === 0) {
+                return;
+            }
+
+            const topicPartitions = assignments?.map(assignment => ({
+                topic: assignment.topic,
+                partition: assignment.partition
+            }));
+
+            this._consumer.pause(topicPartitions);
+            this._log.info('circuitbreaker tripped, consumption paused', {
+                method: 'BackbeatConsumer._pauseAssignments',
+            });
+        } catch (error) {
+            this._log.error('failed to pause partitions', {
+                method: 'BackbeatConsumer._pauseAssignments',
+                error,
+            });
+        }
+    }
+
+    /**
+     * Resumes consumption of paused partitions
+     * @returns {undefined}
+     */
+    _resumePausedPartitions() {
+        try {
+            const assignments = this._consumer?.assignments();
+
+            if (!assignments || assignments.length === 0) {
+                return;
+            }
+
+            const topicPartitions = assignments?.map(assignment => ({
+                topic: assignment.topic,
+                partition: assignment.partition
+            }));
+
+            this._consumer.resume(topicPartitions);
+            this._log.info('Resumed consumption', {
+                method: 'BackbeatConsumer._resumePausedPartitions',
+                reason: this._circuitBreaker.state === BreakerState.Nominal ?
+                    'circuitbreaker state is nominal' : 'partitions unassigned',
+            });
+        } catch (error) {
+            this._log.error('failed to resume partitions', {
+                method: 'BackbeatConsumer._resumePausedPartitions',
+                error,
+            });
+        }
+    }
+
+    /**
+     * Circuit breaker state change handler
+     * Pauses consumption when circuit breaker is tripped
+     * @param {BreakerState} state circuit breaker state
+     * @returns {undefined}
+     */
+    _onCircuitBreakerStateChanged(state) {
+        switch (state) {
+        case BreakerState.Nominal:
+            this._resumePausedPartitions();
+            break;
+        case BreakerState.Tripped:
+            this._pauseAssignments();
+            break;
+        case BreakerState.Stabilizing:
+            // Nothing to do, wait for state
+            // to become nominal
+            break;
+        default:
+            // this should never happen
+            this._log.warn('unsupported circuit breaker state', {
+                method: 'BackbeatConsumer._onCircuitBreakerStateChanged',
+                state,
+            });
+            break;
+        }
+    }
+
+    /**
      * @param {kafka.KafkaError} err Rebalance event
      * @param {TopicPartition[]} assignment List of (un)assigned partitions
      * @returns {void}
@@ -546,6 +629,9 @@ class BackbeatConsumer extends EventEmitter {
 
             try {
                 this._consumer.assign(assignment);
+                if (this._circuitBreaker.state !== BreakerState.Nominal) {
+                    this._pauseAssignments();
+                }
             } catch (e) {
                 // Ignore exceptions if we are not connected
                 const logger = this._consumer.isConnected() ? this._log.error : this._log.debug;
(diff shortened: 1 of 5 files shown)


lib/BackbeatConsumer.js
```diff
@@ -938,7 +1024,7 @@ class BackbeatConsumer extends EventEmitter {
                 if (this._zookeeper) {
                     this._zookeeper.close();
                 }
-                if (this._consumer.isConnected()) {
+                if (this._consumer && this._consumer.isConnected()) {
```
francoisferrand: nit: `this._consumer?.isConnected()`

lib/BackbeatConsumer.js
```diff
@@ -920,7 +1006,7 @@ class BackbeatConsumer extends EventEmitter {
         this._circuitBreaker.stop();
         return async.waterfall([
             next => {
-                if (this._consumer.isConnected()) {
+                if (this._consumer && this._consumer.isConnected()) {
```
francoisferrand: nit: this._consumer?.isConnected()

lib/BackbeatConsumer.js
```diff
@@ -588,12 +670,16 @@ class BackbeatConsumer extends EventEmitter {
...
                         this._consumer.unassign();
+                        this._assignment = [];
                     } catch (e) {
                         // Ignore exceptions if we are not connected
                         const logger = this._consumer.isConnected() ? this._log.error : this._log.info;
                         logger.bind(this._log)('rdkafka.unassign failed', { e: e.toString(), assignment });
+                        return;
                     }
 
+                    this._resumePausedPartitions();
```
francoisferrand: should we not resume _before_ calling `consumer.unassign()`, to be consistent?

lib/BackbeatConsumer.js
```diff
@@ -546,11 +619,20 @@ class BackbeatConsumer extends EventEmitter {
...
                 const logger = this._consumer.isConnected() ? this._log.error : this._log.debug;
                 logger.bind(this._log)('rdkafka.assign failed', { e: e.toString(), assignment });
+                return;
+            }
+
+            if (this._circuitBreaker.state !== BreakerState.Nominal) {
+                this._pauseAssignments();
             }
+
+            this.emit('assign');
```
francoisferrand: this is not used?

lib/BackbeatConsumer.js
```diff
@@ -546,11 +619,20 @@ class BackbeatConsumer extends EventEmitter {
 
             try {
                 this._consumer.assign(assignment);
+                this._assignment = assignment;
```
francoisferrand: do we need to copy the assignment?
kafka consumer has an api which returns the current assigment already, maybe we can use it directly

lib/BackbeatConsumer.js
```diff
@@ -588,12 +670,16 @@ class BackbeatConsumer extends EventEmitter {
                 const doUnassign = () => {
                     try {
                         this._consumer.unassign();
+                        this._assignment = [];
                     } catch (e) {
                         // Ignore exceptions if we are not connected
                         const logger = this._consumer.isConnected() ? this._log.error : this._log.info;
                         logger.bind(this._log)('rdkafka.unassign failed', { e: e.toString(), assignment });
+                        return;
```
francoisferrand: for consistency, should we not resume in that case?

lib/BackbeatConsumer.js
```diff
@@ -148,10 +149,19 @@ class BackbeatConsumer extends EventEmitter {
...
+        /** @type {Map<string,Kafka.TopicPartition[]>} */
+        // Key format is "topic:partition"
+        this._pausedPartitions = new Map();
+
         this._circuitBreaker = new CircuitBreaker(circuitBreaker);
         this._circuitBreakerMetrics = circuitBreakerMetrics;
+        this._circuitBreaker.on('state-changed', this._onCircuitBreakerStateChanged.bind(this));
 
-        this._init();
+        if (init) {
```
francoisferrand: not fond of this flag, it hints that there is a design flaw here.

can we find a better to test (so we don't need this flag), and/or a different way to integrate the code so that the "pause/resume" logic can be unit tested separately (since this is what you try to test with this flag, I gather)?

francoisferrand: note: this (kind of ugly) hack may do the trick...

```
consumer = new BackbeatConsumer({bootstrap: true, ...});
consumer._bootstrapping = true;
```

francoisferrand: another apporach may be to introduce a wrapper/proxy for KafkaConsumer, which can change internal state depending on circuit breaker + wrap the assign/unassign methods to pause/resume when needed

lib/BackbeatConsumer.js
welansari: Since we always resume when unassigning, i think this is not needed ?

lib/BackbeatConsumer.js
welansari: Used by the tests as we need to wait for the partitions to be assigned before sending the circuit breaking event, but i guess we can wait by manually calling `consumer.assignements()`

KillianG (approved):

lib/BackbeatConsumer.js
```diff
@@ -550,7 +621,13 @@ class BackbeatConsumer extends EventEmitter {
                 // Ignore exceptions if we are not connected
                 const logger = this._consumer.isConnected() ? this._log.error : this._log.debug;
                 logger.bind(this._log)('rdkafka.assign failed', { e: e.toString(), assignment });
+                return;
+            }
+
+            if (this._circuitBreaker.state !== BreakerState.Nominal) {
```
francoisferrand: instead of having the `return`, this can be added in the `try` block to be more readable
(and may catch exceptions if `pause()` is called while disconnected, right after `assign()` succeeded?)

lib/BackbeatConsumer.js
francoisferrand: why do we need the `return` ?
does not seem related to this change...

lib/BackbeatConsumer.js
```diff
@@ -535,6 +540,72 @@ class BackbeatConsumer extends EventEmitter {
...
     }
 
+    /**
+     * Pauses consumption of assigned partitions
+     * @returns {undefined}
+     */
+    _pauseAssignments() {
+        const assignments = this._consumer?.assignments();
+        // removing offset from assignments
+        const topicPartitions = assignments?.map(({ topic, partition }) => ({ topic, partition }));
```
francoisferrand: nit: may be clearer by not using destructuring (the function looks like a "nop")

```suggestion
        const topicPartitions = assignments?.map(assignment => ({ topic: assignment.topic, partition: assignment.partition }));
```

lib/BackbeatConsumer.js
```diff
@@ -535,6 +540,72 @@ class BackbeatConsumer extends EventEmitter {
...
+        if (!topicPartitions || topicPartitions.length === 0) {
+            return;
+        }
+        this._consumer.pause(topicPartitions);
+        topicPartitions.forEach(({ topic, partition }) => {
+            this._pausedPartitions.set(`${topic}:${partition}`, { topic, partition });
```
francoisferrand: why do we keep track of the "paused" partitions in a separate variable: can't we just simply retrieve the (current) assingments every time?

lib/BackbeatConsumer.js
```diff
@@ -535,6 +540,72 @@ class BackbeatConsumer extends EventEmitter {
...
+
+    /**
+     * Resumes consumption of paused partitions
+     * @returns {undefined}
+     */
+    _resumePausedPartitions() {
+        if (this._pausedPartitions.size === 0) {
+            return;
+        }
+        const topicPartitions = Array.from(this._pausedPartitions.values());
```
francoisferrand: why do we store a map, forcing us to re-convert to array here?

tests/unit/backbeatConsumer.js
```diff
@@ -23,9 +33,169 @@ describe('backbeatConsumer', () => {
...
+                { topic: 'my-test-topic', partition: 0 },
+                { topic: 'my-test-topic', partition: 1 },
+                { topic: 'my-test-topic', partition: 2 },
+            ];
+            consumer._consumer.assignments = () => assignments;
+
+            consumer._pauseAssignments();
+            consumer._pausedPartitions.has('my-test-topic:0');
+            consumer._pausedPartitions.has('my-test-topic:1');
+            consumer._pausedPartitions.has('my-test-topic:2');
```
francoisferrand: not needed, this is internal state: the "exposed" state is tested already through the mock consumer

lib/BackbeatConsumer.js
welansari: Yeah it doesn't make sense to store it separately since we resume before unassigning. The current assignments  should always contain the full list of paused partitions.

francoisferrand (approved):

PR #2487, opened 2023-12-18
Fix late lifecycle scan alerts

francoisferrand:
- Fix lifecycle late scan alert firing
- Cleanup alert jobs
- Release 8.6.32

Issue: BB-478

### DIFF
--- a/monitoring/lifecycle/alerts.yaml
+++ b/monitoring/lifecycle/alerts.yaml
@@ -43,10 +43,14 @@ groups:
   - alert: LifecycleLateScanWarning
     Expr: |
       (
-        time()
-        - max(s3_lifecycle_latest_batch_start_time{
-             namespace="${namespace}", job="${job_lifecycle_producer}"
-           }) / 1000
+        time() - (
+          max(last_over_time(
+            s3_lifecycle_latest_batch_start_time{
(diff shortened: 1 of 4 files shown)


welansari (approved):

rachedbenmustapha (approved):

monitoring/lifecycle/alerts.yaml
```diff
@@ -43,10 +43,14 @@ groups:
...
-        - max(s3_lifecycle_latest_batch_start_time{
-             namespace="${namespace}", job="${job_lifecycle_producer}"
-           }) / 1000
+        time() - (
+          max(last_over_time(
+            s3_lifecycle_latest_batch_start_time{
+              namespace="${namespace}", job="${job_lifecycle_producer}"
+            }[${lifecycle_latency_warning_threshold}s]
+          ))
+          > 0 or max(kube_service_created{namespace="${namespace}", service="${job_lifecycle_producer}"})
```
jonathan-gramain, line 52: Sorry to come after the battle, but I'm not sure I understand why we do `> 0 or ...` (supposed to return a boolean value?), then `/ 1000` (should be applied to the `max()` only?)

PR #2489, opened 2024-01-09
Restart connectors on task failure

welansari:
Restart Kafka-Connect connectors when they are in a failed state or if their tasks are in a failed state.

Issue: BB-480

### DIFF
--- a/extensions/oplogPopulator/OplogPopulatorMetrics.js
+++ b/extensions/oplogPopulator/OplogPopulatorMetrics.js
@@ -58,6 +58,11 @@ class OplogPopulatorMetrics {
             help: 'Total number of connector configuration submissions to kafka-connect',
             labelNames: ['connector', 'success'],
         });
+        this.connectorRestarts = ZenkoMetrics.createCounter({
+            name: 's3_oplog_populator_connector_restarts',
+            help: 'Total number of connector restarts',
+            labelNames: ['connector'],
+        });
     }
 
     /**
@@ -169,6 +174,24 @@ class OplogPopulatorMetrics {
         } catch (error) {
             this._logger.error('An error occured while pushing metrics', {
                 method: 'OplogPopulatorMetrics.onConnectorReconfiguration',
+                error: error.message,
+            });
+        }
+    }
+
+    /**
+     * updates s3_oplog_populator_connector_restarts metric
+     * @param {string} connector connector name
+     * @returns {undefined}
+     */
+    onConnectorRestart(connector) {
+        try {
+            this.connectorRestarts.inc({
+                connector: connector.name,
+            });
+        } catch (error) {
+            this._logger.error('An error occured while pushing metric', {
+                method: 'OplogPopulatorMetrics.onConnectorRestarted',
                 error: error.message,
             });
         }
--- a/extensions/oplogPopulator/modules/Connector.js
+++ b/extensions/oplogPopulator/modules/Connector.js
@@ -199,6 +199,32 @@ class Connector {
     }
 
     /**
+     * Restarts the Kafka-connect mongo connector
+     * @returns {Promise|undefined} undefined
+     * @throws {InternalError}
+     */
+    async restart() {
+        if (!this._isRunning) {
+            this._logger.error('tried restarting a destroyed connector', {
+                method: 'Connector.restart',
+                connector: this._name,
+            });
+            return;
+        }
+        try {
+            // only restarting failed instances of tasks and connector
+            await this._kafkaConnect.restartConnector(this._name, true, true);
+        } catch (err) {
+            this._logger.error('Error while restarting connector', {
+                method: 'Connector.restart',
+                connector: this._name,
+                error: err.description || err.message,
+            });
+            throw errors.InternalError.customizeDescription(err.description);
+        }
+    }
+
+    /**
      * Add bucket to this connector
      * Connector is updated with the new bucket list
      * @param {string} bucket bucket to add
--- a/extensions/oplogPopulator/modules/ConnectorsManager.js
+++ b/extensions/oplogPopulator/modules/ConnectorsManager.js
@@ -274,6 +274,9 @@ class ConnectorsManager {
                 updated: null,
             };
             try {
+                // check if connector is in a failed state and restart it
+                await this._validateConnectorState(connector);
+
                 // check if we need to spawn/despawn the connector
                 // - connector is destroyed if no buckets are configured
                 // - connector is spawned when buckets are configured on it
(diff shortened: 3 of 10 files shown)


extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -58,6 +58,11 @@ class OplogPopulatorMetrics {
             help: 'Total number of connector configuration submissions to kafka-connect',
             labelNames: ['connector', 'success'],
         });
+        this.connectorRestarts = ZenkoMetrics.createCounter({
+            name: 's3_oplog_populator_connector_restarts',
+            help: 'Total number of connector restarts',
+            labelNames: ['connector'],
```
francoisferrand, line 64: do we need the connector name in metrics?

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -296,14 +291,59 @@ class ConnectorsManager {
...
+            } catch (err) {
+                this._logger.error('Could not check or reset connector state', {
+                    method: 'ConnectorsManager._validateConnectorsState',
+                    connector: connector.name,
+                    error: err.description || err.message,
+                });
+            }
         });
+        if (restartedConnectors.length > 0) {
+            this._logger.info('Successfully restarted connectors', {
```
francoisferrand: restart should seldom happen : may be simpler to just log each restart on success (e.g. line 331)

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -312,6 +352,7 @@ class ConnectorsManager {
      */
     scheduleConnectorUpdates() {
         schedule.scheduleJob(this._cronRule, async () => {
+            await this._validateConnectorsState();
```
francoisferrand: does it make sense to check the status of a connector we update? (both just before or after the update)
→ maybe we should only check for restart in the `else` case of `updateConnectors` (line 283)?

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -312,6 +352,7 @@ class ConnectorsManager {
      */
     scheduleConnectorUpdates() {
         schedule.scheduleJob(this._cronRule, async () => {
```
francoisferrand: should `restarts` be checked on the same schedule as updates?
1 api call per second is not huge (if the api is indeed lightweight), but at the same time we certainly can tolerate more delay before restarting (and maybe best to leave some time for kafka-connect to try and recover by itself, or for previous restart command to "complete")

monitoring/oplog-populator/alerts.yaml
```diff
@@ -51,3 +51,14 @@ groups:
       description: "Oplog populator metastore change stream lag is too big"
       summary: "Oplog populator configuration lag is above threshold"
 
+  - alert: KafkaConnectFailedConnectors
+    Expr: |
+        sum by(connector) (increase(s3_oplog_populator_connector_restarts{job="${oplog_populator_job}",namespace="${namespace}"}[5m]))
+        > 0
+    For: "1m"
```
francoisferrand: timing should probably be the other way:

```suggestion
    Expr: |
        sum by(connector) (increase(s3_oplog_populator_connector_restarts{job="${oplog_populator_job}",namespace="${namespace}"}[1m]))
        > 0
    For: "5m"
```

→ check that we restart every minute in the last 5 minutes (instead of checking for 1min that we restarted in the last 5 min)

* the `1m` time range should match the periodicity of the restart check
* not sure what duration to use for the `5m` grace period of the alert (`For: 5m`) : is 5m enough? should we increase to 10m to avoid 'flaky' alerts?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -198,6 +198,32 @@ class Connector {
...
+                method: 'Connector.restart',
+                connector: this._name,
+            });
+            return;
+        }
+        try {
+            // only restarting failed instances of tasks and connector
+            await this._kafkaConnect.restartConnector(this._name, true, true);
+        } catch (err) {
+            this._logger.error('Error while restarting connector', {
```
francoisferrand, line 218: missing test for error handling

extensions/oplogPopulator/modules/ConnectorsManager.js
welansari: Checking each connector state before updating it makes more sense yes, but i think it should be done systematically, no just when we don't update the connector

extensions/oplogPopulator/modules/ConnectorsManager.js
welansari: I think the connectors should be restarted as soon as possible.
We want to avoid letting the connector updates be what restarts the connectors as we might loose events when buckets have been removed from the pipeline.

francoisferrand (approved):

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -296,14 +294,56 @@ class ConnectorsManager {
...
+    async _validateConnectorState(connector) {
+        if (!connector.isRunning) {
+            return;
+        }
+
+        try {
+            const connectorStatus = await this._kafkaConnect.getConnectorStatus(connector.name);
+            const isConnectorFailed = connectorStatus?.connector?.state === 'FAILED';
+            const areTasksFailed = connectorStatus?.tasks?.some(task => {
+                if (task.state === 'FAILED') {
```
francoisferrand, line 320: what is the state during "startup"?
is there no "intermediate" state where we could restart the task inappropriately?

welansari: Doesn't look like there is an intermediate state, the only states possible according to [kafka connect docs](https://docs.confluent.io/platform/current/connect/monitoring.html#:~:text=The%20following%20states%20are%20possible,task%20has%20been%20administratively%20paused.) are : 

- UNASSIGNED: The connector/task has not yet been assigned to a worker.
- RUNNING: The connector/task is running.
- PAUSED: The connector/task has been administratively paused.
- FAILED: The connector/task has failed (usually by raising an exception, which is reported in the status output).

tests/unit/oplogPopulator/ConnectorsManager.js
```diff
@@ -299,5 +304,76 @@ describe('ConnectorsManager', () => {
...
+                            state: 'RUNNING',
+                        },
+                    ]
+                });
+            connector1._isRunning = true;
+            await connectorsManager._validateConnectorState(connector1);
+            assert(getStatusStub.called);
+            assert(connectorRestartStub.notCalled);
+        });
+        it('should do nothing when connector is not spawned', async () => {
```
KillianG: nit
```suggestion
        });

        it('should do nothing when connector is not spawned', async () => {
```

KillianG (approved):

PR #2490, opened 2024-01-09
BB-484 update error handler when GC object is not found 

KillianG:
When we want to GC an object, if this one is not found, it means that it already has been deleted, so there is not point continuing to try deleting it, changeing the behavior to  make this specific error commitable

### DIFF
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -267,7 +267,11 @@ class GarbageCollectorTask extends BackbeatTask {
             },
         ], err => {
             if (err) {
-                // if error occurs, do not commit offset
+                // if error occurs, do not commit offset unless the error is ObjNotFound
+                // because it means the object has been deleted by other means and we don't need to retry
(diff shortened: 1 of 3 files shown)


extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -267,7 +267,11 @@ class GarbageCollectorTask extends BackbeatTask {
             },
         ], err => {
             if (err) {
-                // if error occurs, do not commit offset
+                // if error occurs, do not commit offset unless the error is ObjNotFound
+                // because it means the object has been deleted by other means and we don't need to retry
+                if (err.code === 'ObjNotFound') {
+                    return done(err, { committable: true });
```
welansari, line 273: Not sure we should return the error here ?

KillianG: The fact that the error is committable doesnt change the fact that there is an error, i think we should still return it, to log it etc..

welansari (approved):

francoisferrand (approved):

PR #2497, opened 2024-02-12
fix double notifications when having multiple destinations using the …

welansari:
…shared internal topic

When having multiple notification destinations, Kafka messages for the notification are pushed as many times as we have destinations into the same shared internal topic.

Issue: BB-500

### DIFF
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -125,8 +125,16 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                 key,
                 eventType,
             });
+            const pushedToTopic = new Map();
             // validate and push kafka message foreach destination topic
             this.notificationConfig.destinations.forEach(destination => {
+                const topic = destination.internalTopic ||
+                    this.notificationConfig.topic;
+                // avoid pushing a message multiple times to the
+                // same internal topic
+                if (pushedToTopic[topic]) {
+                    return undefined;
+                }
                 // get destination specific notification config
                 const destBnConf = config.queueConfig.find(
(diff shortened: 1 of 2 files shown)


extensions/notification/NotificationQueuePopulator.js
```diff
@@ -125,8 +125,14 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                 key,
                 eventType,
             });
+            let wasPushedToSharedTopic = false;
             // validate and push kafka message foreach destination topic
             this.notificationConfig.destinations.forEach(destination => {
+                // avoid pushing the same message multiple times to the
+                // internal shared topic
```
francoisferrand: This may be a partial fix: eg if the same "external" topic is specified in multiple targets' config
→ either we need to handle that, or to detect/prevent (or at least document) this case/limitation

welansari: that's easy to handle, better do that

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -125,8 +125,16 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                 key,
                 eventType,
             });
+            const pushedToTopic = new Map();
             // validate and push kafka message foreach destination topic
             this.notificationConfig.destinations.forEach(destination => {
+                const internalTopic = destination.internalTopic ||
```
francoisferrand: nit: this variable is not always "internal" topic, but really just the topic : so may be renamed...

francoisferrand (approved):

anurag4DSB (approved):

PR #2499, opened 2024-02-20
Add accountID to GC message

KillianG:
Issue: BB-503

### DIFF
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -23,16 +23,16 @@ class GarbageCollectorTask extends BackbeatTask {
     // account id information is missing.
     // TODO BB-367: replace once replication services uses assume role
     _getAccountId(entry, log, cb) {
-        const { accountId, ownerId } = entry.getAttribute('target');
+        const { accountId, owner } = entry.getAttribute('target');
 
         if (accountId) {
             return process.nextTick(cb, null, accountId);
         }
 
         log.debug('unable to find account id in entry; performing vault request', {
-            ownerId,
+            owner,
         });
-        return this.getAccountId(ownerId, log, cb);
+        return this.getAccountId(owner, log, cb);
     }
 
     _getMetadata(entry, log, done) {
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -271,7 +271,7 @@ class GarbageCollectorTask extends BackbeatTask {
             if (err) {
                 // if error occurs, do not commit offset unless the error is ObjNotFound
                 // because it means the object has been deleted by other means and we don't need to retry
-                if (err.code === 'ObjNotFound') {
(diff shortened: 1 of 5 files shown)


welansari (approved):

extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
```diff
@@ -126,7 +126,8 @@ class LifecycleUpdateExpirationTask extends BackbeatTask {
               })
               .setAttribute('source', entry.getAttribute('source'))
               .setAttribute('serviceName', 'lifecycle-transition')
-              .setAttribute('target.locations', locations);
+              .setAttribute('target.locations', locations)
+              .setAttribute('target.accountId', accountId);
```
francoisferrand: should also (try to) pass the `ownerId`, for good measure (like in the other cases)

tests/unit/gc/GarbageCollectorTask.spec.js
```diff
@@ -199,4 +199,33 @@ describe('GarbageCollectorTask', () => {
...
+                  versionId: version,
+              })
+              .setAttribute('serviceName', 'lifecycle-transition')
+              .setAttribute('target.oldLocation', 'old-location')
+              .setAttribute('target.newLocation', 'new-location')
+              .setAttribute('target.bucket', bucket)
+              .setAttribute('target.key', version)
+              .setAttribute('target.version', key)
+              .setAttribute('target.accountId', accountId)
+              .setAttribute('target.owner', owner);
```
francoisferrand, line 221: should it not be `target.ownerId` ?

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -36,7 +37,7 @@ class LifecycleColdStatusArchiveTask extends LifecycleUpdateTransitionTask {
               .setAttribute('target.key', key)
               .setAttribute('target.version', version)
               .setAttribute('target.accountId', accountId)
-              .setAttribute('target.owner', owner);
+              .setAttribute('target.ownerId', ownerId);
```
francoisferrand: this is a partial fix only.

Looking at backbeat (lifecycle) overall, there are _lots_ places where we set `target.owner` field : so either we should change them all, or change GC to extract the `owner` field from message.

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -23,16 +23,16 @@ class GarbageCollectorTask extends BackbeatTask {
     // account id information is missing.
     // TODO BB-367: replace once replication services uses assume role
     _getAccountId(entry, log, cb) {
-        const { accountId, ownerId } = entry.getAttribute('target');
+        const { accountId, owner } = entry.getAttribute('target');
```
francoisferrand, line 26: also needs to be changed in https://github.com/scality/backbeat/blob/c790d6de0e52b1ba701dd1fe7103362fb97ed093/extensions/replication/tasks/UpdateReplicationStatus.js#L279

francoisferrand (approved):

PR #2500, opened 2024-02-21
Support per workflow circuit breaking in lifecycle

welansari:
Issue: BB-492

### DIFF
--- a/extensions/lifecycle/CircuitBreakerGroup.js
+++ b/extensions/lifecycle/CircuitBreakerGroup.js
@@ -1,0 +1,286 @@
+'use-strict';
+/* eslint no-template-curly-in-string: 0 */
+
+const { CircuitBreaker, BreakerState } = require('breakbeat').CircuitBreaker;
+const { startCircuitBreakerMetricsExport } = require('../../lib/CircuitBreaker');
+
+/**
+ * circuit breaker configuration adapted for the bucket processor
+ * @typedef {Object} BucketProcessorCircuitBreakerConfig
+ * @property {Object} global - global circuit breaker configurations (circuit breaks everything)
+ * @property {CircuitBreakerGroup} circuitBreakerGroup - circuit breakers per workflow and location/topic
+ */
+
+class CircuitBreakerGroup {
+    constructor(cbGlobalConf, circuitBreakers) {
+        if (circuitBreakers) {
+            this.circuitBreakers = circuitBreakers;
+            return;
+        }
+
+        this._cbGlobalConf = cbGlobalConf;
+
+        // breaker id used for the metrics
+        this._latestBreakerId = 0;
+
+        this.circuitBreakers = {
+            transition: {
+                topic: {},
+                location: {},
+                global: [],
+            },
+            expiration: {
+                topic: {},
+                location: {},
+                global: [],
+            },
+        };
+    }
+
+    _append(array, element) {
+        if (!array) {
+            return [element];
+        }
+        return [...array, element];
+    }
+
+    _addWorkflowCircuitBreaker(circuitBreaker, workflowType, topic, location) {
+        if (!topic && !location) {
+            this.circuitBreakers[workflowType].global.push(circuitBreaker);
+        }
+        if (topic) {
+            this.circuitBreakers[workflowType].topic[topic] =
+                this._append(this.circuitBreakers[workflowType].topic[topic], circuitBreaker);
+        }
+        if (location) {
+            this.circuitBreakers[workflowType].location[location] =
+                this._append(this.circuitBreakers[workflowType].location[location], circuitBreaker);
+        }
+    }
+
+    exportCircuitBreakerMetric(circuitBreaker, metricsSuffix) {
+        let metricsName = `lifecycle_bucket_processor_${this._latestBreakerId}`;
+        if (metricsSuffix) {
+            metricsName += `_${metricsSuffix}`;
+        }
+        this._latestBreakerId += 1;
+        startCircuitBreakerMetricsExport(
+            circuitBreaker,
+            metricsName
+        );
+    }
+
+    addCircuitBreaker(probeConf, query, isTransition, isExpiration, topic, location, metricsSuffix) {
+        const circuitBreaker = new CircuitBreaker({
+            ...this._cbGlobalConf,
+            probes: [{
+                ...probeConf,
+                query,
+            }],
+        });
+        circuitBreaker.start();
+        this.exportCircuitBreakerMetric(circuitBreaker, metricsSuffix);
+        if (isTransition) {
+            this._addWorkflowCircuitBreaker(circuitBreaker, 'transition', topic, location);
+        }
+        if (isExpiration) {
+            this._addWorkflowCircuitBreaker(circuitBreaker, 'expiration', topic, location);
+        }
+    }
+
+    /**
+     * Checks if circuit breakers appropriate to our action are triggered
+     * @param {string} worfklowType transition or expiration
+     * @param {string} location location name of object
+     * @param {string} topic topic we want to push to
+     * @returns {bool} true if it should circuit break
+     */
+    tripped(worfklowType, location, topic) {
+        const breakers = [];
+        ['expiration', 'transition'].forEach(type => {
+            if (worfklowType && type !== worfklowType) {
+                return;
+            }
+            if (location && this.circuitBreakers[type].location[location]) {
+                breakers.push(
+                    ...this.circuitBreakers[type].location[location],
+                );
+            }
+            if (topic && this.circuitBreakers[type].topic[topic]) {
+                breakers.push(
+                    ...this.circuitBreakers[type].topic[topic],
+                );
+            }
+            breakers.push(
+                ...this.circuitBreakers[type].global,
+            );
+        });
+        return breakers.some(breaker => breaker.state !== BreakerState.Nominal);
+    }
+}
+
+/**
+ * returns a list of relevant topics to the bucket processor
+ * @param {Object} lcConfig lifecycle configuration
+ * @param {Object} repConfig replication configuration
+ * @param {Object} locations  location configuration
+ * @returns {string[]} list of topics
+ */
+function getAllTopics(lcConfig, repConfig, locations) {
+    const topics = [
+        lcConfig.objectTasksTopic,
+        repConfig.dataMoverTopic,
+    ];
+    Object.keys(locations).forEach(locationName => {
+        if (locations[locationName].isCold) {
+            topics.push(lcConfig.coldStorageArchiveTopicPrefix + locationName);
+        }
+    });
+    return topics;
+}
+
+/**
+ * Adapts the templated circuit breaker configuration to the bucket processor
+ * Invalid probes are ignored
+ * @param {Object} cbConf templated circuit breaker configuration
+ * @param {Object} lcConfig lifecycle configuration
+ * @param {Object} repConfig replication configuration
+ * @param {Object} locations location configuration
+ * @param {Logger} logger logger instance
+ * @returns {BucketProcessorCircuitBreakerConfig} circuit breaker configuration
+ */
+function extractBucketProcessorCircuitBreakerConfigs(cbConf, lcConfig, repConfig, locations, logger) {
+    if (!cbConf || !cbConf.probes) {
+        return {
+            global: {},
+            circuitBreakerGroup: new CircuitBreakerGroup({}, null),
+        };
+    }
+
+    const { probes, ...globalCircuitBreakerConf } = cbConf;
+
+    const circuitBreakerGroup = new CircuitBreakerGroup(globalCircuitBreakerConf, null);
+
+    // config off global circuit breaker
+    const global = {
+        ...globalCircuitBreakerConf,
+        probes: [],
+    };
+
+    probes.forEach(probe => {
+        if (probe.type !== 'prometheusQuery') {
+            global.probes.push(probe);
+            return;
+        }
+        // when clause is optional and can be used to specify the filtering on
+        // the topic and location, along with specifying if a probe should be
+        // used for transition and expiration.
+        // Query syntax:
+        // "when({topic=topicName,location=locationName,transition=true,expiration=true}) and ANY"
+        // Note: should either specify the topic or the location, not both
+        // Behaviour:
+        // - no when clause + templated topic: creates a circuit breaker for each topic
+        // - no when clause + templated location: creates a circuit breaker for each location
+        // - when clause with a specific topic + templated topic: creates a circuit breaker for the specified topic
+        // - when clause with a specific location + templated location: creates a circuit breaker for the location
+        // - when clause with topic and location + topic and location template: creates a circuit breaker for the
+        // location and topic
+        // - topic specified + location template : creates a circuit breaker for each location and references them in
+        // the topic
+        // - location specified + topic template : creates a circuit breaker for each topic and references them in
+        // the location
+        const whenClause = probe.query.match(/^when\s?\(\{(.*?)\}\)\sand\s/);
+
+        const topic = whenClause && whenClause[1].match(/topic="(.*?)"(,|$)/);
+        const location = whenClause && whenClause[1].match(/location="(.*?)"(,|$)/);
+
+        const query = whenClause ? probe.query.replace(whenClause[0], '') : probe.query;
+
+        const withTopicTemplate = query.includes('${topic}');
+        const withLocationTemplate = query.includes('${location}');
+
+        const transition = whenClause && whenClause[1].match(/transition="(true|false)"(,|$)/);
+        const expiration = whenClause && whenClause[1].match(/expiration="(true|false)"(,|$)/);
+        // By default all probes are used for both transition and expiration
+        const useForTransition = (transition && transition[1] === 'true') || !transition;
+        const useForExpiration = (expiration && expiration[1] === 'true') || !expiration;
+
+        if (!useForExpiration && !useForTransition) {
+            logger.error('Invalid circuit breaker probe config', {
+                reason: 'both lifecycle workflows disabled',
+                method: 'extractBucketProcessorCircuitBreakerConfigs',
+                query: probe.query,
+            });
+            return;
+        }
+
+        if (withTopicTemplate && withLocationTemplate) {
+            logger.error('Invalid circuit breaker probe config', {
+                reason: 'using multiple unspecified templates is not supported',
+                method: 'extractBucketProcessorCircuitBreakerConfigs',
+                query: probe.query,
+            });
+            return;
+        }
+
+        if (!withTopicTemplate && !withLocationTemplate) {
+            if (!whenClause) {
+                global.probes.push(probe);
+                return;
+            }
+            circuitBreakerGroup.addCircuitBreaker(
+                probe,
+                query,
+                useForTransition,
+                useForExpiration,
+                topic ? topic[1] : '',
+                location ? location[1] : '',
+                '',
+            );
+            return;
+        }
+
+        const topics = topic ? [topic[1]] : getAllTopics(lcConfig, repConfig, locations);
+        const locationNames = location ? [location[1]] : Object.keys(locations);
+
+        if (withTopicTemplate) {
+            topics.forEach(tp => {
+                circuitBreakerGroup.addCircuitBreaker(
+                    probe,
+                    query.replace('${topic}', tp).replace('${location}', location ? location[1] : ''),
+                    useForTransition,
+                    useForExpiration,
+                    tp,
+                    location ? location[1] : '',
+                    tp,
+                );
+            });
+            return;
+        }
+
+        if (withLocationTemplate) {
+            locationNames.forEach(loc => {
+                circuitBreakerGroup.addCircuitBreaker(
+                    probe,
+                    query.replace('${location}', loc).replace('${topic}', topic ? topic[1] : ''),
+                    useForTransition,
+                    useForExpiration,
+                    topic ? topic[1] : '',
+                    loc,
+                    loc,
+                );
+            });
+            return;
+        }
+    });
+
+    return {
+        global,
+        circuitBreakerGroup,
+    };
+}
+
+module.exports = {
+    extractBucketProcessorCircuitBreakerConfigs,
+    CircuitBreakerGroup,
+};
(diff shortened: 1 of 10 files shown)


lib/CircuitBreaker.js
```diff
@@ -5,6 +5,66 @@ const { ZenkoMetrics } = require('arsenal').metrics;
...
+ * Expiration specific probes should follow this format "&type=expiration;&query=XXX" (order is important)
+ * Transition specific probes should follow this format "&type=transition;&location=YYY;&query=XXX" (order is important)
+ * Probes that have no &type are considered default and should only contain normal queries
+ * Invalid probes are ignored
+ * @param {Object} cbConf templated circuit breaker configuration
+ * @returns {BucketProcessorCircuitBreakerConfig} circuit breaker configuration
+ */
+function extractBucketProcessorCircuitBreakerConfigs(cbConf) {
+    const configs = {
+        default: [],
```
francoisferrand: `global` may be better `default`

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+        // If no topic template is used the probe is added as is to the topic probes
+        if (topic || withTopicTemplate) {
+            const topics = topic ? [topic[1]] : getAllTopics(lcConfig, repConfig, locations);
+            topics.forEach(topic => {
+                const probeConfig = {
+                    config: {
+                        ...probe,
+                    },
+                    transition: isTransition,
+                    expiration: isExpiration,
```
francoisferrand: I think it would be easier to read if the `transition` and `expiration` were renamed to better indicate that it means _"rule is enabled for transition"_

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+    cbConf.probes.forEach(probe => {
+        if (probe.type !== 'prometheusQuery') {
+            configs.global.push(probe);
+            return;
+        }
+        // with clause is optional and can be used to specify the filtering on
+        // the topic and location, along with specifying if a probe should be
+        // used for transition and expiration.
+        // Query syntax:
+        // "with(topic=topicName,location=locationName,transition=true,expiration=true) and
```
francoisferrand: `with(foo=bar)` is not very prometheus like: the "label filter expression" is only supported in braces... (and it may eventually allow supporting the other operator : `~`, `!=`, `!~`)

→ `when({foo="bar"})` (actually, `when` may be more appropriate than `with` here)

(there may be alternative approaches: using a "fake" metric name instead of function `when{}` or even no metric name, but this is all pretty similar: just maybe to see what fits best with the `and` operator)

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+        //      PROM_METRIC_NAME{topic='${topic}',location='${location}'}"
+        // Note: should either specify the topic or the location, not both
+        // Behaviour:
+        // - no with clause + templated topic: creates a circuit breaker for each topic
```
francoisferrand: + same for location

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+        // - with clause with a specific topic + templated topic: creates a circuit breaker for the specified topic
+        // - with clause with a specific location + templated location: creates a circuit breaker for the location
```
francoisferrand: can we (do we need?) to specify a _location_ filter (location=dmf) & templated _topic_ expression?
(e.g. i want to a specific object-topic lag limit for one location)

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+        if (probe.type !== 'prometheusQuery') {
+            configs.global.push(probe);
+            return;
+        }
+        // with clause is optional and can be used to specify the filtering on
+        // the topic and location, along with specifying if a probe should be
+        // used for transition and expiration.
+        // Query syntax:
+        // "with(topic=topicName,location=locationName,transition=true,expiration=true) and
+        //      PROM_METRIC_NAME{topic='${topic}',location='${location}'}"
```
francoisferrand: nit: the "rest" of the expression may be much more than one metric name, it may be anyy (very complex)

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+
+        const withTopicTemplate = probe.query.includes('${topic}');
+        const withLocationTemplate = probe.query.includes('${location}');
+
+        const topic = withClause && withClause[1].match(/topic=(.*?)(,|$)/);
+        const location = withClause && withClause[1].match(/location=(.*?)(,|$)/);
+
+        // Ignore ambiguous configuration
+        if (
+            (topic && location) ||
```
francoisferrand: is this one really an issue?

lib/CircuitBreaker.js
francoisferrand: (I wonder if it would not be simpler to just create as many CircuitBreaker as there are rules (one query = one circuit breaker x location|topic), then go through the "whole" list to consider only which one apply... to possibly optimize (?) or at least simplify the code, we may have a list of CircuitBreaker+matcher function, and dynamically create the matcher functions as needed on startup ; and/or we can pre-filter/duplicate the list to reference the same breakers by workflow/topic/location)

@rachedbenmustapha is there an extra cost to having multiple CircuitBreaker instances, vs a single one with multiple queries ?

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+        // used for transition and expiration.
+        // Query syntax:
+        // "with(topic=topicName,location=locationName,transition=true,expiration=true) and
+        //      PROM_METRIC_NAME{topic='${topic}',location='${location}'}"
+        // Note: should either specify the topic or the location, not both
+        // Behaviour:
+        // - no with clause + templated topic: creates a circuit breaker for each topic
+        // - with clause with a specific topic + templated topic: creates a circuit breaker for the specified topic
+        // - with clause with a specific location + templated location: creates a circuit breaker for the location
+        const withClause = probe.query.match(/with\((.*?)\)\sand\s/);
```
francoisferrand: should we match at the beginning only, to further limit risk?

```suggestion
        const withClause = probe.query.match(/^with\((.*?)\)\sand\s/);
```

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+
+        const withTopicTemplate = probe.query.includes('${topic}');
+        const withLocationTemplate = probe.query.includes('${location}');
+
+        const topic = withClause && withClause[1].match(/topic=(.*?)(,|$)/);
+        const location = withClause && withClause[1].match(/location=(.*?)(,|$)/);
```
francoisferrand: ideally we would want to properly parse/validate the "content" of the withClause : i.e.
* split by comma to get the individual "filters"
* check each filter key is valid : topic, location, transition, expiration
* check the operator (`=` for now)

though it may do for now, kind of in a pinch... can you create a ticket nonetheless to address this?

lib/CircuitBreaker.js
rachedbenmustapha: Each breaker is pretty low-overhead in itself, so it shouldn't change much.

What would have an impact for future-proofing (for example if we implement breaker state caching later on) would be having prom queries that yield consistent results and that are as reusable as possible at runtime.

Another aspect (though not here specifically) would be query performance: keeping queries scope narrow to avoid expensive computation, if we go with lots of queries.

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
 'use strict'; // eslint-disable-line
 /* eslint no-param-reassign: 0 */
+/* eslint no-template-curly-in-string: 0 */
 
 const { ZenkoMetrics } = require('arsenal').metrics;
 
 const collectDefaultMetricsIntervalMs = 10000;
 
+/**
+ * circuit breaker configuration adapted for the bucket processor
```
rachedbenmustapha: Since this file not does not contain any extension-specific code, I'd recommend moving this functionality to `extensions/lifecycle`.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -161,10 +164,32 @@ class LifecycleTask extends BackbeatTask {
...
+            return;
+        }
+
+        const topicCircuitBreaker = this.perTopicCircuitBreakers[this.objectTasksTopic];
+        if (
+            topicCircuitBreaker &&
+            topicCircuitBreaker.expiration &&
+            topicCircuitBreaker.circuitBreaker.state !== BreakerState.Nominal
+        ) {
+            cb(errorLocationPaused);
```
rachedbenmustapha: This error might be confusing or surprising in the logs or metrics.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -29,6 +30,8 @@ const errorReplicationInProgress = errors.InternalError.
     customizeDescription('replication of the object is currently in progress');
 const errorLocationPaused = errors.InternalError.
     customizeDescription('lifecycle events to location have been paused');
+const errorCircuitBreakerTripped = errors.InternalError.
```
rachedbenmustapha: `InternalError` seems like it will cause some false data, `errors.Throttling` would be more accurate.

lib/CircuitBreaker.js
```diff
@@ -1,10 +1,150 @@
...
+        const location = withClause && withClause[1].match(/location=(.*?)(,|$)/);
+
+        // Ignore ambiguous configuration
+        if (
+            (topic && location) ||
+            (withTopicTemplate && withLocationTemplate) ||
+            (location && withTopicTemplate) ||
+            (topic && withLocationTemplate)
+        ) {
+            return;
```
rachedbenmustapha: If ignoring, I'd crash the process or at least print a loud log error.

rachedbenmustapha (commented):
Not sure how/if exposing status for each new breaker is planned, as breaker metrics are automatically exported by `BackbeatConsumer` and a few other places - with the new ones the export loop would need to be started manually.

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,315 @@
...
+            expiration: {
+                topic: {},
+                location: {},
+                global: [],
+            },
+        };
+    }
+
+    addTransitionCircuitBreaker(circuitBreaker, topic, location) {
+        if (!topic && ! location) {
```
rachedbenmustapha:
```suggestion
        if (!topic && !location) {
```

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,315 @@
...
+                [...this.circuitBreakers.transition.topic[topic], circuitBreaker] : [circuitBreaker];
+        }
+        if (location) {
+            this.circuitBreakers.transition.location[location] =  this.circuitBreakers.transition.location[location] ?
+                [...this.circuitBreakers.transition.location[location], circuitBreaker] : [circuitBreaker];
+        }
+    }
+
+    addExpirationCircuitBreaker(circuitBreaker, topic, location) {
+        if (!topic && ! location) {
```
rachedbenmustapha:
```suggestion
        if (!topic && !location) {
```

rachedbenmustapha (approved):

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,317 @@
...
+        };
+    }
+
+    addTransitionCircuitBreaker(circuitBreaker, topic, location) {
+        if (!topic && !location) {
+            this.circuitBreakers.transition.global.push(circuitBreaker);
+        }
+        if (topic) {
+            this.circuitBreakers.transition.topic[topic] =  this.circuitBreakers.transition.topic[topic] ?
+                [...this.circuitBreakers.transition.topic[topic], circuitBreaker] : [circuitBreaker];
```
francoisferrand: nit: could introduce an `_appendCb(list, cb)` method to simplify this

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,317 @@
...
+            },
+            expiration: {
+                topic: {},
+                location: {},
+                global: [],
+            },
+        };
+    }
+
+    addTransitionCircuitBreaker(circuitBreaker, topic, location) {
```
francoisferrand: nit: "internal" methods should start with an `_` (i.e. most methods except `addCircuitBreaker`)

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,317 @@
...
+                );
+            }
+            if (topic && this.circuitBreakers.expiration.topic[topic]) {
+                breakers.push(
+                    ...this.circuitBreakers.expiration.topic[topic],
+                );
+            }
+            breakers.push(
+                ...this.circuitBreakers.expiration.global,
+            );
```
francoisferrand: nit: this could be deduplicated (same thing on `.transition` & `.expiration`, either with a loop or a function

tests/unit/lifecycle/CircuitBreakerGroup.spec.js
```diff
@@ -0,0 +1,1297 @@
...
+        query: 'when{topic=cold-archive-req-location-dmf-v1} and s3_sorbet_is_throttled{"location"="${location}"}',
+        threshold: 100,
+        prometheus: {
+            endpoint: 'http://prometheus:9090',
+        },
+    };
+
+    const locationTemplatedProbe = {
+        type: 'prometheusQuery',
+        query: 's3_sorbet_is_throttled{"location"="${location}"}',
```
francoisferrand: typo in query:

```suggestion
        query: 's3_sorbet_is_throttled{location="${location}"}',
```

tests/unit/lifecycle/CircuitBreakerGroup.spec.js
```diff
@@ -0,0 +1,1297 @@
...
+        query: 'when{topic=cold-archive-req-location-dmf-v1} and kafka_consumergroup_group_lag{}',
+        threshold: 100,
+        prometheus: {
+            endpoint: 'http://prometheus:9090',
+        },
+    };
+
+    const topicSpecificLocationTemplateProbe = {
+        type: 'prometheusQuery',
+        query: 'when{topic=cold-archive-req-location-dmf-v1} and s3_sorbet_is_throttled{"location"="${location}"}',
```
francoisferrand: typo in query:

```suggestion
        query: 'when{topic="cold-archive-req-location-dmf-v1"} and s3_sorbet_is_throttled{location="${location}"}',
```

tests/unit/lifecycle/CircuitBreakerGroup.spec.js
```diff
@@ -0,0 +1,1297 @@
...
+        query: 's3_sorbet_is_throttled{"location"="${location}"}',
+        threshold: 100,
+        prometheus: {
+            endpoint: 'http://prometheus:9090',
+        },
+    };
+
+    const locationSpecificTemplatedProbe = {
+        type: 'prometheusQuery',
+        query: 'when{location=dmf-v1} and s3_sorbet_is_throttled{"location"="${location}"}',
```
francoisferrand: typo in query:

```suggestion
        query: 'when{location="dmf-v1"} and s3_sorbet_is_throttled{location="${location}"}',
```

tests/unit/lifecycle/CircuitBreakerGroup.spec.js
```diff
@@ -0,0 +1,1297 @@
...
+        query: 'when{location=dmf-v1} and kafka_consumergroup_group_lag{}',
+        threshold: 100,
+        prometheus: {
+            endpoint: 'http://prometheus:9090',
+        },
+    };
+
+    const locationSpecificTopicTemplateProbe = {
+        type: 'prometheusQuery',
+        query: 'when{location=dmf-v1} and kafka_consumergroup_group_lag{"topic"="${topic}"}',
```
francoisferrand:
```suggestion
        query: 'when{location="dmf-v1"} and kafka_consumergroup_group_lag{topic="${topic}"}',
```

tests/unit/lifecycle/CircuitBreakerGroup.spec.js
```diff
@@ -0,0 +1,1297 @@
...
+        query: 'when{topic=cold-archive-req-location-dmf-v1,location=dmf-v1} and kafka_consumergroup_group_lag{}',
+        threshold: 100,
+        prometheus: {
+            endpoint: 'http://prometheus:9090',
+        },
+    };
+
+    const invalidDoubleTemplate = {
+        type: 'prometheusQuery',
+        query: 's3_sorbet_is_throttled{"location"="${location}","topic"="${topic}"}',
```
francoisferrand:
```suggestion
        query: 's3_sorbet_is_throttled{location="${location}", topic="${topic}"}',
```

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,317 @@
...
+        // the topic
+        // - location specified + topic template : creates a circuit breaker for each topic and references them in
+        // the location
+        const whenClause = probe.query.match(/^when\{(.*?)\}\sand\s/);
+
+        const withTopicTemplate = probe.query.includes('${topic}');
+        const withLocationTemplate = probe.query.includes('${location}');
+
+        const topic = whenClause && whenClause[1].match(/topic=(.*?)(,|$)/);
+        const location = whenClause && whenClause[1].match(/location=(.*?)(,|$)/);
```
francoisferrand: should require a quoted string value:

```suggestion
        const topic = whenClause && whenClause[1].match(/topic="(.*?)"(,|$)/);
        const location = whenClause && whenClause[1].match(/location="(.*?)"(,|$)/);
```

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,317 @@
...
+        // - no when clause + templated location: creates a circuit breaker for each location
+        // - when clause with a specific topic + templated topic: creates a circuit breaker for the specified topic
+        // - when clause with a specific location + templated location: creates a circuit breaker for the location
+        // - when clause with topic and location + topic and location template: creates a circuit breaker for the
+        // location and topic
+        // - topic specified + location template : creates a circuit breaker for each location and references them in
+        // the topic
+        // - location specified + topic template : creates a circuit breaker for each topic and references them in
+        // the location
+        const whenClause = probe.query.match(/^when\{(.*?)\}\sand\s/);
```
francoisferrand: may have spaces between function a brace

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,317 @@
...
+
+        const withTopicTemplate = probe.query.includes('${topic}');
+        const withLocationTemplate = probe.query.includes('${location}');
+
+        const topic = whenClause && whenClause[1].match(/topic=(.*?)(,|$)/);
+        const location = whenClause && whenClause[1].match(/location=(.*?)(,|$)/);
+
+        const transition = whenClause && whenClause[1].match(/transition=(.*?)(,|$)/);
+        const expiration = whenClause && whenClause[1].match(/expiration=(.*?)(,|$)/);
```
francoisferrand: should be quoted here as well: prometheus does not support `bool` literals
```suggestion
        const transition = whenClause && whenClause[1].match(/transition="(.*?)"(,|$)/);
        const expiration = whenClause && whenClause[1].match(/expiration="(.*?)"(,|$)/);
```

maybe we can also be more strict?
 ```suggestion
        const transition = whenClause && whenClause[1].match(/transition="(true|false)"(,|$)/);
        const expiration = whenClause && whenClause[1].match(/expiration="(true|false)"(,|$)/);
```

File: extensions/lifecycle/CircuitBreakerGroup.js [hunk 27]
```diff
@@ -0,0 +1,312 @@
...
+        }
+        if (topic) {
+            this.circuitBreakers.expiration.topic[topic] =
+                this._append(this.circuitBreakers.expiration.topic[topic], circuitBreaker);
+        }
+        if (location) {
+            this.circuitBreakers.expiration.location[location] =
+                this._append(this.circuitBreakers.expiration.location[location], circuitBreaker);
+        }
+    }
```
francoisferrand: nit:

```suggestion
    _addCircuitBreaker(circuitBreaker, topic, location) {
        if (!topic && !location) {
            this.circuitBreakers[kind].global.push(circuitBreaker);
        }
        if (topic) {
            this.circuitBreakers[kind].topic[topic] =
                this._append(this.circuitBreakers[kind].topic[topic], circuitBreaker);
        }
        if (location) {
            this.circuitBreakers[kind].location[location] =
                this._append(this.circuitBreakers[kind].location[location], circuitBreaker);
        }
    }
```

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,312 @@
...
+
+    exportCircuitBreakerMetric(circuitBreaker, metricsSuffix) {
+        let metricsName = 'lifecycle_bucket_processor';
+        if (metricsSuffix) {
+            metricsName += `_${metricsSuffix}`;
+        }
+        metricsName += `_${this._latestBreakerId}`;
```
francoisferrand: the 'number' should come first IMO, may be easier to match with the index in the config?

extensions/lifecycle/CircuitBreakerGroup.js
```diff
@@ -0,0 +1,312 @@
...
+    probes.forEach(probe => {
+        if (probe.type !== 'prometheusQuery') {
+            global.probes.push(probe);
+            return;
+        }
+        // when clause is optional and can be used to specify the filtering on
+        // the topic and location, along with specifying if a probe should be
+        // used for transition and expiration.
+        // Query syntax:
+        // "when{topic=topicName,location=locationName,transition=true,expiration=true} and ANY"
```
francoisferrand: nit: not as implemented anymore...

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -161,10 +163,22 @@ class LifecycleTask extends BackbeatTask {
...
+            return;
+        }
+
         LifecycleMetrics.onLifecycleTriggered(this.log, 'bucket',
-            entry.getActionType() === 'deleteMPU' ? 'expiration:mpu' : 'expiration',
-            entry.getAttribute('details.dataStoreName'),
-            Date.now() - entry.getAttribute('transitionTime'));
+        entry.getActionType() === 'deleteMPU' ? 'expiration:mpu' : 'expiration',
+        location,
+        Date.now() - entry.getAttribute('transitionTime'));
```
francoisferrand: bad indent

francoisferrand (approved):

KillianG (approved):

PR #2501, opened 2024-02-27
include accountId in cold delete kafka messages

welansari:
Cold delete Kafka messages need to contain the accountId like for every other cold operation.

Issue: BB-508

### DIFF
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -295,6 +295,7 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
 
             if (err) {
                 this.log.error('unable to get account', {
+                    method: 'LifecycleQueuePopulator._handleRestoreOp',
                     ownerId,
                     err,
                 });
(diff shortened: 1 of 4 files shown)


extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -422,48 +422,60 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
 
-        this.log.trace(
-            'publishing object delete entry',
-            { bucket: entry.bucket, key: entry.key, version: value.versionId },
-        );
-
-        const topic = `${coldStorageGCTopicPrefix}${locationName}`;
-        const key = `${entry.bucket}/${entry.key}`;
+        const ownerId = value['owner-id'];
+        this.vaultClientWrapper.getAccountId(ownerId, (err, accountId) => {
```
francoisferrand, line 427: i don't think we can/want to afford the extra call to Vault for each message, when the value is actually not used?

→ One way would be to pass trough the `ownerId`, as fallback for `accountId` (both here & in Kafka), leaving all options open... but it may be a bit slow/unsafe to implement now?
→ Or, as a stop-gap measure, we may simply make the field required only for `archive` status messages

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -422,48 +423,63 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+            // In case of faillure, we still send the delete request to Sorbet forwarder
```
rachedbenmustapha:
```suggestion
            // In case of failure, we still send the delete request to Sorbet forwarder
```

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -422,48 +423,63 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+            // without the accountId. The accountId is not used for delete operations. It is
+            // only included for consistency with other operations.
```
rachedbenmustapha, line 430: This may or may not remain true in the future, and this comment might become inaccurate if for some reason we start needing to update some other piece of metadata after GC.

rachedbenmustapha (approved):

francoisferrand (approved):

PR #2502, opened 2024-03-08
simplify handling of replication failure notification

welansari:
Instead of duplicating the notification logic we set the correct originOp when updating the metadata to trigger the replication failure notification. (Bucket notification listens to the oplog and sends notification based on the originOp field in the metadata)

At the end of a failed replication we update the metadata with the failure status that ends up in the oplog topic. So this does not add any overhead in terms of extra Kafka messages being pushed to any of the topics.

This also fixes the duplicate notifications bug where we send as many notifications as destinations we have in case they use the same internal topic (bug already fixed in notification code).

Issue: BB-507

### DIFF
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -2,7 +2,7 @@ const assert = require('assert');
 const util = require('util');
 
 const { isMasterKey } = require('arsenal').versioning;
-const { usersBucket, mpuBucketPrefix } = require('arsenal').constants;
+const { usersBucket, mpuBucketPrefix, supportedNotificationEvents } = require('arsenal').constants;
 const configUtil = require('./utils/config');
 const safeJsonParse = require('./utils/safeJsonParse');
 const messageUtil = require('./utils/message');
@@ -92,6 +92,18 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
     }
 
     /**
+     * Notification rules are normally verified when setting the notification
+     * configuration (even for wildcards), however we need an explicit check at
+     * this level to filter out non standard events that might be valid for one
+     * of the wildcard rules set.
+     * @param {String} eventType - notification event type
+     * @returns {boolean} - true if notification is supported
+     */
+    _isNotificationEventSupported(eventType) {
+        return supportedNotificationEvents.has(eventType);
+    }
+
+    /**
      * Process object entry from the log
      *
      * @param {String} bucket - bucket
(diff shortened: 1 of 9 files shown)


lib/models/ObjectQueueEntry.js
```diff
@@ -193,21 +194,24 @@ class ObjectQueueEntry extends ObjectMD {
         return this.clone()
             .setAccountId(this.getAccountId())
             .setReplicationSiteStatus(site, 'COMPLETED')
-            .setReplicationStatus(this._getGlobalReplicationStatus());
+            .setReplicationStatus(this._getGlobalReplicationStatus())
+            .setOriginOp('');
```
francoisferrand: should we not set a value anyway?

it could (eventually) be used to generate non-standard notification (or make it easier/safer to refactor), and will be helpful already for troubleshooting: as we can directly see what caused the metadata change...

(note: need to make sure it does not match one of the "wildcard" notification rules, though...)

lib/models/ObjectQueueEntry.js
```diff
@@ -193,21 +194,24 @@ class ObjectQueueEntry extends ObjectMD {
...
+            .setOriginOp(replicationFailedEvent);
     }
 
     toPendingEntry(site) {
         return this.clone()
             .setAccountId(this.getAccountId())
             .setReplicationSiteStatus(site, 'PENDING')
-            .setReplicationStatus(this._getGlobalReplicationStatus());
+            .setReplicationStatus(this._getGlobalReplicationStatus())
+            .setOriginOp('');
```
francoisferrand: same

KillianG (approved):

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -91,6 +91,18 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+     * Notification rules are normally verified when setting the notification
+     * configuration (even for wildcards), however we need an explicit check at
+     * this level to filter out non standard events that might be valid for one
+     * of the wildcard rules set.
+     * @param {String} eventType - notification event type
+     * @returns {boolean} - true if notification is supported
+     */
+    _isNotificationEventSupported(eventType) {
+        return supportedNotificationEvents.has(eventType);
+    }
```
francoisferrand, line 104: not really an issue, but we may (at some point, just maybe) want to allow explicitely registering for these non-standard notifications (even though we probably don't want those to be picked up automatically by wildcard configs indeed)
→ we should remember this, may trick us later... (but certainly fine for now)

tests/unit/replication/UpdateReplicationStatus.js
```diff
@@ -88,6 +86,20 @@ describe('UpdateReplicationStatus._getUpdatedSourceEntry()', () => {
               ._getUpdatedSourceEntry({ sourceEntry, refreshedEntry }, log);
         assert.strictEqual(updatedSourceEntry, null);
     });
+
+    it('should return a PENDING entry when site status is FAILED', () => {
```
francoisferrand: .... `and replay count not reached` ?

(with this title it looks like a bug, and it seems we also have the case where it returns FAILED...)

francoisferrand (approved):

tests/unit/replication/UpdateReplicationStatus.js
welansari: Removed the test as it's a case that can never happen and can be confusing

PR #2505, opened 2024-03-25
bucket processor: avoid listing buckets when no lifecycle rules enabled

welansari:
Issue: BB-511

### DIFF
--- a/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
+++ b/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
@@ -3,7 +3,6 @@
 const async = require('async');
 const { Logger } = require('werelogs');
 const { errors } = require('arsenal');
-const { supportedLifecycleRules } = require('arsenal').constants;
 
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
@@ -15,6 +14,7 @@ const safeJsonParse = require('../util/safeJsonParse');
 const ClientManager = require('../../../lib/clients/ClientManager');
 const { authTypeAssumeRole } = require('../../../lib/constants');
 const LocationStatusStream = require('../../utils/LocationStatusStream');
+const { getFormattedSupportedLifecycleRules } = require('../util/rules');
 const {
     updateCircuitBreakerConfigForImplicitOutputQueue,
 } = require('../../../lib/CircuitBreaker');
@@ -85,17 +85,7 @@ class LifecycleBucketProcessor {
         this._producer = null;
         this._kafkaBacklogMetrics = null;
 
-        this._supportedRulesObject = {};
-        this._hasSupportedRules = Array.isArray(supportedLifecycleRules) &&
-            supportedLifecycleRules.length > 0;
-
-        if (this._hasSupportedRules) {
-            supportedLifecycleRules.forEach(rule => {
-                this._supportedRulesObject[rule] = { enabled: true };
-            });
-        } else {
-            this._log.debug('no lifecycle rules enabled');
-        }
+        this._supportedRules = getFormattedSupportedLifecycleRules();
 
         this._producerReady = false;
         this._consumerReady = false;
@@ -184,7 +174,6 @@ class LifecycleBucketProcessor {
         return {
             producer: this._producer,
             bootstrapList: this._repConfig.destination.bootstrapList,
-            enabledRules: this._supportedRulesObject,
             bucketTasksTopic: this._lcConfig.bucketTasksTopic,
             objectTasksTopic: this._lcConfig.objectTasksTopic,
             transitionTasksTopic: this._lcConfig.transitionTasksTopic,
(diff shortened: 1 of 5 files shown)


extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -236,7 +236,7 @@ class LifecycleBucketProcessor {
      * @return {Boolean} Whether the config should be processed
      */
     _shouldProcessConfig(config) {
-        if (config.Rules.length === 0) {
+        if (config.Rules.some(rule => rule.Status === 'Enabled')) {
```
francoisferrand: should we also check that the location associated with the (transition) rule is not paused?

tests/unit/lifecycle/LifecycleBucketProcessor.spec.js
```diff
@@ -38,4 +38,64 @@ describe('Lifecycle Bucket Processor', () => {
         lbp._resumeServiceForLocation('new-location');
         assert(!lbp._pausedLocations.has('new-location'));
     });
+
+    describe('_shouldProcessConfig', () => {
+        [
+            {
+                it: 'should return false if no lifecycle rules were set',
```
francoisferrand: nit: `it` → `title` ?

KazToozs (approved):

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -236,14 +236,29 @@ class LifecycleBucketProcessor {
...
+                return false;
+            }
+            if (rule.Transition &&
+                this._pausedLocations.has(rule.Transition.StorageClass)) {
+                return false;
+            }
+            if (rule.NoncurrentVersionTransition &&
+                this._pausedLocations.has(rule.NoncurrentVersionTransition.StorageClass)) {
+                return false;
+            }
```
francoisferrand: i wonder if it is not possible to have multiple "kinds" (transition, ncvTransiton, Expiration...) in the same rule : and thus we may have one transition to paused location but another one to non-paused location...

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -236,14 +236,29 @@ class LifecycleBucketProcessor {
...
+            return true;
+        });
+
+        if (rulesEnabled) {
+            return this._hasSupportedRules;
```
francoisferrand: not sure what this is for (seems fishy) : we should check what it was, and if we can clean this...

also, it is used to build a map of "supported kind of rules" : which we should probably use here, so that we don't scan if the rule is of unsupported kind... (this will be more relevant after unification of backbeat...)

welansari: I've checked it's dead code left after [switching to the lifecycle helper class in Arsenal](https://github.com/scality/backbeat/commit/7f66e415dca7de956d0f0f5e342e5b7bd44b9f61#diff-a171b708c40459f15b15e8656c2c1c52480522e07f44285483dac0458b1e10c8). It's not referenced anywhere else.

Reusing the Arsenal array of supported rules is a bit ugly however as the transition rule is written in the plural form.

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -85,17 +85,13 @@ class LifecycleBucketProcessor {
...
-        this._supportedRulesObject = {};
-        this._hasSupportedRules = Array.isArray(supportedLifecycleRules) &&
-            supportedLifecycleRules.length > 0;
-
-        if (this._hasSupportedRules) {
-            supportedLifecycleRules.forEach(rule => {
-                this._supportedRulesObject[rule] = { enabled: true };
+        this._supportedRules = !Array.isArray(supportedLifecycleRules) ? [] :
+            supportedLifecycleRules.map(rule => {
+                if (rule === 'transitions') {
```
francoisferrand: why do we need this mapping?
at the very least it requires a comment to explain (and remember) how this happens and why we need to do this mapping...

welansari: Supported lifecycle rules are written slightly differently in Arsenal than what we have in the lifecycle config.
I added a separate function to deal with that

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -236,14 +225,21 @@ class LifecycleBucketProcessor {
...
+                return false;
+            }
+            return Object.keys(rule).some(key => {
+                if (!this._supportedRules.includes(key)) {
+                    return false;
+                }
+                if (!rule[key].StorageClass || !this._pausedLocations.has(rule[key].StorageClass)) {
+                    return true;
+                }
+                return false;
```
francoisferrand: nit: seems weird this way...

either check false explicitely when the target location is paused (e.g. try every condition to 'skip'):
```suggestion
                if (rule[key].StorageClass && this._pausedLocations.has(rule[key].StorageClass)) {
                    return false;
                }
                return true;
```

or just return this value:
```suggestion
return !rule[key].StorageClass || !this._pausedLocations.has(rule[key].StorageClass)
```

extensions/lifecycle/util/rules.js
```diff
@@ -193,6 +194,25 @@ function rulesToParams(versioningStatus, currentDate, bucketLCRules, bucketData,
...
 
+/**
+ * Formats the array of supported lifecycle rules in Arsenal so that
+ * they correspond to how they are written in the lifecycle configuration.
+ * @param {string[]} supportedLifecycleRules list of supported lifecycle rules
+ * @returns {string[]} formatted supported lifecycle rules
+ */
+function getFormattedSupportedLifecycleRules() {
+    if (!Array.isArray(supportedLifecycleRules)) {
+        return [];
```
francoisferrand: can this case actually happen?
we should always have a (non-empty) array I guess.... (and it is not tested/testable?)

francoisferrand (approved):

PR #2507, opened 2024-04-02
Fix lifecycle rules name

welansari:
Use proper lifecycle rule names defined here https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getBucketLifecycleConfiguration-property

Issue: BB-516

### DIFF
--- a/extensions/lifecycle/util/rules.js
+++ b/extensions/lifecycle/util/rules.js
@@ -202,14 +202,26 @@ function rulesToParams(versioningStatus, currentDate, bucketLCRules, bucketData,
  */
 function getFormattedSupportedLifecycleRules() {
    return supportedLifecycleRules.map(rule => {
-        if (rule === 'transitions') {
-            return 'Transition';
+        if (rule === 'noncurrentVersionTransition') {
+            return 'NoncurrentVersionTransitions';
         }
         return rule.charAt(0).toUpperCase() + rule.slice(1);
     });
 }
 
+/**
+ * Check if ruleName is an expiration rule
+ * @param {string} ruleName lifecycle rule name
+ * @returns {bool} true if ruleName is an expiration rule
+ */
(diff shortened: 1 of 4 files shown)


KillianG (approved):

tests/unit/lifecycle/LifecycleBucketProcessor.spec.js
```diff
@@ -74,7 +74,7 @@ describe('Lifecycle Bucket Processor', () => {
                         },
                         {
                             Status: 'Disabled',
-                            Transition: {
+                            Transitions: {
```
francoisferrand: Should this not be an array, as per AWS specs?

tests/unit/lifecycle/LifecycleBucketProcessor.spec.js
```diff
@@ -104,7 +104,7 @@ describe('Lifecycle Bucket Processor', () => {
                         },
                         {
                             Status: 'Enabled',
-                            Transition: {
+                            Transitions: {
```
francoisferrand: Same here

extensions/lifecycle/util/rules.js
```diff
@@ -202,14 +202,26 @@ function rulesToParams(versioningStatus, currentDate, bucketLCRules, bucketData,
...
         return rule.charAt(0).toUpperCase() + rule.slice(1);
     });
 }
 
+/**
+ * Check if ruleName is an expiration rule
+ * @param {string} ruleName lifecycle rule name
+ * @returns {bool} true if ruleName is an expiration rule
+ */
+function isExpirationRules(ruleName) {
```
francoisferrand: nit: `isExpirationRule`

```suggestion
function isExpirationRule(ruleName) {
```

francoisferrand (approved):

PR #2510, opened 2024-04-16
Properly finish waterfall

KillianG:
Call done callback with error when an error occurs during a listBucket

Process was blocked because callback was not called when an error occurs during a listMongodbBuckets

Issue: BB-579

### DIFF
--- a/tests/unit/lifecycle/LifecycleConductor.spec.js
+++ b/tests/unit/lifecycle/LifecycleConductor.spec.js
@@ -49,7 +49,7 @@ describe('Lifecycle Conductor', () => {
         // timeout set to 4000 to account for 2s for async ops + 1s for bucket queue completion check interval
         this.timeout(4000);
 
-        it('should call callback with error when mongodbclient returns an error', done => {
+        it('should return error when mongodbclient returns an error', done => {
             conductor._mongodbClient = { getIndexingJobs: () => {}, getCollection: () => (
                 { find: () => (
                     { project: () => (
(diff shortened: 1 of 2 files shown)


tests/unit/lifecycle/LifecycleConductor.spec.js
```diff
@@ -48,6 +48,32 @@ describe('Lifecycle Conductor', () => {
...
+                        })
+                    })
+                })
+            }) };
+            conductor._zkClient = { getData: () => {} };
+
+            sinon.stub(conductor, '_controlBacklog')
+                .callsFake(cb => cb(null));
+            sinon.stub(conductor._mongodbClient, 'getIndexingJobs')
+                .callsFake((_, cb) => cb(null, ['job1', 'job2']));
```
francoisferrand: this stub is redundant with defniition on line 53 (`getIndexingJobs: () => {}`), should keep only one of the two...

more generally, looking at the kind of stubs you are defining here: either use sinon stubs or create non-configurable callbacks manually, but avoid mixing the two approaches...

tests/unit/lifecycle/LifecycleConductor.spec.js
```diff
@@ -48,6 +48,32 @@ describe('Lifecycle Conductor', () => {
...
+        it('should call callback with error when mongodbclient returns an error', done => {
+            conductor._mongodbClient = { getIndexingJobs: () => {}, getCollection: () => (
+                { find: () => (
+                    { project: () => (
+                        { hasNext: () => new Promise((resolve, reject) => {
+                            reject(new Error('error'));
+                        })
+                    })
+                })
+            }) };
```
francoisferrand:
```suggestion
            conductor._mongodbClient = {
                getIndexingJobs: () => {},
                getCollection: () => ({
                    find: () => ({
                        project: () => ({
                            hasNext: () => new Promise((resolve, reject) => {
                                reject(new Error('error'));
                            })
                        })
                    })
                })
            };
```

tests/unit/lifecycle/LifecycleConductor.spec.js
KillianG: In that case I can't use `sinon.stub` for `zkClient` and `mongoDbClient` because I can't stub non-existent property

francoisferrand (approved):

welansari (approved):

PR #2512, opened 2024-05-01
Allow replication of null master version

nicolas2bert:
- if it is a "standalone null master key": created when an object is placed in a non-versioned bucket, which is then converted to a versioned bucket. If no new versioned objects are added for that object, it appears as a standalone null master key with no version id.
- if it is a master suspended null version: created when a bucket versioning gets suspended and then an object is put. 

_NOTE 1: Null versions are not being replicated unless triggered by CrrExistingObject s3utils script (or OOB Ingested in Zenko/Artesca)_
_NOTE 2: Backporting existing Artesca logic (c.f. cherry-picks) with an end goal to merge Artesca and Ring branches._

### DIFF
--- a/extensions/replication/ReplicationQueuePopulator.js
+++ b/extensions/replication/ReplicationQueuePopulator.js
@@ -19,10 +19,7 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
         if (entry.bucket === usersBucket) {
             return this._filterBucketOp(entry);
         }
-        if (!isMasterKey(entry.key)) {
-            return this._filterVersionedKey(entry);
-        }
-        return undefined;
+        return this._filterKeyOp(entry);
     }
 
     _filterBucketOp(entry) {
@@ -40,7 +37,7 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
                      entry.bucket, JSON.stringify(publishedEntry));
     }
 
-    _filterVersionedKey(entry) {
+    _filterKeyOp(entry) {
         if (entry.type !== 'put') {
             return;
         }
@@ -51,8 +48,6 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
         if (sanityCheckRes) {
             return;
         }
-        // Allow a non-versioned object if being replicated from an NFS bucket.
-        // Or if the master key is of a non versioned object
         if (!this._entryCanBeReplicated(queueEntry)) {
             return;
         }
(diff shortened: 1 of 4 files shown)


nicolas2bert:
ping

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -78,6 +78,31 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+    /**
+     * Filter if the entry is considered a valid master key entry.
+     * There is a case where a single null entry looks like a master key and
+     * will not have a duplicate versioned key. They are created when you have a
+     * non-versioned bucket with objects, and then convert bucket to versioned.
+     * If no new versioned objects are added for given object(s), they look like
+     * standalone master keys. The `isNull` case is undefined for these entries.
+     * Null versions which are objects created after suspending versioning are allowed,
+     * these only have a master object that has an internal versionId and a 'isNull' flag.
+     * @param {ObjectQueueEntry} entry - raw queue entry
```
jonathan-gramain: (now raw since we parsed it and transformed it into an `ObjectQueueEntry`)

```suggestion
     * @param {ObjectQueueEntry} entry - queue entry
```

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -78,6 +78,31 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+     * @return {Boolean} true if we should filter entry
```
jonathan-gramain: `filter` is ambiguous, I'd actually more naturally think of the opposite of what is meant here (filter = reject as in "filter out").

```suggestion
     * @return {Boolean} true if we should accept entry
```

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -78,6 +78,31 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
                      `${queueEntry.getBucket()}/${queueEntry.getObjectKey()}`,
                      JSON.stringify(publishedEntry));
     }
+
+    /**
+     * Filter if the entry is considered a valid master key entry.
```
jonathan-gramain:
```suggestion
     * Accept the entry if considered a valid master key entry.
```

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -78,6 +78,31 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+        const isMaster = isMasterKey(entry.getObjectVersionedKey());
+        // single null entries will have a version id as undefined or null.
+        // do not filter single null entries
+        const isNonVersionedMaster = entry.getVersionId() === undefined;
+        const isNullVersionedMaster = entry.getIsNull();
+        if (isMaster && !isNonVersionedMaster && !isNullVersionedMaster) {
+            this.log.trace('skipping master key entry');
+            return false;
+        }
+        return true;
```
jonathan-gramain: Minor readability suggestion, I tend to think that reversing the logic is easier to reason about, because it transforms into a "either of" rather than a combination of three conditions that all have to be true at once (but up to you to change or not):

```suggestion
        if (!isMaster || isNonVersionedMaster || isNullVersionedMaster) {
            return true;
        }
        this.log.trace('skipping master key entry');
        return false;
```

tests/unit/replication/ReplicationQueuePopulator.spec.js
```diff
@@ -149,4 +149,68 @@ describe('replication queue populator', () => {
         // should not throw
         rqp._filterBucketOp(entry);
     });
+
+    // A "standalone null master key" is created when an object is placed in a non-versioned bucket,
+    // which is then converted to a versioned bucket. If no new versioned objects are added for that object,
+    // it appears as a standalone null master key with no version id.
+    it('should replicate standalone null master key', () => {
+        const customKafkaValiue = {
```
jonathan-gramain: Here and in other tests where this typo can be found:

```suggestion
        const customKafkaValue = {
```

jonathan-gramain (approved):
Good with the logic so approving now, just a few typos and suggestions

BourgoisMickael (approved):
Should the fixes rather be applied on artesca branch so there is no diff between the branches ?

KazToozs (approved):

welansari (approved):

nicolas2bert:
@bert-e approve

PR #2516, opened 2024-06-19
Improvement/bb 580 circuit breaker alert

benzekrimaha:
Linked Issue : 
https://scality.atlassian.net/browse/BB-580

This is what the alert looks like on the platform :
image

### DIFF
--- a/lib/CircuitBreaker.js
+++ b/lib/CircuitBreaker.js
@@ -46,14 +46,24 @@ const circuitBreakerGauge = ZenkoMetrics.createGauge({
     labelNames: ['type'],
 });
 
+const circuitBreakerCounter = ZenkoMetrics.createCounter({
+    name: 's3_circuit_breaker_errors_count',
+    help: 'total number of circuit breaker errors',
+    labelNames: [],
+});
+
 function startCircuitBreakerMetricsExport(cb, cbType, intervalMs = collectDefaultMetricsIntervalMs) {
     const type = cbType || 'generic';
     setInterval(() => {
+        if (cb.failedProbes) {
+            circuitBreakerCounter.inc();
+        }
         circuitBreakerGauge.set({ type }, cb.state);
     }, intervalMs);
 }
 
 module.exports = {
+    circuitBreakerCounter,
     circuitBreakerGauge,
     startCircuitBreakerMetricsExport,
     updateCircuitBreakerConfigForImplicitOutputQueue,
--- a/monitoring/lifecycle/alerts.test.yaml
+++ b/monitoring/lifecycle/alerts.test.yaml
@@ -611,3 +611,25 @@ tests:
       - alertname: KafkaConsumerRebalanceTimeouts
         eval_time: 4m
         exp_alerts: []
+
+  - name: CircuitBreakerError
+    interval: 1m
+    input_series:
+      - series: s3_circuit_breaker_errors_count{namespace="zenko"}
+        values: 0 0 1 1 1 1 1 1 1 1
+    alert_rule_test:
+      - alertname: CircuitBreakerError
+        eval_time: 0m
+        exp_alerts: []
+      - alertname: CircuitBreakerError
+        eval_time: 5m
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: zenko
+            exp_annotations:
+              summary: "Circuit breaker is unable to query prometheus"
+              description: "Circuit breaker is unable to query prometheus."
+      - alertname: CircuitBreakerError
+        eval_time: 10m
+        exp_alerts: []
--- a/monitoring/lifecycle/alerts.yaml
+++ b/monitoring/lifecycle/alerts.yaml
@@ -334,3 +334,15 @@ groups:
         Kafka rebalance has timed out for pod `{{ $labels.pod }}`, which indicates that the consumer
         is not working anymore, and should be restarted.
       summary: "Kafka consumer has stopped consuming messages"
+
+- name: Circuit Breaker
+  rules:
+  - alert: CircuitBreakerError
+    Expr: |
+      increase(s3_circuit_breaker_errors_count{namespace="${namespace}"}[5m]) > 0
+    Labels:
+     severity: warning
+    Annotations:
+      description: >-
+        Circuit breaker is unable to query prometheus.
+      summary: "Circuit breaker is unable to query prometheus"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "8.6.42",
+  "version": "8.6.43",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "async": "^2.3.0",
     "aws-sdk": "^2.938.0",
     "backo": "^1.1.0",
-    "breakbeat": "scality/breakbeat#v1.0.2",
+    "breakbeat": "scality/breakbeat#v1.0.3",
     "bucketclient": "scality/bucketclient#8.1.9",
     "commander": "^2.11.0",
     "fcntl": "github:scality/node-fcntl#0.2.2",
--- a/tests/unit/lib/util/circuitBreaker.spec.js
+++ b/tests/unit/lib/util/circuitBreaker.spec.js
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const {
+    circuitBreakerCounter,
     circuitBreakerGauge,
     startCircuitBreakerMetricsExport,
     updateCircuitBreakerConfigForImplicitOutputQueue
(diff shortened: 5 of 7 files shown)


lib/CircuitBreaker.js
```diff
@@ -43,13 +43,13 @@ function updateCircuitBreakerConfigForImplicitOutputQueue(cbConf, groupId, topic
 const circuitBreakerGauge = ZenkoMetrics.createGauge({
     name: 's3_circuit_breaker',
     help: 'Circuit Breaker State',
-    labelNames: ['type'],
+    labelNames: ['type', 'promErr'],
```
francoisferrand: the label should not be specific to prometheus : any error could happen...

lib/CircuitBreaker.js
```diff
@@ -43,13 +43,13 @@ function updateCircuitBreakerConfigForImplicitOutputQueue(cbConf, groupId, topic
...
 });
```
francoisferrand: Adding a label to exisiting metric or a new metric is basically the same thing : it will just create another series.

In this case, the value of the metric does not really make sense when the `error` label is present (since it is fundamentally more like an higher-order state of the circuit breaker)

→ this hints that this approach may not be the best way to model the information

* maybe an `s3_circuit_breaker_errors` **counter**? Since we continue polling, this counter would keep growing when there is an error in the config ; but would also cleanly catching "intermitent" errors, e.g. prometheus being temporarily not available...
* if we consider only the "validation of the configuration" (which was our initial problem), a completely different approach may be to let circuit breaker report the number of probes succesfully configured and the number of probes which failed to be configured
* following the same idea of just exposing the errors as a gauge, this may be done with a separate `s3_circuit_breaker_errors` gauge which reports the number of errors, with metrics exporting looking like:
```
circuitBreakerErrorsGauge.set({ type }, cb.failedProbes);   
circuitBreakerGauge.set({ type }, cb.state);
```

francoisferrand: or maybe we can actually have a "valid" circuit breaker state in combination with some errors, e.g. if only some probes are failing?

lib/CircuitBreaker.js
```diff
@@ -43,13 +43,13 @@ function updateCircuitBreakerConfigForImplicitOutputQueue(cbConf, groupId, topic
...
 
 function startCircuitBreakerMetricsExport(cb, cbType, intervalMs = collectDefaultMetricsIntervalMs) {
     const type = cbType || 'generic';
     setInterval(() => {
-        circuitBreakerGauge.set({ type }, cb.state);
+        circuitBreakerGauge.set({ type, promErr: cb.promErr }, cb.state);
```
francoisferrand: this is actually not correct. there is only "time serie" for each set of labels, and it does not expire.
this is not really an issue for counters (i.e. the number of "events" which happened), but a gauge represents the current number of something, and must thus be reset when that number drops...

here, looking over time:

```
# T0 : circuit breaker ok
s3_circuit_breaker{type="xxx", promErr="false"} 0

# T1 : circuit breaker reported an error (still state 0)
s3_circuit_breaker{type="xxx", promErr="false"} 0  # ← still present !
s3_circuit_breaker{type="xxx", promErr="true"}  0  # ← set by the timer

# T2 : circuit breaker still in error, state update to 2
s3_circuit_breaker{type="xxx", promErr="false"} 0  # ← still present !
s3_circuit_breaker{type="xxx", promErr="true"}  2  # ← set by the timer

# T3 : circuit breaker back to normal (no error, state 0)
s3_circuit_breaker{type="xxx", promErr="false"} 0  # ← set by the timer
s3_circuit_breaker{type="xxx", promErr="true"}  2  # ← still present !
```

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -153,3 +153,18 @@ groups:
       zenko_service: backbeat-ingestion-producer
       description: "Errors in publishing messages to ingestion topic has increased to 5% in the last 5 minutes"
       summary: "Ingestion producer error publishing messages"
+
+- name: Circuit Breaker
+  rules:
+  
+  - alert: CircuitBreakerError
+    Expr: |
+      count(s3_circuit_breaker{namespace="${namespace}", promErr="true"}) by(job) > 0
```
francoisferrand: this alert definition will match any of the backbeat pods, so no need to duplicate it.
duplication is needed only if we "specialize" each altering rule, e.g. one rule where the expression filters on the job label

benzekrimaha: The alert was kept in lifecycle only

lib/CircuitBreaker.js
benzekrimaha: I went for a counter : https://github.com/scality/backbeat/pull/2516/commits/73b8545ac169e6e00dc3e1ce637db01656c487f2#diff-07825dc8c7adf67b6bbea289468e3446d7646ec1c8f5b9c361b381807fb2ce28R58-R60
and the alert is now triggered based on the increase : https://github.com/scality/backbeat/pull/2516/commits/73b8545ac169e6e00dc3e1ce637db01656c487f2#diff-794b759f3f971e96a7be74227bbf5af68323e43bfb5dbbe2bc66fbe5dfb5f521R342

lib/CircuitBreaker.js
benzekrimaha: I went for a counter : https://github.com/scality/backbeat/commit/73b8545ac169e6e00dc3e1ce637db01656c487f2#diff-07825dc8c7adf67b6bbea289468e3446d7646ec1c8f5b9c361b381807fb2ce28R58-R60

lib/CircuitBreaker.js
benzekrimaha: The label has been removed and a new counter metric has been introduced here : https://github.com/scality/backbeat/pull/2516/commits/73b8545ac169e6e00dc3e1ce637db01656c487f2#diff-07825dc8c7adf67b6bbea289468e3446d7646ec1c8f5b9c361b381807fb2ce28R46-R52

lib/CircuitBreaker.js
```diff
@@ -46,14 +46,24 @@ const circuitBreakerGauge = ZenkoMetrics.createGauge({
     labelNames: ['type'],
 });
 
+const circuitBreakerCounter = ZenkoMetrics.createCounter({
+    name: 's3_circuit_breaker_err',
```
francoisferrand: s3_circuit_breaker_errors_count

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -153,3 +153,5 @@ groups:
       zenko_service: backbeat-ingestion-producer
       description: "Errors in publishing messages to ingestion topic has increased to 5% in the last 5 minutes"
       summary: "Ingestion producer error publishing messages"
+
```
francoisferrand: extra modification → should not be in the change

monitoring/lifecycle/alerts.test.yaml
```diff
@@ -611,3 +611,25 @@ tests:
...
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: zenko
+            exp_annotations:
+              summary: "Circuit breaker is unable to query prometheus"
+              description: "Circuit breaker is unable to query prometheus."
+      - alertname: CircuitBreakerError
+        eval_time: 10m
+        exp_alerts: []
```
francoisferrand: missing new line at EOF

monitoring/lifecycle/alerts.test.yaml
```diff
@@ -611,3 +611,25 @@ tests:
...
+        eval_time: 5m
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: zenko
+            exp_annotations:
+              summary: "Circuit breaker is unable to query prometheus"
+              description: "Circuit breaker is unable to query prometheus."
+      - alertname: CircuitBreakerError
+        eval_time: 10m
```
francoisferrand, line 634: the time series in this test does not reach 10m : so you are not actually testing that the alert stops when things go back to normal, but instead testing that the alert disappear when there is no an (s3_circuit_breaker_err) metric anymore...

tests/unit/lifecycle/CircuitBreakerGroup.spec.js
```diff
@@ -967,7 +967,7 @@ describe('extractBucketProcessorCircuitBreakerConfigs', () => {
                     transition: {
                         location: {},
                         topic: {},
-                        global: [ ],
+                        global: [],
```
benzekrimaha, line 970: Please note that this change is just to fix a lint warning , will be separated in a specific commit post reviews

monitoring/lifecycle/alerts.test.yaml
benzekrimaha: The test has been adapted to check the 10m interval https://github.com/scality/backbeat/pull/2516/commits/32be5a866e3f8db3da1d9222236041b735ef32e8#diff-08e5453052d3ff64a7ec5f78d40faf7811de2787e942a284b0263f118de6a13aR619

monitoring/lifecycle/alerts.yaml
```diff
@@ -334,3 +334,15 @@ groups:
         Kafka rebalance has timed out for pod `{{ $labels.pod }}`, which indicates that the consumer
         is not working anymore, and should be restarted.
       summary: "Kafka consumer has stopped consuming messages"
+- name: Circuit Breaker
+  rules:
+
+  - alert: CircuitBreakerError
```
francoisferrand: nit:

```suggestion

- name: Circuit Breaker
  rules:
  - alert: CircuitBreakerError
```

yarn.lock
```diff
@@ -1452,9 +1452,9 @@ brace-expansion@^1.1.7:
     balanced-match "^1.0.0"
     concat-map "0.0.1"
 
-breakbeat@scality/breakbeat#v1.0.2:
+breakbeat@scality/breakbeat#ca4687064a42c0cdca6bda864a6a94c01d0add5c:
```
francoisferrand, line 1455: reminder to bump before merge

tests/unit/lib/util/circuitBreaker.spec.js
```diff
@@ -165,12 +166,14 @@ describe('updateCircuitBreakerConfigForImplicitOutputQueue', () => {
 
 describe('startCircuitBreakerMetricsExport', () => {
     it('should export circuit breaker state', done => {
```
francoisferrand: should also have test with probesFailed == false

package.json
```diff
@@ -50,7 +50,7 @@
     "async": "^2.3.0",
     "aws-sdk": "^2.938.0",
     "backo": "^1.1.0",
-    "breakbeat": "scality/breakbeat#v1.0.2",
+    "breakbeat": "scality/breakbeat#ca4687064a42c0cdca6bda864a6a94c01d0add5c",
```
francoisferrand: reminder to bump before merge

tests/unit/lib/util/circuitBreaker.spec.js
benzekrimaha: https://github.com/scality/backbeat/pull/2516/commits/9a8e5128073c30bca4eda2e005675bf42a57fe5a#diff-b2d322d1781a863d778ef5651dadf9baaea5ec28278dbe21abcde66f89e12c5dR168-R189

welansari (approved):

lib/CircuitBreaker.js
```diff
@@ -46,14 +46,24 @@ const circuitBreakerGauge = ZenkoMetrics.createGauge({
     labelNames: ['type'],
 });
 
+const circuitBreakerCounter = ZenkoMetrics.createCounter({
+    name: 's3_circuit_breaker_errors_count',
+    help: 'circuitBreker errors',
```
francoisferrand: typo
```suggestion
    help: 'total number of circuit breaker errors',
```

francoisferrand (approved):

benzekrimaha:
@bert-e create_pull_requests

PR #2520, opened 2024-07-18
Improvement/bb 587 retry on s3 errors

benzekrimaha:
Linked issue : https://scality.atlassian.net/browse/BB-587

### DIFF
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -17,6 +17,10 @@ class GarbageCollectorTask extends BackbeatTask {
         super();
         const gcState = gc.getStateVars();
         Object.assign(this, gcState);
+        const retryParams = gcState.gcConfig?.consumer.retry;
+        if (retryParams) {
+            this.retryParams = retryParams;
+        }
     }
 
     // helper method needed for replication generated entries in which the
@@ -134,7 +138,7 @@ class GarbageCollectorTask extends BackbeatTask {
         });
     }
 
-    _executeDeleteData(entry, log, done) {
+    _executeDeleteDataOnce(entry, log, done) {
         const { locations } = entry.getAttribute('target');
         const ruleType = entry.getContextAttribute('ruleType');
         const params = {
@@ -184,7 +188,24 @@ class GarbageCollectorTask extends BackbeatTask {
         });
     }
 
-    _deleteArchivedSourceData(entry, log, done) {
+    _executeDeleteData(entry, log, done) {
+        const logFields = {
+            bucket: entry.getAttribute('source.bucket'),
+            objectKey: entry.getAttribute('source.objectKey'),
+            storageClass: entry.getAttribute('source.storageClass'),
+            ruleType: entry.getContextAttribute('ruleType'),
+        };
+
+        this.retry({
+            actionDesc: 'execute delete data',
+            logFields,
+            actionFunc: cb => this._executeDeleteDataOnce(entry, log, cb),
+            shouldRetryFunc: err => err.retryable,
+            log,
+        }, done);
+    }
+
+    _deleteArchivedSourceDataOnce(entry, log, done) {
         const { bucket, key, version, oldLocation, newLocation } = entry.getAttribute('target');
 
         async.waterfall([
@@ -254,7 +275,7 @@ class GarbageCollectorTask extends BackbeatTask {
                     .setUserMetadata({
                         'x-amz-meta-scal-s3-transition-attempt': undefined,
                     });
-                return this._putMetadata(entry, objMD, log, err => {
+                this._putMetadata(entry, objMD, log, err => {
                     GarbageCollectorMetrics.onS3Request(log, 'putMetadata', 'archive', err);
                     if (!err) {
                         log.end().info('completed expiration of archived data',
@@ -278,6 +299,17 @@ class GarbageCollectorTask extends BackbeatTask {
             }
             return done();
         });
+    }
+
+    _deleteArchivedSourceData(entry, log, done) {
+        this.retry({
+            actionDesc: 'delete archived source data',
+            logFields: { bucket: entry.getAttribute('target').bucket,
+                key: entry.getAttribute('target').key, version: entry.getAttribute('target').version },
+            actionFunc: cb => this._deleteArchivedSourceDataOnce(entry, log, cb),
+            shouldRetryFunc: err => err.retryable,
+            log,
+        }, done);
     }
 
     /**
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "8.6.45",
+  "version": "8.6.46",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
@@ -17,7 +17,7 @@
     "oplog_populator": "node extensions/oplogPopulator/OplogPopulatorTask.js",
     "mongo_queue_processor": "node extensions/mongoProcessor/mongoProcessorTask.js",
     "garbage_collector": "node extensions/gc/service.js",
-    "test": "mocha --recursive tests/unit",
+    "test": "mocha --recursive tests/unit --timeout 30000",
     "cover:test": "nyc --clean --silent yarn run test && nyc report --report-dir ./coverage/test --reporter=lcov",
     "ft_test": "mocha --recursive $(find tests/functional -name '*.js') --timeout 30000",
     "ft_test:notification": "mocha --recursive $(find tests/functional/notification -name '*.js') --timeout 30000",
--- a/tests/unit/gc/GarbageCollectorTask.spec.js
+++ b/tests/unit/gc/GarbageCollectorTask.spec.js
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const werelogs = require('werelogs');
+const sinon = require('sinon');
 const { ObjectMD } = require('arsenal').models;
 
 const GarbageCollectorTask = require('../../../extensions/gc/tasks/GarbageCollectorTask');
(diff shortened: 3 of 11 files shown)


extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -269,8 +288,6 @@ class GarbageCollectorTask extends BackbeatTask {
             },
         ], err => {
             if (err) {
-                // if error occurs, do not commit offset unless the error is ObjNotFound
```
KillianG: Why do you remove this comment ?

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -254,7 +267,13 @@ class GarbageCollectorTask extends BackbeatTask {
                     .setUserMetadata({
                         'x-amz-meta-scal-s3-transition-attempt': undefined,
                     });
-                return this._putMetadata(entry, objMD, log, err => {
+                return this.retry({
+                    actionDesc: 'put metadata',
+                    logFields: { bucket, key, version },
+                    actionFunc: cb => this._putMetadata(entry, objMD, log, cb),
```
KillianG: Should probably set maxRetries on all retries

tests/unit/gc/GarbageCollectorTask.spec.js
```diff
@@ -171,6 +172,42 @@ describe('GarbageCollectorTask', () => {
...
+        mdObj.setLocation(loc)
+            .setDataStoreName('old-location')
+            .setAmzStorageClass('old-location')
+            .setTransitionInProgress(true);
+        backbeatMetadataProxyClient.setMdObj(mdObj);
+    
+        const batchDeleteDataSpy = sinon.spy(gcTask, '_batchDeleteData');
+    
+        gcTask.processActionEntry(entry, (err, commitInfo) => {
+            assert(batchDeleteDataSpy.callCount > 1);
```
KillianG: Maybe you can check that callcount is equal to maxRetries ? As the function always returns retryable error here

tests/unit/gc/GarbageCollectorTask.spec.js
```diff
@@ -171,6 +172,42 @@ describe('GarbageCollectorTask', () => {
...
+            batchDeleteDataSpy.restore();
+
```
KillianG: Check that it still returns an error at the end of the retries

extensions/gc/tasks/GarbageCollectorTask.js
benzekrimaha: Max retries are set to 2 ( 3 calls will be occurring with the first one into account )

tests/unit/gc/GarbageCollectorTask.spec.js
benzekrimaha: Done here : https://github.com/scality/backbeat/pull/2520/files#diff-c8fffb759b180f07e916e3e53512ba1af73162a2d0fd4d530df4aa68db717009R204

tests/unit/gc/GarbageCollectorTask.spec.js
benzekrimaha: done here : https://github.com/scality/backbeat/pull/2520/files#diff-c8fffb759b180f07e916e3e53512ba1af73162a2d0fd4d530df4aa68db717009R205-R206

extensions/gc/tasks/GarbageCollectorTask.js
benzekrimaha: It was just an oversight, nice catch

extensions/gc/tasks/GarbageCollectorTask.js
KillianG: have you been able to understand why the default value is not used ?

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -188,7 +188,14 @@ class GarbageCollectorTask extends BackbeatTask {
...
         async.waterfall([
-            next => this._getMetadata(entry, log, (err, objMD) => {
+            next => this.retry({
+                actionDesc: 'get metadata',
+                logFields: { bucket, key, version },
+                maxRetries: 2,
+                actionFunc: cb => this._getMetadata(entry, log, cb),
+                shouldRetryFunc: err => err.retryable,
+                log,
+            }, (err, objMD) => {
```
welansari: I think a better approach for this would be having a global retry for the whole function and not a retry per operation. Something like this :
https://github.com/scality/backbeat/blob/e19a1e5a8b9fffc09d536a04fa4176ef2ec13fb4/extensions/replication/tasks/MultipleBackendTask.js#L195-L247

So we rename `_deleteArchivedSourceData` to `_deleteArchivedSourceDataOnce` and `_deleteArchivedSourceData` becomes a new wrapper function that implements the retry on `_deleteArchivedSourceDataOnce`

tests/unit/gc/GarbageCollectorTask.spec.js
```diff
@@ -171,6 +172,43 @@ describe('GarbageCollectorTask', () => {
         });
     });
 
+    it('should retry delete operation if gc failed with retryable error', done => {
```
KillianG, line 184: I think you are missing tests for put metadata and get metadata

extensions/gc/tasks/GarbageCollectorTask.js
welansari: btw, should we also retry `_executeDeleteData` ?

benzekrimaha: Done here : https://github.com/scality/backbeat/pull/2520/commits/8dbcf1115f3227232abe57a90fd3983798cbb46a

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "8.6.43",
+  "version": "8.6.44",
```
welansari: need to rebase

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -184,7 +184,25 @@ class GarbageCollectorTask extends BackbeatTask {
...
+            bucket: entry.getAttribute('source.bucket'),
+            objectKey: entry.getAttribute('source.objectKey'),
+            storageClass: entry.getAttribute('source.storageClass'),
+            ruleType: entry.getContextAttribute('ruleType'),
+        };
+
+        this.retry({
+            actionDesc: 'execute delete data',
+            logFields,
+            maxRetries: 2,
```
welansari: No need to specify `maxRetries`, that way it uses the value set in the config and which is configured by ZKOP.

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -280,6 +298,18 @@ class GarbageCollectorTask extends BackbeatTask {
         });
     }
 
+    _deleteArchivedSourceData(entry, log, done) {
+        this.retry({
+            actionDesc: 'delete archived source data',
+            logFields: { bucket: entry.getAttribute('target').bucket,
+                key: entry.getAttribute('target').key, version: entry.getAttribute('target').version },
+            maxRetries: 2,
```
welansari: same here

extensions/gc/tasks/GarbageCollectorTask.js
benzekrimaha: done here : https://github.com/scality/backbeat/pull/2520/commits/1e91524b7b2b2b6de6067258f1b53101da49daec#diff-7beaf0ffcd262dad5a78c1b25cb7d54b9ac577441b372167f96d8c182be8fcf7R20-R24

extensions/gc/tasks/GarbageCollectorTask.js
benzekrimaha: done here : https://github.com/scality/backbeat/pull/2520/commits/1e91524b7b2b2b6de6067258f1b53101da49daec#diff-7beaf0ffcd262dad5a78c1b25cb7d54b9ac577441b372167f96d8c182be8fcf7R20-R24

package.json
benzekrimaha: Will do once the PR approved

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -17,6 +17,11 @@ class GarbageCollectorTask extends BackbeatTask {
         super();
         const gcState = gc.getStateVars();
         Object.assign(this, gcState);
+        const retryParams =
+        gcState.gcConfig?.consumer.retry;
+        if (retryParams) {
+            Object.assign(this, retryParams);
```
welansari: Not sure this will work. I think `Object.assign` will put the fields/values that are inside `retryParams` and add then to `this` instead of adding `retryParams` as a field. So instead of having
```
{
    retryParams: {
      "backoff": {
        "factor": "1.5",
        "jitter": "0.1",
        "max": 300000,
        "min": 1000
      },
      "maxRetries": 5,
      "timeoutS": 300
    }
}
```
we get
```
{
    "backoff": {
        "factor": "1.5",
        "jitter": "0.1",
        "max": 300000,
        "min": 1000
    },
    "maxRetries": 5,
    "timeoutS": 300
}
```

benzekrimaha: Indeed fixed here : https://github.com/scality/backbeat/pull/2520/commits/c80b93d63bf745e80314515d17a6ad8aa74347cf#diff-7beaf0ffcd262dad5a78c1b25cb7d54b9ac577441b372167f96d8c182be8fcf7R22

welansari (approved):

KillianG (approved):

benzekrimaha:
@bert-e reset

benzekrimaha:
@bert-e create_pull_requests

PR #2533, opened 2024-09-02
add missing Kafka log source in error

benzekrimaha:
Issue : BB-530

### DIFF
--- a/lib/queuePopulator/QueuePopulator.js
+++ b/lib/queuePopulator/QueuePopulator.js
@@ -358,7 +358,7 @@ class QueuePopulator {
                 break;
             default:
                 throw new Error("bad 'logSource' config value: expect 'bucketd,'" +
-                    ` mongo or 'dmd', got '${this.qpConfig.logSource}'`);
+                    ` 'mongo' or 'kafka' or 'dmd', got '${this.qpConfig.logSource}'`);
         }
     }
 


lib/queuePopulator/QueuePopulator.js
```diff
@@ -350,7 +350,7 @@ class QueuePopulator {
                 break;
             default:
                 throw new Error("bad 'logSource' config value: expect 'bucketd,'" +
-                    ` mongo or 'dmd', got '${this.qpConfig.logSource}'`);
+                    ` mongo or 'kafka' or 'dmd', got '${this.qpConfig.logSource}'`);
```
ghost:
```suggestion
                throw new Error("bad 'logSource' config value: expect 'bucketd'," +
                    ` 'mongo', 'kafka' or 'dmd', got '${this.qpConfig.logSource}'`);
```

ghost (approved):

lib/queuePopulator/QueuePopulator.js
ghost: Also the ticket says "_setupLogSources(): missing Kafka log source in the error we throw when no log source corresponds."
but it was here before right? Are we missing something?

francoisferrand (approved):

lib/queuePopulator/QueuePopulator.js
benzekrimaha: yes it is already in the error message, what was meant in the ticket is only that kafka was not listed as one of the expected values from what I understood

PR #2534, opened 2024-09-03
Remove useless config elements

benzekrimaha:
Issue : BB-520

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -113,14 +113,6 @@ class Config extends EventEmitter {
                 { host: '127.0.0.1', port: 6379 });
         }
 
-        if (parsedConfig.localCache) {
-            this.localCache = {
-                host: config.localCache.host,
-                port: config.localCache.port,
-                password: config.localCache.password,
-            };
-        }
-
         // additional certs checks
         if (parsedConfig.certFilePaths) {
             parsedConfig.https = this._parseCertFilePaths(
--- a/conf/config.json
+++ b/conf/config.json
@@ -324,14 +324,6 @@
         "host": "127.0.0.1",
         "port": 8900
     },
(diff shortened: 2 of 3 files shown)


lib/Config.js
ghost, line 1: Just saw that, but do we still use the 8.5 branch? I see 8.6 in Zenko 2.6

ghost: (affects the other pr, although it's harmless, maybe now is the right time to archive the 8.5 branch, if not used anymore)

lib/Config.js
```diff
@@ -106,14 +106,6 @@ class Config extends EventEmitter {
                 { host: '127.0.0.1', port: 6379 });
         }
 
-        if (parsedConfig.localCache) {
```
ghost, line 116: Should be removed from the json file as well, maybe also from the operator (https://github.com/scality/zenko-operator/blob/development/1.6/pkg/controller/zenko/reconcile_backbeat_config.go#L982)

lib/Config.js
benzekrimaha: yes , going to target the 8.7 instead in all PRs

lib/Config.js
benzekrimaha: done here : https://github.com/scality/backbeat/pull/2534/commits/1ad8a8219e5107e9c1340432457538387a5833ec#diff-10001999fea4d2498276120abf7902b81510153910d8314c7493ddafa6e19e09L324-L331, for the operator i'm going to check then create a ticket if necessary

benzekrimaha: ticket created here : https://scality.atlassian.net/browse/ZKOP-408

ghost (approved):

francoisferrand (approved):

PR #2535, opened 2024-09-03
Fix BackbeatProducer event.error handlers set twice.

benzekrimaha:
Issue : BB-523

### DIFF
--- a/lib/BackbeatProducer.js
+++ b/lib/BackbeatProducer.js
@@ -36,7 +36,6 @@ class BackbeatProducer extends EventEmitter {
         this._producer.on('event', event => this._log.info('rdkafka.event', { event }));
         this._producer.on('event.log', log => this._log.info('rdkafka.log', { log }));
         this._producer.on('warning', warning => this._log.warn('rdkafka.warning', { warning }));
-        this._producer.on('event.error', err => this._log.error('rdkafka.error', { err }));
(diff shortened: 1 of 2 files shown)


lib/BackbeatProducer.js
```diff
@@ -36,7 +36,6 @@ class BackbeatProducer extends EventEmitter {
         this._producer.on('event', event => this._log.info('rdkafka.event', { event }));
         this._producer.on('event.log', log => this._log.info('rdkafka.log', { log }));
         this._producer.on('warning', warning => this._log.warn('rdkafka.warning', { warning }));
-        this._producer.on('event.error', err => this._log.error('rdkafka.error', { err }));
         this._producer.on('event.throttle', throttle => this._log.info('rdkafka.throttle', { throttle }));
```
francoisferrand, line 39: i think this is more subtle than this.

we set 2 event handlers (which is fine) : this one for logs, always ; the other one to emit an error (but also log) only from a certain point in the state machine

→ we propably want to avoid the duplicate logs, but we still want to have the logs all the time ; without changing the current logic of the automaton...

benzekrimaha: Went for a flag to avoid duplicate logs here: https://github.com/scality/backbeat/pull/2535/commits/82315d227ce17bded9d4adaad8571433476ccbb9

ghost (approved):

KillianG (approved):

PR #2537, opened 2024-09-06
Replication deployment with single location per instance

benzekrimaha:
Issue : BB-495

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -36,6 +36,8 @@ class Config extends EventEmitter {
          * By default, the config file is "config.json" at the root.
          * It can be overridden using the BACKBEAT_CONFIG_FILE environment var.
          */
+        // Note: used to allow access to the class itself from an instance.
+        this.Config = Config;
         this._basePath = `${__dirname}/../conf`;
         if (process.env.BACKBEAT_CONFIG_FILE !== undefined) {
             this._configPath = process.env.BACKBEAT_CONFIG_FILE;
@@ -81,7 +83,10 @@ class Config extends EventEmitter {
         } else {
             this.bootstrapList = [];
         }
-
+        const shouldRestrictToSite = process.env.BOOTSTRAP_SITE_NAME;
+        if (shouldRestrictToSite) {
+            this.bootstrapList = this.bootstrapList.filter(item => item.site === shouldRestrictToSite);
+        }
         // NOTE: used to store ingestion bucket information
         this.ingestionBuckets = [];
 
--- a/tests/unit/conf/Config.js
+++ b/tests/unit/conf/Config.js
@@ -1,13 +1,13 @@
 'use strict'; // eslint-disable-line
 
 const assert = require('assert');
-
 const joi = require('joi');
-
 const config = require('../../../lib/Config');
+const { Config } = require('../../../lib/Config');
 const { authJoi, inheritedAuthJoi } = require('../../../lib/config/configItems.joi');
 
 describe('backbeat config parsing and validation', () => {
+
     it('should parse correctly the default config', () => {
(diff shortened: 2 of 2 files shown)


lib/Config.js
```diff
@@ -82,6 +82,9 @@ class Config extends EventEmitter {
             this.bootstrapList = [];
         }
 
+        if (process.env.SITE_NAME) {
```
ghost: I think the environment variable should better explain what it does exactly (if I understand correctly, we want to restrict the list of locations to be managed, so we could have something like `MANAGED_SITE_NAME` (but please challenge that)

lib/Config.js
```diff
@@ -82,6 +82,9 @@ class Config extends EventEmitter {
...
+            this.bootstrapList = this.bootstrapList.filter(item => item.site === process.env.SITE_NAME);
+        }
```
ghost: Accessing process.env variable is slow in general (it executes a C code at the system level), in general we tend to avoid accessing the same env variable multiple times:

```suggestion
        const shouldRestrictToSite = process.env.SITE_NAME;
        if (!!shouldRestrictToSite) {
            this.bootstrapList = this.bootstrapList.filter(item => item.site === shouldRestrictToSite);
        }
```

tests/unit/conf/Config.js
```diff
@@ -1,13 +1,14 @@
 'use strict'; // eslint-disable-line
 
-const assert = require('assert');
+process.env.SITE_NAME = 'aws-location';
```
ghost: Doing this at the top-level will always filter the bootstrap list, don't we want to test both cases? We also want tests to be idempotent, so if we need to unset this value in the test, if someday we run them in a random order, it will not work...

If always testing with filtering is fine, then we would prefer having this env variable in the github action step, so we do not need to alter the config at runtime (depending on the order of imports, it may also not always work).

lib/Config.js
```diff
@@ -292,4 +295,7 @@ class Config extends EventEmitter {
     }
 }
 
-module.exports = new Config();
+module.exports = {
```
benzekrimaha: this has been done to ease the testing

francoisferrand: yes, but it breaks code coverage (and 10x more changes than what you are trying to change...)

Alternatively, kind of a hack to make it work may be this:

```
class Config {
   constructor() {
       this.Config = Config; // store the `class` as an attribute
       [...]
   }
}

module.exports = new Config();
```

* When imported as before, we get the singleton "config" object
   ```
   const config = require('lib/Config.js')
   ```
* When destructured imports are used, we get `Config` field of the "config" singletion object, which is actually the class of the object: and can thus be used to create other config object
   ```
   const { Config } = require('lib/Config.js')
   ```

lib/Config.js
```diff
@@ -81,7 +81,10 @@ class Config extends EventEmitter {
         } else {
             this.bootstrapList = [];
         }
-
+        const shouldRestrictToSite = process.env.SITE_NAME;
```
francoisferrand: maybe `BOOTSTRAP_SITE_NAME` would be more explicit, conveying that this does indeed affect the bootstrap list...

lib/Config.js
```diff
@@ -36,6 +36,7 @@ class Config extends EventEmitter {
          * By default, the config file is "config.json" at the root.
          * It can be overridden using the BACKBEAT_CONFIG_FILE environment var.
          */
+        this.Config = Config;
```
welansari, line 40: maybe we should add a small comment to explain why this was added ?

welansari (approved):

ghost (approved):

PR #2539, opened 2024-09-10
Support limiting bucket number in connectors and mongodb 6

ghost:
Observations:

- Add a security to not accept more than N buckets in a single pipeline, to avoid exceeding the BSON size limit.
- Support mongo6 with immutable strategy for connectors.
- Update the allocation strategy to ensure that we do not remove buckets from immutable connectors.
- Upgrade handling is not needed: previous pipelines will not be updated if the pipelines are immutable.
- Fix an issue where a bucket would be temporarily handled by two connectors: if we remove a bucket from the in-memory Set of the connector, it may be propagated to the pipeline after 5s, so the same bucket might be assigned to another connector in-between, causing some event to be processed twice.
- Add metrics on the number of buckets exceed.
- Add an alert (warning) when we exceed the configured number of connectors.

image

### DIFF
--- a/.github/workflows/alerts.yaml
+++ b/.github/workflows/alerts.yaml
@@ -81,4 +81,5 @@ jobs:
             namespace=zenko
             oplog_populator_job=artesca-data-backbeat-oplog-populator-headless
             oplogPopulatorChangeStreamLagThreshold=10
+            oplogPopulatorChangeStreamLimit=10
           github_token: ${{ secrets.GIT_ACCESS_TOKEN }}
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -287,11 +287,11 @@ class OplogPopulator {
             });
             // For now, we always use the RetainBucketsDecorator
             // so, we map the events from the classes
-            this._connectorsManager.on('connector-updated', connector =>
+            this._connectorsManager.on(constants.connectorUpdatedEvent, connector =>
                 this._allocationStrategy.onConnectorUpdatedOrDestroyed(connector));
-            this._allocator.on('bucket-removed', (bucket, connector) =>
+            this._allocator.on(constants.bucketRemovedFromConnectorEvent, (bucket, connector) =>
                 this._allocationStrategy.onBucketRemoved(bucket, connector));
-            this._connectorsManager.on('connectors-reconciled', bucketsExceedingLimit => {
+            this._connectorsManager.on(constants.connectorsReconciledEvent, bucketsExceedingLimit => {
                 this._metricsHandler.onConnectorsReconciled(
                     bucketsExceedingLimit,
                     this._allocationStrategy.retainedBucketsCount,
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -51,7 +51,7 @@ class OplogPopulator {
         this._logger = params.logger;
         this._changeStreamWrapper = null;
         this._allocator = null;
-        this._connectorsManager  = null;
+        this._connectorsManager = null;
         // contains OplogPopulatorUtils class of each supported extension
         this._extHelpers = {};
         // MongoDB related
@@ -78,9 +78,9 @@ class OplogPopulator {
     async _setupMongoClient() {
         try {
             const client = await MongoClient.connect(this._mongoUrl, {
-                 replicaSet: this._replicaSet,
-                 useNewUrlParser: true,
-                 useUnifiedTopology: true,
+                replicaSet: this._replicaSet,
+                useNewUrlParser: true,
+                useUnifiedTopology: true,
             });
             // connect to metadata DB
             this._mongoClient = client.db(this._database, {
@@ -291,13 +291,13 @@ class OplogPopulator {
             this._logger.info('OplogPopulator setup complete', {
                 method: 'OplogPopulator.setup',
             });
-       } catch (err) {
+        } catch (err) {
             this._logger.error('An error occured when setting up the OplogPopulator', {
                 method: 'OplogPopulator.setup',
                 error: err.description || err.message,
             });
             throw errors.InternalError.customizeDescription(err.description);
-       }
+        }
     }
 
     /**
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -288,10 +288,16 @@ class OplogPopulator {
             });
             // For now, we always use the RetainBucketsDecorator
             // so, we map the events from the classes
-            this._connectorsManager.on('connector-updated', connector =>
-                allocationStrategy.onConnectorUpdated(connector));
+            this._connectorsManager.on('connector-destroyed', connector =>
+                allocationStrategy.onConnectorDestroyed(connector));
             this._allocator.on('bucket-removed', (bucket, connector) =>
                 allocationStrategy.onBucketRemoved(bucket, connector));
+            this._allocator.on('connectors-reconciled', bucketsExceedingLimit => {
+                this._metricsHandler.onConnectorsReconciliation(
+                    bucketsExceedingLimit,
+                    allocationStrategy.retainedBucketsNb,
+                );
+            });
             // get currently valid buckets from mongo
             const validBuckets = await this._getBackbeatEnabledBuckets();
             // listen to valid buckets
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -266,7 +266,6 @@ class OplogPopulator {
             this._allocationStrategy = this.initStrategy();
             this._connectorsManager = new ConnectorsManager({
                 nbConnectors: this._config.numberOfConnectors,
-                maximumBucketsPerConnector: this._maximumBucketsPerConnector,
                 database: this._database,
                 mongoUrl: this._mongoUrl,
                 oplogTopic: this._config.topic,
@@ -288,14 +287,14 @@ class OplogPopulator {
             });
             // For now, we always use the RetainBucketsDecorator
             // so, we map the events from the classes
-            this._connectorsManager.on('connector-destroyed', connector =>
-                this._allocationStrategy.onConnectorDestroyed(connector));
+            this._connectorsManager.on('connector-updated', connector =>
+                this._allocationStrategy.onConnectorUpdatedOrDestroyed(connector));
             this._allocator.on('bucket-removed', (bucket, connector) =>
                 this._allocationStrategy.onBucketRemoved(bucket, connector));
             this._connectorsManager.on('connectors-reconciled', bucketsExceedingLimit => {
-                this._metricsHandler.onConnectorsReconciliation(
+                this._metricsHandler.onConnectorsReconciled(
                     bucketsExceedingLimit,
-                    this._allocationStrategy.retainedBucketsNb,
+                    this._allocationStrategy.retainedBucketsCount,
                 );
             });
             // get currently valid buckets from mongo
@@ -340,7 +339,6 @@ class OplogPopulator {
             // words, we cannot alter an existing pipeline. In this
             // case, the strategy is to allow a maximum of one
             // bucket per kafka connector.
-            this._maximumBucketsPerConnector = 1;
             strategy = new ImmutableConnector({
                 logger: this._logger,
                 metricsHandler: this._metricsHandler,
@@ -351,12 +349,10 @@ class OplogPopulator {
             // kafka connector. However, we want to proactively
             // ensure that the pipeline will be accepted by
             // mongodb.
-            this._maximumBucketsPerConnector = constants.maxBucketsPerConnector;
             strategy = new LeastFullConnector({
                 logger: this._logger,
                 metricsHandler: this._metricsHandler,
                 connectorsManager: this._connectorsManager,
-                maximumBucketsPerConnector: this._maximumBucketsPerConnector,
             });
         }
         return new RetainBucketsDecorator(
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -263,7 +263,7 @@ class OplogPopulator {
             this._loadOplogHelperClasses();
             // initialize mongo client
             await this._setupMongoClient();
-            const allocationStrategy = this.initStrategy();
+            this._allocationStrategy = this.initStrategy();
             this._connectorsManager = new ConnectorsManager({
                 nbConnectors: this._config.numberOfConnectors,
                 maximumBucketsPerConnector: this._maximumBucketsPerConnector,
@@ -276,26 +276,26 @@ class OplogPopulator {
                 kafkaConnectHost: this._config.kafkaConnectHost,
                 kafkaConnectPort: this._config.kafkaConnectPort,
                 metricsHandler: this._metricsHandler,
-                allocationStrategy,
+                allocationStrategy: this._allocationStrategy,
                 logger: this._logger,
             });
             await this._initializeConnectorsManager();
             this._allocator = new Allocator({
                 connectorsManager: this._connectorsManager,
                 metricsHandler: this._metricsHandler,
-                allocationStrategy,
+                allocationStrategy: this._allocationStrategy,
                 logger: this._logger,
             });
             // For now, we always use the RetainBucketsDecorator
             // so, we map the events from the classes
             this._connectorsManager.on('connector-destroyed', connector =>
-                allocationStrategy.onConnectorDestroyed(connector));
+                this._allocationStrategy.onConnectorDestroyed(connector));
             this._allocator.on('bucket-removed', (bucket, connector) =>
-                allocationStrategy.onBucketRemoved(bucket, connector));
-            this._allocator.on('connectors-reconciled', bucketsExceedingLimit => {
+                this._allocationStrategy.onBucketRemoved(bucket, connector));
+            this._connectorsManager.on('connectors-reconciled', bucketsExceedingLimit => {
                 this._metricsHandler.onConnectorsReconciliation(
                     bucketsExceedingLimit,
-                    allocationStrategy.retainedBucketsNb,
+                    this._allocationStrategy.retainedBucketsNb,
                 );
             });
             // get currently valid buckets from mongo
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -11,6 +11,9 @@ const { ZenkoMetrics } = require('arsenal').metrics;
 const OplogPopulatorMetrics = require('./OplogPopulatorMetrics');
 const { OplogPopulatorConfigJoiSchema } = require('./OplogPopulatorConfigValidator');
 const { mongoJoi } = require('../../lib/config/configItems.joi');
+const ImmutableConnector = require('./allocationStrategy/ImmutableConnector');
+const RetainBucketsDecorator = require('./allocationStrategy/RetainBucketsDecorator');
+const LeastFullConnector = require('./allocationStrategy/LeastFullConnector');
 
 const paramsJoi = joi.object({
     config: OplogPopulatorConfigJoiSchema.required(),
@@ -242,35 +245,53 @@ class OplogPopulator {
         this._changeStreamWrapper.start();
     }
 
+    _arePipelinesImmutable() {
+        return semver.gte(this._mongoVersion, constants.mongodbVersionWithImmutablePipelines);
+    }
+
+    async _initializeConnectorsManager() {
+        return this._connectorsManager.initializeConnectors();
+    }
+
     /**
      * Sets the OplogPopulator
      * @returns {Promise|undefined} undefined
      * @throws {InternalError}
      */
     async setup() {
-       try {
-           this._loadOplogHelperClasses();
-           this._connectorsManager = new ConnectorsManager({
-               nbConnectors: this._config.numberOfConnectors,
-               database: this._database,
-               mongoUrl: this._mongoUrl,
-               oplogTopic: this._config.topic,
-               cronRule: this._config.connectorsUpdateCronRule,
-               prefix: this._config.prefix,
-               heartbeatIntervalMs: this._config.heartbeatIntervalMs,
-               kafkaConnectHost: this._config.kafkaConnectHost,
-               kafkaConnectPort: this._config.kafkaConnectPort,
-               metricsHandler: this._metricsHandler,
-               logger: this._logger,
+        try {
+            this._loadOplogHelperClasses();
+            // initialize mongo client
+            await this._setupMongoClient();
+            const allocationStrategy = this.initStrategy();
+            this._connectorsManager = new ConnectorsManager({
+                nbConnectors: this._config.numberOfConnectors,
+                maximumBucketsPerConnector: this._maximumBucketsPerConnector,
+                database: this._database,
+                mongoUrl: this._mongoUrl,
+                oplogTopic: this._config.topic,
+                cronRule: this._config.connectorsUpdateCronRule,
+                prefix: this._config.prefix,
+                heartbeatIntervalMs: this._config.heartbeatIntervalMs,
+                kafkaConnectHost: this._config.kafkaConnectHost,
+                kafkaConnectPort: this._config.kafkaConnectPort,
+                metricsHandler: this._metricsHandler,
+                allocationStrategy,
+                logger: this._logger,
             });
-            await this._connectorsManager.initializeConnectors();
+            await this._initializeConnectorsManager();
             this._allocator = new Allocator({
                 connectorsManager: this._connectorsManager,
                 metricsHandler: this._metricsHandler,
+                allocationStrategy,
                 logger: this._logger,
             });
-            // initialize mongo client
-            await this._setupMongoClient();
+            // For now, we always use the RetainBucketsDecorator
+            // so, we map the events from the classes
+            this._connectorsManager.on('connector-updated', connector =>
+                allocationStrategy.onConnectorUpdated(connector));
+            this._allocator.on('bucket-removed', (bucket, connector) =>
+                allocationStrategy.onBucketRemoved(bucket, connector));
             // get currently valid buckets from mongo
             const validBuckets = await this._getBackbeatEnabledBuckets();
             // listen to valid buckets
@@ -298,6 +319,44 @@ class OplogPopulator {
             });
             throw errors.InternalError.customizeDescription(err.description);
         }
+    }
+
+    /**
+     * Init the allocation strategy
+     * @returns {RetainBucketsDecorator} extended allocation strategy
+     * handling retained buckets
+     */
+    initStrategy() {
+        let strategy;
+        if (this._arePipelinesImmutable()) {
+            // In this case, mongodb does not support reusing a
+            // resume token from a different pipeline. In other
+            // words, we cannot alter an existing pipeline. In this
+            // case, the strategy is to allow a maximum of one
+            // bucket per kafka connector.
+            this._maximumBucketsPerConnector = 1;
+            strategy = new ImmutableConnector({
+                logger: this._logger,
+                metricsHandler: this._metricsHandler,
+                connectorsManager: this._connectorsManager,
+            });
+        } else {
+            // In this case, we can have multiple buckets per
+            // kafka connector. However, we want to proactively
+            // ensure that the pipeline will be accepted by
+            // mongodb.
+            this._maximumBucketsPerConnector = constants.maxBucketsPerConnector;
+            strategy = new LeastFullConnector({
+                logger: this._logger,
+                metricsHandler: this._metricsHandler,
+                connectorsManager: this._connectorsManager,
+                maximumBucketsPerConnector: this._maximumBucketsPerConnector,
+            });
+        }
+        return new RetainBucketsDecorator(
+            strategy,
+            { logger: this._logger },
+        );
     }
 
     /**
--- a/extensions/oplogPopulator/OplogPopulatorMetrics.js
+++ b/extensions/oplogPopulator/OplogPopulatorMetrics.js
@@ -33,6 +33,10 @@ class OplogPopulatorMetrics {
             help: 'Total number of buckets per connector',
             labelNames: ['connector'],
         });
+        this.bucketsExceedingLimit = ZenkoMetrics.createGauge({
+            name: 's3_oplog_populator_connector_buckets_exceeding_limit',
+            help: 'Total number of buckets exceeding the limit for all connectors',
+        });
         this.retainedBuckets = ZenkoMetrics.createGauge({
             name: 's3_oplog_populator_connector_retained_buckets',
             help: 'Current number of buckets still listened to by immutable connectors despite intended removal',
@@ -202,16 +206,20 @@ class OplogPopulatorMetrics {
     }
 
     /**
-     * updates s3_oplog_populator_connector_retained_buckets metric
-     * @param {number} count number of buckets retained
+     * updates metrics when the connectors are reconcilied
+     * @param {number} bucketsExceedingLimit number of buckets above the limit
+     * for all connectors
+     * @param {number} retainedBuckets number of buckets still listened to by
+     * immutable connectors despite intended removal
      * @returns {undefined}
      */
-    onRetainedBuckets(count) {
+    onConnectorsReconciliation(bucketsExceedingLimit, retainedBuckets) {
         try {
-            this.retainedBuckets.set(count);
+            this.bucketsExceedingLimit.set(bucketsExceedingLimit);
+            this.retainedBuckets.set(retainedBuckets);
         } catch (error) {
             this._logger.error('An error occured while pushing metric', {
-                method: 'OplogPopulatorMetrics.onRetainedBuckets',
+                method: 'OplogPopulatorMetrics.onConnectorsReconciliation',
                 error: error.message,
             });
         }
--- a/extensions/oplogPopulator/OplogPopulatorMetrics.js
+++ b/extensions/oplogPopulator/OplogPopulatorMetrics.js
@@ -33,6 +33,10 @@ class OplogPopulatorMetrics {
             help: 'Total number of buckets per connector',
             labelNames: ['connector'],
         });
+        this.retainedBuckets = ZenkoMetrics.createGauge({
+            name: 's3_oplog_populator_connector_retained_buckets',
+            help: 'Current number of buckets still listened to by immutable connectors despite intended removal',
+        });
         this.requestSize = ZenkoMetrics.createCounter({
             name: 's3_oplog_populator_connector_request_bytes_total',
             help: 'Total size of kafka connect request in bytes',
@@ -192,6 +196,22 @@ class OplogPopulatorMetrics {
         } catch (error) {
             this._logger.error('An error occured while pushing metric', {
                 method: 'OplogPopulatorMetrics.onConnectorRestarted',
+                error: error.message,
+            });
+        }
+    }
+
+    /**
+     * updates s3_oplog_populator_connector_retained_buckets metric
+     * @param {number} count number of buckets retained
+     * @returns {undefined}
+     */
+    onRetainedBuckets(count) {
+        try {
+            this.retainedBuckets.set(count);
+        } catch (error) {
+            this._logger.error('An error occured while pushing metric', {
+                method: 'OplogPopulatorMetrics.onRetainedBuckets',
                 error: error.message,
             });
         }
--- a/extensions/oplogPopulator/OplogPopulatorMetrics.js
+++ b/extensions/oplogPopulator/OplogPopulatorMetrics.js
@@ -206,20 +206,20 @@ class OplogPopulatorMetrics {
     }
 
     /**
-     * updates metrics when the connectors are reconcilied
+     * updates metrics when the connectors are reconciled
      * @param {number} bucketsExceedingLimit number of buckets above the limit
      * for all connectors
      * @param {number} retainedBuckets number of buckets still listened to by
      * immutable connectors despite intended removal
      * @returns {undefined}
      */
-    onConnectorsReconciliation(bucketsExceedingLimit, retainedBuckets) {
+    onConnectorsReconciled(bucketsExceedingLimit, retainedBuckets) {
         try {
             this.bucketsExceedingLimit.set(bucketsExceedingLimit);
             this.retainedBuckets.set(retainedBuckets);
         } catch (error) {
             this._logger.error('An error occured while pushing metric', {
-                method: 'OplogPopulatorMetrics.onConnectorsReconciliation',
+                method: 'OplogPopulatorMetrics.onConnectorsReconciled',
                 error: error.message,
             });
         }
--- a/extensions/oplogPopulator/OplogPopulatorMetrics.js
+++ b/extensions/oplogPopulator/OplogPopulatorMetrics.js
@@ -34,11 +34,11 @@ class OplogPopulatorMetrics {
             labelNames: ['connector'],
         });
         this.bucketsExceedingLimit = ZenkoMetrics.createGauge({
-            name: 's3_oplog_populator_connector_buckets_exceeding_limit',
+            name: 's3_oplog_populator_buckets_exceeding_limit',
             help: 'Total number of buckets exceeding the limit for all connectors',
         });
         this.retainedBuckets = ZenkoMetrics.createGauge({
-            name: 's3_oplog_populator_connector_retained_buckets',
+            name: 's3_oplog_populator_retained_buckets',
             help: 'Current number of buckets still listened to by immutable connectors despite intended removal',
         });
         this.requestSize = ZenkoMetrics.createCounter({
--- a/extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
+++ b/extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
@@ -24,10 +24,17 @@ class AllocationStrategy {
 
     /**
      * Assess if a pipeline can be updated
-     * @param {Connector} connector connector
      * @returns {Boolean} true if the connector can be updated
      */
-    canUpdate(connector) { // eslint-disable-line no-unused-vars
+    canUpdate() {
+        throw errors.NotImplemented;
+    }
+
+    /**
+     * Getter for the maximum number of buckets per connector
+     * @returns {Number} maximum number of buckets per connector
+     */
+    get maximumBucketsPerConnector() {
         throw errors.NotImplemented;
     }
 }
--- a/extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
+++ b/extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
@@ -4,21 +4,32 @@ class AllocationStrategy {
 
     /**
      * @constructor
-     * @param {Logger} logger logger object
+     * @param {Object} params params
+     * @param {Logger} params.logger logger object
      */
-    constructor(logger) {
-        this._logger = logger;
+    constructor(params) {
+        this._logger = params.logger;
     }
 
     /**
-     * Get best connector for assigning a bucket
-     * @param {Connector[]} connectors available connectors
-     * @returns {Connector} connector
+     * Get best connector to assign a bucket to.
+     * If no connector is available, null is returned.
+     * @param {Array<Connector>} connectors connectors
+     * @param {String} bucket bucket name
+     * @returns {Connector | null} connector
      */
-    getConnector(connectors) { // eslint-disable-line no-unused-vars
+    getConnector(connectors, bucket) { // eslint-disable-line no-unused-vars
         throw errors.NotImplemented;
     }
 
+    /**
+     * Assess if a pipeline can be updated
+     * @param {Connector} connector connector
+     * @returns {Boolean} true if the connector can be updated
+     */
+    canUpdate(connector) { // eslint-disable-line no-unused-vars
+        throw errors.NotImplemented;
+    }
 }
 
 module.exports = AllocationStrategy;
--- a/extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
+++ b/extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
@@ -22,11 +22,18 @@ class ImmutableConnector extends AllocationStrategy {
     /**
      * Assess if a pipeline can be updated. With the immutable
      * strategy, a connector cannot be updated.
-     * @param {Connector} connector connector
      * @returns {false} false
      */
-    canUpdate(connector) { // eslint-disable-line no-unused-vars
+    canUpdate() {
         return false;
+    }
+
+    /**
+     * Getter for the maximum number of buckets per connector
+     * @returns {Number} maximum number of buckets per connector
+     */
+    get maximumBucketsPerConnector() {
+        return 1;
     }
 }
 
--- a/extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
+++ b/extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
@@ -1,0 +1,33 @@
+const AllocationStrategy = require('./AllocationStrategy');
+
+/**
+ * @class ImmutableConnector
+ *
+ * @classdesc ImmutableConnector is an allocation
+ * strategy that always requires a new connector to be
+ * created for each bucket.
+ */
+class ImmutableConnector extends AllocationStrategy {
+    /**
+     * Get best connector to assign a bucket to.
+     * If no connector is available, null is returned.
+     * @param {Array<Connector>} connectors connectors
+     * @param {String} bucket bucket name
+     * @returns {Connector | null} connector
+     */
+    getConnector(connectors, bucket) { // eslint-disable-line no-unused-vars
+        return null;
+    }
+
+    /**
+     * Assess if a pipeline can be updated. With the immutable
+     * strategy, a connector cannot be updated.
+     * @param {Connector} connector connector
+     * @returns {false} false
+     */
+    canUpdate(connector) { // eslint-disable-line no-unused-vars
+        return false;
+    }
+}
+
+module.exports = ImmutableConnector;
--- a/extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
+++ b/extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
@@ -1,3 +1,4 @@
+const constants = require('../constants');
 const AllocationStrategy = require('./AllocationStrategy');
 
 /**
@@ -11,17 +12,6 @@ const AllocationStrategy = require('./AllocationStrategy');
 class LeastFullConnector extends AllocationStrategy {
 
     /**
-     * @constructor
-     * @param {Object} params params
-     * @param {Number} params.maximumBucketsPerConnector maximum number of buckets per connector
-     * @param {Logger} params.logger logger object
-     */
-    constructor(params) {
-        super(params);
-        this._maximumBucketsPerConnector = params.maximumBucketsPerConnector;
-    }
-
-    /**
      * Get best connector to assign a bucket to.
      * If no connector is available, null is returned.
      * @param {Array<Connector>} connectors connectors
@@ -33,7 +23,7 @@ class LeastFullConnector extends AllocationStrategy {
             return null;
         }
         const connector = connectors.reduce((prev, elt) => (elt.bucketCount < prev.bucketCount ? elt : prev));
-        if (connector.buckets.length >= this._maximumBucketsPerConnector) {
+        if (connector.buckets.length >= this.maximumBucketsPerConnector) {
             return null;
         }
         return connector;
@@ -41,11 +31,18 @@ class LeastFullConnector extends AllocationStrategy {
 
     /**
      * Assess if a pipeline can be updated.
-     * @param {Connector} connector connector
      * @returns {true} true
      */
-    canUpdate(connector) { // eslint-disable-line no-unused-vars
+    canUpdate() {
         return true;
+    }
+
+    /**
+     * Getter for the maximum number of buckets per connector
+     * @returns {Number} maximum number of buckets per connector
+     */
+    get maximumBucketsPerConnector() {
+        return constants.maxBucketsPerConnector;
     }
 }
 
--- a/extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
+++ b/extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
@@ -13,19 +13,39 @@ class LeastFullConnector extends AllocationStrategy {
     /**
      * @constructor
      * @param {Object} params params
+     * @param {Number} params.maximumBucketsPerConnector maximum number of buckets per connector
      * @param {Logger} params.logger logger object
      */
     constructor(params) {
-        super(params.logger);
+        super(params);
+        this._maximumBucketsPerConnector = params.maximumBucketsPerConnector;
     }
 
     /**
-     * Get best connector for assigning a bucket
-     * @param {Connector[]} connectors available connectors
-     * @returns {Connector} connector
+     * Get best connector to assign a bucket to.
+     * If no connector is available, null is returned.
+     * @param {Array<Connector>} connectors connectors
+     * @param {String} bucket bucket name
+     * @returns {Connector | null} connector
      */
-    getConnector(connectors) {
-        return connectors.reduce((prev, elt) => (elt.bucketCount < prev.bucketCount ? elt : prev));
+    getConnector(connectors, bucket) { // eslint-disable-line no-unused-vars
+        if (!connectors.length) {
+            return null;
+        }
+        const connector = connectors.reduce((prev, elt) => (elt.bucketCount < prev.bucketCount ? elt : prev));
+        if (connector.buckets.length >= this._maximumBucketsPerConnector) {
+            return null;
+        }
+        return connector;
+    }
+
+    /**
+     * Assess if a pipeline can be updated.
+     * @param {Connector} connector connector
+     * @returns {true} true
+     */
+    canUpdate(connector) { // eslint-disable-line no-unused-vars
+        return true;
     }
 }
 
--- a/extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
+++ b/extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
@@ -18,7 +18,7 @@ class RetainBucketsDecorator extends AllocationStrategy {
         super(params);
         this._strategy = strategy;
 
-        // Stores buckets that should are removed from the connector
+        // Stores buckets that should be removed from the connector
         // but still in use
         this._retainedBuckets = new Map();
     }
--- a/extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
+++ b/extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
@@ -27,7 +27,7 @@ class RetainBucketsDecorator extends AllocationStrategy {
      * Get the number of retained buckets
      * @returns {Number} number of retained buckets
      */
-    get retainedBucketsNb() { return this._retainedBuckets.size; }
+    get retainedBucketsCount() { return this._retainedBuckets.size; }
 
     /**
      * Callback when a bucket is removed from a connector
@@ -43,20 +43,12 @@ class RetainBucketsDecorator extends AllocationStrategy {
     }
 
     /**
-     * Callback when a connector is destroyed.
+     * Callback when a connector is destroyed or
+     * updated.
      * @param {Connector} connector connector
      * @returns {undefined}
      */
-    onConnectorDestroyed(connector) {
-        this._cleanupRetainedBucket(connector);
-    }
-
-    /**
-     * Cleanup retained buckets for a connector
-     * @param {Connector} connector connector
-     * @returns {undefined}
-     */
-    _cleanupRetainedBucket(connector) {
+    onConnectorUpdatedOrDestroyed(connector) {
         // When a connector is updated or destroyed, the retained
         // buckets are removed from the connector
         this._retainedBuckets.forEach((conn, bucket) => {
@@ -76,25 +68,24 @@ class RetainBucketsDecorator extends AllocationStrategy {
      * @returns {Connector | null} connector
      */
     getConnector(connectors, bucket) {
-        if (this._retainedBuckets.has(bucket)) {
-            return this._retainedBuckets.get(bucket);
-        }
-
-        return this._strategy.getConnector(connectors, bucket);
+        return this._retainedBuckets.get(bucket) ||
+            this._strategy.getConnector(connectors, bucket);
     }
 
     /**
-     * Assess if a pipeline can be updated. If the connector can
-     * be updated, the bucket is added as retained.
-     * @param {Connector} connector connector
+     * Assess if a pipeline can be updated.
      * @returns {Boolean} true if the connector can be updated
      */
-    canUpdate(connector) {
-        const res = this._strategy.canUpdate();
-        if (res) {
-            this._cleanupRetainedBucket(connector);
-        }
-        return res;
+    canUpdate() {
+        return this._strategy.canUpdate();
+    }
+
+    /**
+     * Getter for the maximum number of buckets per connector
+     * @returns {Number} maximum number of buckets per connector
+     */
+    get maximumBucketsPerConnector() {
+        return this._strategy.maximumBucketsPerConnector;
     }
 }
 
--- a/extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
+++ b/extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
@@ -1,0 +1,86 @@
+const AllocationStrategy = require('./AllocationStrategy');
+
+/**
+ * @class RetainBucketsDecorator
+ *
+ * @classdesc RetainBucketsDecorator is a decorator
+ * that retains buckets that are removed from the
+ * connector but still in use.
+ */
+class RetainBucketsDecorator extends AllocationStrategy {
+    /**
+     * @constructor
+     * @param {AllocationStrategy} strategy the strategy to decorate
+     * @param {Object} params params
+     * @param {Logger} params.logger logger object
+     */
+    constructor(strategy, params) {
+        super(params);
+        this._strategy = strategy;
+
+        // Stores buckets that should are removed from the connector
+        // but still in use
+        this._retainedBuckets = new Map();
+    }
+
+    /**
+     * Callback when a bucket is removed from a connector
+     * @param {String} bucket bucket name
+     * @param {Connector} connector connector
+     * @returns {undefined}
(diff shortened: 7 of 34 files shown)


monitoring/oplog-populator/alerts.yaml
```diff
@@ -36,6 +42,17 @@ groups:
...
+  - alert: KafkaConnectReplicasAboveThreshold
+    Expr: |
+      sum(up{namespace="${namespace}",job="${kafka_connect_job}"}) > ${oplogPopulatorChangeStreamSoftLimit}
+    For: "5m"
+    Labels:
+     severity: warning
+    Annotations:
+      zenko_service: backbeat-oplog-populator
+      description: "Kafka connect replica count is above the soft limit"
+      summary: "Kafka connect replica count is above the soft limit. If the number of buckets is too high, consider using a single replica for your cluster"
```
williamlardier, line 54: We may want to improve this message

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -260,7 +274,19 @@ class Connector {
...
+                    method: 'Connector.removeBucket',
+                    connector: this._name,
+                    bucket,
+                });
+            } else if (this._isPipelineImmutable) {
+                // If the pipeline is immutable and only one bucket is left,
+                // we can destroy the connector, so it will be recreated with
+                // a new bucket later.
+                return this.destroy();
+            }
```
williamlardier, line 288: The reason why we do this here is explained in the PR message

If we can safely delete a running connector, then we can decide to have this logic in the loop, to avoid deleting connectors possibly too frequently...

.github/workflows/alerts.yaml
```diff
@@ -81,4 +81,5 @@ jobs:
             namespace=zenko
             oplog_populator_job=artesca-data-backbeat-oplog-populator-headless
             oplogPopulatorChangeStreamLagThreshold=10
+            oplogPopulatorChangeStreamSoftLimit=10
```
francoisferrand: should be just `oplogPopulatorChangeStreamLimit`, as it is "soft" only in some case

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -242,35 +242,71 @@ class OplogPopulator {
...
+            if (this._config.singleChangeStream) {
+                this._maximumBucketsPerConnector = Infinity;
+                this._config.numberOfConnectors = 1;
+            }
+
+            this._connectorsManager = new ConnectorsManager({
+                nbConnectors: this._config.numberOfConnectors,
+                singleChangeStream: this._config.singleChangeStream,
+                maximumBucketsPerConnector: this._maximumBucketsPerConnector,
+                isPipelineImmutable: this._isPipelineImmutable(),
```
francoisferrand: is this needed?
we already "handle" immutability via setting `maximumBucketsPerConnector`, best minimize impact

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -242,35 +242,71 @@ class OplogPopulator {
...
+                // mongodb.
+                this._maximumBucketsPerConnector = constants.maxBucketPerConnector;
+            }
+            // If the flag useSingleChangeStream is set to true, we
+            // set the max number to infinity, and the number of connectors
+            // to 1.
+            if (this._config.singleChangeStream) {
+                this._maximumBucketsPerConnector = Infinity;
+                this._config.numberOfConnectors = 1;
+            }
```
francoisferrand: the last option (`singleChangeStream`) overrides the first ones, so should be more list:

```suggestion
            if (this._config.singleChangeStream) {
                // If the flag useSingleChangeStream is set to true, we
                // set the max number to infinity, and the number of connectors
                // to 1.
                this._maximumBucketsPerConnector = Infinity;
                this._config.numberOfConnectors = 1;
            } else if (this._isPipelineImmutable()) {
                // In this case, mongodb does not support reusing a
                // resume token from a different pipeline. In other
                // words, we cannot alter an existing pipeline. In this
                // case, the strategy is to allow a maximum of one
                // bucket per kafka connector.
                this._maximumBucketsPerConnector = 1;
            } else {
                // In this case, we can have multiple buckets per
                // kafka connector. However, we want to proactively
                // ensure that the pipeline will be accepted by
                // mongodb.
                this._maximumBucketsPerConnector = constants.maxBucketPerConnector;
            }
```

extensions/oplogPopulator/OplogPopulatorConfigValidator.js
```diff
@@ -10,6 +10,7 @@ const joiSchema = joi.object({
     probeServer: probeServerJoi.default(),
     connectorsUpdateCronRule: joi.string().default('*/1 * * * * *'),
     heartbeatIntervalMs: joi.number().default(10000),
+    singleChangeStream: joi.boolean().default(false),
```
francoisferrand: i think we should find a better name for this option.

this option is _impemented_ as a single change stream, but the semantic is really to listen to the whole oplog (instead of picking only the minimal relevant buckets). And in the future, we may/could/will be able to improve that case to still ingest the whole oplog, but over multiple change stream (e.g. inventing a "sharded change stream") : which should probably still be the same config option...

→ maybe `wholeOplog` (false by default) ? or an option defaulting to true, if we find a better name in that direction (like, but better than, `bucketFiltering` or something like that)
→ an alternative could be to activate this mode when `numberOfConnectors` is 0. Not really explicit, but avoids having "cascaded"/incompatible options (e.g. when this new mode is active, `numberOfConnectors` actually becomes irrelevant, and 1 is the only supported value). If we keep both option, should add an extra validation that `numberOfConnectors` is optional or must be equal to 1 when the new option is set (but is required and must be > 0 otherwise)
→ if possible, may be best to split support of this new mode in a separate PR, to possibly allow merging the other "required" option earlier

extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
```diff
@@ -6,26 +6,22 @@ const AllocationStrategy = require('./AllocationStrategy');
...
     /**
      * Get best connector for assigning a bucket
      * @param {Connector[]} connectors available connectors
      * @returns {Connector} connector
      */
     getConnector(connectors) {
-        return connectors.reduce((prev, elt) => (elt.bucketCount < prev.bucketCount ? elt : prev));
+        const connector = connectors.reduce((prev, elt) => (elt.bucketCount < prev.bucketCount ? elt : prev));
+        if (connector.shouldBeDestroyed || connector.buckets.length >= this._maximumBucketsPerConnector) {
+            return this._addConnector();
```
francoisferrand: I need to dig deeper into the `AllocationStrategy` interface and how it is used, but it feels like a break of the abstraction: i fear it may either break, create corner case or at least make things harder to understand...

→ At the very least, it is completely unexpected that a `getter` would actually have side effects.
→ Just looking at this specific code (again, would need to go back to look at the interface and its use), a better abstraction would be to let the caller of the strategy handle this condition ; and possibly (if needed) introduce a new kind of result for `getConnector`, e.g. if the result is `null` then a new connector would be used.

NOTE: just an idea, when `maxPerBucket` is set to 1, we actually don't really care about the existing bucket.... it is really about creating new buckets every time: so maybe creating a separate (and possibly trivial) `AllocationStrategy` (`ImmutableAllocationStrategy`) would be the clean way, and avoid the repeated search in a long list of connectors....
(though having the limit is handy in the old strategy as well, to handle the 262000 bucket limit of mongo query)

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -233,6 +256,9 @@ class Connector {
      * @throws {InternalError}
      */
     async addBucket(bucket, doUpdate = false) {
+        if (this._buckets.size >= this._maximumBucketsPerConnector) {
+            throw errors.InternalError.customizeDescription('Connector reached maximum number of buckets');
```
francoisferrand: won't having this check here break handling of upgrades?

e.g. the goal here is to keep existing connectors configured as they are, even if the have more than the expected number of buckets...

the problem of mongo 6 is not having more than 1 bucket per connector, but really that connectors must be immutable. A multi-bucket connector (pre-mongo 6) is fine afterwards as well, but we must simply:
- not try to add more buckets to it (if there are extra buckets, they should be allocated to new connectors)
- not removed buckets from it (after upgrade, we may listen to more buckets than "optimal")

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -260,7 +286,16 @@ class Connector {
         this._buckets.delete(bucket);
         this._updateConnectorState(true);
         try {
-            await this.updatePipeline(doUpdate);
+            if (this._isPipelineImmutable && this._buckets.size >= 1) {
+                this._logger.warn('Removing a bucket from an immutable pipeline', {
```
francoisferrand: if immutable, we must not remove a bucket from the list : we must really keep the existing buckets, unless we don't need to listen to any bucket and we can remove the connector.

francoisferrand: we may want to avoid warning here, as the oplog populator is reconciling very few seconds, and we expect this may stay for a long time... so it could be lots of logs for not much...

since the connectors will now be immutable, having a log only on startup ("connector has more than expected number of bucket") as it will stay that way almost forever (or until the happy case where we stop listening to every bucket).

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -260,7 +286,16 @@ class Connector {
...
+                    method: 'Connector.removeBucket',
+                    connector: this._name,
```
francoisferrand: I think we already have one case where we detect that we do not "respect" the config (e.g. when the number of connectors is reduced); to handle/report this case, we have a dedicated metrics I think: may be nice here to add some metric so that it becomes visible that there is a deviation (e.g. we listen to more buckets than expected, and/or we have connectors with more changestreams than expected)

We should also double check that we don't have metrics with a `connector` label, as it may explose if there are too many buckets to listen to... Probably no fix needed yet (kafka-connect will probably not scale to 262000 connectors anyway, in our current setup), but we should create a ticket...

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -274,11 +309,16 @@ class Connector {
...
-     * buckets assigned to this connector
+     * buckets assigned to this connector. If the
+     * singleChangeStream parameter is set to true,
+     * returns a pipeline that listens to all collections.
      * @param {string[]} buckets list of bucket names
      * @returns {string} new connector pipeline
      */
     _generateConnectorPipeline(buckets) {
+        if (this._singleChangeStream) {
+            return '[]';
```
francoisferrand: may still need to do some filtering, to pick up only the "bucket" collections (and no mpu shadow bucket)
not sure if the metastore needs to be in the stream as well (i got conflicting memories, not sure if we changed these to use dedicated change streams, or if oplog populator adds kind of a hardcoded bucket for the metastore...)

extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
francoisferrand: properly handling this will probably also help avoid the `shouldBeDestroyed` connector status, which looks weird (why would the connector indicate to itself that it should be destroyed, instead of destroying itself...) : maybe that handling would be better for the `Allocator`? (same, need to look more deeply at the code)

francoisferrand (changes requested):
The approach seems to go against in the "design" of the code, which makes it much harder to separate the concern and understand ("proof") that it is working well... And it may create corner cases.

There may obviously be more in the details, but from afar this change should:
* Adding a new `ImmutableAllocationStrategy` (maybe not needed at first, but would avoid going through the now possibly long list of connectors ; it could also help to handle the "histeresis" that we can have more than 1 bucket if the connector is already like that, but we must always allocate "new" buckets to new connectors)
* Updating the Allocator (and AllocationStrategy "contract") to support the new situations (e.g. need to allocate a new connector)
* Connector is class to manage a connector: it should not be affected, simply we will never reach a situation where we _change_ the list of buckets (thanks to the new allocator/strategy)
* Maybe some change to ConnectorsManager, as it bridges Allocator and Connectors, not sure

extensions/oplogPopulator/modules/Connector.js
ghost: We do have a dedicated changestream listening to the metastore collection, yes (in addition to the ones we create here)

extensions/oplogPopulator/modules/Connector.js
ghost: > won't having this check here break handling of upgrades?

Not for old connectors, we recreate a connector with a predefined list of buckets, so we do not "addBucket" in it

extensions/oplogPopulator/modules/Connector.js
ghost: All oplog populator metrics but two have a `connector` label... I created [BB-604](https://scality.atlassian.net/browse/BB-604).

ghost: > I think we already have one case where we detect that we do not "respect" the config (e.g. when the number of connectors is reduced)

We have the `s3_oplog_populator_connectors` metric that we decrement when removing a connector, so we can see that we do not respect the config indeed. If we want to have metrics do detect that we have a deviation, we will need to have a metric per connector, otherwise, it'll just be a counter of the number of "not modified" connectors (something like `s3_oplog_populator_connector_reconfiguration_skipped`). Do we have any preference?

extensions/oplogPopulator/OplogPopulatorConfigValidator.js
ghost: I created a dedicated PR: https://github.com/scality/backbeat/pull/2540
I went with the second option, I think this one is better as we do not need to introduce yet another config in the CR (we can udpate the number of connector's documentation though), in addition to the advantages you listed.

extensions/oplogPopulator/modules/Connector.js
francoisferrand: we could have a _gauge_ on the number of buckets "unexpectedly listened to", i.e. present in multi-bucket connector config from earlier version but not present anymore.
→ i.e. just count the number of buckets we want to remove, but cannot because the connector is immutable and not "empty"

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -242,35 +242,61 @@ class OplogPopulator {
         this._changeStreamWrapper.start();
     }
 
+    _isPipelineImmutable() {
```
francoisferrand: nit: there may be more than one pipeline, and every one of them is immutable

```suggestion
    _arePipelinesImmutable() {
```

extensions/oplogPopulator/constants.js
```diff
@@ -1,6 +1,11 @@
 const constants = {
     bucketMetastore: '__metastore',
     defaultConnectorName: 'source-connector',
+    // Max length in a pipeline is equal to the MongoDB BSON max document size,
+    // so 16MB. To allow for other parameters in the pipeline, we round the max
+    // to 16 MB (16777216B) / 64 (max length of a bucket name) ~= 260000
+    maxBucketPerConnector: 260000,
```
francoisferrand: nit:
```suggestion
    maxBucketsPerConnector: 260000,
```

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -224,6 +240,14 @@ class Connector {
         }
     }
 
+    /**
+     * Detects if the current connector is immutable
+     * @returns {boolean} true if immutable
+     */
+    _isImmutable() {
```
francoisferrand: this is not a property of the connector, does not need to be here

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -253,14 +281,23 @@ class Connector {
...
      * @throws {InternalError}
      */
     async removeBucket(bucket, doUpdate = false) {
         this._buckets.delete(bucket);
+        // Here, we must avoid removing the bucket from the
+        // bucketsInUse set, as the connector might still be
+        // using it in a deployed pipeline.
         this._updateConnectorState(true);
         try {
+            if (this._isImmutable()) {
```
francoisferrand: seems to break abstraction, this is not the role of the `Connector`...

Allocator/... decides what to do, and the Connector should just do it

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -29,28 +34,18 @@ class Allocator {
         joi.attempt(params, paramsJoi);
         this._connectorsManager = params.connectorsManager;
         this._logger = params.logger;
-        this._allocationStrategy = new LeastFullConnector({
-            logger: params.logger,
-        });
+        if (params.maximumBucketsPerConnector === 1) {
+            this._allocationStrategy = new ImmutableConnector({
+                logger: params.logger,
+                maximumBucketsPerConnector: params.maximumBucketsPerConnector,
```
francoisferrand: not needed, this is always `1`

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -60,7 +55,16 @@ class Allocator {
      * listening to the bucket
      */
     has(bucket) {
-        return this._bucketsToConnectors.has(bucket);
+        return this._connectorsManager.connectors.some(conn => conn.bucketsInUse.has(bucket));
```
francoisferrand: * this is inefficient, should keep the map
* it is the allocator's job to allocate, it should own it's data (respect levels of abstraction)

extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+class ImmutableConnector extends AllocationStrategy {
+    /**
+     * Get best connector for assigning a bucket
+     * @param {Connector[]} connectors available connectors
+     * @returns {Connector} connector
+     */
+    // eslint-disable-next-line no-unused-vars
+    getConnector(connectors) {
+        return null;
```
francoisferrand, line 19: code-wise, it may be a good place to return the "existing" connector (if any) which still had the extraneous connector

(but not sure, it may complexify a lot the removal of "empty" connector after every bucket was un-listened to)

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -60,7 +55,16 @@ class Allocator {
...
+    }
+
+    /**
+     * Get connector assigned to a bucket
+     * @param {string} bucket bucket name
+     * @returns {Connector | undefined} connector
+     */
+    get(bucket) {
+        return this._connectorsManager.connectors.find(conn => conn.bucketsInUse.has(bucket));
```
francoisferrand: same, not efficient

extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
ghost: Isn't this already done by the allocator? We keep track of the buckets->connectors, and here we try to rely on the real list (not the "in memory" list), so we already do that

extensions/oplogPopulator/modules/Connector.js
ghost: Moved to the Allocator.

extensions/oplogPopulator/modules/Allocator.js
ghost: Closing & keeping only the discussion above.

extensions/oplogPopulator/modules/Allocator.js
ghost: I did not see a simple solution for this. To effectively detect if a connector has a bucket, we must use the "inUse" Set, **not** the in-memory Set. But, due to the current architecture, we have:
- The Allocator that uses private variables from the ConnectorsManager to access the list of connector, and to create them
- The ConnectorsManager that creates the Connectors and periodically update the pipeline
- The Connector that holds the information about whether the bucket is in use or not

As a result, the Allocator has no way to know, using simple maps, if a bucket is currently in use or not: we consider it in use as soon as it is added, but it stops being in use after 0-5s once we remove it from the in-memory set. This operation being done outside of the Allocator, having a Map would require breaking the abstraction levels, or keeping all the buckets in memory forever...

---

So I went with this solution:

- We remove a bucket from the map only after each update of the pipelines, using an event from the ConnectorsManager. I tried to have something resilient against race conditions.
- We improve the `get` and `has` functions to detect if the bucket is in use, as the Allocator only needs to detect buckets in use: either it's in `.buckets` (pending deployment or in use), or not (the bucket was removed from memory, pending removal from the connector).

extensions/oplogPopulator/modules/Connector.js
ghost: I created the gauge, and added as a top-line info, along with the other numbers: https://github.com/scality/backbeat/pull/2539/commits/3965666c96af0b614aa473527019b71ca1283836

extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
francoisferrand: i mean the `Allocator` keeps track of already "allocated" buckets (not the really applied once) - synchronously ; but it could be the strategy's role to remember that the bucket (even though not allocated anymore) is still "pending" in some connector.

`getConnector()` is called when we need to find the connector to use for a 'new' bucket (i.e. one we don't already know/track), so it would be fitting to return the existing connector if any.

→ instead of modifying the Allocator, it would be very nice if we could just use a "wrapper" Strategy, which returns the existing connector is the bucket was "pending" deallocation; and just passes the call to underlying connector otherwise.

ghost: If I understand correctly, we want to avoid the new logic in the Allocator, and always remove the bucket from the map (i.e., we do not consider buckets "in use" in the Allocator), and we then rely on the strategy to automatically get the current connector, the one where the bucket is still pending, or a new connector? (maybe not a wrapper but a new method?)

While it would simplify the code, I do not see the difference in terms of complexity with [the comment above](https://github.com/scality/backbeat/pull/2539#discussion_r1760697035), as we would need to parse all the connectors to detect if one has the bucket in it?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -59,6 +68,7 @@ class Connector {
             kafkaConnectPort: params.kafkaConnectPort,
             logger: this._logger,
         });
+        this._maximumBucketsPerConnector = params.maximumBucketsPerConnector;
```
welansari: not used ?

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -33,6 +33,10 @@ class OplogPopulatorMetrics {
             help: 'Total number of buckets per connector',
             labelNames: ['connector'],
         });
+        this.bucketsNotDeletable = ZenkoMetrics.createGauge({
```
francoisferrand: nit: `retainedBuckets`

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -242,35 +242,61 @@ class OplogPopulator {
...
+                // In this case, we can have multiple buckets per
+                // kafka connector. However, we want to proactively
+                // ensure that the pipeline will be accepted by
+                // mongodb.
+                this._maximumBucketsPerConnector = constants.maxBucketsPerConnector;
+            }
+
+            this._connectorsManager = new ConnectorsManager({
+                nbConnectors: this._config.numberOfConnectors,
+                maximumBucketsPerConnector: this._maximumBucketsPerConnector,
```
francoisferrand: needed? should this not be in Allocator only?

extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
```diff
@@ -4,10 +4,13 @@ class AllocationStrategy {
...
-     * @param {Logger} logger logger object
+     * @param {Object} params params
+     * @param {Number} params.maximumBucketsPerConnector maximum number of buckets per connector
+     * @param {Logger} params.logger logger object
      */
-    constructor(logger) {
-        this._logger = logger;
+    constructor(params) {
+        this._logger = params.logger;
+        this._maximumBucketsPerConnector = params.maximumBucketsPerConnector;
```
francoisferrand: does not seem like a general parameter of AllocationStrategy, it is used only in the `LeastFullStrategy` ?
→ should be moved there

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -29,13 +34,36 @@ class Allocator {
...
         this._metricsHandler = params.metricsHandler;
         // Stores connector assigned for each bucket
         this._bucketsToConnectors = new Map();
         this._initConnectorToBucketMap();
+        // Stores the buckets that are to be removed from the map
+        // after the next connectorsManager update
+        this._bucketsToConnectorsToClear = new Set();
+        this._connectorsManager.on('update', () => {
+            const bucketsToRemove = new Set(this._bucketsToConnectorsToClear);
+            this._bucketsToConnectorsToClear.clear();
```
francoisferrand: why duplicate the `Set` ?

there is no race condition (this is javascript & synchronous code), and we can just `clear()` or reset (`= newSet()`) the object attribute after processing the list of buckets to clear...

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -111,16 +164,27 @@ class Allocator {
...
-                    bucket,
-                    connector: connector.name,
-                });
+                // If the pipelines are immutable, we only remove the bucket
+                // from the connector if it is the single bucket in use. In
+                // this case, the connector will be either deleted by the
+                // connectorManager, or reused, if a bucket with the same
+                // name is created. Otherwise, the connector would be
+                // reconfigured during the next connectorsManager update.
+                if (!this._isImmutable() || connector.bucketsInUse.size === 1) {
```
francoisferrand: this condition (`!this._isImmutable() || connector.bucketsInUse.size === 1`) looks like it should be part of the `AllocatorStrategy` : should extend the interface to abstract this away

monitoring/oplog-populator/alerts.yaml
```diff
@@ -36,6 +39,18 @@ groups:
       description: "Oplog populator failed to configure connector"
       summary: "Oplog populator couldn't update kafka connect connector"
 
+  - alert: KafkaConnectReplicasAboveThreshold
```
welansari: I feel like the name of the alert might be missleading, as the user might think we are talking about the k8s replicas of kafka connect ?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -253,11 +273,15 @@ class Connector {
...
-     * @returns {Promise|undefined} undefined
+     * @returns {Promise<boolean>} true if bucket was removed from
+     * the bucket
      * @throws {InternalError}
      */
     async removeBucket(bucket, doUpdate = false) {
         this._buckets.delete(bucket);
+        // Here, we must avoid removing the bucket from the
+        // bucketsInUse set, as the connector might still be
+        // using it in a deployed pipeline.
```
francoisferrand: this looks very complex : we manage the "not removing" in many places (Allocator, AllocatorStrategy, and Connector)... really the goal is to divide the problem, and end-up with the things in a single place, to make them easy to reason with.

In the current state of the PR, we have the logic in here (`bucketInUse`) as well as in Allocator (`_bucketsToConnectorsToClear`) which seems fishy, w/ not so "symetric" add/remove...

- As Connector really manages a connector, it seems that in case of immutable, we should basically "just" not call `removeBucket` (unless it is the "last" bucket) nor `addBucket` (unless it is the first one)...
- To integrate in the current code, we "simply" need to inject some logic in between Allocator and Connector to delay the removal from the connector (i.e. keep a separate map connector -> "retained buckets")
- That same info can be used in reverse to re-use the same connector if the bucket was retained (→ may be a good fit in a Strategy)

but not sure if there is a deeper issue I don't see...

extensions/oplogPopulator/OplogPopulator.js
ghost: We just need it for the warning/metric when we handle old connectors; but not for logic-wise indeed. I am keeping it for the metric.

monitoring/oplog-populator/alerts.yaml
ghost: I renamed it as `KafkaConnectConnectorsAboveThreshold`, the description also helps understanding, what do you think?

extensions/oplogPopulator/modules/Allocator.js
ghost: I updated the strategy class, the new method is `canRemoveBucket`, it always returns true for the LeastFullConnector, and true if we only have one bucket, for the Immutable one (Immutable connectors may have more than one bucket when we retrieve the old connectors on startup).

monitoring/oplog-populator/alerts.yaml
ghost: Finally it is now `KafkaConnectTooManyConnectors`.

extensions/oplogPopulator/modules/Connector.js
ghost: Here is my attempt after our discussion: https://github.com/scality/backbeat/pull/2539/commits/d4bce59e27e11458a178160a5b7f92e9fde8d670 (edited!)

-> I basically moved the responsibility to "allocate" to the strategies,; as having the connectorsManager & list of retained buckets means we basically have access to everything. So it includes the map of bucket-> connectors, notification from the connectorsManager that the pipelines are updated to clear the retained buckets, and handling of the logic to listen/stop listen to a bucket. This way, the Allocator is very simple and wan work with any strategy. the strategies are now a bit more complex, but it is very linear... Please let me know what you think.

extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
francoisferrand: we don't want to "avoid the new logic in the Allocator" ; we want to leverage the interface to abstract implementation details, so that we add additional, composable logic in separate class instead of making a bigger and bigger class which covers too many aspects to be easily understandable.

Each class should end up with their own responsability (somewhat as they already have), and clear contract between them.

so there may be some logic added in Allocator (typically the "create connector if getConnector returns null" logic is what would be expected: the contract has been extended to support a bit more) ; and maybe the contract needs to be extended to cover the whole semantic

extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
```diff
@@ -6,26 +6,69 @@ const AllocationStrategy = require('./AllocationStrategy');
...
-        return connectors.reduce((prev, elt) => (elt.bucketCount < prev.bucketCount ? elt : prev));
+    async getOrAssignBucketConnector(bucket) {
+        let connector;
+        if (this._bucketsToConnectors.has(bucket)) {
+            return this._bucketsToConnectors.get(bucket);
+        }
+        if (this._retainedBuckets.has(bucket)) {
+            connector = this._retainedBuckets.get(bucket);
+        } else if (this._connectorsManager.connectors.length === 0) {
+            connector = this._connectorsManager.addConnector();
```
francoisferrand: missing `await`, not sure what we would end up with...

extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
```diff
@@ -0,0 +1,57 @@
...
+            return this._bucketsToConnectors.get(bucket);
+        } else if (this._retainedBuckets.has(bucket)) {
+            connector = this._retainedBuckets.get(bucket);
+        } else if (!connector) {
+            connector = await this._connectorsManager.addConnector();
+        }
+        this._bucketsToConnectors.set(bucket, connector);
+        await connector.addBucket(bucket);
+        return connector;
+    }
```
francoisferrand: Lots of duplication between the implementations, the only difference is really when we chose to reuse the connector

--> the older interface seemed better, allowing to keep only the "choice" in the Strategy.

---

Specifically for `_retainedBuckets`, it breaks layers as it is really managed in the base class (which is not really the role of an interface) but it is used here.

The cleaner way to do this is to add a separate `RetainBucket` "proxy" AllocationStrategy, which would look something like this:

```
class RetainBuckets extends AllocationStrategy {
    constructor(strategy, params) {
        this._strategy = strategy;
        this._retainedBuckets = new Map();

        params.connectorsManager.on('update', () => {
            this._retainedBuckets.forEach(bucket => {
                if (!this._bucketsToConnectors.has(bucket)) {
                    this._retainedBuckets.delete(bucket);
                }
            });
            this._retainedBuckets.clear();
        });
    }

    getBucketConnector(bucket) {
        if (this._retainedBuckets.has(bucket)) {
            return this._retainedBuckets.get(bucket);
        }

        return this._strategy.getBucketConnector(bucket)
    }

    // the interface expects to return TRUE if the connector must be removed, or FALSE to keep it
    removeBucketFromConnector(bucket, connector) {   
        const res = this._strategy.removeBucketFromConnector(bucket, connector);
        if res {
            this._retainedBuckets.set(bucket, connector);
        }
        return res;
    }
}
```

This way we mananged the retained buckets in a single place, and the "retaining" code is neither duplicated nor intermingled with the rest of the code.

extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Delete a bucket from a connector, if it can be
+     * deleted
+     * @param {string} bucket bucket name
+     * @returns {Promise<Connector | null>} the modified connector returned if modified
+     */
+    async removeBucketFromConnector(bucket) {
```
francoisferrand: the only difference between both implementations is this line condition: `connector?.buckets.length === 1` vs `connector != nil`

--> this indicates the abstraction for this part of the interface is not at the right level

Having a simpler (non async) `removeBucketFromConnector(bucket, connector)` which simply returns true/false to indicate if the connector should be removed would do the job in a simpler way.

extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
```diff
@@ -0,0 +1,57 @@
...
+        // If the pipelines are immutable, we only remove the bucket
+        // from the connector if it is the single bucket in use. In
+        // this case, the connector will be either deleted by the
+        // connectorManager, or reused, if a bucket with the same
+        // name is created. Otherwise, the connector would be
+        // reconfigured during the next connectorsManager update.
+        const connector = this._bucketsToConnectors.get(bucket);
+        if (connector?.buckets.length === 1) {
```
francoisferrand: if we have an "old" connector with multiple buckets, it will never be removed with this implementation?

not sure this is a _practical_ issue, but it leaves a unexpected behavior, possibly hitting us in the back later, when we least expect it...

extensions/oplogPopulator/modules/Connector.js
francoisferrand: Too much duplication is a sign that we have the bad level of abstraction...

With your current approach this can be easily fixed by having 2 methods, 1 "generic" in `AllocationStrategy` (which does the allocation) and 1 "simple" the actual strategy implementations.... However this means the AllocationStrategy is not an interface anymore, but really an Allocator...

C.f. another comment on how to deal with the retained bucket "in isolation". And looking at the code I did not see a place where the strategies needed the bucket->connector mapping.

extensions/oplogPopulator/OplogPopulator.js
francoisferrand: could the metric be managed by the `Allocator` ?
as it is fundamentally a metric which verifies that the allocator did its job...

extensions/oplogPopulator/allocationStrategy/ImmutableConnector.js
francoisferrand: thinking on this, it seems we do not have the right "hooks" in the interface for handling removal:
* the `Retain` strategy needs to _know_ when a bucket is removed from the connector, just like it needs to know when the connectors configs are actually "applied"; but the other strategies do not need it --> this should just be another event (or call) from `ConnectorsManager`, only in this specific strategy
* what is part of the strategy is dedicing when to actually "change" the connector configuration, i.e. if the actual update should be skipped (`connector.updatePipeline()`) : so maybe we should simply have a `canUpdate(connector) -> bool` method instead, which gets called from whereever appropriate (conenctorsManager maybe?)

NOTE:
* Indeed, it makes the `AllocationStrategy` not deal only with allocation, but also with the actual applying of configuration... but should be a single strategy to ensure consistency, I think... This however means the strategy should probably be a parameter of the Allocator, and the "choice" of strategy moved to oplogPopulator
* The "extra retained buckets" metric may thus be nicely generated in this new `canUpdate()` metric: as the function will naturally see when it prevents actually 'removing' buckets from a connector, and we will thus listen to more buckets than needed...

extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
```diff
@@ -4,21 +4,68 @@ class AllocationStrategy {
...
+        this._initConnectorToBucketMap();
+
+        this._connectorsManager.on('update', () => {
+            this._retainedBuckets.forEach(bucket => {
+                if (!this._bucketsToConnectors.has(bucket)) {
+                    this._retainedBuckets.delete(bucket);
+                }
+            });
+            this._retainedBuckets.clear();
+        });
```
francoisferrand: I don't think this covers all cases.

- When we have an "old" connector, listening to multiple buckets : it may be kept longer, while the bucket will have been removed from `_bucketsToConnectors` already
- The (multiple) connectors are not updated synchronously, so we cannot clear everything atomically

--> should probably have an event when each connector is actually updated, to clear the retained buckets map accordingly:

```suggestion
        this._connectorsManager.on('update', updatedConnector => {
            this._retainedBuckets.forEach((bucket, connector) => {
                if (updatedConnector === connector) {
                    this._retainedBuckets.delete(bucket);
                }
            });
        });
```

monitoring/oplog-populator/dashboard.py
```diff
@@ -182,6 +183,17 @@ class Metrics:
...
+    title='Retained Buckets',
+    dataSource='${DS_PROMETHEUS}',
+    gradientMode='opacity',
+    lineInterpolation='smooth',
+    legendDisplayMode='hidden',
+    fillOpacity=10,
+    spanNulls=True,
+    targets=[Target(
+        expr='sum(' + Metrics.BUCKETS_RETAINED.raw() + ')',
+    )])
```
francoisferrand: do we really want to _display_ that information?
this should not be a nominal case, just having the alert is probably enough IMHO...

monitoring/oplog-populator/dashboard.py
```diff
@@ -336,6 +348,7 @@ class Metrics:
                 total_number_buckets,
                 max_pipeline_size,
                 bandwidth,
+                retained_buckets,
```
francoisferrand: grafana layout does not divide by 7, so if we keep this extra panel we need to manually set some panel size to nicely fill the line.

monitoring/oplog-populator/alerts.yaml
```diff
@@ -36,6 +39,18 @@ groups:
...
+    Expr: |
+        sum(s3_oplog_populator_connectors{job="${oplog_populator_job}",namespace="${namespace}"})
+        > ${oplogPopulatorChangeStreamLimit}
+    For: "5m"
+    Labels:
+     severity: warning
+    Annotations:
+      zenko_service: backbeat-oplog-populator
+      description: "Kafka connector count is above the soft limit"
+      summary: "Kafka connector count is above the soft limit. If the number of buckets is too high, consider using a single replica for your cluster"
```
francoisferrand: this is the other way round. `Summary` is a (short) summary, while `Description` should give the details.
```suggestion
      summary: "Kafka connector count is above the soft limit"
      description: "Kafka connector count is above the soft limit. If the number of buckets is too high, consider using a single replica for your cluster"
```

monitoring/oplog-populator/alerts.yaml
```diff
@@ -36,6 +39,18 @@ groups:
...
+
```
francoisferrand, line 53: after an upgrade, we may have buckets we unexpectedly listen to (not just as "temporary" retained), but longer term : e.g. we remove a workflow rule, but the connector is used for multiple buckets

maybe we want to alert when the number of retained buckets is stable for "some time" (a few minutes).

this also means we may not want to update the metric continuously, but possibly only at the end of the "reconciliation" : having retained bucket in the middle of the reconciliation is expected and normal (though unlikely), having them after the reconciliation means we listen to more buckets than we should...

(or maybe we can just ignore it, since this is kind of normal and not really an issue? the only "risk" i see if a user tries to disable workflow on a high-traffic bucket to lower the load via oplog, and this kind of situation prevents the change to be effective....)

extensions/oplogPopulator/OplogPopulator.js
ghost: I don't think so (without changing the current logic) - the handling of old bucket is fully done outside of the Allocator today, it is managed by OplogPopulator, this is also why it works using the immutable strategy and still create connectors with more than 1 bucket...

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -33,6 +33,10 @@ class OplogPopulatorMetrics {
             help: 'Total number of buckets per connector',
             labelNames: ['connector'],
         });
+        this.retainedBuckets = ZenkoMetrics.createGauge({
+            name: 's3_oplog_populator_connector_retained_buckets',
+            help: 'Total number of buckets still listened to by immutable connectors despite intended removal',
```
francoisferrand: nit: this is not the "total number" since it is a gauge.
More like the "current number"

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -235,19 +255,23 @@ class ConnectorsManager {
...
-                    connector: connector.name
+                    connector: connector.name,
                 });
                 return true;
-            } else if (connector.isRunning) {
-                return connector.updatePipeline(true);
+            } else if (connector.isRunning && this._allocationStrategy.canUpdate(connector)) {
+                const isConnectorUpdated = connector.updatePipeline(true);
+                if (isConnectorUpdated) {
+                    this._allocationStrategy.onConnectorUpdated(connector);
```
francoisferrand: this is kind of a stretch : `onBucketRemoved` and `onConnectorUpdated` is not actually part of `AllocationStrategy` interface, but only of `RetainBucketsDecorator` implementation.

It works only because this is Javascript (would not build with typescript), and because RetainBucketsDecorator is the "first" strategy in the chain (i.e. the abstraction is broken, and we have coupling between the "choice" of strategy and its "users")

So need either:
* Apply inversion of control: e.g. Allocator and ConnectorsManager emit an event or allow registering a callback, so that `RetainBucketsDecorator` can register for these 'events'
* Extend the AllocationStrategy interface to have the extra events (even though it seems a bit of a stretch to make them part of the interface)
* Introduce one (or two) interfaces for this/these 2 events. This interface would be implemented by `RetainBucketsDecorator` only (as today), the only difference would be that ConnectorsManager and Allocator would get 2 objects: one strategy and the object having these interfaces... (in typepscript, we could materialize with multiple interfaces; in other languages with multiple inheritance)
* Similar to the last approach above, these classes could take `onConnectorUpdated` and `onBucketRemoved` callbacks in their constructor, which would be implemented by binding them to the `RetainBucketsDecorator` methods...

monitoring/oplog-populator/alerts.yaml
```diff
@@ -8,6 +8,9 @@ x-inputs:
 - name: oplogPopulatorChangeStreamLagThreshold
   type: config
   value: 10
+- name: oplogPopulatorChangeStreamLimit
+  type: config
+  value: 0
```
francoisferrand, line 13: default value should be the value that we used in default zenko value (10), so that we get somewhat consistent default if there is some mismatch (like older zkop w/ newer backbeat)

monitoring/oplog-populator/dashboard.json
francoisferrand, line 1: some mistake has been introduced here, the metrics should not have been renamed (in the json dashboard) to strip the s3 prefix...

extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,86 @@
...
+     * Assess if a pipeline can be updated. If the connector can
+     * be updated, the bucket is added as retained.
+     * @param {String} bucket bucket name
+     * @param {Connector} connector connector
+     * @returns {Boolean} true if the connector can be updated
+     */
+    canUpdate(bucket, connector) {
+        const res = this._strategy.canUpdate();
+        if (res) {
+            this._retainedBuckets.set(bucket, connector);
```
francoisferrand: i don't think this is the right place : the bucket should be marked as retained as soon as it is "removed" from the connector, i.e. in `onBucketRemoved`

extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Callback when a bucket is removed from a connector
+     * @param {String} bucket bucket name
+     * @returns {undefined}
+     */
+    onBucketRemoved(bucket) {
+        // When a bucket is removed, it is not immediately removed
+        // from the connector. This allows to retain the bucket
+        // in the connector until it is removed.
+        this._retainedBuckets.delete(bucket);
```
francoisferrand: when bucket is removed from the connector, it actually _becomes_ retained

```suggestion
        this._retainedBuckets.set(bucket, connector);
```

extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Callback when a connector is updated
+     * @param {Connector} connector connector
+     * @returns {undefined}
+     */
+    onConnectorUpdated(connector) {
```
francoisferrand: this should also be called when the connector is actually _destroyed_

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -235,19 +255,23 @@ class ConnectorsManager {
                 this._metricsHandler.onConnectorDestroyed();
```
francoisferrand, line 252: should call `RetainBucketsDecorator.onConnectorUpdated(connector)` around here, so that the buckets marked as retained can be cleared

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -165,6 +177,15 @@ class ConnectorsManager {
                     kafkaConnectHost: this._kafkaConnectHost,
                     kafkaConnectPort: this._kafkaConnectPort,
                 });
```
francoisferrand, line 177: * what you compute here is not the number of retained buckets (i.e. the ones which are unexpectedly listened to), but only the deviation from the max number of buckets (which may not be an issue it itself)
* the number of "extra" buckets is not `buckets.length`, but the sum of extra buckets for each connector

---

typically, if the platform was setup to listen more than 10 buckets:
* after upgrade to mongo 6 & al, indeed we will have one or more connector which listen to more than 1 bucket --> not very important, just not the "new" nominal state
* eventually, the user stop listening to some bucket --> now we will have "real" inefficiency, as we will keep listening to that bucket

--> The first info may be represented as the number of connector which have more buckets than expected ; or the "total" number of buckets which do not fit in the list (e.g. sum(buckets.length - _maximumBucketsPerConnector) for every connector)
--> The seocnd info is the number entries in retainBuckets map at the end of the periodic update of all connectors

extensions/oplogPopulator/modules/ConnectorsManager.js
ghost: As discussed, I chose to bind the events from the classes from the oplog populator, as we know there that we use the RetainBucket strategy (may change someday, but for now, we only do that).
I chose events because it allows the not alter the classes: they will just emit events.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -231,24 +253,27 @@ class ConnectorsManager {
     async _spawnOrDestroyConnector(connector) {
         try {
             if (connector.isRunning && connector.bucketCount === 0) {
+                this.emit('connector-updated', connector);
```
francoisferrand: this is more a 'connector-destroyed' event

extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,86 @@
...
+        this._retainedBuckets.set(bucket, connector);
+    }
+
+    /**
+     * Callback when a connector is updated
+     * @param {Connector} connector connector
+     * @returns {undefined}
+     */
+    onConnectorUpdated(connector) {
+        // When a connector is updated or destroyed, the retained
```
francoisferrand: there is a single code, but the way it is integrated can cause confusion...
may be more readable to have the "public" calkback (onConnectorDestroy), for binding in oplog populator ; and an internal function to do the actual removal on destroy or update

```suggestion
    onConnectorDestroy(connector) {
        this._cleanupRetainedBuclet(connector)
    }

    _cleanupRetainedBuclet(connector) {
        // When a connector is updated or destroyed, the retained
```

extensions/oplogPopulator/modules/ConnectorsManager.js
ghost: I updated to add these two metrics

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -242,35 +245,59 @@ class OplogPopulator {
...
                 logger: this._logger,
             });
-            // initialize mongo client
-            await this._setupMongoClient();
+            // For now, we always use the RetainBucketsDecorator
+            // so, we map the events from the classes
+            this._connectorsManager.on('connector-destroyed', connector =>
+                allocationStrategy.onConnectorDestroyed(connector));
+            this._allocator.on('bucket-removed', (bucket, connector) =>
+                allocationStrategy.onBucketRemoved(bucket, connector));
```
francoisferrand: nit: can we do this?

```suggestion
            this._connectorsManager.on('connector-destroyed', allocationStrategy.onConnectorDestroyed);
            this._allocator.on('bucket-removed', allocationStrategy.onBucketRemoved);
```

extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,101 @@
...
+        // Stores buckets that should are removed from the connector
+        // but still in use
+        this._retainedBuckets = new Map();
+    }
+
+    /**
+     * Get the number of retained buckets
+     * @returns {Number} number of retained buckets
+     */
+    get retainedBucketsNb() { return this._retainedBuckets.size; }
```
francoisferrand: nit: retainedBucketsCount

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -165,12 +180,22 @@ class ConnectorsManager {
...
+                        allowed: this._maximumBucketsPerConnector,
+                    });
+                }
                 this._logger.debug('Successfully retreived old connector', {
                     method: 'ConnectorsManager._getOldConnectors',
                     connector: connector.name
                 });
                 return connector;
             }));
+            this.emit('connectors-reconciled', bucketsExceedingLimit);
```
francoisferrand: `_getOldConnectors` is only called once on startup, from `initializeConnectors()`.

the actual reconcilization is done in `_updateConnectors()`

--> the event must be called at the end of `_updateConnectors()`, and maybe (but not sure it is strictly needed?) at the end of `initializeConnectors()`

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -196,6 +204,26 @@ class OplogPopulatorMetrics {
...
+
+    /**
+     * updates metrics when the connectors are reconcilied
+     * @param {number} bucketsExceedingLimit number of buckets above the limit
+     * for all connectors
+     * @param {number} retainedBuckets number of buckets still listened to by
+     * immutable connectors despite intended removal
+     * @returns {undefined}
+     */
+    onConnectorsReconciliation(bucketsExceedingLimit, retainedBuckets) {
```
francoisferrand: nit: `onConnectorsReconciled`

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -196,6 +204,26 @@ class OplogPopulatorMetrics {
             });
         }
     }
+
+    /**
+     * updates metrics when the connectors are reconcilied
```
francoisferrand: typo

```suggestion
     * updates metrics when the connectors are reconciled
```

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -231,24 +255,27 @@ class ConnectorsManager {
...
                     method: 'ConnectorsManager._spawnOrDestroyConnector',
-                    connector: connector.name
+                    connector: connector.name,
                 });
                 return true;
-            } else if (connector.isRunning) {
-                return connector.updatePipeline(true);
+            } else if (connector.isRunning && this._allocationStrategy.canUpdate(connector)) {
+                const isConnectorUpdated = connector.updatePipeline(true);
+                return isConnectorUpdated;
```
francoisferrand: keep as before:

```suggestion
                return connector.updatePipeline(true);
```

extensions/oplogPopulator/OplogPopulator.js
ghost: No, it breaks the unit tests

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -242,35 +245,59 @@ class OplogPopulator {
...
-               logger: this._logger,
+        try {
+            this._loadOplogHelperClasses();
+            // initialize mongo client
+            await this._setupMongoClient();
+            this._allocationStrategy = this.initStrategy();
+            this._connectorsManager = new ConnectorsManager({
+                nbConnectors: this._config.numberOfConnectors,
+                maximumBucketsPerConnector: this._maximumBucketsPerConnector,
+                database: this._database,
```
francoisferrand: nit: may actually not be needed to pass `maximumBucketsPerConnector`, it actually really comes from the strategy... (i.e. add in the interface API, and just return the mongo limit from LeastFullConnector)

tests/unit/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,112 @@
...
+                decorator._retainedBuckets.set('bucket1', connector1);
+                decorator._retainedBuckets.set('bucket2', connector2);
+                decorator.onConnectorDestroyed(connector2);
+                assert.strictEqual(decorator._retainedBuckets.size, 1);
+                assert.strictEqual(decorator._retainedBuckets.get('bucket1'), connector1);
+            });
+        });
+
+        describe('canUpdate', () => {
+            it('Should return the strategy result', async () => {
```
francoisferrand:
```suggestion
            it('should return the strategy result', async () => {
```

tests/unit/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,112 @@
...
+        describe('onBucketRemoved', () => {
+            it('should add bucket from retained buckets', () => {
+                decorator.onBucketRemoved('bucket1', connector1);
+                assert.strictEqual(decorator._retainedBuckets.size, 1);
+                assert.strictEqual(decorator._retainedBuckets.get('bucket1'), connector1);
+            });
+        });
+
+        describe('getConnector', () => {
+            it('Should return the retrained bucket\'s connector if it exists', () => {
```
francoisferrand:
```suggestion
            it('should return the retrained bucket\'s connector if it exists', () => {
```

tests/unit/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,112 @@
...
+                decorator._retainedBuckets.set('bucket1', connector1);
+                const result = decorator.getConnector([connector1, connector2], 'bucket1');
+                assert.strictEqual(result, connector1);
+            });
+
+            it('Should call the strategy if the bucket is not retained', () => {
```
francoisferrand:
```suggestion
            it('should call the strategy if the bucket is not retained', () => {
```

tests/unit/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,112 @@
...
+            });
+        });
+
+        describe('canUpdate', () => {
+            it('Should return the strategy result', async () => {
+                const result = await decorator.canUpdate(connector1);
+                assert.strictEqual(result, strategy.canUpdate(connector1));
+            });
+
+            it('Should remove from retained buckets if the strategy allows', async () => {
```
francoisferrand:
```suggestion
            it('should remove from retained buckets if the strategy allows', async () => {
```

extensions/oplogPopulator/modules/ConnectorsManager.js
ghost: done, also adapted the logic

extensions/oplogPopulator/allocationStrategy/LeastFullConnector.js
```diff
@@ -47,6 +47,14 @@ class LeastFullConnector extends AllocationStrategy {
     canUpdate(connector) { // eslint-disable-line no-unused-vars
         return true;
     }
+
+    /**
+     * Getter for the maximum number of buckets per connector
+     * @returns {Number} maximum number of buckets per connector
+     */
+    get maximumBucketsPerConnector() {
+        return this._maximumBucketsPerConnector;
```
francoisferrand: should just return the constant, it does not need to be parameter?
(or if it needs to customizable, e.g. for test : then the default should be the constant from mongo, and remove the variable from oplogPopulator.js as well)

extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,109 @@
...
+     * @param {String} bucket bucket name
+     * @returns {Connector | null} connector
+     */
+    getConnector(connectors, bucket) {
+        if (this._retainedBuckets.has(bucket)) {
+            return this._retainedBuckets.get(bucket);
+        }
+
+        return this._strategy.getConnector(connectors, bucket);
+    }
```
francoisferrand: nit: avoid duplicate Map lookup

```suggestion
        return this._retainedBuckets.get(bucket) || this._strategy.getConnector(connectors, bucket);
    }
```

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -92,6 +102,9 @@ class Allocator {
                 });
             }
         } catch (err) {
+            if (this._bucketsToConnectors.get(bucket) === connector) {
+                this._bucketsToConnectors.delete(bucket);
+            }
```
francoisferrand, line 107: ?

extensions/oplogPopulator/modules/Allocator.js
```diff
@@ -114,6 +127,7 @@ class Allocator {
             const connector = this._bucketsToConnectors.get(bucket);
             if (connector) {
                 await connector.removeBucket(bucket);
+                this.emit('bucket-removed', bucket, connector);
```
francoisferrand: similarly, this should be done the other way to avoid race conditions:

```suggestion
                this.emit('bucket-removed', bucket, connector);
                await connector.removeBucket(bucket);
```

francoisferrand: (to avoid these, maybe we should [in a later PR] refactor the code to make these calls non-async, since currently they do not perform the change...)

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -231,24 +246,30 @@ class ConnectorsManager {
...
                 this._logger.info('Successfully spawned a connector', {
                     method: 'ConnectorsManager._spawnOrDestroyConnector',
-                    connector: connector.name
+                    connector: connector.name,
                 });
                 return true;
-            } else if (connector.isRunning) {
-                return connector.updatePipeline(true);
+            } else if (connector.isRunning && this._allocationStrategy.canUpdate()) {
+                const isPipelineUpdated = connector.updatePipeline(true);
```
francoisferrand: the function is async, so need to await:
```suggestion
                const isPipelineUpdated = await connector.updatePipeline(true);
```

...however this makes the event be generated after the update was performed, which is not what we want: we need to somehow send the event from the connector (e.g. connector becomes an event generator, or ConnectorManager provides an event generator to the connector)

extensions/oplogPopulator/modules/Allocator.js
ghost: Done

extensions/oplogPopulator/modules/ConnectorsManager.js
ghost: I added the event in the connector, and I eventually moved both in it, so we can handle all events (destroy & update pipeline) from the connector class, and the event is then propagated to the oplog populator through the connectors manager.
I also used a constant to avoid duplication of the string.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -237,7 +240,7 @@ class ConnectorsManager extends EventEmitter {
     }
 
     /**
-     * Spawns a connector when buckets are configured for it and is not running,
+     * Spawns a connector when buckets aremit(e configured for it and is not running,
```
francoisferrand: typo

```suggestion
     * Spawns a connector when buckets are configured for it and is not running,
```

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -263,11 +265,7 @@ class ConnectorsManager extends EventEmitter {
                 });
                 return true;
             } else if (connector.isRunning && this._allocationStrategy.canUpdate()) {
-                const isPipelineUpdated = connector.updatePipeline(true);
-                if (isPipelineUpdated) {
-                    this.emit('connector-updated', connector);
-                }
-                return isPipelineUpdated;
+                return connector.updatePipeline(true);
```
francoisferrand, line 267: still need to `await`, otherwise the return value is not a bool ?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -185,6 +188,7 @@ class Connector {
         }
         try {
             await this._kafkaConnect.deleteConnector(this._name);
+            this.emit(constants.connectorUpdatedEvent, this);
```
francoisferrand, line 190: should be sent before actually deleting the connector

extensions/oplogPopulator/modules/ConnectorsManager.js
ghost: No need to await a function we return in an async function, if we await it (the parent function), then this line will be awaited and return a bool

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -121,6 +128,10 @@ class ConnectorsManager {
             kafkaConnectHost: this._kafkaConnectHost,
             kafkaConnectPort: this._kafkaConnectPort,
         });
+        connector.on(constants.connectorUpdatedEvent, connector => {
+            return this.emit(constants.connectorUpdatedEvent, connector);
+        });
```
francoisferrand:
```suggestion
        connector.on(constants.connectorUpdatedEvent,
        	connector => this.emit(constants.connectorUpdatedEvent, connector));
```

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -242,35 +245,58 @@ class OplogPopulator {
...
+                allocationStrategy: this._allocationStrategy,
                 logger: this._logger,
             });
-            // initialize mongo client
-            await this._setupMongoClient();
+            // For now, we always use the RetainBucketsDecorator
+            // so, we map the events from the classes
+            this._connectorsManager.on(constants.connectorUpdatedEvent, connector =>
+                this._allocationStrategy.onConnectorUpdatedOrDestroyed(connector));
+            this._allocator.on('bucket-removed', (bucket, connector) =>
```
welansari: why not define the other two event types as constants ?

extensions/oplogPopulator/modules/Connector.js
ghost: Either way will have flaws, if we send it before, and deletion fails, we loose track of the retained bucket (and no easy way to handle it as we work with event (same with callbacks))

francoisferrand (approved):

extensions/oplogPopulator/modules/Connector.js
ghost: (I will change it before, but just sharing for visibility)

extensions/oplogPopulator/OplogPopulator.js
ghost: indeed, it is now done

extensions/oplogPopulator/allocationStrategy/RetainBucketsDecorator.js
```diff
@@ -0,0 +1,92 @@
...
+     * @constructor
+     * @param {AllocationStrategy} strategy the strategy to decorate
+     * @param {Object} params params
+     * @param {Logger} params.logger logger object
+     */
+    constructor(strategy, params) {
+        super(params);
+        this._strategy = strategy;
+
+        // Stores buckets that should are removed from the connector
```
welansari:
```suggestion
        // Stores buckets that should be removed from the connector
```

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -33,6 +33,14 @@ class OplogPopulatorMetrics {
             help: 'Total number of buckets per connector',
             labelNames: ['connector'],
         });
+        this.bucketsExceedingLimit = ZenkoMetrics.createGauge({
+            name: 's3_oplog_populator_connector_buckets_exceeding_limit',
```
francoisferrand:
```suggestion
            name: 's3_oplog_populator_buckets_exceeding_limit',
```

extensions/oplogPopulator/OplogPopulatorMetrics.js
```diff
@@ -33,6 +33,14 @@ class OplogPopulatorMetrics {
...
+            help: 'Total number of buckets exceeding the limit for all connectors',
+        });
+        this.retainedBuckets = ZenkoMetrics.createGauge({
+            name: 's3_oplog_populator_connector_retained_buckets',
```
francoisferrand:
```suggestion
            name: 's3_oplog_populator_retained_buckets',
```

extensions/oplogPopulator/OplogPopulator.js
welansari: `connectors-reconciled` too ?

ghost: yes...

welansari (approved):

PR #2540, opened 2024-09-13
Support a single change stream to listen to all op log events

ghost:
- Listen to all collections but the special ones and the mpu shadow buckets.
- Introduce pipeline factories, that will handle all the logic related to pipelines, to simplify other files.
- Introduce the UniqueConnector strategy, to handle the new type of pipeline and allocations in a standard way.
- Update the handling of old connector: the strategies' & their pipeline factory are used to know if an old bucket should be removed, or kept. This means, if we switch between wildcard and non-wildcard pipelines, we ensure a given bucket is never being listened to more than once.

Issue: BB-602

### DIFF
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -326,7 +326,16 @@ class OplogPopulator {
      */
     initStrategy() {
         let strategy;
-        if (this._arePipelinesImmutable()) {
+        let pipelineFactory;
+        if (this._config.numberOfConnectors === 0) {
+            // If the number of connector is set to 0, then we
+            // use a single connector to listen to the whole DB.
+            pipelineFactory = new WildcardPipelineFactory();
+            strategy = new UniqueConnector({
+                logger: this._logger,
+                pipelineFactory,
+            });
+        } else if (this._arePipelinesImmutable()) {
             // In this case, mongodb does not support reusing a
             // resume token from a different pipeline. In other
             // words, we cannot alter an existing pipeline. In this
@@ -348,13 +357,7 @@ class OplogPopulator {
                 pipelineFactory,
             });
         }
-        return new RetainBucketsDecorator(
-            strategy,
-            {
-                logger: this._logger,
-                pipelineFactory,
-            },
-        );
+        return new RetainBucketsDecorator(strategy);
     }
 
     /**
--- a/extensions/oplogPopulator/OplogPopulator.js
+++ b/extensions/oplogPopulator/OplogPopulator.js
@@ -14,6 +14,9 @@ const { mongoJoi } = require('../../lib/config/configItems.joi');
 const ImmutableConnector = require('./allocationStrategy/ImmutableConnector');
 const RetainBucketsDecorator = require('./allocationStrategy/RetainBucketsDecorator');
 const LeastFullConnector = require('./allocationStrategy/LeastFullConnector');
+const UniqueConnector = require('./allocationStrategy/UniqueConnector');
+const WildcardPipelineFactory = require('./pipeline/WildcardPipelineFactory');
+const MultipleBucketsPipelineFactory = require('./pipeline/MultipleBucketsPipelineFactory');
 
 const paramsJoi = joi.object({
     config: OplogPopulatorConfigJoiSchema.required(),
@@ -285,7 +288,8 @@ class OplogPopulator {
                 allocationStrategy: this._allocationStrategy,
                 logger: this._logger,
             });
-            this._allocationStrategy.bindConnectorEvents(this._connectorsManager, this._metricsHandler);
+            this._allocationStrategy.bindConnectorEvents(
+                this._connectorsManager, this._allocator, this._metricsHandler);
             // get currently valid buckets from mongo
             const validBuckets = await this._getBackbeatEnabledBuckets();
             // listen to valid buckets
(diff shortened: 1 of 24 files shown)


extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -261,7 +261,11 @@ class OplogPopulator {
             // initialize mongo client
             await this._setupMongoClient();
 
-            if (this._isPipelineImmutable()) {
+            if (this._config.numberOfConnectors === 0) {
+                // If the number of connector is set to 0, then we
+                // use a single connector to listen to the whole DB.
+                this._maximumBucketsPerConnector = Infinity;
```
francoisferrand: i wonder if it is safer to use this `Infinity` value (vs using `0` as in the config)

semantically, it is more correct indeed ; however since it requires custom handling anyway, maybe having 0 reduces the risk of "missing something" or behaving unexpectedly (e.g. with 0, we will sure it will just fail if we try to go through the regular allocation process, etc...)

francoisferrand (commented):
I wonder if this is the right approach... This change is trying to fit this single-change-stream approach in the existing code flow (computing which connector should handle the bucket, etc...), while in this approach there is mostly nothing to do : it is really about setting up one connector on system startup, and just call it a day...

(would still need to keep the reconcile loop, to ensure we restart failed connectors, though... and we may need to consider what to do when changing the overall mode)

I don't know if we should change or how, but I think we really need to dig a bit deeper here.

ghost:
@francoisferrand shouldn't we also handle "old" connectors on startup? We can detect them and close them to recreate a single connector, or leave them untouched and handle any new connector in a single one. If we only create one connector, I'm not sure this will work for existing deployments.

Edit: as discussed the current code should cover our tests needs. If a customer needs this approach, we will handle it manually, so no need to handle old connectors here.

extensions/oplogPopulator/pipeline/WildcardPipelineFactory.js
```diff
@@ -0,0 +1,46 @@
...
+     * @param {string[] | undefined} buckets buckets assigned to this connector
+     * @returns {string} new connector pipeline
+     */
+    getPipeline(buckets) { // eslint-disable-line no-unused-vars
+        return JSON.stringify([
+            {
+                $match: {
+                    'ns.coll': {
+                        $not: {
+                            $regex: `^(${constants.mpuBucketPrefix}|__).*`,
```
Kerkesni, line 37: Isn't the wildcard element `.*` instead of just `*` ?
Also, do we really need to listen to internal collections ? These are not needed for Backbeat.

welansari (approved):

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -134,33 +135,29 @@ class ConnectorsManager extends EventEmitter {
...
             const connectors = await Promise.all(connectorNames.map(async connectorName => {
                 // get old connector config
                 const oldConfig = await this._kafkaConnect.getConnectorConfig(connectorName);
                 // extract buckets from old connector config
-                const buckets = this._extractBucketsFromConfig(oldConfig);
+                const buckets = this._allocationStrategy.getOldConnectorBucketList(oldConfig);
+                if (!buckets) {
+                    await this._kafkaConnect.deleteConnector(connectorName);
+                    this._logger.warn('Removed old connector', {
+                        method: 'ConnectorsManager._processOldConnectors',
```
welansari: If i understood correctly, when using a single connector to listen to everything, we delete all previous connectors and create a new one with the wildcard element. This might cause us to lose some events or process some events multiple times in case we decide to set a resume time in the config of the connector. Not really sure how we should handle this ?

ghost: Yes, this is "known" and the idea is to document, in a separate work, a procedure to handle this case, so likely, pause the workflows, update the config, and re-enabling them. The main use case for now is testing, or clients with too many buckets/workflows, so the impact should be limited...

extensions/oplogPopulator/allocationStrategy/UniqueConnector.js
```diff
@@ -0,0 +1,44 @@
...
+     * If no connector is available, null is returned.
+     * @param {Array<Connector>} connectors connectors
+     * @param {String} bucket bucket name
+     * @returns {Connector | null} connector
+     */
+    getConnector(connectors, bucket) { // eslint-disable-line no-unused-vars
+        if (!connectors.length) {
+            return null;
+        }
+        return connectors.find(conn => conn.config.pipeline?.includes(constants.wildCardForAllBuckets)) || null;
```
welansari, line 24: Nit: Should we fail or crash when we don't find the connector ? This should never happen looking at the code. Also if we fall in that case and return `null` a new "wildcard" connector will be created but we're not certain no other connectors are laying around which will be a problem, maybe we should check that ?

KillianG (approved):

extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
```diff
@@ -37,6 +39,30 @@ class AllocationStrategy {
...
+     * @param {Object} oldConfig old configuration
+     * @returns {string[] | null} old configuration if valid
+     */
+    getOldConnectorBucketList(oldConfig) {
+        const bucketList = this.pipelineFactory.extractBucketsFromConfig(oldConfig);
+        if (this.pipelineFactory.isValid(bucketList)) {
+            return bucketList;
+        }
+        return null;
+    }
```
francoisferrand: this function does not have to do with the allocation strategy : the "validation" could be moved in the pipeline factory

extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
```diff
@@ -5,9 +5,11 @@ class AllocationStrategy {
     /**
      * @constructor
      * @param {Object} params params
+     * @param {PipelineFactory} params.pipelineFactory pipeline factory
      * @param {Logger} params.logger logger object
      */
     constructor(params) {
+        this._pipelineFactory = params.pipelineFactory;
```
francoisferrand: is the pipeline factory really *a part of* the strategy?
it seems is is only used as a vessel, to avoid passing an extra parameter to ConnectorManager : but may be cleaner to just keep things separate (lower coupling), and add the parameter where it should be.

francoisferrand (approved):

extensions/oplogPopulator/allocationStrategy/UniqueConnector.js
ghost: So we are supposed to delete any non-wildcard connector if this mode is used, and here we should detect any existing wildcard connector, which should cover all cases... If we do not detect the connector, we need to create it.

extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
ghost: i moved it to the pipeline factory

extensions/oplogPopulator/allocationStrategy/AllocationStrategy.js
ghost: Indeed, along with the above comment, I also updated the allocation strategy so it does not hold a reference to the pipeline factory, but we pass it to the ConnectorManager, and it does the same, with simpler code...

PR #2542, opened 2024-09-19
Fallback into checking the forceLegacyListing parameter when the listing type is not included in the Kafka message

welansari:
The check of forceLegacyListing in the bucket processor was moved to the conductor as the listing type also depends on the MongoDB indexes, which is not something used in S3C.

Lifecycle conductor in 7.x doesn't set the listing type in the bucketProcessor messages. So we now fallback to the forceLegacyListing param inside the config when encountering this case during the transition phase from backbeat 7.x to 8.x/9.x.

Also changed the default value of forceLegacyListing to how it's set in Backbeat 7.x. This is done to avoid any behaviour change in S3C in case the value is not set (which should in theory not be the case in RING10 but just in case)

Issue: BB-550

### DIFF
--- a/tests/unit/lifecycle/LifecycleBucketProcessor.spec.js
+++ b/tests/unit/lifecycle/LifecycleBucketProcessor.spec.js
@@ -2,6 +2,7 @@
 
 const assert = require('assert');
 const sinon = require('sinon');
+const async = require('async');
 
 const LifecycleBucketProcessor = require(
     '../../../extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor');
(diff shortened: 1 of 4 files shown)


tests/unit/lifecycle/LifecycleBucketProcessor.spec.js
```diff
@@ -197,4 +205,73 @@ describe('Lifecycle Bucket Processor', () => {
...
+
+                lbp._internalTaskScheduler = [];
+
+                lbp._processBucketEntry(opts.kafkaMessage, err => {
+                    assert.ifError(err);
+                    assert.strictEqual(lbp._internalTaskScheduler.length, 1);
+                    assert(opts.validateTask(lbp._internalTaskScheduler[0].task));
+                });
+            });
+        });
```
ghost: This is not testing the function at all, as we do the checks in the callback

```suggestion
            it(opts.it, done => {
                if (opts.forceLegacyListing !== null) {
                    lbp._lcConfig.forceLegacyListing = opts.forceLegacyListing;
                }

                lbp.clientManager = {
                    getS3Client: () => sinon.stub(),
                    getBackbeatMetadataProxy: () => sinon.stub(),
                };

                sinon.stub(lbp, '_getBucketLifecycleConfiguration').yields(null, {
                    Rules: {
                        Status: 'Enabled',
                        Transitions: [{
                            Days: 10,
                            StorageClass: 'azure'
                        }],
                        ID: 'dac36d89-0005-4c78-8e00-7e9ace06a9c4'
                    }
                });

                lbp._internalTaskScheduler = [];

                lbp._processBucketEntry(opts.kafkaMessage, err => {
                    if (err) {
                        return done(err);
                    }
                    assert.strictEqual(lbp._internalTaskScheduler.length, 1);
                    assert(opts.validateTask(lbp._internalTaskScheduler[0].task));
                    done();
                });
            });
        });
```

note: it also seems like the stub is not working for some reason...

ghost (changes requested):

tests/unit/lifecycle/LifecycleBucketProcessor.spec.js
welansari: oh yeah good catch !

ghost (approved):

KillianG (approved):

PR #2544, opened 2024-09-24
forward port workarround for queue populator batch getting stuck

welansari:
7.x fix was not forward ported to avoid introducing memory leaks into the code. Instead of just unblocking the next batch processing call and leaving the stuck one alive, now we shutdown the whole process.

7.x fix commit: https://github.com/scality/backbeat/commit/dfcc2f5786ab6da8bd3e2fc893d897aed0a77fe8

Issue: BB-526

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -54,6 +54,7 @@ class LogReader {
         this._zkMetricsHandler = params.zkMetricsHandler;
 
         this._batchTimeoutSeconds = parseInt(process.env.BATCH_TIMEOUT_SECONDS, 10) || 300;
+        this._batchTimedOut = false;
     }
 
     _setEntryBatch(entryBatch) {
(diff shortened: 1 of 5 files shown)


KillianG (approved):

lib/queuePopulator/LogReader.js
```diff
@@ -291,6 +293,15 @@ class LogReader {
...
             logger: this.log.newRequestLogger(),
         };
+        // When using the RaftLogReader log consumer, The batch
+        // processing can get stuck sometimes for unknown reasons.
+        // As a temporary fix we set a timeout to kill the process.
+        const batchTimeoutTimer = setTimeout(() => {
+            this.log.error('queue populator batch timeout', {
+                logStats: batchState.logStats,
+            });
+            process.emit('SIGTERM');
```
francoisferrand: should we really "crash" ?
could we not just fail healthcheck? (may not be supported at the moment in s3c, but will need to anyway...)

welansari: As far as i know, S3C (precisely supervisord) doesn't support restarting a process on healthcheck failure.
I added logic for failing the healthcheck, but i'm still keeping the crashing behaviour.

ghost (approved):

francoisferrand (approved):

PR #2545, opened 2024-09-25
remove GC producer from ObjectProcessor

benzekrimaha:
Issue : BB-554

### DIFF
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
@@ -1,12 +1,10 @@
 'use strict'; // eslint-disable-line
 
-const async = require('async');
 const { EventEmitter } = require('events');
 const Logger = require('werelogs').Logger;
 
 const BackbeatConsumerManager = require('../../../lib/BackbeatConsumerManager');
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
-const GarbageCollectorProducer = require('../../gc/GarbageCollectorProducer');
 const ClientManager = require('../../../lib/clients/ClientManager');
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 
@@ -56,7 +54,6 @@ class LifecycleObjectProcessor extends EventEmitter {
         this._lcConfig = lcConfig;
         this._processConfig = this.getProcessConfig(this._lcConfig);
         this._consumers = null;
-        this._gcProducer = null;
 
         this.clientManager = new ClientManager({
             id: this.getProcessorType(),
@@ -131,13 +128,7 @@ class LifecycleObjectProcessor extends EventEmitter {
     start(done) {
         this.clientManager.initSTSConfig();
         this.clientManager.initCredentialsManager();
-        async.parallel([
-            done => this._setupConsumers(done),
-            done => {
-                this._gcProducer = new GarbageCollectorProducer();
-                this._gcProducer.setupProducer(done);
-            },
-        ], done);
+        this._setupConsumers(done);
     }
 
     /**
@@ -228,7 +219,6 @@ class LifecycleObjectProcessor extends EventEmitter {
                 this.clientManager.getBackbeatClient.bind(this.clientManager),
             getBackbeatMetadataProxy:
                 this.clientManager.getBackbeatMetadataProxy.bind(this.clientManager),
-            gcProducer: this._gcProducer,
             logger: this._log,
         };
     }
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
@@ -1,5 +1,6 @@
 'use strict'; // eslint-disable-line
 const assert = require('assert');
+const async = require('async');
 
 const ColdStorageStatusQueueEntry = require('../../../lib/models/ColdStorageStatusQueueEntry');
 const LifecycleObjectProcessor = require('./LifecycleObjectProcessor');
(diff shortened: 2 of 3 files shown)


extensions/lifecycle/objectProcessor/LifecycleObjectExpirationProcessor.js
```diff
@@ -33,9 +33,10 @@ class LifecycleObjectExpirationProcessor extends LifecycleObjectProcessor {
      * @param {Number} s3Config.port - s3 endpoint port
      * @param {String} [transport="http"] - transport method ("http"
      *  or "https")
+     * @param {Object} gcConfig - gc congifuration object
      */
-    constructor(zkConfig, kafkaConfig, lcConfig, s3Config, transport = 'http') {
-        super(zkConfig, kafkaConfig, lcConfig, s3Config, transport);
+    constructor(zkConfig, kafkaConfig, lcConfig, s3Config, transport = 'http', gcConfig) {
```
francoisferrand: the "expiration processor" is actually what is running in the object processor service : where indeed there is no use for GC (as it is used for transitions), and the gc producer should not be created...

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -134,8 +136,12 @@ class LifecycleObjectProcessor extends EventEmitter {
         async.parallel([
             done => this._setupConsumers(done),
             done => {
-                this._gcProducer = new GarbageCollectorProducer();
-                this._gcProducer.setupProducer(done);
+                if (this._gcConfig) {
+                    this._gcProducer = new GarbageCollectorProducer();
+                    this._gcProducer.setupProducer(done);
+                } else {
+                    done();
```
francoisferrand: please exercise this branch

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -47,15 +47,17 @@ class LifecycleObjectProcessor extends EventEmitter {
      * @param {Number} s3Config.port - s3 endpoint port
      * @param {String} [transport="http"] - transport method ("http"
      *  or "https")
+     * @param {Object} gcConfig - gc congifuration object
      */
-    constructor(zkConfig, kafkaConfig, lcConfig, s3Config, transport = 'http') {
+    constructor(zkConfig, kafkaConfig, lcConfig, s3Config, transport = 'http', gcConfig) {
```
francoisferrand: gcConfig should probably be added before the transport field : otherwise we may have issues using the default value for that field (and it seems more consistent to keep the config fields together)

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -134,8 +136,12 @@ class LifecycleObjectProcessor extends EventEmitter {
         async.parallel([
             done => this._setupConsumers(done),
             done => {
-                this._gcProducer = new GarbageCollectorProducer();
-                this._gcProducer.setupProducer(done);
+                if (this._gcConfig) {
```
francoisferrand: this `gcConfig` is actually not used : so it should not be a `config`, but really just a boolean flag...

....and better yet, this the `gcProducer` is used only in the transition processor : it should actually be moved into that class, there is (theorically) no point keeping it in the base class...

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -134,8 +136,12 @@ class LifecycleObjectProcessor extends EventEmitter {
...
+                    this._gcProducer = new GarbageCollectorProducer();
+                    this._gcProducer.setupProducer(done);
+                } else {
+                    done();
+                }
             },
         ], done);
```
welansari: I think maybe a cleaner fix for this would be to move the gcProducer creation logic to `LifecycleObjectTransitionProcessor`, this will avoid having to pass arround the gc config everywhere

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -57,7 +58,13 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
                 return done(err);
             }
-            return this.setupProducer(done);
+            this._gcProducer = new GarbageCollectorProducer();
+            return this._gcProducer.setupProducer(err => {
+                if (err) {
+                    return done(err);
+                }
+                return this.setupProducer(done);
+            });
```
welansari: Maybe better to use `async.waterfall` to avoid too many nested layers ?

welansari (approved):

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -53,12 +55,14 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
-        super.start(err => {
-            if (err) {
-                return done(err);
-            }
-            return this.setupProducer(done);
-        });
+        async.waterfall([
+            next => super.start(next),
+            next => {
+                this._gcProducer = new GarbageCollectorProducer();
```
francoisferrand, line 61: these used to be done in parallel, is there a (significant) impact on startup time to doing this asynchronously?

francoisferrand (approved):

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
benzekrimaha: this is done using waterfall because the start method needs to complete successfully before setting up the producers ( this is what was done initially )

PR #2546, opened 2024-09-26
add missing overhead field used in S3C

welansari:
Metadata adds `versionId` as an extra field in the oplog's delete events as they originally contain no metadata. These are needed for Bucket notification.

Note: Updating Bucket Notifications to use this field will be addressed in another ticket.

Issue: BB-535

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -499,6 +499,7 @@ class LogReader {
             const overheadFields = {
                 commitTimestamp: record.timestamp,
                 opTimestamp: entry.timestamp,
+                versionId: entry.overhead?.versionId,
             };
             const entryToFilter = {
                 type: entry.type,
--- a/tests/unit/lib/queuePopulator/LogReader.spec.js
+++ b/tests/unit/lib/queuePopulator/LogReader.spec.js
@@ -110,6 +110,7 @@ describe('LogReader', () => {
             overheadFields: {
                 commitTimestamp: record.timestamp,
                 opTimestamp: '2023-11-29T15:05:57.065Z',
+                versionId: undefined,
             },
         };
         assert(mockExtension.filter.firstCall.calledWith(expectedArgs));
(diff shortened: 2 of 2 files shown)


lib/queuePopulator/LogReader.js
```diff
@@ -499,6 +499,7 @@ class LogReader {
             const overheadFields = {
                 commitTimestamp: record.timestamp,
                 opTimestamp: entry.timestamp,
+                versionId: entry.overhead && entry.overhead.versionId,
```
francoisferrand: Nit: ˋentry.overhead?.versionId`

tests/unit/lib/queuePopulator/LogReader.spec.js
```diff
@@ -244,4 +246,44 @@ describe('LogReader', () => {
             });
         });
     });
+
+    describe('_processLogEntry', () => {
+        it('should pass the proper overhead fields', done => {
```
francoisferrand: Should also test (explicitely) the case where versionId is missing

lib/queuePopulator/LogReader.js
ghost: If this is supported by backbeat...
```suggestion
                versionId: entry.overhead?.versionId,
```

ghost (approved):

lib/queuePopulator/LogReader.js
welansari: it's not :(

francoisferrand:
I wonder if we should not push the change bit further, to abstract the difference : eg store the `versionId` (and possibly other fields) in the same place whatever the backend, so that we limit the difference to existing adapters (like LogReader implementations) and have a single code path to deal with in extensions.

Not sure it fits in this ticket though, and/or not be fully possible as we need to handle "existing" objects already in kafka as well

francoisferrand (approved):

lib/queuePopulator/LogReader.js
francoisferrand: Why? How can/could we make it supported?

ghost: To me optional chaining should work as it's supported in node v14+, and I see it in other files

welansari: Oh my bad, it does work, i remembered it not working a while back.

PR #2547, opened 2024-09-26
Support setting a site name for Kafka consumers

welansari:
Code in `backbeatConsumer` already supports setting the `client.rack` consumer parameter from the config. So here i just allow passing that field in the config as we do in S3C

Issue: BB-521

### DIFF
--- a/lib/config.joi.js
+++ b/lib/config.joi.js
@@ -29,6 +29,7 @@ const joiSchema = joi.object({
             intervalS: joi.number().default(60),
         },
         maxRequestSize: joi.number().default(KAFKA_PRODUCER_MESSAGE_MAX_BYTES),
+        site: joi.string(),
     },
     transport: transportJoi,
     s3: hostPortJoi.required(),


KillianG (approved):

benzekrimaha (approved):

lib/config.joi.js
```diff
@@ -29,6 +29,7 @@ const joiSchema = joi.object({
             intervalS: joi.number().default(60),
         },
         maxRequestSize: joi.number().default(KAFKA_PRODUCER_MESSAGE_MAX_BYTES),
+        site: joi.string(),
```
francoisferrand, line 32: - should we not rename this parameter to `rack`, for consistency?
- what is the purpose of this parameter, how is it used? Like should we support having different values for different consumers (esp. in context of Zenko), or is it enough to have different config files for each service?

welansari: > should we not rename this parameter to rack, for consistency?

We could rename it as `rack` yes, this will mean changing the config in federation.

> what is the purpose of this parameter, how is it used? Like should we support having different values for different consumers (esp. in context of Zenko), or is it enough to have different config files for each service?

Specifying the rack in both the broker config (which is done in S3C) and the client config allows us to configure an [assignment strategy](https://kafka.apache.org/documentation/#streamsconfigs_rack.aware.assignment.strategy) in Kafka which can optimise traffic between the clients and brokers.

In Zenko, this could be useful in a 3-nodes or more configuration, but this won't work looking at how the services are deployed as they can be scheduled on a different node when restarted. Maybe we should pass the node name as an env var that references the CR field `nodeName` and overwrite the config with that value.

```
env:
- name: RACK
  valueFrom:
    fieldRef:
      fieldPath: spec.nodeName
```

Also, not sure how Kafka is deployed, we might need to enforce the node where each replica is deployed to be able to properly specify the rack in the configuration.

francoisferrand: so how does in work in s3c, is the same value passed to every backbeat container? or does the value somewhat derive from the node the container is running on or the actual service (bucket notification, ...)? or is it maybe used in stretch scenarios, to talk to the "local" kafka brokers first (if kafka is stretched)?

welansari: Seems we have a stretched Kafka deployment. And yes this param is only relevant on a multi-site/stretched deployment

Deployment of S3C is done per site where we deploy a set of backbeat containers + a Kafka container on every site. Containers from different sites don't share the same config. And so Backbeat containers of the same site specify the site name in their Kafka config to try and preferably target the Kafka deployed on that site. Same is done for Kafka config where `broker.rack` is specified and is different on each site.

francoisferrand: so nothing to do in Zenko _for now_, but we should take a note somewhere : so we rememember about this when we work on "stretched" Zenko in the future.

PR #2548, opened 2024-10-01
Improvement/bb 514 handle lifecycle rules

benzekrimaha:
Issue : BB-514

### DIFF
--- a/extensions/lifecycle/LifecycleConfigValidator.js
+++ b/extensions/lifecycle/LifecycleConfigValidator.js
@@ -8,6 +8,8 @@ const {
     retryParamsJoi,
 } = require('../../lib/config/configItems.joi');
 
+const { supportedLifecycleRules }  = require('../../lib/constants');
+
 const joiSchema = joi.object({
     zookeeperPath: joi.string().required(),
     bucketTasksTopic: joi.string().required(),
@@ -73,6 +75,9 @@ const joiSchema = joi.object({
     coldStorageRestoreAdjustTopicPrefix: joi.string().default('cold-restore-adjust-req-'),
     coldStorageGCTopicPrefix: joi.string().default('cold-gc-req-'),
     coldStorageStatusTopicPrefix: joi.string().default('cold-status-'),
+    supportedLifecycleRules: joi.array().items(
+        joi.string().valid(...supportedLifecycleRules)
+    ).default(supportedLifecycleRules),
 });
 
 function configValidator(backbeatConfig, extConfig) {
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -3,7 +3,6 @@
 const async = require('async');
 const { errors, versioning } = require('arsenal');
 const { ObjectMD } = require('arsenal').models;
-const { supportedLifecycleRules } = require('arsenal').constants;
 const {
     LifecycleDateTime,
     LifecycleUtils,
@@ -97,10 +96,10 @@ class LifecycleTask extends BackbeatTask {
         });
 
         this._lifecycleUtils = new LifecycleUtils(
-            supportedLifecycleRules,
+            this.supportedRules,
             this._lifecycleDateTime
         );
-        this._supportedRules = supportedLifecycleRules;
+        this._supportedRules = this.supportedRules;
         this._totalRetries = 0;
     }
 
--- a/extensions/lifecycle/util/rules.js
+++ b/extensions/lifecycle/util/rules.js
@@ -2,7 +2,6 @@ const { RulesReducer } = require('./RulesReducer');
 const { lifecycleListing: { NON_CURRENT_TYPE, CURRENT_TYPE, ORPHAN_DM_TYPE } } = require('../../../lib/constants');
 
 const { s3middleware } = require('arsenal');
-const { supportedLifecycleRules } = require('arsenal').constants;
 const { scaleMsPerDay } = s3middleware.objectUtils;
 
 // Default max AWS limit is 1000 for both list objects and list object versions
@@ -198,11 +197,12 @@ function rulesToParams(versioningStatus, currentDate, bucketLCRules, bucketData,
 /**
  * Formats the array of supported lifecycle rules in Arsenal so that
  * they correspond to how they are written in the lifecycle configuration.
- * @param {string[]} supportedLifecycleRules list of supported lifecycle rules
+ * @param {Object} lcConfig - Lifecycle configuration object
+ * @param {string[]} lcConfig.supportedLifecycleRules - List of supported lifecycle rules
  * @returns {string[]} formatted supported lifecycle rules
  */
-function getFormattedSupportedLifecycleRules() {
-   return supportedLifecycleRules.map(rule => {
+function formatSupportedLifecycleRules(lcConfig) {
+   return lcConfig.supportedLifecycleRules.map(rule => {
         if (rule === 'noncurrentVersionTransition') {
             return 'NoncurrentVersionTransitions';
         }
@@ -228,7 +228,7 @@ function rulesSupportTransition(lcRules) {
 
 module.exports = {
     rulesToParams,
-    getFormattedSupportedLifecycleRules,
+    formatSupportedLifecycleRules,
     isExpirationRule,
     rulesSupportTransition,
 };
--- a/lib/config/configItems.joi.js
+++ b/lib/config/configItems.joi.js
@@ -158,6 +158,7 @@ const qpKafkaJoi = kafkaJoi.append({
     replication: extensionKafkaJoi,
 });
 
+
 module.exports = {
     hostPortJoi,
     transportJoi,
@@ -172,5 +173,5 @@ module.exports = {
     probeServerPerSite,
     stsConfigJoi,
     mongoJoi,
-    qpKafkaJoi,
+    qpKafkaJoi
 };
--- a/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
+++ b/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
@@ -15,7 +15,7 @@ const ClientManager = require('../../../lib/clients/ClientManager');
 const { authTypeAssumeRole } = require('../../../lib/constants');
 const LocationStatusStream = require('../../utils/LocationStatusStream');
 const {
-    getFormattedSupportedLifecycleRules,
+    formatSupportedLifecycleRules,
     isExpirationRule
 } = require('../util/rules');
 const {
@@ -88,7 +88,7 @@ class LifecycleBucketProcessor {
         this._producer = null;
         this._kafkaBacklogMetrics = null;
 
-        this._supportedRules = getFormattedSupportedLifecycleRules();
+        this._supportedRules = formatSupportedLifecycleRules(lcConfig);
 
         this._producerReady = false;
         this._consumerReady = false;
@@ -186,6 +186,7 @@ class LifecycleBucketProcessor {
             lcOptions: this._lcOptions,
             circuitBreakers: this._circuitBreakers,
             log: this._log,
+            supportedRules: this._lcConfig.supportedLifecycleRules,
         };
     }
 
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -54,7 +54,14 @@ const constants = {
                 },
             ],
         },
-    }
+    },
+    supportedLifecycleRules: [
+        'expiration',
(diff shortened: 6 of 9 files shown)


conf/config.json
```diff
@@ -272,7 +272,14 @@
             "coldStorageArchiveTopicPrefix": "cold-archive-req-",
             "coldStorageRestoreTopicPrefix": "cold-restore-req-",
             "coldStorageGCTopicPrefix": "cold-gc-req-",
-            "coldStorageStatusTopicPrefix": "cold-status-"
+            "coldStorageStatusTopicPrefix": "cold-status-",
+            "supportedLifecycleRules": [
```
francoisferrand: not needed, this is the default (and the default should generally be used, to ensure we don't miss new features)

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -3,7 +3,6 @@
 const async = require('async');
 const { errors, versioning } = require('arsenal');
 const { ObjectMD } = require('arsenal').models;
```
francoisferrand, line 5: we need to create a ticket to remove this from Arsenal, eventually
(i.e. after we have migrated all branches to this new backbeat)

tests/config.json
```diff
@@ -222,7 +222,14 @@
             "coldStorageRestoreTopicPrefix": "cold-restore-req-",
             "coldStorageGCTopicPrefix": "cold-gc-req-",
             "coldStorageStatusTopicPrefix": "cold-status-",
-            "coldStorageTopics": []
+            "coldStorageTopics": [],
+            "supportedLifecycleRules": [
```
francoisferrand: not needed, this is the default

extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -67,6 +67,13 @@ const joiSchema = joi.object({
     coldStorageRestoreAdjustTopicPrefix: joi.string().default('cold-restore-adjust-req-'),
     coldStorageGCTopicPrefix: joi.string().default('cold-gc-req-'),
     coldStorageStatusTopicPrefix: joi.string().default('cold-status-'),
+    supportedLifecycleRules: joi.array().items(joi.string()).default([
```
francoisferrand: we may also want to validate each item from this list : i.e. to ensure each value is indeed of the potentially supported rule type

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: Ticket created here => https://scality.atlassian.net/browse/ARSN-435

extensions/lifecycle/LifecycleConfigValidator.js
benzekrimaha: done here : https://github.com/scality/backbeat/pull/2548/commits/1d605348937a9adcde59b729c1708b49d714167c#diff-4ddf07d1cd40d3d68a85bcf7ab41f36f43094af8cfc0e93424067c48bd03b623R70-R78

tests/config.json
```diff
@@ -222,7 +222,7 @@
             "coldStorageRestoreTopicPrefix": "cold-restore-req-",
             "coldStorageGCTopicPrefix": "cold-gc-req-",
             "coldStorageStatusTopicPrefix": "cold-status-",
-            "coldStorageTopics": []
+            "coldStorageTopics": []
```
ghost:
```suggestion
            "coldStorageTopics": []
```

extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -67,6 +67,21 @@ const joiSchema = joi.object({
...
+            'transitions',
+            'noncurrentVersionTransition'
+        )
+    ).default([
+        'expiration',
+        'noncurrentVersionExpiration',
+        'abortIncompleteMultipartUpload',
+        'transitions',
+        'noncurrentVersionTransition',
+    ]),
```
ghost: Could this array be a constant in a single place?
Something like this (not tested)

```suggestion
        Joi.string().valid(...supportedLifecycleRules)
    ).default(supportedLifecycleRules),
```

lib/config/configItems.joi.js
```diff
@@ -150,6 +151,10 @@ const qpKafkaJoi = kafkaJoi.append({
     replication: extensionKafkaJoi,
 });
 
+const supportedLifecycleRulesJoi =  joi.array().items(
```
ghost:
```suggestion
const supportedLifecycleRulesJoi = joi.array().items(
```

ghost: (maybe not very useful to store this one here if we only use it once, but the array can stay here)

ghost (approved):

lib/config/configItems.joi.js
```diff
@@ -150,6 +151,10 @@ const qpKafkaJoi = kafkaJoi.append({
...
+    joi.string().valid(...supportedLifecycleRules))
```
welansari: Not sure we want to limit the valid choices to only what's in `supportedLifecycleRules`. We might support more later ?

lib/config/configItems.joi.js
```diff
@@ -150,6 +151,10 @@ const qpKafkaJoi = kafkaJoi.append({
...
+    .default(supportedLifecycleRules);
```
welansari: Since this is specific to lifecycle, we might want to define it directly in `LifecycleConfigValidator` ?

extensions/lifecycle/util/rules.js
```diff
@@ -201,7 +201,7 @@ function rulesToParams(versioningStatus, currentDate, bucketLCRules, bucketData,
  * @returns {string[]} formatted supported lifecycle rules
  */
 function getFormattedSupportedLifecycleRules() {
-   return supportedLifecycleRules.map(rule => {
+   return config.extensions.lifecycle.supportedLifecycleRules.map(rule => {
         if (rule === 'noncurrentVersionTransition') {
```
welansari: This function is used in the bucket processor where lifecycle config is available [here](https://github.com/scality/backbeat/blob/0a32e66773027f99b5057bf9159bf1ec1f5fc3c0/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js#L84), i think it's better if we pass the supported rules as an argument.

We then should also change the function name to `formatSupportedLifecycleRules`.

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -96,10 +95,10 @@ class LifecycleTask extends BackbeatTask {
...
 
         this._lifecycleUtils = new LifecycleUtils(
-            supportedLifecycleRules,
+            config.extensions.lifecycle.supportedLifecycleRules,
             this._lifecycleDateTime
         );
-        this._supportedRules = supportedLifecycleRules;
+        this._supportedRules = config.extensions.lifecycle.supportedLifecycleRules;
         this._totalRetries = 0;
     }
```
welansari: Might be better to pass the supported rules through `lp.getStateVars()`, which is called at the beginning of the constructor. This function [lets you pass values from the bucket procesor](https://github.com/scality/backbeat/blob/0a32e66773027f99b5057bf9159bf1ec1f5fc3c0/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js#L176-L190) which has access to the lifecycle config.

lib/config/configItems.joi.js
francoisferrand: if we support more, the new rules must be added to `supportedLifecycleRules`.

(maybe later we will need to have a different default, e.g. if some rules must not be enabled by default ; but for now we don't need this, so the whole list is also the default)

extensions/lifecycle/util/rules.js
```diff
@@ -201,7 +201,7 @@ function rulesToParams(versioningStatus, currentDate, bucketLCRules, bucketData,
...
             return 'NoncurrentVersionTransitions';
```
francoisferrand, line 207: is this function still needed?

could we remove `getFormattedSupportedLifecycleRules` altogether, and use the rules name from `lifecycle configuration` (NoncurrentVersionTransitions, Expiration....)

welansari: We have to keep it for now as the class `LifecycleUtils` that we use in `LifecycleTask` uses the values as defined in Arsenal.

francoisferrand: `LifecycleUtils` is in arsenal? If it is in backbeat, we could update it?

welansari: yes the class is in Arsenal

francoisferrand: Do we know why it is needed in arsenal, could we move it back?
If we could (or don't know), please create ticket to clean this up (move back + remove the duplicate list). Ideally this should be done before the 8.7 release, to avoid having to change the configuration.

KillianG:
we might also want to add the new field to [conf/config.json](https://github.com/scality/backbeat/blob/development/8.6/conf/config.json#L252) no ?

extensions/lifecycle/util/rules.js
welansari: The class is also used in Cloudserver, which gives it the values in Arsenal constants https://github.com/scality/cloudserver/blob/ab5d49a975e0ac8d9b1e4ddde94ae0b59ce37f25/lib/api/apiUtils/object/expirationHeaders.js#L22

extensions/lifecycle/util/rules.js
benzekrimaha: done here : https://github.com/scality/backbeat/pull/2548/commits/08e817a420a339682e58d7d7776932e2d98e5f75#diff-c560baf5a4758706eaf47282c6fd3b381b38bc8b0eb355aa3983e7e675a8fc09R203-R204

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: done here : https://github.com/scality/backbeat/pull/2548/commits/a2b5caaaa07e99197759d1dbc195f1ff4d56358e#diff-0251eb0ff96b3559c00aac5d7e592f81df2eece3b644f9278307237dd8f6bbbcR189

lib/config/configItems.joi.js
```diff
@@ -5,7 +5,7 @@ const constants = require('../../extensions/replication/constants');
 const {
     authTypeAccount,
     authTypeService,
-    authTypeAssumeRole,
+    authTypeAssumeRole
 } = require('../constants');
```
welansari: No need to update this file, we should keep trailing commas to be consistent with the rest of the code.

extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -73,6 +75,8 @@ const joiSchema = joi.object({
     coldStorageRestoreAdjustTopicPrefix: joi.string().default('cold-restore-adjust-req-'),
     coldStorageGCTopicPrefix: joi.string().default('cold-gc-req-'),
     coldStorageStatusTopicPrefix: joi.string().default('cold-status-'),
+    supportedLifecycleRules: joi.array().items(
+        joi.string().valid(...supportedLifecycleRules)).default(supportedLifecycleRules),
 });
```
welansari:
```suggestion
    supportedLifecycleRules: joi.array().items(
        joi.string().valid(...supportedLifecycleRules)
    ).default(supportedLifecycleRules),
});
```
nit

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -96,10 +95,10 @@ class LifecycleTask extends BackbeatTask {
         });
 
         this._lifecycleUtils = new LifecycleUtils(
-            supportedLifecycleRules,
+            lpState.supportedRules,
             this._lifecycleDateTime
         );
-        this._supportedRules = supportedLifecycleRules;
+        this._supportedRules = lpState.supportedRules;
         this._totalRetries = 0;
```
welansari:
```suggestion
        this._lifecycleUtils = new LifecycleUtils(
            this.supportedRules,
            this._lifecycleDateTime
        );
        this._supportedRules = this.supportedRules;
        this._totalRetries = 0;
```

lib/config/configItems.joi.js
ghost: not having comma does not follow Scality guidelines https://github.com/scality/Guidelines/blob/development/8.2/index.js#L19C10-L19C22

It will be enforced anyway when we upgrade the dependencies...

welansari (approved):

PR #2549, opened 2024-10-07
Bucket Notifications Unification

welansari:
- Added support for specifying multiple destination kafka hosts in bucket notification (ported logic from 7.x)
- Ported NotificationConfigManager from 7.x with tests
- Added an abstract NotificationConfigManager class to support both MongoDB and Zookeeper backends
- Ported code from 7.x that handles updating notification configs in the Zookeeper backend
- NotificationConfigManager now returns 7.x config structure, this was changed to keep the code functional with previously stored configs in Zookeeper on S3C.
- Support S3C delete notifications. 
- Made metastore collection name configurable as the name is different between S3c and Zenko.
- Fixed config not getting removed from zookeeper when a bucket is deleted.

**Note:** To ease reviews, please review commit by commit.

Issue: BB-497

### DIFF
--- a/extensions/notification/NotificationConfigManager.js
+++ b/extensions/notification/NotificationConfigManager.js
@@ -10,7 +10,7 @@ class NotificationConfigManager {
 
     constructor(params) {
         const {
-            mongoConfig, bucketMetastore, maxCachedConfigs, zkClient, zkConfig, zkPath, logger,
+            mongoConfig, bucketMetastore, maxCachedConfigs, zkClient, zkConfig, zkPath, zkConcurrency, logger,
         } = params;
         if (mongoConfig) {
             this._configManagerBackend = new MongoConfigManager({
@@ -25,6 +25,7 @@ class NotificationConfigManager {
                 zkClient,
                 zkConfig,
                 zkPath,
+                zkConcurrency,
                 logger,
             });
         }
--- a/extensions/notification/NotificationConfigManager.js
+++ b/extensions/notification/NotificationConfigManager.js
@@ -1,283 +1,79 @@
-const joi = require('joi');
-const semver = require('semver');
-
-const { ZenkoMetrics } = require('arsenal').metrics;
-const LRUCache = require('arsenal').algorithms
-    .cache.LRUCache;
-const MongoClient = require('mongodb').MongoClient;
-const ChangeStream = require('../../lib/wrappers/ChangeStream');
-const constants = require('./constants');
-const { constructConnectionString, getMongoVersion } = require('../utils/MongoUtils');
-
-const paramsJoi = joi.object({
-    mongoConfig: joi.object().required(),
-    logger: joi.object().required(),
-}).required();
-
-const MAX_CACHED_ENTRIES = Number(process.env.MAX_CACHED_BUCKET_NOTIFICATION_CONFIGS)
-    || 1000;
-
-// should equal true if config manager's cache was hit during a get operation
-const CONFIG_MANAGER_CACHE_HIT = 'cache_hit';
-// Type of operation performed on the cache
-const CONFIG_MANAGER_OPERATION_TYPE = 'op';
-
-const cacheUpdates = ZenkoMetrics.createCounter({
-    name: 's3_notification_config_manager_cache_updates_total',
-    help: 'Total number of cache updates',
-    labelNames: [
-        CONFIG_MANAGER_OPERATION_TYPE,
-    ],
-});
-
-const configGetLag = ZenkoMetrics.createHistogram({
-    name: 's3_notification_config_manager_config_get_seconds',
-    help: 'Time it takes in seconds to get a bucket notification config from MongoDB',
-    labelNames: [
-        CONFIG_MANAGER_CACHE_HIT,
-    ],
-    buckets: [0.001, 0.01, 1, 10, 100, 1000],
-});
-
-const cachedBuckets = ZenkoMetrics.createGauge({
-    name: 's3_notification_config_manager_cached_buckets_count',
-    help: 'Total number of cached buckets in the notification config manager',
-});
-
-function onConfigManagerCacheUpdate(op) {
-    cacheUpdates.inc({
-        [CONFIG_MANAGER_OPERATION_TYPE]: op,
-    });
-    if (op === 'add') {
-        cachedBuckets.inc({});
-    } else if (op === 'delete') {
-        cachedBuckets.dec({});
-    }
-}
-
-function onConfigManagerConfigGet(cacheHit, delay) {
-    configGetLag.observe({
-        [CONFIG_MANAGER_CACHE_HIT]: cacheHit,
-    }, delay);
-}
+const MongoConfigManager = require('./configManager/MongoConfigManager');
+const ZookeeperConfigManager = require('./configManager/ZookeeperConfigManager');
 
 /**
  * @class NotificationConfigManager
  *
- * @classdesc Manages bucket notification configurations, the configurations
- * are directly retrieved from the metastore, and are locally cached. Cache
- * is invalidated using MongoDB change streams.
+ * @classdesc Manages bucket notification configurations
  */
 class NotificationConfigManager {
-    /**
-     * @constructor
-     * @param {Object} params - constructor params
-     * @param {Object} params.mongoConfig - mongoDB config
-     * @param {Logger} params.logger - logger object
-     */
+
     constructor(params) {
-        joi.attempt(params, paramsJoi);
-        this._logger = params.logger;
-        this._mongoConfig = params.mongoConfig;
-        this._cachedConfigs = new LRUCache(MAX_CACHED_ENTRIES);
-        this._mongoClient = null;
-        this._metastore = null;
-        this._metastoreChangeStream = null;
-    }
-
-    /**
-     * Connects to MongoDB using the MongoClientInterface
-     * and retreives the metastore collection
-     * @param {Function} cb callback
-     * @returns {undefined}
-     */
-    _setupMongoClient(cb) {
-        const mongoUrl = constructConnectionString(this._mongoConfig);
-        MongoClient.connect(mongoUrl, {
-            replicaSet: this._mongoConfig.replicaSet,
-            useNewUrlParser: true,
-        },
-        (err, client) => {
-            if (err) {
-                this._logger.error('Could not connect to MongoDB', {
-                    method: 'NotificationConfigManager._setupMongoClient',
-                    error: err.message,
-                });
-                return cb(err);
-            }
-            this._logger.debug('Connected to MongoDB', {
-                method: 'NotificationConfigManager._setupMongoClient',
+        const { mongoConfig, bucketMetastore, zkClient, zkConfig, zkPath, logger } = params;
+        if (mongoConfig) {
+            this._configManagerBackend = new MongoConfigManager({
+                mongoConfig,
+                bucketMetastore,
+                logger,
             });
-            try {
-                this._mongoClient = client.db(this._mongoConfig.database, {
-                    ignoreUndefined: true,
-                });
-                this._metastore = this._mongoClient.collection(constants.bucketMetastore);
-                // get mongodb version
-                getMongoVersion(this._mongoClient, (err, version) => {
-                    if (err) {
-                        this._logger.error('Could not get MongoDB version', {
-                            method: 'NotificationConfigManager._setupMongoClient',
-                            error: err.message,
-                        });
-                        return cb(err);
-                    }
-                    this._mongoVersion = version;
-                    return cb();
-                });
-                return undefined;
-            } catch (error) {
-                return cb(error);
-            }
-        });
-    }
-
-    /**
-     * Handler for the change stream "change" event.
-     * Invalidates cached bucket configs based on the change.
-     * @param {ChangeStreamDocument} change Change stream change object
-     * @returns {undefined}
-     */
-    _handleChangeStreamChangeEvent(change) {
-        // invalidating cached notification configs
-        const cachedConfig = this._cachedConfigs.get(change.documentKey._id);
-        const bucketNotificationConfiguration = change.fullDocument ? change.fullDocument.value.
-            notificationConfiguration : null;
-        switch (change.operationType) {
-            case 'delete':
-                if (cachedConfig) {
-                    this._cachedConfigs.remove(change.documentKey._id);
-                    onConfigManagerCacheUpdate('delete');
-                }
-                break;
-            case 'replace':
-            case 'update':
-                if (cachedConfig) {
-                    // add() replaces the value of an entry if it exists in cache
-                    this._cachedConfigs.add(change.documentKey._id, bucketNotificationConfiguration);
-                    onConfigManagerCacheUpdate('update');
-                }
-                break;
-            default:
-                this._logger.debug('Skipping unsupported change stream event', {
-                    method: 'NotificationConfigManager._handleChangeStreamChange',
-                });
-                break;
+        } else {
+            this._usesZookeeperBackend = true;
+            this._configManagerBackend = new ZookeeperConfigManager({
+                zkClient,
+                zkConfig,
+                zkPath,
+                logger,
+            });
         }
-        this._logger.debug('Change stream event processed', {
-            method: 'NotificationConfigManager._handleChangeStreamChange',
-        });
-    }
-
-    /**
-     * Initializes a change stream on the metastore collection
-     * Only document delete and update/replace operations are
-     * taken into consideration to invalidate cache.
-     * Newly created buckets (insert operations) are not cached
-     * as queue populator instances read from different kafka
-     * partitions and so don't need the configs for all buckets
-     * @returns {undefined}
-     */
-    _setMetastoreChangeStream() {
-        /**
-         * To avoid processing irrelevant events
-         * we filter by the operation types and
-         * only project the fields needed
-         */
-        const changeStreamPipeline = [
-            {
-                $match: {
-                    $or: [
-                        { operationType: 'delete' },
-                        { operationType: 'replace' },
-                        { operationType: 'update' },
-                    ]
-                }
-            },
-            {
-                $project: {
-                    '_id': 1,
-                    'operationType': 1,
-                    'documentKey._id': 1,
-                    'fullDocument._id': 1,
-                    'fullDocument.value.notificationConfiguration': 1
-                },
-            },
-        ];
-        this._metastoreChangeStream = new ChangeStream({
-            logger: this._logger,
-            collection: this._metastore,
-            pipeline: changeStreamPipeline,
-            handler: this._handleChangeStreamChangeEvent.bind(this),
-            throwOnError: false,
-            useStartAfter: semver.gte(this._mongoVersion, '4.2.0'),
-        });
-        // start watching metastore
-        this._metastoreChangeStream.start();
-    }
-
-    /**
-     * Sets up the NotificationConfigManager by
-     * connecting to mongo and initializing the
-     * change stream
-     * @param {Function} cb callback
-     * @returns {undefined}
-     */
-    setup(cb) {
-        this._setupMongoClient(err => {
-            if (err) {
-                this._logger.error('An error occured while setting up mongo client', {
-                    method: 'NotificationConfigManager.setup',
-                });
-                return cb(err);
-            }
-            try {
-                this._setMetastoreChangeStream();
-            } catch (error) {
-                this._logger.error('An error occured while establishing the change stream', {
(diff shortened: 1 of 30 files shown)


extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -490,6 +490,10 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
      * @return {undefined}
      */
     filter(entry) {
+        if (!entry.value) {
```
francoisferrand: can't this happen with `delete` entries?

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,283 +1,87 @@
...
-                this._logger.error('Could not connect to MongoDB', {
-                    method: 'NotificationConfigManager._setupMongoClient',
-                    error: err.message,
-                });
-                return cb(err);
-            }
-            this._logger.debug('Connected to MongoDB', {
-                method: 'NotificationConfigManager._setupMongoClient',
+    constructor(params) {
+        this._usesZookeeperBackend = false;
```
francoisferrand: seems like an anti-pattern: we already have abstracted away zookeeper/mongo through `_configManagerBackend`, so either
* this needs to be moved to the backend
* every backend needs to have a new method, called at the right time, to trigger this zookeeper-specific handling

extensions/lifecycle/LifecycleQueuePopulator.js
welansari: We used to ignore these in the logReader, i moved that logic to the actual extensions as S3c uses true delete events for bucket notifications. In lifecycle we use the update we do before each delete as we need the metadata.

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,283 +1,87 @@
...
+    getConfig(bucket, cb) {
+        const val = this._configManagerBackend.getConfig(bucket);
+        if (!cb) {
+            return val;
         }
-        this._logger.debug('Change stream event processed', {
-            method: 'NotificationConfigManager._handleChangeStreamChange',
-        });
+        if (val instanceof Promise) {
+            return val.then(res => cb(null, res)).catch(err => cb(err));
```
francoisferrand, line 47: not needed or not tested...

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,283 +1,87 @@
...
-            collection: this._metastore,
-            pipeline: changeStreamPipeline,
-            handler: this._handleChangeStreamChangeEvent.bind(this),
-            throwOnError: false,
-            useStartAfter: semver.gte(this._mongoVersion, '4.2.0'),
-        });
-        // start watching metastore
-        this._metastoreChangeStream.start();
+    setConfig(bucket, config) {
+        if (this._usesZookeeperBackend) {
```
francoisferrand: not needed : implement the method in mongo backend as well, and just return false

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,283 +1,87 @@
...
-            } catch (error) {
-                this._logger.error('An error occured while establishing the change stream', {
-                    method: 'NotificationConfigManager._setMetastoreChangeStream',
-                });
-                return cb(error);
-            }
-            return cb();
-        });
+    removeConfig(bucket) {
+        if (this._usesZookeeperBackend) {
```
francoisferrand: not needed : implement the method in mongo backend as well, and just return false

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -200,24 +312,29 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
      * @return {undefined} Promise|undefined
      */
     filterAsync(entry, cb) {
-        const { bucket, key, overheadFields } = entry;
+        if (this.notificationConfig.ignoreEmptyEvents && !entry.value) {
+            return cb();
```
francoisferrand: why do we we have a config `ignoreEmptyEvents` ?
- either this can happen & there is nothing to do → ignore them
- or it can happen & but there is nothing to do → ignore them
- or it can happen & there is something to do → we must not ignore them...

welansari: Empty events have to be ignored in Zenko, else we'll get duplicate delete notifications. We use the update we do before each delete for delete notifications.
In S3C empty delete events are the only event available for delete notifications.

extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -0,0 +1,291 @@
...
+    if (op === 'add') {
+        cachedBuckets.inc({});
+    } else if (op === 'delete') {
+        cachedBuckets.dec({});
+    }
+}
+
+function onConfigManagerConfigGet(cacheHit, delay) {
+    configGetLag.observe({
+        [CONFIG_MANAGER_CACHE_HIT]: cacheHit,
```
francoisferrand, line 62:
```suggestion
        CONFIG_MANAGER_CACHE_HIT: cacheHit,
```

extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -0,0 +1,291 @@
...
+});
+
+const cachedBuckets = ZenkoMetrics.createGauge({
+    name: 's3_notification_config_manager_cached_buckets_count',
+    help: 'Total number of cached buckets in the notification config manager',
+});
+
+function onConfigManagerCacheUpdate(op) {
+    cacheUpdates.inc({
+        [CONFIG_MANAGER_OPERATION_TYPE]: op,
```
francoisferrand, line 51:
```suggestion
        CONFIG_MANAGER_OPERATION_TYPE: op,
```

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -141,6 +165,7 @@ class QueueProcessor extends EventEmitter {
      */
     start(options, done) {
         async.series([
+            next => this._setupZookeeper(next),
```
francoisferrand: can't we setup zookeper inside the `ZookeeperConfigManager` ?

extensions/notification/utils/config.js
```diff
@@ -87,9 +87,14 @@ function validateEntry(bnConfig, entry) {
         return { isValid: false, matchingConfig: null };
     }
 
+    if (bucket !== bnConfig.bucket) {
+        return { isValid: false, matchingConfig: null };
```
francoisferrand, line 91: should be tested

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -84,14 +86,34 @@ class QueueProcessor extends EventEmitter {
...
         this.logger = new Logger('Backbeat:Notification:QueueProcessor');
     }
 
+    _setupZookeeper(done) {
+        if (this.mongoConfig) {
+            done();
+            return;
+        }
+
+        const populatorZkPath = this.notifConfig.zookeeperPath;
```
francoisferrand: can we add some tests for this file?

KillianG (approved):

extensions/notification/queueProcessor/QueueProcessor.js
welansari: The queue populator creates a zookeeper client for all queue populator extensions, we kind of have to do it this way in the queue processor.

extensions/notification/configManager/MongoConfigManager.js
welansari: We have to use the brackets else the label literally becomes `CONFIG_MANAGER_OPERATION_TYPE`

extensions/notification/configManager/MongoConfigManager.js
welansari: We have to use the brackets else the label literally becomes `CONFIG_MANAGER_CACHE_HIT `

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,283 +1,79 @@
...
     }
 
     /**
-     * Sets up the NotificationConfigManager by
-     * connecting to mongo and initializing the
-     * change stream
-     * @param {Function} cb callback
-     * @returns {undefined}
+     * Remove bucket notification configuration
+     * Only supported in the Zookeeper backend.
```
francoisferrand: not really true : we have implemented an abstraction, the method is supported on every backend... (even if some backends actually have an _empty_ implementation)

extensions/notification/NotificationConfigManager.js
```diff
@@ -1,283 +1,79 @@
...
     /**
-     * Initializes a change stream on the metastore collection
-     * Only document delete and update/replace operations are
-     * taken into consideration to invalidate cache.
-     * Newly created buckets (insert operations) are not cached
-     * as queue populator instances read from different kafka
-     * partitions and so don't need the configs for all buckets
-     * @returns {undefined}
+     * Add/update bucket notification configuration.
+     * Only supported in the Zookeeper backend.
```
francoisferrand: not really true : we have implemented an abstraction, the method is supported on every backend... (even if some backends actually have an empty implementation)

extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -0,0 +1,311 @@
...
+                bucket,
+                error: err.message,
+            });
+            return undefined;
+        }
+    }
+
+    /**
+     * Set bucket notification configuration
+     * (Not implemted in the MongoDB backend)
```
francoisferrand: it is not "not implemented", more like it is not "needed" : because mongo backend automatically configures itself (using a change stream)

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -240,19 +262,15 @@ class QueueProcessor extends EventEmitter {
...
-                    if (!destBnConf.length) {
+                    // get destination specific notification config
+                    const queueConfig = notifConfig.notificationConfiguration.queueConfig.filter(
+                            c => c.queueArn.split(':').pop() === this.destinationId
+                    );
+                    if (!queueConfig.length) {
                         // skip, if there is no config for the current
                         // destination resource
-                        return done();
+                        return undefined;
```
francoisferrand, line 251: should this not call `done()` as before?

extensions/notification/queueProcessor/QueueProcessor.js
francoisferrand: I don't understand why.
* this specific attribute (`zkClient`) is initialized here, once for every call to `start()`
* it seems it is only used on the next function (`_setupNotificationConfigManager`)
* this function does not store zkClient, it just passes it to notificationManager constructor, which passes it only to the zookeerConfigManager...

--> It seems to me we may as well initialize the zkClient inside `zookeeperConfigManager` constructor (and avoid an extra zkClient in Zenko context) 

(but we do indeed need to pass zkConfig)

extensions/notification/NotificationQueuePopulator.js
francoisferrand: is this filtering (in zenko) not already done in the mongo abstraction layers (e.g. either MongoClientInterface or LogReader), where we also do generate a "delete" event from an update with "deleted=true" ?

It seems this would really be the proper place to handle it, keeping proper layering & abstractions.

francoisferrand: Reading further, I see you removed this from `LogReader` : it seems to me this should not be "moved up" to QueuePopulator/..., but instead "moved down" into the specific implementations of `LogReader` (i.e. maybe a virtual function in LogReader, defined differently in each implementation ; or a flag from LogReader initialized by each subclass...)

extensions/notification/queueProcessor/QueueProcessor.js
francoisferrand: ok, i see what you mean: issue is not here, but in QueuePopulator where we want to reuse the one we already have (even in Zenko).

- However, this current implementation creates a useless zkClient in QueueProcessor on Zenko : which should be avoided.
- One way to avoid this may be to pass both zkConfig and an (optional) zkClient to `NotificationConfigManager` : so that if provided with a client, it can use it (QueuePopulator case) but it will otherwise create one from the config (QueueProcessor case).

lib/queuePopulator/QueuePopulator.js
```diff
@@ -417,6 +417,8 @@ class QueuePopulator {
             try {
                 this.bnConfigManager = new NotificationConfigManager({
                     mongoConfig: this.qpConfig.mongo,
+                    bucketMetastore: this.extConfigs.notification.bucketMetastore,
+                    zkClient: this.qpConfig.mongo ? null : this.zkClient,
```
francoisferrand: why filter the parameters?
NotificationConfigManager takes multiple optional params (mongo, zkClient...) and deals with it as needed, no point doing it for him here.

extensions/notification/configManager/ZookeeperConfigManager.js
```diff
@@ -0,0 +1,446 @@
...
+    _getConfigDataFromBuffer(data) {
+        const { error, result } = safeJsonParse(data);
+        if (error) {
+            this.log.error('invalid config', { error, config: data });
+            return undefined;
+        }
+        return result;
+    }
+
+    _getBucketNotifConfig(bucket, cb) {
```
anurag4DSB, line 144: its worth adding some logs here for easier debugging.
Same for methods, `_removeBucketNotifConfigNode`, `_setBucketNotifConfig`

extensions/notification/configManager/ZookeeperConfigManager.js
```diff
@@ -0,0 +1,446 @@
...
+                    error,
+                });
+                this._callbackHandler(cb, error);
+            }
+            this._callbackHandler(cb, null, buckets);
+        });
+    }
+
+    _updateLocalStore(buckets, cb) {
+        async.eachSeries(buckets, (bucket, next) => {
```
anurag4DSB: have you thought about using  `async.eachLimit` with a configurable limit instead of `eachSeries` in `_updateLocalStore` for improved performance on large datasets(controlled parallel processing of tasks)

extensions/notification/configManager/ZookeeperConfigManager.js
```diff
@@ -0,0 +1,446 @@
...
+        return undefined;
+    }
+
+    _setConfigListener(bucket, config) {
+        this.log.debug('ZookeeperConfigManager.emitter.setConfig', {
+            event: 'setConfig',
+            bucket,
+            config,
+        });
+        this._setBucketNotifConfig(bucket, JSON.stringify(config), err => {
```
anurag4DSB, line 67: Should we add a retry logic to handle transient network issues?
same for `_getBucketNotifConfig`

extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -0,0 +1,312 @@
...
+                notificationConfiguration
+            };
+        } catch (err) {
+            this._logger.error('An error occured when getting notification ' +
+                'configuration of bucket', {
+                method: 'MongoConfigManager.getConfig',
+                bucket,
+                error: err.message,
+            });
+            return undefined;
```
anurag4DSB: should we return the err here?

extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -0,0 +1,312 @@
...
+        });
+    }
+
+    /**
+     * Get bucket notification configuration
+     *
+     * @param {String} bucket - bucket
+     * @return {Object|undefined} - configuration if available or undefined
+     */
+    async getConfig(bucket) {
```
anurag4DSB, line 256: can we add some debug logs for cache hit-and-miss scenarios to improve observability of cache performance and usage

extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -0,0 +1,312 @@
...
+const ChangeStream = require('../../../lib/wrappers/ChangeStream');
+const constants = require('../constants');
+const { constructConnectionString, getMongoVersion } = require('../../utils/MongoUtils');
+
+const paramsJoi = joi.object({
+    mongoConfig: joi.object().required(),
+    logger: joi.object().required(),
+}).required();
+
+const MAX_CACHED_ENTRIES = Number(process.env.MAX_CACHED_BUCKET_NOTIFICATION_CONFIGS)
```
anurag4DSB: can this be a config item as well?

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -42,6 +42,75 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
     }
 
+    /**
+     * Get bucket attributes from log entry
+     *
+     * @param {Object} value - log entry object
+     * @return {Object|undefined} - bucket attributes if available
+     */
+    _getBucketAttributes(value) {
+        if (value && value.attributes) {
```
anurag4DSB, line 53: should we cache this in `this._bucketAttributes` or something?(Lazy Initialization of _getBucketAttributes)

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -42,6 +42,75 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
...
+            = this._getBucketNotificationConfiguration(value);
+        if (notificationConfiguration &&
+            Object.keys(notificationConfiguration).length > 0) {
+            const bnConfig = {
+                bucket: bucketName,
+                notificationConfiguration,
+            };
+            // bucket notification config is available, update node
+            this.bnConfigManager.setConfig(bucketName, bnConfig);
+            return undefined;
```
anurag4DSB, line 109: is it possible to keep the return consistent with the below, either returning the defined below or changing here?

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -3,6 +3,7 @@ const util = require('util');
 
 const { isMasterKey } = require('arsenal').versioning;
 const { usersBucket, mpuBucketPrefix, supportedNotificationEvents } = require('arsenal').constants;
+const VID_SEP = require('arsenal').versioning.VersioningConstants.VersionId.Separator;
```
anurag4DSB: NIT: Worth writing `VID_SEPERATOR`, I am okay with VID :)

anurag4DSB (commented):
I added some comments; I am curious how we are testing this code as it includes s3c-specific changes.

extensions/notification/configManager/ZookeeperConfigManager.js
welansari: The Zookeeper library we use already handles retries on network issues. Only needed to enable it by configuring the max number of retries.

extensions/notification/configManager/MongoConfigManager.js
welansari: We could use this metric for that `s3_notification_config_manager_config_get_seconds`. We also have a metric for how many times we update the cache `s3_notification_config_manager_cache_updates_total`

welansari:
> I am curious how we are testing this code as it includes s3c-specific changes

@anurag4DSB I'm manually testing everything on a RING with custom images and modified federation steps to have the proper configs for Backbeat.

extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -0,0 +1,312 @@
...
+     * @return {Object|undefined} - configuration if available or undefined
+     */
+    async getConfig(bucket) {
+        const startTime = Date.now();
+        // return cached config for bucket if it exists
+        const cachedConfig = this._cachedConfigs.get(bucket);
+        if (cachedConfig) {
+            const delay = (Date.now() - startTime) / 1000;
+            onConfigManagerConfigGet(true, delay);
+            return {
```
francoisferrand, line 263: what is the point of returning the `bucket`?
it seems we are not using it (kind of logical, since we already have it to get the config...)

extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -0,0 +1,312 @@
...
+}
+
+/**
+ * @class MongoConfigManager
+ *
+ * @classdesc Manages bucket notification configurations, the configurations
+ * are directly retrieved from the metastore, and are locally cached. Cache
+ * is invalidated using MongoDB change streams.
+ */
+class MongoConfigManager {
```
francoisferrand: it seems to me we are missing a "base" class, if only to document the interface these must conform to.

francoisferrand (approved):

extensions/notification/configManager/MongoConfigManager.js
welansari: The old config manager used to store the config in zookeeper this way. To avoid having to migrate the old data stored in zookeeper or having to deal with the old vs new format i decided to keep the old one.

PR #2550, opened 2024-10-09
S3C supports having different queueProcessor probeserver configuratio…

KillianG:
S3C supports having different queueProcessor probeserver configuration per site.

Joi configuration supports having multiple types for a single config parameter, this will allow supporting passing both an array or a single object containing the probeserver config.

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -71,7 +71,10 @@ const joiSchema = joi.object({
         concurrency: joi.number().greater(0).default(10),
         mpuPartsConcurrency: joi.number().greater(0).default(10),
         minMPUSizeMB: joi.number().greater(0).default(20),
-        probeServer: probeServerJoi.default(),
+        probeServer: probeServerJoi.alternatives().try(
+            probeServerJoi,
+            probeServerPerSite
+        ).default(),
         circuitBreaker: joi.object().optional(),
     }).required(),
     replicationStatusProcessor: {
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -71,7 +71,7 @@ const joiSchema = joi.object({
         concurrency: joi.number().greater(0).default(10),
         mpuPartsConcurrency: joi.number().greater(0).default(10),
         minMPUSizeMB: joi.number().greater(0).default(20),
-        probeServer: probeServerJoi.alternatives().try(
+        probeServer: joi.alternatives().try(
             probeServerJoi,
             probeServerPerSite
         ).default(),
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -1,7 +1,7 @@
 const fs = require('fs');
 const joi = require('joi');
 const { hostPortJoi, transportJoi, bootstrapListJoi, adminCredsJoi,
-        retryParamsJoi, probeServerJoi } =
+        retryParamsJoi, probeServerJoi, probeServerPerSite } =
     require('../../lib/config/configItems.joi');
 
 const qpRetryJoi = joi.object({
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -74,7 +74,7 @@ const joiSchema = joi.object({
         probeServer: joi.alternatives().try(
             probeServerJoi,
             probeServerPerSite
-        ).default({bindAddress: 'localhost', port: ""}),
+        ).default({ bindAddress: 'localhost', port: '' }),
         circuitBreaker: joi.object().optional(),
     }).required(),
     replicationStatusProcessor: {
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -73,7 +73,7 @@ const joiSchema = joi.object({
         minMPUSizeMB: joi.number().greater(0).default(20),
         probeServer: joi.alternatives().try(
             probeServerJoi,
-            probeServerPerSite
+            probeServerPerSite,
         ).default({ bindAddress: 'localhost', port: 4042 }),
         circuitBreaker: joi.object().optional(),
     }).required(),
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -74,7 +74,7 @@ const joiSchema = joi.object({
         probeServer: joi.alternatives().try(
             probeServerJoi,
             probeServerPerSite
-        ).default({ bindAddress: 'localhost', port: '' }),
+        ).default({ bindAddress: 'localhost', port: 4042 }),
         circuitBreaker: joi.object().optional(),
     }).required(),
     replicationStatusProcessor: {
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -74,7 +74,7 @@ const joiSchema = joi.object({
         probeServer: joi.alternatives().try(
             probeServerJoi,
             probeServerPerSite
-        ).default(),
+        ).default({bindAddress: 'localhost', port: ""}),
         circuitBreaker: joi.object().optional(),
     }).required(),
     replicationStatusProcessor: {
--- a/extensions/replication/queueProcessor/task.js
+++ b/extensions/replication/queueProcessor/task.js
@@ -243,23 +243,24 @@ function initAndStart(zkClient) {
          * Get probe config will pull the configuration for the probe server based on
          * the provided site key, and if the probe server config is not an array, it will
          * return the global probe server config.
+         * 
+         *  Get the probe server config for the first site in the site names list,
+         *  as if the probeConfig is an array there is only one element in site names
          *
          * @param {Object} queueProcessorConfig - Configuration of the queue processor that
          *      holds the probe server configs for all sites
-         * @param {string} site - Name of the site we are processing
+         * @param {Array<String>} siteNames - List of site names
          * @returns {ProbeServerConfig} Config for site or global config
          */
-        function getProbeConfig(queueProcessorConfig, site) {
-            if (Array.isArray(queueProcessorConfig.probeServer) && site) {
-                return queueProcessorConfig.probeServer.filter(probe => probe.site === site)[0];
+        function getProbeConfig(queueProcessorConfig, siteNames) {
+            if (Array.isArray(queueProcessorConfig.probeServer) && siteNames[0]) {
+                return queueProcessorConfig.probeServer.filter(probe => probe.site === siteNames[0])[0];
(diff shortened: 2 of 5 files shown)


extensions/replication/queueProcessor/task.js
```diff
@@ -239,8 +239,27 @@ function initAndStart(zkClient) {
...
+                return queueProcessorConfig.probeServer.filter(probe => probe.site === site)[0];
+            }
+            return queueProcessorConfig.probeServer;
+        }
+
         startProbeServer(
-            repConfig.queueProcessor.probeServer,
+            // get the probe server config for the first site in the bootstrap list,
+            // as if the probeConfig is an array there is only one element in bootstrap list
+            getProbeConfig(repConfig.queueProcessor, bootstrapList[0].site),
```
welansari: `bootstrapList` can be empty when no location was configured. This will make the code crash.

extensions/replication/queueProcessor/task.js
```diff
@@ -239,8 +239,28 @@ function initAndStart(zkClient) {
...
+         *
+         *  Get the probe server config for the first site in the site names list,
+         *  as if the probeConfig is an array there is only one element in site names
+         *
+         * @param {Object} queueProcessorConfig - Configuration of the queue processor that
+         *      holds the probe server configs for all sites
+         * @param {Array<String>} siteNames - List of site names
+         * @returns {ProbeServerConfig} Config for site or global config
+         */
+        function getProbeConfig(queueProcessorConfig, siteNames) {
```
francoisferrand: this function should not be in this "bootstrap" script : there is already too much here, let's not add to the mess.
and this will allow testing this function (even we do not test -yet- the startup script itself)

extensions/replication/queueProcessor/task.js
```diff
@@ -239,8 +239,28 @@ function initAndStart(zkClient) {
...
+            if (Array.isArray(queueProcessorConfig.probeServer) && siteNames[0]) {
+                return queueProcessorConfig.probeServer.filter(probe => probe.site === siteNames[0])[0];
```
francoisferrand: this does not handle all cases.
* we may be started with an array of proble configs and a single siteName (s3c style) → ok
* we may be started with a single proble config and (one or more) site names (zenko style) → ok
* but we may also be started with an array of probe configs and multiple sites : in that case we should create multiple proble servers, one for each site... or at least generate an error, if this case is not supported

extensions/replication/queueProcessor/task.js
```diff
@@ -239,8 +239,28 @@ function initAndStart(zkClient) {
...
+         *  Get the probe server config for the first site in the site names list,
+         *  as if the probeConfig is an array there is only one element in site names
+         *
+         * @param {Object} queueProcessorConfig - Configuration of the queue processor that
+         *      holds the probe server configs for all sites
+         * @param {Array<String>} siteNames - List of site names
+         * @returns {ProbeServerConfig} Config for site or global config
+         */
+        function getProbeConfig(queueProcessorConfig, siteNames) {
+            if (Array.isArray(queueProcessorConfig.probeServer) && siteNames[0]) {
```
francoisferrand: it may be an empty array, must be managed

lib/config/configItems.joi.js
```diff
@@ -112,6 +112,14 @@ const probeServerJoi = joi.object({
     port: joi.number().required(),
 });
 
+const probeServerPerSite = joi.array().items(
```
francoisferrand: there is not validation to 'reject' empty array: not sure if this is by design, or just an omission.

* can this happen in s3c, e.g. maybe in case there is no `site` (no crr configured) → must be supported
* if not, would be better to add the required `joi` config to validate the array is not empty

extensions/replication/queueProcessor/task.js
KillianG: > but we may also be started with an array of probe configs and multiple sites : in that case we should create multiple proble servers, one for each site... or at least generate an error, if this case is not supported

Normally this can't happen as Maha did a PR for that : https://github.com/scality/backbeat/pull/2537
Or am I missing something ?

lib/config/configItems.joi.js
KillianG: I added the `joi` config to verify array is not empty, as if we have an array it shouldnt be empty

If CRR is not configured then probeserver is supposed to use the "normal" config which is `probeServerJoi`
as done here: 
```
probeServer: joi.alternatives().try(
            probeServerJoi,
            probeServerPerSite
```

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -71,7 +71,10 @@ const joiSchema = joi.object({
         concurrency: joi.number().greater(0).default(10),
         mpuPartsConcurrency: joi.number().greater(0).default(10),
         minMPUSizeMB: joi.number().greater(0).default(20),
-        probeServer: probeServerJoi.default(),
+        probeServer: joi.alternatives().try(
+            probeServerJoi,
+            probeServerPerSite
+        ).default({ bindAddress: 'localhost', port: '' }),
```
francoisferrand: do we need to separate the 2 definitions (probeServerJoi and probeServerPerSite) ?

```suggestion
probeServerJoi = joi.alternatives().try(
            <probeServerJoi>,
            probeServerPerSite
        ).default({ bindAddress: 'localhost', port: '' })
```

the default looks strange, should have a port (which should be an integer?)

lib/util/probe.js
```diff
@@ -60,8 +60,39 @@ function observeKafkaStats(msg) {
...
+    if (siteNames.length === 0) {
+        if (!Array.isArray(queueProcessorConfig.probeServer)) {
+            return queueProcessorConfig.probeServer;
+        }
+        return undefined;
+    }
+
+    if (Array.isArray(queueProcessorConfig.probeServer)) {
+        if (siteNames.length !== 1) {
+            return undefined;
```
francoisferrand: returning `undefined` here will generate a strange error : "configuration for probe server is missing" (when starting probe server with undefined probe)

We may want to add an error log here to point to the issue...

extensions/replication/ReplicationConfigValidator.js
KillianG: We need both because we want to have probeServerPersite available only for replication, so I can't set the alternative to probeServerJoi directly..

lib/config/configItems.joi.js
```diff
@@ -112,6 +112,14 @@ const probeServerJoi = joi.object({
     port: joi.number().required(),
 });
 
+const probeServerPerSite = joi.array().min(1).items(
+    joi.object({
+        bindAddress: joi.string().default('localhost'),
+        port: joi.number().required(),
+        site: joi.string().required()
```
ghost, line 119:
```suggestion
        site: joi.string().required(),
```

lib/util/probe.js
```diff
@@ -60,8 +64,50 @@ function observeKafkaStats(msg) {
...
+                queueProcessorConfig,
+            });
+            return undefined;
+        }
+
+        const siteConfig = queueProcessorConfig.probeServer.find(config => config.site === siteNames[0]);
+        return siteConfig || undefined;
+    }
+
+    return undefined;
```
welansari: Shouldn't this return `queueProcessorConfig.probeServer` ? We fall in this case when having multiple sites and only one probeserver config, which is how it's deployed in Artesca.

lib/util/probe.js
```diff
@@ -60,8 +64,50 @@ function observeKafkaStats(msg) {
...
+ * the provided site name. If siteNames is empty, it returns the global probe server config
+ * only if it's a single object.
+ *
+ * @param {Object} queueProcessorConfig - Configuration of the queue processor that
+ *      holds the probe server configs for all sites
+ * @param {Array<String>} siteNames - List of site names (should contain at most one element)
+ * @returns {Object|undefined} Config for site or global config, undefined if no match found or invalid config
+ */
+function getProbeConfig(queueProcessorConfig, siteNames) {
+    const logger = new Logger('Backbeat:Probe:getProbeConfig');
```
welansari: Looking at how and where this function is called, we might want get the logger from the function's arguments. The `task` file already initializes a logger. And to be consistent with how we use the logger in backbeat we might want to add `method: "getProbeConfig"` in the logs.

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -71,7 +71,10 @@ const joiSchema = joi.object({
         concurrency: joi.number().greater(0).default(10),
         mpuPartsConcurrency: joi.number().greater(0).default(10),
         minMPUSizeMB: joi.number().greater(0).default(20),
-        probeServer: probeServerJoi.default(),
+        probeServer: joi.alternatives().try(
+            probeServerJoi,
+            probeServerPerSite
```
ghost:
```suggestion
            probeServerPerSite,
```

lib/util/probe.js
```diff
@@ -60,8 +60,50 @@ function observeKafkaStats(msg) {
...
+ *
+ * @param {Object} queueProcessorConfig - Configuration of the queue processor that
+ *      holds the probe server configs for all sites
+ * @param {Array<String>} siteNames - List of site names (should contain at most one element)
+ * @param {Logger} logger - Logger instance
+ * @returns {Object|undefined} Config for site or global config, undefined if no match found or invalid config
+ */
+function getProbeConfig(queueProcessorConfig, siteNames, logger) {
+
+    if (siteNames.length === 0) {
```
ghost:
```suggestion
function getProbeConfig(queueProcessorConfig, siteNames, logger) {
    if (siteNames.length === 0) {
```

lib/util/probe.js
```diff
@@ -60,8 +60,50 @@ function observeKafkaStats(msg) {
...
+        if (siteNames.length !== 1) {
+            logger.error('Process configured for more than one site', {
+                siteNames,
+                queueProcessorConfig,
+            });
+            return undefined;
+        }
+
+        const siteConfig = queueProcessorConfig.probeServer.find(config => config.site === siteNames[0]);
+        return siteConfig || undefined;
```
ghost: Should we log with a warn here, if we are not able to find the config for a specific site?

ghost (approved):

lib/util/probe.js
```diff
@@ -60,8 +60,56 @@ function observeKafkaStats(msg) {
...
+            logger.warn('Probe server configuration for site not found', {
+                siteName: siteNames[0],
+                queueProcessorConfig,
+            });
+        }
+
+        return siteConfig;
+    }
+
+    return queueProcessorConfig.probeServer;
```
welansari: I think this could be simplified. We don't actually need the first check. We only care about whether or not the probeserver config is an array.

-> If it's an array -> return the proper config OR undefined when we have multiple sitenames or none.
The way it is deployed in S3C, we always have 1 site name. if there are no sites no process is spawned.

-> Otherwise we just return the probeserver config.
It doesn't matter how many sites we have in this case, as it's the Zenko case, where we can have 0 or N sites and in all cases we only have one config

welansari (approved):

PR #2551, opened 2024-10-09
condition start method on bucketprocessor

benzekrimaha:
Issue : BB-549

### DIFF
--- a/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
+++ b/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
@@ -456,12 +456,18 @@ class LifecycleBucketProcessor {
             done => this._initKafkaBacklogMetrics(done),
             done => this._setupConsumer(done),
             done => {
-                this._locationStatusStream = new LocationStatusStream('lifecycle',
-                    this._mongoConfig,
-                    this._pauseServiceForLocation.bind(this),
-                    this._resumeServiceForLocation.bind(this),
-                    this._log);
-                this._locationStatusStream.start(done);
+                if (this._mongoConfig) {
+                    this._locationStatusStream = new LocationStatusStream(
+                        'lifecycle',
+                        this._mongoConfig,
+                        this._pauseServiceForLocation.bind(this),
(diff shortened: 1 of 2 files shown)


extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -451,19 +451,24 @@ class LifecycleBucketProcessor {
     start(done) {
         this.clientManager.initSTSConfig();
         this.clientManager.initCredentialsManager();
-        async.series([
+        const tasks = [
```
welansari: No need to take this out of the `async.series`, you can just call `done()` when `this._mongoConfig` is not configured.

benzekrimaha: done here : https://github.com/scality/backbeat/pull/2551/commits/f910139f25a006237f08a45c711dc7d4b087fe5b#diff-0251eb0ff96b3559c00aac5d7e592f81df2eece3b644f9278307237dd8f6bbbcR459-R471

welansari (approved):

KillianG (approved):

extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js
```diff
@@ -456,12 +456,18 @@ class LifecycleBucketProcessor {
             done => this._initKafkaBacklogMetrics(done),
             done => this._setupConsumer(done),
             done => {
-                this._locationStatusStream = new LocationStatusStream('lifecycle',
-                    this._mongoConfig,
-                    this._pauseServiceForLocation.bind(this),
-                    this._resumeServiceForLocation.bind(this),
-                    this._log);
-                this._locationStatusStream.start(done);
+                if (this._mongoConfig) {
```
francoisferrand, line 459: note: eventually (once we support pause/resume on s3c), we _will_ need to have an abstraction here, and always create some sort of "locationStatusStream" : listening to mongo (as today in zenko) or to something else (maybe zookeeper?)

but indeed we don't have this now, this will be done later.

francoisferrand (approved):

PR #2553, opened 2024-10-10
Improvement/bb 541 avoid initializing mongodb client

benzekrimaha:
Issue : BB-541

### DIFF
--- a/lib/api/BackbeatAPI.js
+++ b/lib/api/BackbeatAPI.js
@@ -1260,40 +1260,10 @@ class BackbeatAPI {
             this._healthcheck = new Healthcheck(this._repConfig, this._zkClient,
                 this._crrProducer, this._crrStatusProducer,
                 this._metricProducer);
-            this._locationStatusManager = new LocationStatusManager(
-                this._mongoClient,
-                this._zkClient,
-                this._redisPublisher,
-                {
-                    crr: {
-                        namespace: zookeeperReplicationNamespace,
-                        statePath: zkReplicationStatePath,
-                        topic: this._crrTopic,
-                        isMongo: false,
-                    },
-                    ingestion: {
-                        namespace: zookeeperIngestionNamespace,
-                        statePath: zkIngestionStatePath,
-                        topic: this._ingestionTopic,
-                        isMongo: false,
-                    },
-                    lifecycle: {
-                        isMongo: true,
-                    }
-                },
-                this._logger,
-            );
-            return this._locationStatusManager._setupLocationStatusStore(err => {
-                if (err) {
-                    this._logger.error('error setting up location status manager', {
-                        method: 'BackbeatAPI.setupInternals',
-                        error: err.message,
-                    });
-                    return cb(err);
-                }
-                this._logger.info('BackbeatAPI setup ready');
-                return cb();
-            });
+            if (this._queuePopulator.mongo) {
+                return this._setupLocationStatusManager(cb);
+            }
+            return cb();
         });
     }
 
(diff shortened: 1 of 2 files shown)


lib/api/BackbeatAPI.js
```diff
@@ -1260,40 +1260,12 @@ class BackbeatAPI {
...
-                    this._logger.error('error setting up location status manager', {
-                        method: 'BackbeatAPI.setupInternals',
-                        error: err.message,
-                    });
-                    return cb(err);
-                }
-                this._logger.info('BackbeatAPI setup ready');
+            if (this._config.queuePopulator.mongo) {
+                return this._setupLocationStatusManager(cb);
+            } else {
```
welansari: we still need to initialize the `LocationStatusManager` as the pause/resume of crr uses redis/zookeeper and not mongodb and so we need to keep it. This case should be handled inside the `LocationStatusManager` class.

welansari: Actually, pause/resume was not available on S3C, maybe we should keep as is for now without adding new features.

lib/api/BackbeatAPI.js
```diff
@@ -1335,28 +1307,73 @@ class BackbeatAPI {
...
+
+        return this._locationStatusManager._setupLocationStatusStore(err => {
+            if (err) {
+                this._logger.error('error setting up location status manager', {
+                    method: 'BackbeatAPI.setupInternals',
+                    error: err.message,
+                });
+                return cb(err);
+            }
+            this._logger.info('BackbeatAPI setup ready');
```
welansari: Does it really make sense to keep this log inside this function ?

lib/api/BackbeatAPI.js
```diff
@@ -1335,28 +1307,73 @@ class BackbeatAPI {
...
+                    isMongo: true,
+                }
+            },
+            this._logger,
+        );
+
+        return this._locationStatusManager._setupLocationStatusStore(err => {
+            if (err) {
+                this._logger.error('error setting up location status manager', {
+                    method: 'BackbeatAPI.setupInternals',
```
welansari:
```suggestion
                    method: 'BackbeatAPI._setupLocationStatusManager',
```

lib/api/BackbeatAPI.js
```diff
@@ -1335,28 +1307,73 @@ class BackbeatAPI {
      * @returns {undefined} undefined
      */
     _setupMongoClient(cb) {
-        const mongoConfig = this._config.queuePopulator.mongo;
-        const mongoUrl = constructConnectionString(mongoConfig);
-        MongoClient.connect(mongoUrl, {
+        if (this._config.queuePopulator.mongo) {
+            const mongoConfig = this._config.queuePopulator.mongo;
```
welansari:
```suggestion
        const mongoConfig = this._config.queuePopulator.mongo;
        if (!mongoConfig) {
            this._logger.debug('MongoDB configuration not found, skipping MongoDB client setup', {
                method: 'BackbeatAPI._setupMongoClient',
            });
            return cb();
         }
```
Just my opinion, but i think doing it the other way is cleaner ?

lib/api/BackbeatAPI.js
```diff
@@ -1260,40 +1260,12 @@ class BackbeatAPI {
...
-                        method: 'BackbeatAPI.setupInternals',
-                        error: err.message,
-                    });
-                    return cb(err);
-                }
-                this._logger.info('BackbeatAPI setup ready');
+            if (this._config.queuePopulator.mongo) {
+                return this._setupLocationStatusManager(cb);
+            } else {
+                this._logger.debug('BackbeatAPI setup ready without Mongo');
```
welansari: Redundant with the log in `_setupMongoClient`

KillianG (approved):

lib/api/BackbeatAPI.js
```diff
@@ -1260,40 +1260,11 @@ class BackbeatAPI {
...
-            return this._locationStatusManager._setupLocationStatusStore(err => {
-                if (err) {
-                    this._logger.error('error setting up location status manager', {
-                        method: 'BackbeatAPI.setupInternals',
-                        error: err.message,
-                    });
-                    return cb(err);
-                }
-                this._logger.info('BackbeatAPI setup ready');
+            if (this._config.queuePopulator.mongo) {
```
welansari:
```suggestion
            if (this._queuePopulator.mongo) {
```

lib/api/BackbeatAPI.js
```diff
@@ -1260,40 +1260,11 @@ class BackbeatAPI {
...
+                return this._setupLocationStatusManager(cb);
+            } else {
                 return cb();
-            });
+            }
```
welansari:
```suggestion
                return this._setupLocationStatusManager(cb);
            }
            return cb();
```

welansari (approved):

lib/api/BackbeatAPI.js
```diff
@@ -1336,26 +1307,70 @@ class BackbeatAPI {
...
+            this._logger.debug('MongoDB configuration not found, skipping MongoDB client setup', {
+                method: 'BackbeatAPI._setupMongoClient',
+            });
+            return cb();
+        } else {
+            const mongoUrl = constructConnectionString(mongoConfig);
+            return MongoClient.connect(mongoUrl, {
                 replicaSet: mongoConfig.replicaSet,
                 useNewUrlParser: true,
                 useUnifiedTopology: true,
```
ghost, line 1321: Not related to the changes, but I believe this is not used anymore with recent mongodb drivers, but I also see that we are still using mongo driver v3 in Backbeat... Would it make sense to bump it in this PR, as we modify this function? or we can do it when upgrading the deps (but these lines will be easy to miss...)

benzekrimaha: I suggest adding a comment so it is not missed during the dependencies bump

PR #2554, opened 2024-10-10
dataMoverTopic not required anymore and defaults to empty string

KillianG:
Do not create a consumer if no topic provided

Issue: BB-565

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -53,7 +53,11 @@ const joiSchema = joi.object({
         bootstrapList: bootstrapListJoi,
     },
     topic: joi.string().required(),
-    dataMoverTopic: joi.string().default(''),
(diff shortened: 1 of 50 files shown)


extensions/replication/ReplicationConfigValidator.js
```diff
@@ -53,7 +53,7 @@ const joiSchema = joi.object({
         bootstrapList: bootstrapListJoi,
     },
     topic: joi.string().required(),
-    dataMoverTopic: joi.string().required(),
+    dataMoverTopic: joi.string().default(''),
```
welansari: maybe we should just leave it undefined ?

welansari (approved):

benzekrimaha (approved):

extensions/replication/ReplicationConfigValidator.js
francoisferrand: can we make the topic required if lifecycle transition is enabled?

PR #2556, opened 2024-10-15
Conditionally create sts client if stsEndpoint is provided or not

KillianG:
Issue: BB-519

### DIFF
--- a/bin/ensureServiceUser
+++ b/bin/ensureServiceUser
@@ -377,7 +377,11 @@ function wrapAction(actionFunc, serviceName, options) {
 
     const log = new werelogs.Logger(process.argv[1]).newRequestLogger();
     const iamClient = createIAMClient(options.parent);
-    const stsClient = createSTSClient(options.parent);
+    let stsClient;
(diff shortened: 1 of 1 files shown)


bin/ensureServiceUser
```diff
@@ -377,7 +377,11 @@ function wrapAction(actionFunc, serviceName, options) {
 
     const log = new werelogs.Logger(process.argv[1]).newRequestLogger();
     const iamClient = createIAMClient(options.parent);
-    const stsClient = createSTSClient(options.parent);
+    let stsClient;
+
+    if (options.stsEndpoint) {
+        stsClient = createSTSClient(options.parent);
+    }
```
ghost, line 384: Is it enough not to create the client? Shall we return an error here if the policyFile is defined, to avoid crashing in the middle of the `buildServiceUserForInAccountAssumeRoleHandlers` function?

KillianG: I can add a check but this is not supposed to happen anyway

PR #2557, opened 2024-10-15
allow consuming log records without wait when more logs are available

welansari:
The cooldown between batches is removed when more logs are available to consume.

This is mostly needed for S3C to keep good replication rates.

Issue: BB-532

### DIFF
--- a/lib/queuePopulator/QueuePopulator.js
+++ b/lib/queuePopulator/QueuePopulator.js
@@ -524,7 +524,24 @@ class QueuePopulator {
     _processLogEntries(params, done) {
         return async.map(
             this.logReaders,
-            (logReader, cb) => logReader.processLogEntries(params, cb),
+            (logReader, readerDone) => {
+                async.doWhilst(
+                    next => {
+                        logReader.processLogEntries(params, (err, hasMoreLog) => {
+                            if (err) {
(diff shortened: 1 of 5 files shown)


KillianG (approved):

lib/queuePopulator/QueuePopulator.js
```diff
@@ -524,7 +524,21 @@ class QueuePopulator {
...
+                    if (err) {
+                        return readerDone(err);
+                    }
+                    if (!this.qpConfig.exhaustLogSource || this.logReadersUpdate || !hasMoreLog) {
+                        return readerDone();
+                    }
+                    this.log.debug('skipping wait, more logs are available', {
+                        method: 'QueuePopulator._processLogEntries',
+                    });
+                    return logReader.processLogEntries(params, batchCb);
```
ghost: Is the callback recursion done on purpose here? We call batchCb from within the variable. If this goes for too long, without proper limit, we might have a stack overflow

welansari: Good point, although this implementation is how it's done in S3C and it worked for the most part, it's no reason to keep it this way.

ghost (approved):

PR #2558, opened 2024-10-15
Check large objects before replication

KillianG:
Issue: BB-566

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -73,6 +73,7 @@ const joiSchema = joi.object({
         minMPUSizeMB: joi.number().greater(0).default(20),
         probeServer: probeServerJoi.default(),
         circuitBreaker: joi.object().optional(),
+        sourceCheckIfSizeGreaterThanMB: joi.number().positive().default(100),
     }).required(),
     replicationStatusProcessor: {
         groupId: joi.string().required(),
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -16,6 +16,10 @@ const { getAccountCredentials } = require('../../../lib/credentials/AccountCrede
 const RoleCredentials = require('../../../lib/credentials/RoleCredentials');
 const { metricsExtension, metricsTypeQueued, metricsTypeCompleted, replicationStages } = require('../constants');
 
+const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
+
+const errorAlreadyCompleted = {};
+
 function _extractAccountIdFromRole(role) {
     return role.split(':')[4];
 }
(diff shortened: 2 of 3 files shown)


extensions/replication/ReplicationConfigValidator.js
```diff
@@ -73,6 +73,7 @@ const joiSchema = joi.object({
         minMPUSizeMB: joi.number().greater(0).default(20),
         probeServer: probeServerJoi.default(),
         circuitBreaker: joi.object().optional(),
+        sourceCheckIfSizeGreaterThanMB: joi.number().positive().default(100),
```
francoisferrand, line 76: May be a good occasion to review this option:
- can we find a better name (maxObjectSize ?)
- this option is weird, as it breaks crr semantics: what is the purpose of this option? Is it really used? Is there a matching option in cloudserver?

tests/functional/replication/queueProcessor.js
```diff
@@ -929,6 +930,19 @@ describe('queue processor functional tests with mocking', () => {
...
+                        { once: true });
+                    async.parallel([
+                        done => queueProcessorSF.processReplicationEntry(
+                            s3mock.getParam('kafkaEntry'), err => {
+                                assert.deepStrictEqual(err, errors.ObjNotFound);
+                                done();
+                            }),
+                    ], done);
+                });
+
```
francoisferrand: Should have tests verifying that "large" objects are not replicated...

extensions/replication/ReplicationConfigValidator.js
KillianG: AFAIK, This option simply adds a double check before repushing the object to replication, if an object is so large it takes time to replicate to avoid sending multiple times the same object for replication

We can probably find a better name for it though..
Why do you think it breaks crr sementic ? I don't know if it's really used but there is a default value so I guess it is no ?

welansari: > this option is weird, as it breaks crr semantics: what is the purpose of this option? Is it really used? Is there a matching option in cloudserver?

The logic in S3C serves to avoid replicating already replicated objects. We already have the check in Zenko's replication code.

According to [this comment ](https://scality.atlassian.net/browse/S3C-3603?focusedCommentId=270438), the performance overhead of checking the replication state of all objects before replicating is not worth it, as in S3C and contrary to Zenko we do not create a new version in this case but only overwrite the previously replicated one. The check was only added for large objects as they consumed lots of resources and slowed things down when getting replicated.

This option is used, and has a default value of 100Mb in Federation.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -701,6 +748,14 @@ class ReplicateObject extends BackbeatTask {
...
             },
             // Get data from source bucket and put it on the target bucket
+            next => {
+                if (!mdOnly &&
+                    sourceEntry.getContentLength() / 1000000 >=
+                    this.repConfig.queueProcessor.sourceCheckIfSizeGreaterThanMB) {
+                    return this._checkSourceReplication(sourceEntry, log, next);
+                }
+                return next();
+            },
```
welansari:
```suggestion
            next => {
                if (!mdOnly &&
                    sourceEntry.getContentLength() / 1000000 >=
                    this.repConfig.queueProcessor.sourceCheckIfSizeGreaterThanMB) {
                    return this._checkSourceReplication(sourceEntry, log, next);
                }
                return next();
            },
            // Get data from source bucket and put it on the target bucket
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -701,6 +748,14 @@ class ReplicateObject extends BackbeatTask {
                 this._setTargetAccountMd(destEntry, targetRole, log, next);
             },
             // Get data from source bucket and put it on the target bucket
+            next => {
+                if (!mdOnly &&
+                    sourceEntry.getContentLength() / 1000000 >=
+                    this.repConfig.queueProcessor.sourceCheckIfSizeGreaterThanMB) {
+                    return this._checkSourceReplication(sourceEntry, log, next);
+                }
+                return next();
```
welansari, line 756: We should explicitly test that we only do the check for object larger than `sourceCheckIfSizeGreaterThanMB`

welansari (approved):

benzekrimaha (approved):

PR #2561, opened 2024-10-17
Zookeeper Path handling

benzekrimaha:
Issue : BB-542

### DIFF
--- a/tests/unit/api/BackbeatAPI.spec.js
+++ b/tests/unit/api/BackbeatAPI.spec.js
@@ -1,4 +1,6 @@
 const assert = require('assert');
+const redis = require('ioredis');
+const sinon = require('sinon');
 
 const BackbeatAPI = require('../../../lib/api/BackbeatAPI');
 const BackbeatRequest = require('../../../lib/api/BackbeatRequest');
@@ -19,6 +21,14 @@ describe('BackbeatAPI', () => {
     before(() => {
         setupIngestionSiteMock();
         bbapi = new BackbeatAPI(config, fakeLogger, { timer: true });
+
+        sinon.stub(redis.prototype, 'connect').returns(Promise.resolve());
+        sinon.stub(redis.prototype, 'on').returnsThis();
+        sinon.stub(redis.prototype, 'quit').returns(Promise.resolve());
+    });
+
+    after(() => {
+        sinon.restore();
     });
 
     // valid routes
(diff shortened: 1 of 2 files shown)


tests/unit/api/BackbeatAPI.spec.js
```diff
@@ -19,6 +21,14 @@ describe('BackbeatAPI', () => {
     before(() => {
         setupIngestionSiteMock();
         bbapi = new BackbeatAPI(config, fakeLogger, { timer: true });
+
+        sinon.stub(redis.prototype, 'connect').returns(Promise.resolve());
+        sinon.stub(redis.prototype, 'on').returnsThis();
+        sinon.stub(redis.prototype, 'quit').returns(Promise.resolve());
```
ghost, line 27:
```suggestion
        sinon.stub(redis.prototype, 'connect').resolves();
        sinon.stub(redis.prototype, 'on').returnsThis();
        sinon.stub(redis.prototype, 'quit').resolves();
```

tests/unit/api/BackbeatAPI.spec.js
```diff
@@ -210,4 +220,40 @@ describe('BackbeatAPI', () => {
...
+            constructor(url, options, logger) {
+                zkManagerArgs = { url, options, logger };
+                this.once = sinon.stub().callsFake((event, callback) => {
+                    if (event === 'connected') {
+                        callback();
+                    }
+                });
+                this.removeAllListeners = sinon.stub();
+            }
+        }
```
ghost, line 236: I think we should define this class outside of a `describe`

tests/unit/api/BackbeatAPI.spec.js
```diff
@@ -210,4 +220,40 @@ describe('BackbeatAPI', () => {
...
+        it('should use connectionString directly if this._queuePopulator.mongo exists', done => {
+            bbapi._setZookeeper(() => {
+                assert(zkManagerArgs);
+                assert.strictEqual(zkManagerArgs.url, '127.0.0.1:2181');
+                done();
+            }, MockZookeeperManager);
+        });
+
+        it('should append zookeeperPath to connectionString if this._queuePopulator.mongo does not exist', done => {
+            delete bbapi._queuePopulator.mongo;
```
ghost, line 251: As we should not rely on the order ot unit tests for them to pass, I think this should be reverted at the end of the test (or just mock the value)

welansari:
Actually i think there's nothing to do here, as Zookeeper Client in 7.x is only used for heathchecks, meaning that it's only used to check whether or not Zookeeper is up and running and that we are able to connect to it. BackbeatAPI doesn't actually store any state in it. So i think we can safely leave the new path we use in 8.x.

PR #2562, opened 2024-10-18
use deleteObject to expire objects in S3c

welansari:
Zenko supports the "deleteObjectFromExpiration" API, which sets the proper originOp in the metadata to trigger a nortification when an object gets expired. S3C doesn't support that API, so we switch to the standard deleteObject API in this case.

Issue: BB-556

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
+++ b/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
@@ -98,6 +98,27 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
         return done();
     }
 
+    _getS3Action(actionType, accountId) {
+        let reqMethod;
+        if (actionType === 'deleteObject') {
+            const backbeatClient = this.getBackbeatClient(accountId);
+            if (!backbeatClient) {
+                return null;
+            }
+            // Zenko supports the "deleteObjectFromExpiration" API, which
+            // sets the proper originOp in the metadata to trigger a
+            // nortification when an object gets expired.
+            if (typeof backbeatClient.deleteObjectFromExpiration === 'function') {
+                return backbeatClient.deleteObjectFromExpiration.bind(backbeatClient);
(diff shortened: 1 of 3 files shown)


tests/unit/lifecycle/LifecycleDeleteObjectTask.spec.js
```diff
@@ -226,4 +227,79 @@ describe('LifecycleDeleteObjectTask', () => {
...
+        task.processActionEntry(entry, err => {
+            assert.ifError(err);
+            assert.strictEqual(s3Client.calls.abortMultipartUpload, 1);
+            assert.strictEqual(s3Client.calls.deleteObject, 0);
+            assert.strictEqual(backbeatClient.times.deleteObjectFromExpiration, 0);
+            done();
+        });
+    });
+
+    it('should return an error when it can\'t get the BackbeatClinet', done => {
```
ghost:
```suggestion
    it('should return an error when it can\'t get the BackbeatClient', done => {
```

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
ghost, line 1: (picking randomly)

We do not support the `deleteObjectFromExpiration` method in S3C - so changes LGTM. But do we plan to support it in a unified s3 branch, which would solve this issue? Any reason not to support it in S3C?

ghost (approved):

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
welansari: To be honest, i'm not sure we want to support it because of RING use cases, as it may add significant performance overhead with the additional metadata update we do before each expiration.

I guess what we can do if we really want to get rid of this code, is to just create a dummy route that calls the delete API behind the scenes in Cloudserver.

KillianG (approved):

PR #2563, opened 2024-10-18
Log consumer lag metrics

KillianG:
Issue: BB-561

### DIFF
--- a/extensions/notification/queueProcessor/QueueProcessor.js
+++ b/extensions/notification/queueProcessor/QueueProcessor.js
@@ -14,6 +14,7 @@ const NotificationDestination = require('../destination');
 const configUtil = require('../utils/config');
 const messageUtil = require('../utils/message');
 const NotificationConfigManager = require('../NotificationConfigManager');
+const libConstants = require('../../../lib/constants');
 
 const processedEvents = ZenkoMetrics.createCounter({
     name: 's3_notification_queue_processor_events_total',
@@ -92,6 +93,7 @@ class QueueProcessor extends EventEmitter {
         this.kafkaConfig = kafkaConfig;
         this.notifConfig = notifConfig;
         this.destinationId = destinationId;
+        this.serviceName = libConstants.services.notificationQueueProcessor;
         this.destinationConfig
             = notifConfig.destinations.find(dest => dest.resource === destinationId);
         assert(this.destinationConfig, `Invalid destination argument "${destinationId}".` +
--- a/extensions/notification/queueProcessor/QueueProcessor.js
+++ b/extensions/notification/queueProcessor/QueueProcessor.js
@@ -6,7 +6,7 @@ const async = require('async');
 const assert = require('assert');
 const util = require('util');
 const { ZenkoMetrics } = require('arsenal').metrics;
-const { wrapGaugeSet } = require('../../../lib/util/metrics')
+const { wrapGaugeSet } = require('../../../lib/util/metrics');
 const errors = require('arsenal').errors;
 
 const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
--- a/extensions/notification/queueProcessor/QueueProcessor.js
+++ b/extensions/notification/queueProcessor/QueueProcessor.js
@@ -29,7 +29,7 @@ function onQueueProcessorEventProcessed(destination, eventType) {
 }
 
 const kafkaLagMetric = ZenkoMetrics.createGauge({
-    name: 's3_replication_queue_lag',
+    name: 's3_notification_queue_lag',
     help: 'Number of update entries waiting to be consumed from the Kafka topic',
     labelNames: ['origin', 'containerName', 'partition', 'serviceName'],
 });
--- a/extensions/notification/queueProcessor/QueueProcessor.js
+++ b/extensions/notification/queueProcessor/QueueProcessor.js
@@ -27,6 +27,29 @@ function onQueueProcessorEventProcessed(destination, eventType) {
     });
 }
 
+const kafkaLagMetric = ZenkoMetrics.createGauge({
+    name: 's3_replication_queue_lag',
+    help: 'Number of update entries waiting to be consumed from the Kafka topic',
+    labelNames: ['origin', 'containerName', 'partition', 'serviceName'],
+});
+
+
+/**
+ * Contains methods to incrememt different metrics
+ * @typedef {Object} MetricsHandler
+ * @property {CounterInc} dataReplicationStatus - Increments the replication status metric for data operation
+ * @property {CounterInc} metadataReplicationStatus - Increments the replication status metric for metadata operation
+ * @property {CounterInc} dataReplicationBytes - Increments the replication bytes metric for data operation
+ * @property {CounterInc} metadataReplicationBytes - Increments the replication bytes metric for metadata operation
+ * @property {CounterInc} sourceDataBytes - Increments the source data bytes metric
+ * @property {CounterInc} reads - Increments the read metric
+ * @property {CounterInc} writes - Increments the write metric
+ * @property {HistogramObserve} timeElapsed - Observes the time elapsed metric
+ */
+const metricsHandler = {
+    lag: wrapGaugeSet(kafkaLagMetric, defaultLabels),
+};
+
 class QueueProcessor extends EventEmitter {
     /**
      * Create a queue processor object to activate notification from a
@@ -329,6 +352,19 @@ class QueueProcessor extends EventEmitter {
      */
     async handleMetrics(res, log) {
         log.debug('metrics requested');
+
+        if (this.repConfig.queueProcessor.logConsumerMetricsIntervalS && this._consumer) {
+            // consumer stats lag is on a different update cycle so we need to
+            // update the metrics when requested
+            const lagStats = this._consumer.consumerStats.lag;
+            Object.keys(lagStats).forEach(partition => {
+                metricsHandler.lag({
+                    partition,
+                    serviceName: this.serviceName,
+                }, lagStats[partition]);
+            });
+        }
+
         res.writeHead(200, {
             'Content-Type': ZenkoMetrics.asPrometheusContentType(),
         });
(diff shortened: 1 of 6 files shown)


lib/BackbeatConsumer.js
```diff
@@ -241,6 +246,37 @@ class BackbeatConsumer extends EventEmitter {
         return this._consumer.once('ready', () => {
             this._consumerReady = true;
             this._checkIfReady();
+            if (this._logConsumerMetricsIntervalS !== undefined) {
+                this._consumer.on('event.stats', res => {
+                    const statsObj = JSON.parse(res.message);
+                    if (typeof statsObj !== 'object') {
```
ghost: If you successfully parse a string with JSON.parse, it will always be an object. If not, the JSON.parse will throw an error. I think here we should use a safe parse function (IIRC, we have one in arsenal), or we should wrap that in a try/catch

lib/BackbeatConsumer.js
```diff
@@ -241,6 +246,37 @@ class BackbeatConsumer extends EventEmitter {
...
+                        return undefined;
+                    }
+                    const topicStats = statsObj.topics[this._topic];
+                    if (typeof topicStats !== 'object') {
```
ghost:
```suggestion
                    const topicStats = statsObj.topics?.[this._topic];
                    if (typeof topicStats !== 'object') {
```

lib/BackbeatConsumer.js
```diff
@@ -241,6 +247,40 @@ class BackbeatConsumer extends EventEmitter {
         return this._consumer.once('ready', () => {
             this._consumerReady = true;
             this._checkIfReady();
+            if (this._logConsumerMetricsIntervalS !== undefined) {
+                this._consumer.on('event.stats', res => {
+                    const statsObj = safeJsonParse(res.message);
+                    if (statsObj.error) {
+                        this._log.error('error parsing consumer stats', { error: statsObj.error, topic: this._topic,
```
ghost, line 254: Can we implement some tests to test the error paths of this function? As we test only the happy path, we should add tests to ensure we are confident when we get bad input objects (either the message or result.topic)

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -406,6 +406,7 @@ class QueueProcessor extends EventEmitter {
             groupId,
             concurrency: this.repConfig.queueProcessor.concurrency,
             queueProcessor: queueProcessorFunc,
+            logConsumerMetricsIntervalS: this.repConfig.queueProcessor.logConsumerMetricsIntervalS,
```
welansari, line 417: Should also expose the metrics in Prometheus like we do in 7.x =>
https://github.com/scality/backbeat/blob/57fc2357d1bc9fe477b0f38f012f2e3633f68f70/extensions/replication/queueProcessor/QueueProcessor.js#L574-L580

We might want to do that only when `logConsumerMetricsIntervalS` is set (it's only set in federation, not zkop)

lib/BackbeatConsumer.js
KillianG: I am not sure it is that simple to introduce.. neither if it has any interest, if we are not able to parse for some reason, we just don't log the lag offset..

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -1019,6 +1020,18 @@ class QueueProcessor extends EventEmitter {
...
+            // consumer stats lag is on a different update cycle so we need to
+            // update the metrics when requested
+            const lagStats = this._consumer.consumerStats.lag;
+            Object.keys(lagStats).forEach(partition => {
+                metricsHandler.lag({
+                    partition,
+                    serviceName: this.serviceName,
+                }, lagStats[partition]);
+            });
+        }
```
welansari, line 1043: Also done in the statusProcessor :
https://github.com/scality/backbeat/blob/57fc2357d1bc9fe477b0f38f012f2e3633f68f70/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js#L457-L460

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -1019,6 +1020,18 @@ class QueueProcessor extends EventEmitter {
      */
     static async handleMetrics(res, log) {
         log.debug('metrics requested');
+
+        if (this.repConfig.queueProcessor.logConsumerMetricsIntervalS) {
+            // consumer stats lag is on a different update cycle so we need to
+            // update the metrics when requested
+            const lagStats = this._consumer.consumerStats.lag;
+            Object.keys(lagStats).forEach(partition => {
+                metricsHandler.lag({
```
welansari, line 1038: The lag metric needs to be added
https://github.com/scality/backbeat/blob/57fc2357d1bc9fe477b0f38f012f2e3633f68f70/extensions/replication/queueProcessor/QueueProcessor.js#L62-L66

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -79,6 +79,12 @@ const metadataReplicationBytesMetric = ZenkoMetrics.createCounter({
     labelNames: ['origin', 'serviceName', 'location'],
 });
 
+const kafkaLagMetric = new promClient.Gauge({
```
welansari: `promClient` was replaced by `ZenkoMetrics`. We need to adapt the code to that. Same in the `replicationStatusProcessor`.

francoisferrand (approved):
* please rebase on top of latest backbeat, this is somwhat old
* it does not seem like the responsability of backbeat to compute lag: there are many tools doing that already (like we do in Zenko). Please check again with s3c "experts" if we could not go that road: if yes, PR can be dropped ; if no, please create a cleanup ticket so we remove this metrics in the future (in post-S3C branches)

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -1020,6 +1028,18 @@ class QueueProcessor extends EventEmitter {
      */
     static async handleMetrics(res, log) {
         log.debug('metrics requested');
+
+        if (this.repConfig.queueProcessor.logConsumerMetricsIntervalS) {
```
welansari: Should also check if the consumer exists ?

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -552,6 +560,19 @@ class ReplicationStatusProcessor {
     async handleMetrics(res, log) {
         log.debug('metrics requested');
         const metrics = await ZenkoMetrics.asPrometheus();
+
+        if (this.repConfig.queueProcessor.logConsumerMetricsIntervalS) {
```
welansari: same here

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -406,6 +413,7 @@ class QueueProcessor extends EventEmitter {
             groupId,
             concurrency: this.repConfig.queueProcessor.concurrency,
             queueProcessor: queueProcessorFunc,
+            logConsumerMetricsIntervalS: this.repConfig.queueProcessor.logConsumerMetricsIntervalS,
             canary: true,
             circuitBreaker: this.circuitBreakerConfig,
```
welansari, line 419: I just noticed that we also have this in bucket notification's queue processor. I think we should add the same logic there too.

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -329,6 +349,19 @@ class QueueProcessor extends EventEmitter {
...
         log.debug('metrics requested');
+
+        if (this.repConfig.queueProcessor.logConsumerMetricsIntervalS && this._consumer) {
+            // consumer stats lag is on a different update cycle so we need to
+            // update the metrics when requested
+            const lagStats = this._consumer.consumerStats.lag;
+            Object.keys(lagStats).forEach(partition => {
+                metricsHandler.lag({
+                    partition,
+                    serviceName: this.serviceName,
```
welansari, line 362: `this.serviceName` is not defined

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -552,6 +560,19 @@ class ReplicationStatusProcessor {
...
         const metrics = await ZenkoMetrics.asPrometheus();
+
+        if (this.repConfig.queueProcessor.logConsumerMetricsIntervalS && this._consumer) {
+            // consumer stats lag is on a different update cycle so we need to
+            // update the metrics when requested
+            const lagStats = this._consumer.consumerStats.lag;
+            Object.keys(lagStats).forEach(partition => {
+                this.metricHandlers.lag({
+                    partition,
+                    serviceName: this.serviceName,
```
welansari, line 572: same here

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -6,13 +6,15 @@ const async = require('async');
...
 const { ZenkoMetrics } = require('arsenal').metrics;
+const { wrapGaugeSet } = require('../../../lib/util/metrics');
 const errors = require('arsenal').errors;
 
 const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
 const NotificationDestination = require('../destination');
 const configUtil = require('../utils/config');
 const messageUtil = require('../utils/message');
 const NotificationConfigManager = require('../NotificationConfigManager');
+const libConstants = require('../../../lib/constants');
```
welansari:
```suggestion
const { notificationQueueProcessor } = require('../../../lib/constants').services;
```

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -18,13 +18,15 @@ const FailedCRRProducer = require('../failedCRR/FailedCRRProducer');
 const ReplayProducer = require('../replay/ReplayProducer');
 const MetricsProducer = require('../../../lib/MetricsProducer');
 const { http: HttpAgent, https: HttpsAgent } = require('httpagent');
+const libConstants = require('../../../lib/constants');
```
welansari:
```suggestion
const { replicationStatusProcessor } = require('../../../lib/constants').services;
```

welansari (approved):

PR #2565, opened 2024-10-24
snapshot datamover topic offsets only when transition is supported

welansari:
Issue: BB-557


extensions/lifecycle/util/rules.js
```diff
@@ -220,8 +220,13 @@ function isExpirationRule(ruleName) {
         ruleName === 'AbortIncompleteMultipartUpload';
 }
 
+function rulesSupportTransition(lcRules) {
+    return lcRules.some(rule => ['transitions', 'noncurrentVersionTransition'].includes(rule));
```
ghost: We can create a constant to avoid re-creating the array everytime

ghost (approved):
lgtm, just a small comment

benzekrimaha (approved):

PR #2568, opened 2024-10-25
support dedicated replayProcessor probeservers

welansari:
Support dedicated replay processor per site probe config.

Issue: BB-560

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -83,6 +83,9 @@ const joiSchema = joi.object({
         concurrency: joi.number().greater(0).default(10),
         probeServer: probeServerJoi.default(),
     },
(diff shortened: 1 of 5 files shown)


francoisferrand:
In s3c, all "services" are started in the same container: hence they need separate ports.

So they may need separate probe server configurations for queue processor and replay processor?

welansari:
> So they may need separate probe server configurations for queue processor and replay processor?

~~Yes, that's what this PR is for, i'm just using the same place in the config to host both configs.
The probe configs of the replay processors will have a "topicName" field that we can filter on, the queue processor's config won't have it.~~

I guess the logic would be confusing if we merge both configs into one. I added a dedicated config for the replay processor like we currently have in S3C.

KillianG (approved):

benzekrimaha (approved):

PR #2570, opened 2024-11-04
Revert "use deleteObject to expire objects in S3c"

welansari:
This reverts commit e2d61043936ddbd7ccf7ba43160f13cc1b9a14d6 which introduces logic to select which deletion API to use in expiration based on the backbeatClient API definition. This is useless since the unified backbeat will only be using the definition from. 8.x and that includes everything.

Added failback to deleteObject in case we use an older Cloudserver.

Issue: BB-617

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
+++ b/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
@@ -103,25 +103,62 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
         return done();
     }
 
-    _getS3Action(actionType, accountId) {
-        let reqMethod;
-        if (actionType === 'deleteObject') {
-            const backbeatClient = this.getBackbeatClient(accountId);
-            if (!backbeatClient) {
-                return null;
-            }
-            // Zenko supports the "deleteObjectFromExpiration" API, which
-            // sets the proper originOp in the metadata to trigger a
-            // nortification when an object gets expired.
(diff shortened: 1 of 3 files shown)


KillianG (approved):

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -104,24 +104,12 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
...
-            // Zenko supports the "deleteObjectFromExpiration" API, which
-            // sets the proper originOp in the metadata to trigger a
-            // nortification when an object gets expired.
-            if (typeof backbeatClient.deleteObjectFromExpiration === 'function') {
-                return backbeatClient.deleteObjectFromExpiration.bind(backbeatClient);
-            }
-            reqMethod = 'deleteObject';
-        } else {
-            reqMethod = 'abortMultipartUpload';
+            return backbeatClient?.deleteObjectFromExpiration?.bind(backbeatClient);
```
francoisferrand: * this requires the new API being added to Cloudserver? If it is not present (e.g. cloudserver not bumped), what will happen? should we fallback to the regular `deleteObject` on "not implemented" error ?
* what happens if `backbeatClient?.deleteObjectFromExpiration? == null` : is this handled in the code? Either we should handle that (unlikely) scenario, or we should remove the conditionals and test that it does work indeed (e.g. unit test with `assert(_getS3Action('deleteObject) != null)`)

ghost (approved):

PR #2572, opened 2024-11-04
Fix the dataMoverTopic config param always being required

welansari:
Extension configs are validated independently of each other, so validating the dataMoverTopic based on the supportedLifecycleRules param is not currently possible.

Issue: BB-619

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -53,11 +53,7 @@ const joiSchema = joi.object({
         bootstrapList: bootstrapListJoi,
     },
     topic: joi.string().required(),
-    dataMoverTopic: joi.string().when('..lifecycle.supportedLifecycleRules', {
-        is: joi.array().has(joi.string().valid('transition')),
-        then: joi.required(),
-        otherwise: joi.forbidden(),
-    }),
+    dataMoverTopic: joi.string().optional(),
     replicationStatusTopic: joi.string().required(),
     monitorReplicationFailures: joi.boolean().default(true),
     replicationFailedTopic: joi.string().required(),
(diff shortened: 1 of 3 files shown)


extensions/replication/ReplicationConfigValidator.js
```diff
@@ -53,11 +53,7 @@ const joiSchema = joi.object({
         bootstrapList: bootstrapListJoi,
     },
     topic: joi.string().required(),
-    dataMoverTopic: joi.string().when('..lifecycle.supportedLifecycleRules', {
-        is: joi.array().has(joi.string().valid('transition')),
-        then: joi.required(),
-        otherwise: joi.forbidden(),
-    }),
+    dataMoverTopic: joi.string().optional(),
```
francoisferrand, line 56: can we move this validation somewhere else anyway?

KillianG (approved):

tests/unit/lib/config/Config.spec.js
```diff
@@ -6,20 +6,47 @@ const { Config } = require('../../../../lib/Config');
 const backbeatConfig = require('./config.json');
 
 describe('Config', () => {
+    let config;
+    let testConfig;
+
+    beforeEach(() => {
+        config = new Config();
+        // deep copy the config to avoid modifying the original
+        testConfig = JSON.parse(JSON.stringify(backbeatConfig));
```
ghost, line 15: Nit
```suggestion
        testConfig = { ...backbeatConfig };
```

ghost (approved):

tests/unit/lib/config/Config.spec.js
welansari: `{ ...backbeatConfig }` doesn't deep copy nested objets.

PR #2573, opened 2024-11-04
properly pass compressionType and requiredAcks to BackbeatProducer

welansari:
Issue: BB-620


extensions/notification/destination/KafkaNotificationDestination.js
```diff
@@ -52,6 +52,8 @@ class KafkaNotificationDestination extends NotificationDestination {
             topic,
             pollIntervalMs,
             auth,
+            compressionType: 'none',
+            requiredAcks: '1',
```
francoisferrand: these params may need to be added to the destination config, may be done now or later, but need ticket either way (at product or at least zenko level, so that we can also track the remaining work: zkop, ...)

lib/BackbeatProducer.js
```diff
@@ -57,7 +57,7 @@ class BackbeatProducer extends EventEmitter {
                 pollIntervalMs: joi.number().default(DEFAULT_POLL_INTERVAL),
                 maxRequestSize: joi.number().default(KAFKA_PRODUCER_MESSAGE_MAX_BYTES),
                 compressionType: joi.string().default(KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE),
-                requiredAcks: joi.number().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
+                requiredAcks: joi.string().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
```
francoisferrand: why pass a string? this reduces the effectiveness of validation, as any string will be valid...

if we just neet to support some way of passing "empty" param to kafka, i.e. use kafka defaults, we should be able to just do `joi.number().allow("").default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS)` : which (according to the docs) should allow numbers or empty string, and fall back to default value if `undefined` value is provided

(or maybe just `joi.string().regex(/^[0-9]*$/, 'number').default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS)`)

extensions/gc/GarbageCollector.js
```diff
@@ -132,6 +132,8 @@ class GarbageCollector extends EventEmitter {
             kafka: {
                 hosts: this._kafkaConfig.hosts,
                 site: this._kafkaConfig.site,
+                compressionType: this._kafkaConfig.compressionType,
+                requiredAcks: this._kafkaConfig.requiredAcks,
```
francoisferrand, line 136: won't this change the behavior on artesca?

If there are some good values, best to use them indeed;
but not sure if either
- we should rely on defaults from Kafka : maybe they are good (→ should be used as default everywhere, i.e. both s3c and zenko) or maybe we have better values (→ should be set in both products)
- we should actually have default values in backbeat, if both product don't agree
- if the products ahve different defaults, should probably be explicit : i.e. put the values in each product (or if we can define a "correct" default, but don't want to change it in S3C for compatibility: maybe defaut value in backbeat and "override" the default in s3c)

lib/config.joi.js
```diff
@@ -33,7 +33,7 @@ const joiSchema = joi.object({
         maxRequestSize: joi.number().default(KAFKA_PRODUCER_MESSAGE_MAX_BYTES),
         site: joi.string(),
         compressionType: joi.string().default(KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE),
-        requiredAcks: joi.number().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
+        requiredAcks: joi.string().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
```
francoisferrand: same

lib/BackbeatProducer.js
welansari: I passed a string because the default value used in Artesca is 'all' which waits for all brokers to ack. Looks like we can use -1 too for the same purpose, Leaving it as a number.

extensions/gc/GarbageCollector.js
welansari: The default values is what we use in Artesca, so no the behaviour won't change:
- Waiting for all brokers to ack the messages
- Using `Zstd` compression

I've talked with the S3C team and for RING 9.x we want to keep the same config from before :
- Only wait for one broker to ack the messages.
- Use broker level compression, no compression done by the producers.
The custom values will be set in the backbeat config in federation

benzekrimaha (approved):
LGTM

francoisferrand (approved):

PR #2574, opened 2024-11-04
BB-624 Retry connection to kafka if backbeat processes started before

BourgoisMickael:
- Suppress AWS SDK warning in stderr
- Improve container image size from 2.34GB to 1.55GB (layer chown + dockerignore)
```bash
REPOSITORY                   TAG                         IMAGE ID       SIZE         BLOB SIZE
scality/backbeat             7.70.15.rc1-nodesvc-base    ea2e98003485    1.6 GiB      491.7 MiB
scality/backbeat             9.4.0.0                     cf4ecf197272    2.3 GiB      1.8 GiB

```
- Backport / cherrypick some fixes from zenko to ~~crash~~ exit after a timeout on connection error with kafka client (for CRR, Lifecycle, Bucket notification). Even for bucket notification destination.
- Prevent ProbeServer crash on startup by fixing metric handling when the consumer is not set up and delaying it after the queue components are started.

New behavior: components will try to connect for 60s instead of 30s and ~~crash~~ exit and restart if kafka client can't connect with error like:
```javascript
{"name":"Backbeat:QueueProcessor:task","time":1730775546353,"method":"QueueProcessor::task","error":{"origin":"local","message":"broker transport failure","code":-195,"errno":-195,"stack":"Error: Local: Broker transport failure\n    at Function.createLibrdkafkaError [as create] (/home/scality/backbeat/node_modules/node-rdkafka/lib/error.js:454:10)\n    at /home/scality/backbeat/node_modules/node-rdkafka/lib/client.js:350:28"},"level":"error","message":"error during queue processor initialization","hostname":"MDM-RING-46789-store-1","pid":594}
```
```javascript
{"name":"Backbeat:QueueProcessor:task","time":1730776197182,"error":{"origin":"local","message":"timed out","code":-185,"errno":-185,"stack":"Error: Local: Timed out\n    at Function.createLibrdkafkaError [as create] (/home/scality/backbeat/node_modules/node-rdkafka/lib/error.js:454:10)\n    at /home/scality/backbeat/node_modules/node-rdkafka/lib/client.js:350:28"},"method":"MetricsProducer::setupProducer","level":"error","message":"error starting metrics producer for queue processor","hostname":"MDM-RING-46789-store-2","pid":146}
```


BourgoisMickael:
ping

BourgoisMickael:
ping

BourgoisMickael:
@bert-e create_pull_requests

bin/queuePopulator.js
```diff
@@ -52,6 +52,7 @@ const queuePopulator = new QueuePopulator(
     zkConfig, kafkaConfig, qpConfig, httpsConfig, mConfig, rConfig, extConfigs);
```
nicolas2bert, line 52: Nit: commit title should be probing

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -142,7 +142,15 @@ class QueueProcessor extends EventEmitter {
                     queueProcessor: this.processKafkaEntry.bind(this),
                     logConsumerMetricsIntervalS,
                 });
-                this._consumer.on('error', () => { });
+                this._consumer.on('error', err => {
+                    this.logger.error('error starting notification consumer',
+                    { method: 'QueueProcessor.start', error: err.message });
+                    // crash if got error at startup
+                    if (!this.isReady()) {
+                        return next(err);
```
nicolas2bert, line 153: how would it crash here?

lib/BackbeatConsumer.js
```diff
@@ -186,7 +186,18 @@ class BackbeatConsumer extends EventEmitter {
             consumerParams['client.rack'] = this._site;
         }
         this._consumer = new kafka.KafkaConsumer(consumerParams);
-        this._consumer.connect();
+        // For S3C 60s + prevent any side effect with sanitycheck topic
+        this._consumer.connect({ timeout: 60000 }, (err) => {
+            if (err) {
+                this.emit('error', err);
+            }
+            // const opts = { topic: 'backbeat-sanitycheck', timeout: 10000 };
```
nicolas2bert: do we really need that?

lib/BackbeatConsumer.js
```diff
@@ -186,7 +186,18 @@ class BackbeatConsumer extends EventEmitter {
             consumerParams['client.rack'] = this._site;
         }
         this._consumer = new kafka.KafkaConsumer(consumerParams);
-        this._consumer.connect();
+        // For S3C 60s + prevent any side effect with sanitycheck topic
```
nicolas2bert: adjust the comment.

nicolas2bert (commented):
Suggestions:
- add tests for "Prevent ProbeServer crash on startup by delaying it after the queue components are started".
- add tests for Consumer service crashes in the event of a consumer error?

extensions/notification/queueProcessor/QueueProcessor.js
BourgoisMickael: I've got callback error because I've not copied correctly from zenko branch. I'll fix that.

~~Then it's not a crash, it shoud print an error log and after there is no more instructions, if no handles / connection retain the process active, it should stop with exit code 0.~~

I'll add a process.exit(1) then like it's done in zenko

lib/BackbeatConsumer.js
BourgoisMickael: I'll remove

BourgoisMickael:
reset

BourgoisMickael:
> Suggestions:
> 
> * add tests for "Prevent ProbeServer crash on startup by delaying it after the queue components are started".
> * add tests for Consumer service crashes in the event of a consumer error?

@nicolas2bert 
I added a simple test for the crash on probing, but for testing the whole task  file that stops on error it's not easily testable, I'll try to use `vm` module but it might take some time.

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -116,17 +116,20 @@ class QueueProcessor extends EventEmitter {
...
      */
-    start(options) {
+    start(options, done) {
         async.series([
             next => this._setupZookeeper(next),
             next => this._setupNotificationConfigManager(next),
             next => this._setupDestination(this.destinationConfig.type, next),
-            next => this._destination.init(() => {
+            // if connection to destination fails, process will stop & restart
+            next => this._destination.init(next),
```
BourgoisMickael, line 128: About this it looks like on zenko it could crash with a "callback already called" if destination Kafka is not up on startup

BourgoisMickael:
Here is an Integration e2e run: https://github.com/scality/Integration/actions/runs/11699734386/job/32582300296

It should not trigger any exit implemented in this PR as all kafka dependencies are already UP in those tests

extensions/notification/queueProcessor/QueueProcessor.js
```diff
@@ -142,22 +145,31 @@ class QueueProcessor extends EventEmitter {
                     queueProcessor: this.processKafkaEntry.bind(this),
                     logConsumerMetricsIntervalS,
                 });
-                this._consumer.on('error', () => { });
+                this._consumer.on('error', err => {
+                    this.logger.error('error starting notification consumer',
+                    { method: 'QueueProcessor.start', error: err.message });
+                    // crash if got error at startup
+                    if (!this.isReady()) {
```
welansari, line 152: Should we use `this._consumer.isReady()` instead of adding a new method that does the same ?

BourgoisMickael: I copied exactly the required changes coming from ZENKO branches so there would be no conflict with waterflow

BourgoisMickael: It seems simpler like that

nicolas2bert (approved):

welansari (approved):

PR #2575, opened 2024-11-05
Disable assumeRole for the conductor and lifecycle queue populator in S3C

welansari:
Issue: BB-622

### DIFF
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -39,14 +39,14 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         super(params);
         this._authConfig = params.authConfig;
 
-        this.vaultClientWrapper = new VaultClientWrapper(
-            LIFEYCLE_POPULATOR_CLIENT_ID,
-            params.vaultAdmin,
-            this._authConfig,
-            this.log,
-        );
+        if (this._authConfig?.type === authTypeAssumeRole) {
+            this.vaultClientWrapper = new VaultClientWrapper(
+                LIFEYCLE_POPULATOR_CLIENT_ID,
+                params.vaultAdmin,
+                this._authConfig,
+                this.log,
+            );
 
-        if (this._authConfig.type === authTypeAssumeRole) {
             this.vaultClientWrapper.init();
         }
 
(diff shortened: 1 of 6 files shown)


extensions/utils/VaultClientWrapper.js
```diff
@@ -128,7 +141,7 @@ class VaultClientWrapper {
     }
 
     tempCredentialsReady() {
-        if (this._authConfig.type !== authTypeAssumeRole) {
+        if (this._authConfig.type !== authTypeAssumeRole || !this._useTempCredentials) {
```
francoisferrand: can't we rely only on `type !== authTypeAssumeRole`, instead of introducing this extra `useTempCredentials` ?

in S3C it should not be assume role anyway (since it is not supported), so we should be fine...

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -37,13 +37,24 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
         this._authConfig = params.authConfig;
 
+        let vaultAdminUseTempCreds = true;
+        // Falling back to the global lifecycle auth config when in S3C
+        // and disabling AssumeRole
+        if (!this._vaultAdmin) {
+            this._vaultAdmin = params.config.auth.vault;
+            this._authConfig = params.config.auth;
+            vaultAdminUseTempCreds = false;
+        }
```
francoisferrand: in relation with https://github.com/scality/backbeat/pull/2575/files#r1830765541 , would be nicer to have just a fallback in the config:

```suggestin
        this._vaultAdmin = params.vaultAdmin || params.config.auth.vault;
        this._authConfig = params.authConfig || params.config.auth;
```

...or maybe even better update the config in S3C (if we can?)

extensions/utils/VaultClientWrapper.js
welansari: S3C does actually support assumeRole and we do have that in the config of S3C as well. I introduced a new type of authentication (no auth) for the getAccounts API.

extensions/utils/VaultClientWrapper.js
```diff
@@ -21,7 +21,7 @@ class VaultClientWrapper {
     }
 
     init() {
-        if (this._authConfig.type !== authTypeAssumeRole) {
+        if (![authTypeAssumeRole, authTypeNone].includes(this._authConfig.type)) {
```
KillianG, line 24: `[authTypeAssumeRole, authTypeNone]`
Can this be a variable somewhere ? To avoid repetition and ease the maintenance in case we want to add/remove one for some reason?

KillianG (approved):

extensions/utils/VaultClientWrapper.js
welansari: Those are already constants

benzekrimaha (approved):

PR #2581, opened 2024-11-05
use the same queue populator zookeeper path as in 7.x

welansari:
This change would not affect Zenko. Offsets are no longer stored in Zookeeper as they are managed directly by Kafka when using the Kafka log consumer.

Issue: BB-533

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -39,8 +39,7 @@ class LogReader {
         this.zkClient = params.zkClient;
         this.kafkaConfig = params.kafkaConfig;
         this.logConsumer = params.logConsumer;
-        this.pathToLogOffset = `${config.queuePopulator.zookeeperPath}/` +
-            `logState/${params.logId}/logOffset`;
+        this.pathToLogOffset = `/logState/${params.logId}/logOffset`;
         this.logOffset = null;
         this.log = params.logger;
         this._producers = {};


lib/queuePopulator/LogReader.js
```diff
@@ -39,8 +39,7 @@ class LogReader {
         this.zkClient = params.zkClient;
         this.kafkaConfig = params.kafkaConfig;
         this.logConsumer = params.logConsumer;
-        this.pathToLogOffset = `${config.queuePopulator.zookeeperPath}/` +
-            `logState/${params.logId}/logOffset`;
+        this.pathToLogOffset = `/logState/${params.logId}/logOffset`;
```
francoisferrand, line 42: * is `config.queuePopulator.zookeeperPath` used elsewhere ? can't we just set it to an empty string in case of S3C ?
* if `pathToLogOffset` affects only Metadata log reader, should it not be moved to the dedicated LogReader subclass?
* this will affect the (deprecated) mongodb "replicaset" LogReader → not an issue since it is unused, but may be the occasion to remove it (here or in a separate ticket to be created)

KillianG (approved):

lib/queuePopulator/LogReader.js
welansari: > is config.queuePopulator.zookeeperPath used elsewhere ? can't we just set it to an empty string in case of S3C ?

We cannot just set it as empty string as it used in the QueuePopulator path as the base path for zookeeper.

> if pathToLogOffset affects only Metadata log reader, should it not be moved to the dedicated LogReader subclass?

There's only one logReader class actually, the LogConsumer is what's custom per backend. This class however does not deal with offsets at all, it only handles consumption of the logs.

> this will affect the (deprecated) mongodb "replicaset" LogReader → not an issue since it is unused, but may be the occasion to remove it (here or in a separate ticket to be created)

This could be done in BB-481, as the Kafka implementation currently creates error logs when trying to store the offset.

ghost (approved):

PR #2587, opened 2024-11-06
Make probeserver config optional for bucket notification processes in S3C

welansari:
- `validate()` doesn't seem to validate the config at all, replaced it with `joi.attempt()` that we use to validate the extension configs.
- Moved the probeserver validation out of the probeserver startup function and into the joi config schema, as it can now be optional. `undefined` is now returned by the startup function when no probeserver config is provided. This case is already supported in the entrypoint files.
- In S3C, bucket notification processes only contain configuration for the notification extension. We make the probeserver config optional in this case as it is not currently supported in federation.

Issue: BB-623

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -2,6 +2,7 @@
 
 const assert = require('assert');
 const { EventEmitter } = require('events');
+const joi = require('joi');
 
 const fs = require('fs');
 const path = require('path');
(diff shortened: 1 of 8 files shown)


lib/Config.js
```diff
@@ -54,10 +55,16 @@ class Config extends EventEmitter {
...
+        this._parseConfig(config);
+    }
+
+    /**
+     * Parses Backbeat's configuration
+     * @param {Object} config backbeat configuration
+     * @returns {undefined}
+     */
+    _parseConfig(config) {
+        const parsedConfig = joi.attempt(config, backbeatConfigJoi);
```
francoisferrand, line 67: should we not detect & throw an error if the config is not valid?

welansari: `joi.attempt` throws an error when the config is not valid

francoisferrand (approved):

KillianG (approved):

PR #2588, opened 2024-11-06
Improvement/bb 614 remove redis metrics

benzekrimaha:
Issue : BB-614


extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -97,32 +77,15 @@ class MongoQueueProcessor {
      * @return {undefined}
      */
     start() {
-        this.logger.info('starting mongo queue processor');
```
ghost, line 100: We can keep this log, won't hurt

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -97,32 +77,15 @@ class MongoQueueProcessor {
...
-            }),
-        ], error => {
-            if (error) {
+        this._mongoClient.setup(err => {
+            if (err) {
+                this.logger.error('could not connect to MongoDB', {
+                    method: 'MongoQueueProcessor.start',
+                    error: err.message,
+                });
                 this.logger.fatal('error starting mongo queue processor');
```
ghost, line 86: No need to log twice I would say, just log the fatal message with all the info

```suggestion
                this.logger.fatal('could not connect to MongoDB, error starting mongo queue processor', {
                    method: 'MongoQueueProcessor.start',
                    error: err.message,
                });
```

ghost (approved):
the changes as such lgtm, but I am not sure if we have other places to handle, I'll let other reviewers confirm - I'm trusting here the green build

I just see in the `tests/functional/replication/streamedCopy.spec.js` test file some previous mock still there

extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -336,7 +334,6 @@ class ReplicationStatusProcessor {
             sourceHTTPAgent: this.sourceHTTPAgent,
             vaultclientCache: this.vaultclientCache,
             gcProducer: this._gcProducer,
-            mProducer: this._mProducer,
             statsClient: this._statsClient,
```
welansari, line 337: I think `_statsClient` is also part of the "Redis metrics". To be double checked, but i think it also needs to be removed.

welansari: Same in other files.

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -79,11 +76,6 @@ class IngestionPopulator {
         // shared producer across readers
         this._producer = null;
 
-        // metrics clients
-        this._mProducer = null;
-        this._mConsumer = null;
-        this._redis = null;
```
welansari, line 85: Should keep this line

francoisferrand (commented):
In [cloudserver's `reportHandler`](https://github.com/scality/cloudserver/blob/development/8.8/lib/utilities/reportHandler.js), there are queries on CRR metrics : will this not break if we remove these redis metrics?

PR #2596, opened 2024-11-21
Improvement/bb 632 bump mongo db driver 6 1

KillianG:


extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -80,11 +80,13 @@ class OplogPopulator {
...
     async _setupMongoClient() {
         try {
-            const client = await MongoClient.connect(this._mongoUrl, {
+            let client = new MongoClient(this._mongoUrl, {
                 replicaSet: this._replicaSet,
                 useNewUrlParser: true,
                 useUnifiedTopology: true,
+                readPreference: this._mongoConfig.readPreference,
             });
+            client = await client.connect();
```
francoisferrand, line 89: best to use a separate variable, these are not actually the same "type"

francoisferrand:
Replaced with #2599

PR #2601, opened 2024-12-13
BB-640 Add VersioningEnabled header to Backbeat client

nicolas2bert:


francoisferrand (approved):
LGTM (see https://github.com/scality/cloudserver/pull/5707#discussion_r1883731493 about naming though).

Zenko's `MultipleBackendTask` extends `ReplicateObject` however, so we must double-check (in waterflow branches) that this extra check is not unexpectedly added to zenko's multi-backend replication.

welansari (approved):

nicolas2bert:
@bert-e approve

PR #2603, opened 2024-12-18
Ensure lifecycle tasks wait for messages to be pushed

francoisferrand:
lifecycle task pushes new entries to bucket topic, but may commit before the entry is commited : which allows multiple lifeycle iterations to happen in parallel.

Issue: BB-641

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTaskV2.js
+++ b/extensions/lifecycle/tasks/LifecycleTaskV2.js
@@ -1,6 +1,7 @@
 'use strict'; // eslint-disable-line
 
 const async = require('async');
+const util = require('util');
 const { errors } = require('arsenal');
 
 const LifecycleTask = require('./LifecycleTask');
@@ -39,33 +40,33 @@ class LifecycleTaskV2 extends LifecycleTask {
      * @param {array} remainings - array of { prefix, listType, beforeDate }
      * @param {object} bucketData - bucket data
      * @param {Logger.newRequestLogger} log - logger object
+     * @param {function} done - callback(error)
      * @return {undefined}
      */
-    _handleRemainingListings(remainings, bucketData, log) {
-        if (remainings && remainings.length) {
-            remainings.forEach(l => {
-                const {
-                    prefix,
-                    listType,
-                    beforeDate,
-                    storageClass,
-                } = l;
+    _handleRemainingListings(remainings, bucketData, log, done) {
+        async.forEach(remainings || [], (l, cb) => {
+            const {
+                prefix,
+                listType,
+                beforeDate,
+                storageClass,
+            } = l;
 
-                const entry = Object.assign({}, bucketData, {
-                    contextInfo: { requestId: log.getSerializedUids() },
-                    details: { beforeDate, prefix, listType, storageClass },
-                });
-
-                this._sendBucketEntry(entry, err => {
-                    if (!err) {
-                        log.debug(
-                            'sent kafka entry for bucket consumption', {
-                                method: 'LifecycleTaskV2._getVersionList',
-                            });
-                    }
-                });
+            const entry = Object.assign({}, bucketData, {
+                contextInfo: { requestId: log.getSerializedUids() },
+                details: { beforeDate, prefix, listType, storageClass },
             });
-        }
+
+            this._sendBucketEntry(entry, err => {
+                if (!err) {
+                    log.debug(
+                        'sent kafka entry for bucket consumption', {
+                        method: 'LifecycleTaskV2._getVersionList',
+                    });
+                }
+                cb();
+            });
+        }, done);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


extensions/lifecycle/tasks/LifecycleTaskV2.js
```diff
@@ -236,8 +252,10 @@ class LifecycleTaskV2 extends LifecycleTask {
                         );
                     }
 
-                    return done();
-                });
+                    return resolve();
+                })));
+
+            return Promise.allSettled(promises).then(() => done(), done);
```
ghost: With this code, in case of error in `_compareRulesToList`, we are ignoring the returned error, as we would have it in the "then" with the `rejected` status. This looks like a behavior change we don't want? We ignore errors for `_sendBucketEntry` but for `_compareRulesToList`, we actually call the `reject` function...

`Promise.allSettled` will return an array with the status for each promise, including the ones rejected:

```
Promise.allSettled([
  Promise.resolve(33),
  new Promise((resolve) => setTimeout(() => resolve(66), 0)),
  99,
  Promise.reject(new Error("an error")),
]).then((values) => console.log(values), err => console.log('catch', err));

[
    {
        "status": "fulfilled",
        "value": 33
    },
    {
        "status": "fulfilled",
        "value": 66
    },
    {
        "status": "fulfilled",
        "value": 99
    },
    {
        "status": "rejected",
        "reason": {}
    }
]
```

(and no "catch" log).

And if we are aligned, it also means this is something we do not test, or returning the error in the callback has no impact...

francoisferrand: you are right (and I will update the code), but actually `_compareRulesToList` is not upposed to fail either : it calls `_retryEntry`, which ignores the errors. Errors on individual objects are retried a bit ; but eventually we just skip the object (or batch), relying on the periodic scan to eventually process these obejcts.

extensions/lifecycle/tasks/LifecycleTaskV2.js
```diff
@@ -255,7 +257,9 @@ class LifecycleTaskV2 extends LifecycleTask {
                     return resolve();
                 })));
 
-            return Promise.allSettled(promises).then(() => done(), done);
+            return Promise.allSettled(promises).then(results => done(
+                results.findLast(r => r.status === 'rejected')[0]?.reason
```
ghost: Shouldn't it be

```suggestion
                results.findLast(r => r.status === 'rejected')?.reason
```

as it an array of dimension 1, and findLast only returns the first matching element?

francoisferrand: yes, switch from filter to findLast on the last minute...

ghost (approved):

welansari (approved):

nicolas2bert (changes requested):
Waiting for the Kafka message delivery report callback may not be ideal for at least two reasons:

1. The message is already considered as locally consumed even before it reached the queue processor queue.
The lifecycle conductor relies on `KafkaBacklogMetrics` to calculate the lag. 
`KafkaBacklogMetrics` uses `consumer.position()` (for consumers) and `queryWatermarkOffsets()` (for producers/clients) to fetch offset info, then writes those offsets into Zookeeper.
`consumer.position()` in node-rdkafka returns the consumer’s current read position for each assigned partition, i.e. the local offset that the consumer is about to read next (offset of the last consumed message + 1). It does not return the “committed offset” that Kafka itself stores for the consumer group.  A message is considered “locally consumed” as soon as the consumer (via librdkafka under the hood) has handed that message to `BackbeatConsumer`. Basically, right after consumer.consume(...) is called and the message is shown in your callback or event handler, the offset for that message is advanced in the local consumer state.

2. It will slow down our process since `sendToTopic` callback is called when Kafka returns the delivery report for this message. Delivery reports are invoked asynchronously after the producer has handed off messages to Kafka. Ifwe decide to block or wait synchronously for that delivery report every time we send a message, it will impact performance and throughput.

francoisferrand:
> The message is already considered as locally consumed even before it reached the queue processor queue

That is a fair point (and may actually help on another issue), but I don't really see how this is a problem _for this change_: handling an entry by the bucket processor typically takes at least one second already (scanning & checking the state of every object), so we face this discrepancy anyway...

This change is simply about ensuring that the we keep the "slot" until the entry is "fully" processed, instead of leaving many things pending: which can be an issue esp. since we are listing pushing continuation messages.

What am I missing here?

> Ifwe decide to block or wait synchronously for that delivery report every time we send a message, it will impact performance and throughput.

In theory yes;
Practically however, since we are processing up to 1000 entries at a time, I wonder if this makes a real impact: most of the reports would be received in the time we process each entry... (except for very small buckets, in which case throughput may not be so important)

It is certainly a trade off, but consistent processing seems important as well: or do you think it is completely safe to leave all these messages dangling, and already start processing next message(s)?

nicolas2bert:
> What am I missing here?

My understanding was that the goal of this PR is to prevent multiple lifecycle iterations (triggered by Conductor) from running in parallel. I just pointed out that the lag is based on the “locally consumed” offset rather than on a processed or stored offset. So even if we wait for an entry to be fully processed, it won't stop the bucket-lifecycle topic lag from being zero while there are still other bucket messages in the pipeline.

> most of the reports would be received in the time we process each entry

Regarding the internal lifecycle listing, it does not necessarily return a 1000 objects; it only includes those that meet the specified criteria (prefix, age, etc...) from the next 10,000 entries. We might even end up with a listing response containing only a few objects, or none at all. 
NOTE: This 10,000 entry limit helps avoid placing excessive load on Metadata by preventing the evaluation of an unbounded number of entries.

PR #2606, opened 2024-12-27
Support lifecycle transitions on OOB bucket

francoisferrand:
Core idea is to consider that OOB does not sync the whole bucket, but only the objects which are indeed in the OOB location : there may be objects in Zenko which are not in Metadata, as long as they are not marked as present in that location.

This requires just a few changes:
* Delete operations ignore objects whose dataStoreLocation is not the OOB location itself. This fixes most of the issues, and allows transitioning objects from S3C to cold stoage (they get removed from S3C completely at the end of the transition).
* When object is restored, cloudserver will now add an extra user metadata (`x-amz-meta-scal-version-id`): this is used to ensure we do not duplicate the object when it is restored.
* Finally, need to tweak the way _updates_ (not actual ingestions) from S3C are processed, to ensure we keep the extra Zenko fields (esp. transition or cold storage fields), while updating only what is expected: i.e. only tags.

Issue: BB-590

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -227,16 +227,9 @@ class MongoQueueProcessor {
             params.versionId = versionId;
         }
 
+        log.debug('getting zenko object metadata', { bucket, key, versionId, params });
+
         return this._mongoClient.getObject(bucket, key, params, log, (err, data) => {
-            if (err && err.NoSuchKey) {
-                // TODO: this may happen if the object was created from artesca side (e.g. on restore)
-                //       --> in that case, the versionId does not match, and we cannot find the object
-                //       --> yet we have the 'ring' versionId in the (zenko) entry's `location`
-                //           field, so we may try to look it up (get all versions, check the one with
-                //           where location is indeed in the RING and with the 'target' versionId)
-                // ....or we introduce a way to store the RING object with the target metadata
-                return done();
-            }
             if (err) {
                 log.error('error getting zenko object metadata', {
                     method: 'MongoQueueProcessor._getZenkoObjectMetadata',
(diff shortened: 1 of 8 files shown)


welansari (approved):

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -392,43 +393,95 @@ class MongoQueueProcessor {
...
+        async.waterfall([
+            cb => this._getZenkoObjectMetadata(log, sourceEntry, versionId, cb),
+            (zenkoObjMd, cb) => {
+                // Skip if the object is in a different location, i.e. when the delete was caused
+                // by restored-object expiration or transition. It works because the dataStoreName
+                // is updated before actually sending the object to GC to effectively delete the
+                // data.
+                const encode = versionId => (versionId ? VersionID.encode(versionId) : 'null');
+                if (zenkoObjMd.dataStoreName !== location ||
+                    zenkoObjMd.location?.length !== 1 ||
```
nicolas2bert, line 423: > zenkoObjMd.location?.length !== 1

Why this check?
How about MPU with all its data parts stored at the same location?

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -392,43 +393,95 @@ class MongoQueueProcessor {
...
-                    return done(err);
+                    return done();
+                }
+
+                return cb(null, zenkoObjMd);
+            },
+            (zenkoObjMd, cb) => {
+                const options = {};
+
+                // Calling deleteObject with undefined options to use deleteObjectNoVer which is used for
```
nicolas2bert: `options` never seems to be `undefined`.
Has the behavior changed?

extensions/mongoProcessor/MongoQueueProcessor.js
francoisferrand: we want to skip the delete from OOB if the object is not in OOB : so this is mainly the `dataStoreName !== location` tests.

the extra conditions are sanity check, to ensure we "follow" OOB if the object is not in the expected state: an OOB object always has a single "part" (i.e. the s3c object!) which is also in the same location.

> How about MPU with all its data parts stored at the same location?

OOB locations are s3 locations, which is "mpu aware" : so the data is stored as a single multi-part object in the backend (i.e. in s3c), and the metadata only references a single object.

extensions/mongoProcessor/MongoQueueProcessor.js
francoisferrand: code simplification, it behaves the same in [arsenal](https://github.com/scality/Arsenal/blob/e245687f3db9d444efff85b0f599267b21ada067/lib/storage/metadata/mongoclient/MongoClientInterface.ts#L2067):

```
        const _params = Object.assign({}, params);
```

(at least for Mongo, but OOB only work with mongo: do you think it would have a different behavior with Metadata ?)

nicolas2bert: then we need to update or remove the comment.

nicolas2bert (approved):

PR #2610, opened 2025-01-10
Deploy Metadata in Backbeat tests

welansari:
- Deploy a metadata setup in a separate CI job
- Add a minimal test for queue populator that uses the metadata backend

Issue: [BB-496](https://scality.atlassian.net/browse/BB-496)

[BB-496]: https://scality.atlassian.net/browse/BB-496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/dockerfiles/e2e/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+    build:
+      context: ../sproxyd
+    ports:
+      - 8181
+    user: 0:0
+    privileged: true
+  metadata:
+    profiles: ['s3c']
+    image: ${METADATA_IMAGE}
+    network_mode: host
```
francoisferrand: do we really need `host` mode? it strongly limits the situations where this can be used...

normally (without host mode)
* containers can connect to each other "freedly", simply using the container name and port opened in the container (i.e. cloudserver container can connect to `sproxyd:8181`)
* if the port is needed externally (e.g. from test code), the container should have a `ports` entry to define what is exposed : to make it available on the host's 8181 port, only from localhost, the line would be `127.0.0.1:8181:8181`. When specifying only the port, it will be mapped to a random port on the host (which can be retrieved using either docker or docker-compose commands)

.github/dockerfiles/e2e/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
+services:
+  sproxyd:
+    profiles: ['s3c']
+    build:
+      context: ../sproxyd
+    ports:
+      - 8181
```
francoisferrand: is it used "externally" ? if sproxyd is only used from the other containers, this is not needed

.github/dockerfiles/e2e/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+    user: 0:0
+    privileged: true
```
francoisferrand: why is this needed?
should generally be avoided...

.github/dockerfiles/e2e/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
+services:
+  sproxyd:
+    profiles: ['s3c']
```
francoisferrand: profile is set on every container → not needed?

.github/dockerfiles/e2e/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+    depends_on:
+      - sproxyd
+    environment:
+      SPROXYD: sproxyd:8181
+    volumes:
+        - ./metadata-config.json:/mnt/standalone_workdir/config.json:ro
+  cloudserver:
+    profiles: ['s3c']
+    image: ${CLOUDSERVER_IMAGE}
+    network_mode: host
```
francoisferrand: same: should not use `host`, but expose an actual port

.github/dockerfiles/e2e/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+    depends_on:
+      - metadata
+    environment:
+      S3DATA: file
+      S3METADATA: scality
+      S3VAULT: mem
+      BUCKETD_BOOTSTRAP: localhost:9000
```
francoisferrand: `metadata:9000` ?

.github/dockerfiles/e2e/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+      S3_CONFIG_FILE: /conf/config.json
+      REMOTE_MANAGEMENT_DISABLE: 1
+    command: yarn start
+    volumes:
+      - ./cloudserver-config.json:/conf/config.json:ro
+  redis:
+    profiles: ['s3c']
+    image: ${REDIS_IMAGE}
+    network_mode: host
```
francoisferrand: not needed if we simply inject the redis endpoint in cloudserver config (redis:<whatever port>)

.github/dockerfiles/e2e/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+  kafka:
+    profiles: ['s3c']
+    image: ${KAFKA_IMAGE}
+    network_mode: host
```
francoisferrand: should just expose port 9092 instead:
```suggestion
    ports:
      - 127.0.0.1:9092:9092
```

.github/dockerfiles/sproxyd/conf/nginx.conf
francoisferrand, line 1: maybe not for now, but I wonder if we really need thiks proxy : could we simply run sproxyd, and connect directly to it? this would simplify config greatly (no nginx, no supervisor...)

.github/workflows/docker-build.yaml
```diff
@@ -27,6 +27,13 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ github.token }}
 
+      - name: Install Oras
```
francoisferrand: already fixed → rebase

.github/workflows/release.yaml
```diff
@@ -19,14 +19,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      # TODO: remove the following step once Oras CLI 0.13.0 bug https://github.com/oras-project/oras/issues/447 is fixed.
-      - name: Downgrade Oras to 0.12.0
+      - name: Install Oras
```
francoisferrand: already fixed → rebase

.github/workflows/tests.yaml
```diff
@@ -203,3 +203,72 @@ jobs:
...
+      - name: Setup services
+        env:
+          KAFKA_IMAGE: ghcr.io/${{ github.repository }}/ci-kafka:${{ github.sha }}
+          REDIS_IMAGE: redis:alpine
+          METADATA_IMAGE: ghcr.io/scality/metadata:7.70.45-standalone
+          CLOUDSERVER_IMAGE: ghcr.io/scality/cloudserver:${{ matrix.cloudserver_tag }}
+        run: |-
+          set -e -o pipefail;
+          docker compose --profile ${{ matrix.profile }} up -d --quiet-pull
+          bash ../../scripts/wait_for_local_port.bash 8000 120
```
francoisferrand: i think there is now a way to add "health check" in docker compose, which would allow checking that services are "up" (i.e. answers on port) and thus wait on it with a docker (compose) command...

package.json
```diff
@@ -19,6 +19,8 @@
     "garbage_collector": "node extensions/gc/service.js",
     "test": "mocha --recursive tests/unit --timeout 30000",
     "cover:test": "nyc --clean --silent yarn run test && nyc report --report-dir ./coverage/test --reporter=lcov",
+    "e2e": "mocha --recursive $(find tests/e2e -name '*.js') --timeout 30000",
+    "cover:e2e": "nyc --clean --silent yarn run e2e && nyc report --report-dir ./coverage/e2e --reporter=lcov",
```
francoisferrand: to avoid duplication, we may generate a [generic `cover` script](https://github.com/scality/cloudserver/blob/development/8.8/package.json), something like: 

```suggestion
    "e2e": "mocha --recursive $(find tests/e2e -name '*.js') --timeout 30000",
    "cover": "nyc --clean --silent yarn run",
    "postcover": "nyc report --report-dir ./coverage/test --reporter=lcov",
```

which can be run with `yarn run cover test` and `yarn run cover e2e`.
(but we would need to find a way to get the step name in the middle of the `postcover` command, so we can put the coverage in separate folder)

tests/e2e/queuePopulator/config/s3c-config.json
```diff
@@ -0,0 +1,171 @@
+{
+  "zookeeper": {
+    "connectionString": "127.0.0.1:2181/backbeat",
```
francoisferrand: this is used from within a container, or from an app running "natively" the host ?
- if it is, then the connectionString should be `zookeeper:2181`
- if it runs on the host, then indeed may be something like this... (ideally the port should be dynamically allocated, but that can be changed later)

tests/e2e/queuePopulator/config/s3c-config.json
```diff
@@ -0,0 +1,171 @@
...
+    "autoCreateNamespace": true
+  },
+  "kafka": {
+    "hosts": "127.0.0.1:9092",
```
francoisferrand: same

tests/e2e/queuePopulator/config/s3c-config.json
```diff
@@ -0,0 +1,171 @@
...
+  },
+  "s3": {
+    "host": "127.0.0.1",
+    "port": 8000
+  },
+  "extensions": {
+    "replication": {
+      "source": {
+        "s3": {
+          "host": "127.0.0.1",
```
francoisferrand: same

tests/e2e/queuePopulator/config/s3c-config.json
```diff
@@ -0,0 +1,171 @@
...
+    "probeServer": {
+      "bindAddress": "127.0.0.1",
+      "port": 7900
+    }
+  },
+  "metrics": {
+    "topic": "backbeat-metrics"
+  },
+  "s3": {
+    "host": "127.0.0.1",
```
francoisferrand: same

tests/e2e/queuePopulator/config/s3c-config.json
```diff
@@ -0,0 +1,171 @@
...
+      ]
+    },
+    "host": "localhost",
+    "port": 8900
+  },
+  "certFilePaths": {},
+  "redis": {
+      "name": "scality-s3",
+      "password": "",
+      "host": "127.0.0.1",
```
francoisferrand: same

tests/e2e/queuePopulator/config/s3c-config.json
welansari: This config is used in the tests and not inside a container (we initialize the queuePopulator in the tests with this).

.github/dockerfiles/e2e/docker-compose.yaml
welansari: I added it for future work when running the same tests with a Zenko configuration, in which containers like sproxyd and metadata won't be started.

.github/dockerfiles/e2e/docker-compose.yaml
welansari: Not exactly sure why it is needed, but sproxyd doesn't start properly without it.

.github/dockerfiles/cloudserver/Dockerfile
```diff
@@ -0,0 +1,7 @@
+ARG CLOUDSERVER_IMAGE=ghcr.io/scality/cloudserver:7.70.47
+
+FROM ${CLOUDSERVER_IMAGE}
+
+ADD ./config.json /conf/config.json
+
+RUN apt update && apt install -y curl
```
francoisferrand: should also cleanup packages after install (delete apt cache etc)

or we could use `node` directly to run healthcheck:
```
test: [ "node", "-e", 'fetch(`http://localhost:9000/health`).then(()=>process.exitCode = 0).catch(()=>process.exitCode = 1)']
```

.github/dockerfiles/e2e/docker-compose.yaml
francoisferrand: oh, maybe it needs extra permissions to expose status via FUSE (sfused does this, so i am thinking it may have follow the same pattern...)

→ you can try to pass the `--no-sysfs` option, so there is no such need

.github/workflows/tests.yaml
```diff
@@ -130,7 +130,9 @@ jobs:
       run: yarn run --silent lint
 
     - name: Run unit tests
-      run: yarn run cover:test
+      run: |
+        yarn run cover test
+        yarn run postcover ./coverage/test
```
francoisferrand: should be `yarn run postcover test` for consistency

tests/e2e/queuePopulator/queuePopulator.spec.js
francoisferrand, line 1: should these not be just another "functional" test? (folder & `ft_test:queuepopulator`yarn target)

* it is not really "end-2-end", since we still deploy (some) mocks, and do not use the same deployment as in the product
* it is similar to the funcitonal tests, in the sense that we run an "real" backbeat service, and the tests are running in separate process
* it does not replace (or try to) the actual e2e tests we have in Zenko

francoisferrand (approved):

benzekrimaha (approved):

PR #2611, opened 2025-01-15
Improvement/bb 585 bump deps

benzekrimaha:
Issue: BB-585


.github/workflows/docker-build.yaml
```diff
@@ -26,6 +26,14 @@ jobs:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ github.token }}
+      
+      - name: Install Oras
```
francoisferrand: fix already merged

.github/workflows/tests.yaml
```diff
@@ -196,7 +196,7 @@ jobs:
         token: ${{ secrets.CODECOV_TOKEN }}
 
     - name: run ballooning tests for lifecycle conductor
-      run: yarn mocha tests/performance/lifecycle/conductor-check-memory-balloon.js
+      run: yarn mocha tests/performance/lifecycle/conductor-check-memory-balloon.js --exit
```
francoisferrand, line 199: the `--exit` flag is not recomended by mocha :

> TL;DR: If your tests hang after an upgrade to Mocha v4.0.0 or newer, use --exit for a quick (though not necessarily recommended) fix.
>
> Prior to version v4.0.0, by default, Mocha would force its own process to exit once it was finished executing all tests. This behavior enables a set of potential problems; it’s indicative of tests (or fixtures, harnesses, code under test, etc.) which don’t clean up after themselves properly. Ultimately, “dirty” tests can (but not always) lead to false positive or false negative results.
>
> “Hanging” most often manifests itself if a server is still listening on a port, or a socket is still open, etc. It can also be something like a runaway setInterval(), or even an errant Promise that never fulfilled.
>
> The default behavior in v4.0.0 (and newer) is --no-exit, where previously it was --exit.
>
> The easiest way to “fix” the issue is to pass --exit to the Mocha process. It can be time-consuming to debug — because it’s not always obvious where the problem is — but it is recommended to do so.

- why do we need to add it?
- if we don't know why (and don't have the time to debug or prefer to merge this 'large' PR first to avoid more conflicts/...), then we must create a ticket already so we do a better fix ASAP.

README.md
```diff
@@ -1,6 +1,7 @@
 # Zenko Backbeat
 
 ![backbeat logo](res/backbeat-logo.png)
+[![Build Status][badgepriv]][badgepub]
```
francoisferrand: these badge are not correct, the URLs (at the end of this file) should be updated

package.json
```diff
@@ -17,17 +17,17 @@
     "oplog_populator": "node extensions/oplogPopulator/OplogPopulatorTask.js",
     "mongo_queue_processor": "node extensions/mongoProcessor/mongoProcessorTask.js",
     "garbage_collector": "node extensions/gc/service.js",
-    "test": "mocha --recursive tests/unit --timeout 30000",
+    "test": "mocha --recursive tests/unit --timeout 30000 --exit",
```
francoisferrand, line 20: same, it is recommneded not to use the flag, and fix the test(s) that need it...

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -163,8 +163,8 @@ describe('lifecycle task helper methods', () => {
             // Normally, when dealing with multiple objects, we wouldn't be able
             // to sort by just VersionId, since they could be intertwined.
             const expected = [...versions, ...dms].sort((a, b) => {
-                if (a.VersionId < b.VersionId) return -1;
-                if (a.VersionId > b.VersionId) return 1;
+                if (a.VersionId < b.VersionId) {return -1;}
+                if (a.VersionId > b.VersionId) {return 1;}
                 return 0;
```
francoisferrand:
```suggestion
                if (a.VersionId < b.VersionId) {
                    return -1;
                }
                if (a.VersionId > b.VersionId) {
                    return 1;
                }
                return 0;
```

tests/unit/lib/config/Config.spec.js
```diff
@@ -1,4 +1,4 @@
-'use strict'; // eslint-disable-line
+'use strict';
```
francoisferrand: extra spaces at end of line (in many files actually)

```suggestion
'use strict';
```

tests/unit/lib/util/LocationStatusManager.spec.js
```diff
@@ -1,4 +1,4 @@
-'use strict'; // eslint-disable-line
+'use strict';
```
francoisferrand: extra spaces at end of line

.github/workflows/release.yaml
```diff
@@ -19,14 +19,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      # TODO: remove the following step once Oras CLI 0.13.0 bug https://github.com/oras-project/oras/issues/447 is fixed.
-      - name: Downgrade Oras to 0.12.0
+      - name: Install Oras
```
francoisferrand: fix already merged

.github/workflows/docker-build.yaml
benzekrimaha: I am rebasing indeed

.github/workflows/tests.yaml
benzekrimaha: When not added , actually all tests just keep hanging even though they are done , as this may take time to dig into I created the following ticket : https://scality.atlassian.net/browse/BB-650, I actually think that some handles might still be open leading to the tests to keep on hanging...

package.json
benzekrimaha: https://github.com/scality/backbeat/pull/2611#discussion_r1946595405

README.md
```diff
@@ -1,6 +1,7 @@
 # Zenko Backbeat
 
 ![backbeat logo](res/backbeat-logo.png)
+![Build Status](https://github.com/scality/backbeat/actions/workflows/release.yaml/badge.svg)
```
francoisferrand, line 4: should this not be the badge for the "test/build" workflow?

tests/unit/RoleCredentials.js
```diff
@@ -206,6 +206,7 @@ describe('Credentials Manager', () => {
         });
     });
 
+
```
francoisferrand: nit: extra empty line

tests/unit/lifecycle/LifecycleConductor.spec.js
```diff
@@ -242,19 +242,19 @@ describe('Lifecycle Conductor', () => {
     });
 
     describe('_indexesGetOrCreate', () => {
-        it('should return v1 for bucketd bucket sources', () => {
+        it('should return v2 for bucketd bucket sources', () => {
```
francoisferrand: should not have changed, we should have the same behavior as before the node22 bump?

tests/unit/utils/LocationStatusStream.spec.js
```diff
@@ -30,21 +30,12 @@ describe('LocationStatusStream', () => {
                 collection: collectionStub,
                 command: mongoCommandStub,
             });
-            const mongoConnectStub = sinon.stub(MongoClient, 'connect').resolves({
+            sinon.stub(MongoClient.prototype, 'connect').resolves({
                 db: dbStub,
             });
             const lss = new LocationStatusStream('lifecycle', mongoConfig, null, null, FakeLogger);
             lss._setupMongoClient(err => {
                 assert.ifError(err);
```
francoisferrand, line 38: we should still check that it was called with the expected url, i.e
* at least `assert(mongoConnectStub.calledOnce)`
* to partially check params, with `calledOnceWith` is not practical, we may simply assert on (a subset of) the params of the call afterwarsd, something like `assert(mongoConnectStub.call(0).args[1] == mongoUrl)`

eslint.config.mjs
```diff
@@ -37,13 +53,15 @@
...
+    settings: {
+        'import/resolver': {
+            node: {
+                paths: ["/backbeat/node_modules", "node_modules"]
+            }
         }
-      }
-}
+    },
+}];
```
ghost:
```suggestion
}];

```

eslint.config.mjs
```diff
@@ -37,13 +53,15 @@
         "no-undef-init": "off",
         "newline-per-chained-call": "off",
         "no-useless-escape": "off",
+        "no-redeclare":"off",
```
ghost, line 56: Ideally we should not add any custom rule. if some rules do not make sense for our repos, better to update the guidelines repo. maybe worth a follow up ticket? (to remove all these custom rules)?

extensions/index.js
```diff
@@ -10,7 +10,7 @@ fs.readdirSync(__dirname).forEach(moduleDir => {
         try {
             indexStat = fs.statSync(
                 path.join(__dirname, moduleDir, 'index.js'));
-        } catch (err) {} // eslint-disable-line no-empty
+        } catch {} // eslint-disable-line no-empty
```
ghost: There is definitely a better way to implement this function and avoid empty catch:

```js
fs.readdirSync(__dirname).forEach(moduleDir => {
    const extPath = path.join(__dirname, moduleDir);
    const indexPath = path.join(extPath, 'index.js');
    
    if (fs.statSync(extPath).isDirectory() && 
        fs.existsSync(indexPath)) {
        extensions[moduleDir] = require(`./${moduleDir}`);
    }
});
```

lib/BackbeatConsumer.js
```diff
@@ -808,8 +808,8 @@ class BackbeatConsumer extends EventEmitter {
         const self = this;
         let lastBootstrapId;
         let producer; // eslint-disable-line prefer-const
-        let producerTimer; // eslint-disable-line prefer-const
-        let consumerTimer; // eslint-disable-line prefer-const
+        let producerTimer;  
+        let consumerTimer;
```
ghost:
```suggestion
        let producerTimer;
        let consumerTimer;
```

tests/utils/DummyLogger.js
```diff
@@ -42,7 +42,7 @@ class DummyRequestLogger {
         this.counts.fatal += 1;
     }
 
-    getSerializedUids() { // eslint-disable-line class-methods-use-this
+    getSerializedUids() {
```
ghost:
```suggestion
    getSerializedUids() {
```

eslint.config.mjs
welansari: [BB-653](https://scality.atlassian.net/browse/BB-653)

tests/unit/lifecycle/LifecycleConductor.spec.js
welansari: The initial test case was false, there was a bug in the code that was fixed during the unification (with the previous code, S3C never used the v2 listing) the tests were missed as they didn't fail after the change  as they never called done() after `_indexesGetOrCreate` finished executing.

Bumping the "assert" package made the tests fail, not sure what the exact reason is, might be a performance increase or something else but now the assert function returns before the test ends.

README.md
welansari: Not sure it's useful as the `development/x.y` branches dont't run that workflow (only release is run on them), so it would be getting the status from the last run on some dev branch.

tests/unit/lifecycle/LifecycleConductor.spec.js
francoisferrand: OK, but can we double-check with hopocalipse that we have an issue in the tests (i.e. bucketd must indeed use V2), or is there an issue in the lib (i.e. we should use V1 for bucketd, but we use V2) : maybe s3c should not use v2 listing, and it was done on purpose... aka activating v2 listing in s3c should not be a silent improvement thanks to the unification, but a conscious choice of which everybody should be aware.

welansari: In the 7.x branch, listing v1 is only used when forcing it using the `forceLegacyListing` flag in the config. We kept the same behaviour in 9.x

extensions/index.js
```diff
@@ -4,16 +4,12 @@ const path = require('path');
...
-            indexStat = fs.statSync(
-                path.join(__dirname, moduleDir, 'index.js'));
-        } catch (err) {} // eslint-disable-line no-empty
-        if (indexStat) {
-            extensions[moduleDir] = require(`./${moduleDir}`);
-        }
+    const extPath = path.join(__dirname, moduleDir);
+    const indexPath = path.join(extPath, 'index.js');
+    
+    if (fs.statSync(extPath).isDirectory() &&
```
francoisferrand: ~~this may still throw, if the file (or directory!) does not exist: either pass `throwIfNoEntry` option, or wrap with try/catch~~

do we actually need the `statSync` ? can't we just call `existsSync()`?

francoisferrand (approved):

extensions/index.js
welansari: Calling `existsSync` directly works fine

ghost (approved):

PR #2612, opened 2025-01-16
Fix queue populator shared metrics

welansari:
- Queue Populator can be used for a single or multiple extensions at once. Currently the 'origin' label in the metrics is always set to "replication" which is not always true,  we change this to make it hold the list of loaded extensions in the current queue populator instance.
- Some queue populator metrics are shared between all extensions and not specific to replication. We change their prefix to `s3_backbeat_populator` instead of `s3_replication`

Issue: BB-527

### DIFF
--- a/lib/queuePopulator/BucketFileLogReader.js
+++ b/lib/queuePopulator/BucketFileLogReader.js
@@ -6,13 +6,14 @@ const LogReader = require('./LogReader');
 class BucketFileLogReader extends LogReader {
     constructor(params) {
         const { zkClient, kafkaConfig, dmdConfig, logger,
-            extensions, metricsProducer, metricsHandler } = params;
+            extensions, extensionNames, metricsProducer, metricsHandler } = params;
         super({ zkClient, kafkaConfig, logConsumer: null,
             logId: `bucketFile_${dmdConfig.logName}`, logger, extensions,
             metricsProducer, metricsHandler });
 
         this._dmdConfig = dmdConfig;
         this._log = logger;
+        this._extensionNames = extensionNames;
         this._log.info('initializing bucketfile log reader', {
             method: 'BucketFileLogReader.constructor',
             dmdConfig,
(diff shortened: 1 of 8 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -27,6 +27,13 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ github.token }}
 
+      - name: Install Oras
```
francoisferrand: not needed, please rebase...

lib/queuePopulator/BucketFileLogReader.js
```diff
@@ -45,6 +46,7 @@ class BucketFileLogReader extends LogReader {
 
     getMetricLabels() {
         return {
+            origin: this._extensionNames,
```
francoisferrand, line 49: should we not "join" the strings in the array? (here or maybe better in the constructor)

```this._extensionNames.join(',')```

**Review by francoisferrand — APPROVED** [review]
it would be nice to document somewhere what "origin" really is : i.e. not the pod/process which generated the metric, but the kind of workflow/process from which the 'processing' was triggered... (i.e. so that we have replication count either from "replication" rules and "transition")

File: lib/queuePopulator/BucketFileLogReader.js (hunk 2 above)
Comment (welansari) [reply lib/queuePopulator/BucketFileLogReader.js:49]: `_extensionNames ` is a string containing that. The `join` is done when setting up the log source in the queue populator

File: lib/queuePopulator/QueuePopulator.js [hunk 3]
```diff
@@ -23,6 +23,10 @@ const {
     startCircuitBreakerMetricsExport,
 } = require('../CircuitBreaker');
 
+// origin is the name of the workflow that triggered the processing
+// of the log entries. The label can include multiple values when
+// the queue populator is shared between multiple workflows (e.g.
+// replication, lifecycle, notification).
```
francoisferrand: nit:

```suggestion
// `replication,lifecycle,notification`).
```

BourgoisMickael (approved):

PR #2618, opened 2025-02-26
adapt mongo call to new version of lib

welansari:
Issue: BB-657

### DIFF
--- a/lib/util/LocationStatusManager.js
+++ b/lib/util/LocationStatusManager.js
@@ -249,7 +249,9 @@ class LocationStatusManager {
                     value: this._locationStatusStore[location].getValue(),
                 }
             },
-            { upsert: false }, err => {
+            { upsert: false })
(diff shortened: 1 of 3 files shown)


benzekrimaha (approved):
LGTM

lib/util/LocationStatusManager.js
```diff
@@ -249,7 +249,9 @@ class LocationStatusManager {
                     value: this._locationStatusStore[location].getValue(),
                 }
             },
-            { upsert: false }, err => {
+            { upsert: false })
+            .then(() => cb())
+            .catch(err => {
```
ghost, line 254: Can we test that? It should be something we detect at backbeat level

ghost (approved):
I also reviewed all calls to mongodb driver and couldn't find any other call with callbacks instead of promises

PR #2622, opened 2025-03-05
Support AssumeRole in CRR

welansari:
- Support per site replication destination config to allow having a mix of CRR and Zenko Replication locations.
- Remove vault admin API call that get account details, this task will be delegated to the destination Cloudserver that will get the account details and place them into the objectMD.
- Support AssumeRole auth type in CRR.

Issue: BB-647

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -15,7 +15,7 @@ const CRR_FAILURE_EXPIRY = 24 * 60 * 60; // Expire Redis keys after 24 hours.
 const OBJECT_SIZE_METRICS = [66560, 8388608, 68157440];
 
 const destinationAuthJoi = joi.object({
-    type: joi.alternatives().try('account', 'role', 'service')
+    type: joi.alternatives().try('account', 'role', 'service', 'assumeRole')
         .required(),
     account: joi.string()
         .when('type', { is: 'account', then: joi.required() }),
@@ -25,6 +25,8 @@ const destinationAuthJoi = joi.object({
         adminPort: joi.number().greater(0).optional(),
         adminCredentialsFile: joi.string().optional(),
     }).when('type', { is: 'role', then: joi.required() }),
+    sts: hostPortJoi
+        .when('type', { is: 'assumeRole', then: joi.required() }),
 });
 
 const joiSchema = joi.object({
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -14,6 +14,19 @@ const qpRetryJoi = joi.object({
 const CRR_FAILURE_EXPIRY = 24 * 60 * 60; // Expire Redis keys after 24 hours.
 const OBJECT_SIZE_METRICS = [66560, 8388608, 68157440];
 
+const destinationAuthJoi = joi.object({
+    type: joi.alternatives().try('account', 'role', 'service')
+        .required(),
+    account: joi.string()
+        .when('type', { is: 'account', then: joi.required() }),
+    vault: joi.object({
+        host: joi.string().optional(),
+        port: joi.number().greater(0).optional(),
+        adminPort: joi.number().greater(0).optional(),
+        adminCredentialsFile: joi.string().optional(),
+    }).when('type', { is: 'role', then: joi.required() }),
+});
+
 const joiSchema = joi.object({
     source: {
         transport: transportJoi,
@@ -36,22 +49,22 @@ const joiSchema = joi.object({
             }).when('type', { is: 'role', then: joi.required() }),
         }).required(),
     },
-    destination: {
+    destination: joi.object({
         transport: transportJoi,
-        auth: joi.object({
-            type: joi.alternatives().try('account', 'role', 'service')
-                .required(),
-            account: joi.string()
-                .when('type', { is: 'account', then: joi.required() }),
-            vault: joi.object({
-                host: joi.string().optional(),
-                port: joi.number().greater(0).optional(),
-                adminPort: joi.number().greater(0).optional(),
-                adminCredentialsFile: joi.string().optional(),
-            }),
-        }).required(),
+        auth: destinationAuthJoi.optional(),
+        sites: joi.object().pattern(
+            joi.string(), // site name
+            joi.object({
+                transport: transportJoi,
+                auth: destinationAuthJoi,
+            }).required()
+        ).when('auth', {
+            is: joi.exist(),
+            then: joi.optional(),
+            otherwise: joi.required(),
+        }),
         bootstrapList: bootstrapListJoi,
-    },
+    }).required().custom(_validatePerSiteDestinationConfig),
     topic: joi.string().required(),
     dataMoverTopic: joi.string().optional(),
     replicationStatusTopic: joi.string().required(),
@@ -90,6 +103,33 @@ const joiSchema = joi.object({
     objectSizeMetrics: joi.array().items(joi.number()).default(OBJECT_SIZE_METRICS),
 });
 
+/**
+ * When using per site configuration, validate that there is an auth
+ * and transport configuration for each site in the bootstrap list.
+ * @param {Object} authConfig - auth config
+ * @param {joi.CustomHelpers} helpers  - joi helpers
+ * @returns {Object|joi.ErrorReport} - auth config when validated or a joi error
+ */
+function _validatePerSiteDestinationConfig(destination, helpers) {
+    // destination.auth and destination.transport is used as a default for all sites
+    // no need to check destination.transport as it has a default value and is always set.
+    if (destination.auth) {
+        return destination;
+    }
+    const missingConfigs = [];
+    destination.bootstrapList.forEach(b => {
+        if (!destination.sites?.[b.site]?.auth) {
+            missingConfigs.push(b.site);
+        }
+    });
+    if (missingConfigs.length > 0) {
+        return helpers.message({
+            custom: `missing destination configuration for sites: ${missingConfigs.join(',')}`
+        });
+    }
+    return destination;
+}
+
 function _loadAdminCredentialsFromFile(filePath) {
     const adminCredsJSON = fs.readFileSync(filePath);
     const adminCredsObj = JSON.parse(adminCredsJSON);
@@ -103,7 +143,6 @@ function _loadAdminCredentialsFromFile(filePath) {
 function configValidator(backbeatConfig, extConfig) {
     const validatedConfig = joi.attempt(extConfig, joiSchema);
     const { source, destination } = validatedConfig;
-
     if (source.auth.vault) {
         const { adminCredentialsFile } = source.auth.vault;
         if (adminCredentialsFile) {
@@ -111,12 +150,24 @@ function configValidator(backbeatConfig, extConfig) {
                 _loadAdminCredentialsFromFile(adminCredentialsFile);
         }
     }
-    if (destination.auth.vault) {
+    if (destination.auth?.vault) {
         const { adminCredentialsFile } = destination.auth.vault;
         if (adminCredentialsFile) {
             destination.auth.vault.adminCredentials =
                 _loadAdminCredentialsFromFile(adminCredentialsFile);
         }
+    }
+    if (destination.sites) {
+        Object.values(destination.sites).forEach(site => {
+            if (site.auth?.vault) {
+                const { adminCredentialsFile } = site.auth.vault;
+                if (adminCredentialsFile) {
+                    // eslint-disable-next-line no-param-reassign
+                    site.auth.vault.adminCredentials =
+                        _loadAdminCredentialsFromFile(adminCredentialsFile);
+                }
+            }
+        });
     }
     return validatedConfig;
 }
--- a/extensions/replication/queueProcessor/task.js
+++ b/extensions/replication/queueProcessor/task.js
@@ -169,23 +169,31 @@ function initAndStart(zkClient) {
         log.info('management init done');
 
         const bootstrapList = config.getBootstrapList();
+        const siteNames = bootstrapList.map(i => i.site);
 
-        const destConfig = Object.assign({}, repConfig.destination);
-        destConfig.bootstrapList = bootstrapList;
+        // initialize per site destination configs
+        const destConfig = {};
+        siteNames.forEach(site => {
+            destConfig[site] = config.getReplicationSiteDestConfig(site);
+        });
 
         config.on('bootstrap-list-update', () => {
-            destConfig.bootstrapList = config.getBootstrapList();
-
+            const updatedBootstrapList = config.getBootstrapList();
             const activeSites = Object.keys(activeQProcessors);
-            const updatedSites = destConfig.bootstrapList.map(i => i.site);
+            const updatedSites = updatedBootstrapList.map(i => i.site);
             const allSites = [...new Set(activeSites.concat(updatedSites))];
 
             async.each(allSites, (site, next) => {
                 if (updatedSites.includes(site)) {
+                    if (!destConfig[site]) {
+                        destConfig[site] = config.getReplicationSiteDestConfig(site);
+                    } else {
+                        destConfig[site].bootstrapList = updatedBootstrapList;
+                    }
                     if (!activeSites.includes(site)) {
                         const qp = new QueueProcessor(
                             topic, zkConfig, zkClient, kafkaConfig,
-                            sourceConfig, destConfig,
+                            sourceConfig, destConfig[site],
                             repConfig, redisConfig, mConfig,
                             httpsConfig, internalHttpsConfig,
                             site, repConfig.queueProcessor.circuitBreaker);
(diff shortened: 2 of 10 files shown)


extensions/replication/ReplicationConfigValidator.js
```diff
@@ -37,19 +58,27 @@ const joiSchema = joi.object({
...
-                adminCredentialsFile: joi.string().optional(),
-            }),
-        }).required(),
+        transport: joi.alternatives(
+            transportJoi,
+            joi.array()
+                .items(transportObjJoi)
+                .min(joi.ref('bootstrapList.length')),
+        ),
+        auth: joi.alternatives(
```
francoisferrand: Instead of an array, shouldn't we have a map (indexed by the location name) ?
- it would make it easier to access each entry (just indexed access)
- it would avoid the need for each types

The only benefit I see of having an array of config with the extra `site` field is that it would be possible to have "overrides" : i.e. a default config (without `site`), and overrides (which may or may not be specified).

A few extra things:
- Instead of modifying `transport`, we could keep it (as 'default') and add an extra `transports` field (containing the array/map). This may be slightly cleaner and easier to read, support a "default" if that makes sense, and eventually allow removing `transport` (or we could just remove transport altogether, not sure if we need the retro-compatibility?)
- Depending on the semantics, we should either check that there is an entry for each and every site (e.g. unless we implement some sort of default, we MUST have a value for each site) ; or at least verify each entry corresponds to a site in the bootstrap list

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -212,7 +212,15 @@ class QueueProcessor extends EventEmitter {
         this.zkClient = zkClient;
         this.kafkaConfig = kafkaConfig;
         this.sourceConfig = sourceConfig;
-        this.destConfig = destConfig;
+        this.destConfig = {
```
francoisferrand: `destConfig` is a param provided to this QueueProcessor : so it may be the responsability of the caller to "build" the appropriate destConfig for the site.
this also allows decoupling the QueueProcessor implementation from the format of the Config itself.

francoisferrand: this would probably mean introducing some sort of function in the Config `getDestConfig(site)` to handle the mapping.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -99,6 +106,13 @@ class ReplicateObject extends BackbeatTask {
         }
         this._setupDestClients(this.targetRole, log);
 
+        // Destination Vault admin API is not accessible when
+        // using assumeRole i.e when targeting an ARTESCA
```
francoisferrand:
```suggestion
        // using assumeRole i.e when targeting Zenko
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -696,11 +711,38 @@ class ReplicateObject extends BackbeatTask {
     }
 
     _setupDestClients(targetRole, log) {
+        this.destBackbeatHost = this.destHosts.pickHost();
+
+        if (this.destConfig.auth.type === authTypeAssumeRole) {
+            const accountId = _extractAccountIdFromRole(targetRole);
+            const roleName = _extractRoleNameFromRole(targetRole);
+            this.clientManager = new ClientManager({
```
francoisferrand, line 731: should the clientManager not be created just once, in `QueueProcessor` ?

`_setupDestClients` is called multiple times in the sequence of `ReplicateObject` task, so this will repeatedly create `ClientManager`s, instead of just (re)creating (and possibly reusing) clients...

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -696,11 +711,38 @@ class ReplicateObject extends BackbeatTask {
...
+            this.backbeatDest = this.clientManager.getBackbeatClient(accountId);
+            return;
+        }
+
         this.s3destCredentials =
             this._createCredentials('target', this.destConfig.auth,
                 targetRole, log);
 
-        this.destBackbeatHost = this.destHosts.pickHost();
         this.backbeatDest = new BackbeatClient({
```
francoisferrand, line 758: pushing a bit further: shoud we (can we?) use clientManager to create every `BackbeatClient` instance?
Not sure it is possible for the "dest" client, but should be possible for the "source" as well.

- This may be a followup "cleanup" ticket
- As far as `AssumeRole`, your changes addresses the "dest" site only (needed since we cannot use old Vault API anymore) ; but should not as well modernize the "source" side, and use roles to access the data (like in other extensions)? (same, could be a separate improvement)

extensions/replication/tasks/ReplicateObject.js
welansari: There are two sides to this :
- `ClientManager` is made in a way that doesn't support having clients with different credentials, it only initializes the STS credentials once at init. I created this ticket BB-663 for improving that which will allow sharing a single client between multiple `ReplicateObject` tasks for the same destination.
- The multiple calls to `_setupDestClients` are expected to recreate a client each time, those multiple calls are done in a failure case where we switch the host and retry again.

extensions/replication/tasks/ReplicateObject.js
welansari: - Looks like it is possible to use `ClientManager` to initialize source clients, created BB-669.
- Created BB-668 for supporting role authentication on the source side.

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -38,18 +53,18 @@ const joiSchema = joi.object({
...
+            transportJoi
+        ).default({}),
+        auth: joi.alternatives()
+            .try(
+                destinationAuthJoi,
+                joi.object().pattern(
+                    joi.string(), // site name
+                    destinationAuthJoi
+                ).custom(_validatePerSiteAuthConfig),
+            ).error(_overwriteAuthErrorMessages).required(),
```
francoisferrand: seems what we really want is to always have "pairs" transport+auth, i.e. eiher `{transport, auth}` or `map(site_name -> {transport, auth})` : maybe introduce a `destinations` instead/next to destination (but `bootstrapList` list in there...)? or a new `sites` field in destination ?

something like this:
```
destination: {
    transport: transportJoi,  // deprecated?
    auth: destinationAuthJoi, // deprecated?
    sites: joi.object().pattern(
        joi.string(), // site name
        {
            transport: transportJoi,
            auth: destinationAuthJoi,
        }
    ).default({}),
}
```

francoisferrand: btw what is this `destination.bootstrapList` field?
- isn't the `bootstrapList` stored at the "root" of the config?
- why do we need a separate copy of the list of locations in destination?

extensions/replication/queueProcessor/task.js
```diff
@@ -169,23 +169,30 @@ function initAndStart(zkClient) {
...
 
         const bootstrapList = config.getBootstrapList();
+        const siteNames = bootstrapList.map(i => i.site);
 
-        const destConfig = Object.assign({}, repConfig.destination);
-        destConfig.bootstrapList = bootstrapList;
+        // initialize per site destination configs
+        const destConfig = {};
+        siteNames.forEach(site => {
+            destConfig[site] = config.getReplicationSiteDestConfig(site);
```
francoisferrand, line 177: should set the bootstrap list as well, like we do line 191:
```
destConfig[site].bootstrapList = bootstrapList;
```

(or remove it altogether, maybe no point duplicating this global list...)

extensions/replication/ReplicationConfigValidator.js
welansari: I'm not really sure why we keep this list and pass it everywhere (i also see it getting used in ingestion and lifecycle). The locations config is supposed to have a list of the details of all locations, for me that should be sufficient for all our needs. I made [BB-672](https://scality.atlassian.net/browse/BB-672) to investigate its use.

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -36,22 +51,22 @@ const joiSchema = joi.object({
...
-                adminPort: joi.number().greater(0).optional(),
-                adminCredentialsFile: joi.string().optional(),
-            }),
-        }).required(),
+        auth: destinationAuthJoi.optional(),
+        sites: joi.object().pattern(
+            joi.string(), // site name
+            joi.object({
+                transport: transportJoi,
+                auth: destinationAuthJoi.required(),
```
francoisferrand: - since it is required, `destination.auth` will never be used as default
- does it make sense to specify a destination with the same transport but a dedicated auth?

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -36,22 +51,22 @@ const joiSchema = joi.object({
...
-            vault: joi.object({
-                host: joi.string().optional(),
-                port: joi.number().greater(0).optional(),
-                adminPort: joi.number().greater(0).optional(),
-                adminCredentialsFile: joi.string().optional(),
-            }),
-        }).required(),
+        auth: destinationAuthJoi.optional(),
+        sites: joi.object().pattern(
+            joi.string(), // site name
```
francoisferrand, line 58: maybe we want to validate this is in the bootstrap list?

```suggestion
            joi.string().valid(Joi.in('bootstrapList')), // site name
```

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -36,22 +51,22 @@ const joiSchema = joi.object({
...
-                .when('type', { is: 'account', then: joi.required() }),
-            vault: joi.object({
-                host: joi.string().optional(),
-                port: joi.number().greater(0).optional(),
-                adminPort: joi.number().greater(0).optional(),
-                adminCredentialsFile: joi.string().optional(),
-            }),
-        }).required(),
+        auth: destinationAuthJoi.optional(),
+        sites: joi.object().pattern(
```
francoisferrand, line 57: * there is an `authJoi` type in configItems.joi.js, which holds both transport and auth: can't we use it (or extend it) instead of introducting a new destinationAuthJoi type?

* once we have a type with transport+auth, maybe `destination` can extend this type to add bootstrap? (but maybe this creates lots of override, to make things optional/alternative)

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -36,22 +51,22 @@ const joiSchema = joi.object({
             }).when('type', { is: 'role', then: joi.required() }),
         }).required(),
     },
-    destination: {
+    destination: joi.object({
         transport: transportJoi,
```
francoisferrand, line 55: is transport optional or required?

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -103,21 +145,32 @@ function _loadAdminCredentialsFromFile(filePath) {
...
+    if (destination.sites) {
+        Object.values(destination.sites).forEach(site => {
+            if (site.auth.vault) {
+                const { adminCredentialsFile } = site.auth.vault;
+                if (adminCredentialsFile) {
+                    // eslint-disable-next-line no-param-reassign
+                    site.auth.vault.adminCredentials =
+                        _loadAdminCredentialsFromFile(adminCredentialsFile);
+                }
+            }
```
francoisferrand, line 171: set the default vault already, so that "config" handling is done only here (and does not trickle in the "business" code) : i.e. from here on, auth is always specified in each destination' site?

```suggestion
            } else {
                site.auth.vault = destination.auth.vault;
            }
```

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -36,22 +51,22 @@ const joiSchema = joi.object({
...
-            }),
-        }).required(),
+        auth: destinationAuthJoi.optional(),
+        sites: joi.object().pattern(
+            joi.string(), // site name
+            joi.object({
+                transport: transportJoi,
+                auth: destinationAuthJoi.required(),
+            }).required()
+        ).when('auth', {
```
francoisferrand, line 63: should this condition on be on each destination's `auth` field ?
(i.e. the site's auth must be specified if there is no "top level" default)

extensions/replication/queueProcessor/task.js
```diff
@@ -169,23 +169,31 @@ function initAndStart(zkClient) {
...
+            destConfig[site].bootstrapList = bootstrapList;
```
francoisferrand: not needed, already done in `getReplicationSiteDestConfig()`

extensions/replication/queueProcessor/task.js
```diff
@@ -169,23 +169,31 @@ function initAndStart(zkClient) {
...
-            const updatedSites = destConfig.bootstrapList.map(i => i.site);
+            const updatedSites = updatedBootstrapList.map(i => i.site);
             const allSites = [...new Set(activeSites.concat(updatedSites))];
 
             async.each(allSites, (site, next) => {
                 if (updatedSites.includes(site)) {
+                    if (!destConfig[site]) {
+                        destConfig[site] = config.getReplicationSiteDestConfig(site);
+                    }
+                    destConfig[site].bootstrapList = updatedBootstrapList;
```
francoisferrand: boostrapList already set in `getReplicationSiteDestConfig()`, so either it is an else:

```suggestion
                if (!destConfig[site]) {
                    destConfig[site] = config.getReplicationSiteDestConfig(site);
                }
                else {
                    destConfig[site].bootstrapList = updatedBootstrapList;
                }
```

or maybe we should always update the site's configuration:

```suggestion
                destConfig[site] = config.getReplicationSiteDestConfig(site);
```

tests/unit/conf/Config.js
```diff
@@ -89,3 +90,59 @@ describe('Site name', () => {
...
+            process.env.BACKBEAT_CONFIG_FILE = `${__dirname}/configs/replicationMultiDestConfig.json`;
+            const conf = new Config();
+            const destConfig = conf.getReplicationSiteDestConfig('aws3');
+            assert.deepStrictEqual(destConfig, {
+                transport: 'https',
+                auth: {
+                    type: 'service',
+                    account: 'service-replication-3',
+                },
+                bootstrapList: [
```
francoisferrand, line 118: it may/could happen that `BACKBEAT_CONFIG_FILE` is set before the tests begin (and some test may depend on it...) : the really safe way to handle this is to store the previous value of BACKBEAT_CONFIG_FILE before the test (beforeEach), so that eventually we can either delete the env variable if it was not set or reset it to its original value otherwise

tests/unit/conf/Config.js
```diff
@@ -89,3 +90,59 @@ describe('Site name', () => {
...
+            sinon.stub(conf.extensions.replication, 'destination').value({
+                transport: 'https',
+                auth: {
+                    type: 'service',
+                    account: 'service-replication',
+                },
+                bootstrapList: [
+                    { site: 'aws1', type: 'aws_s3' },
+                    { site: 'aws2', type: 'aws_s3' },
+                    { site: 'aws3', type: 'aws_s3' }
```
francoisferrand, line 138: no test for the `auth.vault` handling (_loadAdminCredentialsFromFile)

francoisferrand (approved):

extensions/replication/ReplicationConfigValidator.js
welansari: It's optional, it has a default value.

extensions/replication/ReplicationConfigValidator.js
welansari: This is already validated in `_validatePerSiteDestinationConfig`

extensions/replication/ReplicationConfigValidator.js
welansari: This won't work as the bootstrapList contains objects so we need a more complex custom validator, and i think `_validatePerSiteDestinationConfig` already does the job of validating what we need, no need to have an extra one just for this, sites that are not in the bootstrapList will just be ignored.

extensions/replication/ReplicationConfigValidator.js
welansari: I made it optional, that way we could specify a different transport if needed. The validation for when the global auth is not specified will be done in `_validatePerSiteDestinationConfig`

extensions/replication/ReplicationConfigValidator.js
welansari: Better do this in another ticket as it touches lots of other components that need to be properly tested, i created [BB-674](https://scality.atlassian.net/browse/BB-674)

extensions/replication/ReplicationConfigValidator.js
welansari: There is no guarantee we have all the sites here, so we would still have to check later. We could create and append the config of each site if it doesn't exist to the structure but that would just be redundant information, especially since we already go through the config class to get the destination site's config so it's already abstracted from the business logic.

extensions/replication/queueProcessor/task.js
welansari: We shouldn't update the whole site's configuration every time as that will make the pre-existent QueueProcessors loose reference to it as it's a new object. I added an else case.

ghost (approved):

PR #2624, opened 2025-03-07
BB-664 use one release process

nicolas2bert:


.github/workflows/release-federation.yaml
BourgoisMickael, line 1: I'm not sure the action will use the file from the hotfix branch, it might just pick the release workflow from the default branch

dvasilas: Yes I think that's right. You might have to merge to the default branch and then launch it with the hotfix branch.

BourgoisMickael: Apparently with `gh` CLI you can run a workflow that's in another branch if it doesn't show up on the UI but was already ran before.

I did 

```
gh workflow run release-federation.yaml -f tag=micktest -r hotfix/7.70.15
```

Now the workflow came back in the UI with all the history https://github.com/scality/backbeat/actions/workflows/release-federation.yaml

nicolas2bert: It looks like it is using the action will use the file from the hotfix branch and not the development/9.0
https://github.com/scality/backbeat/actions/runs/13703780339/job/38365886056

francoisferrand: The workflow from the target branch will be used, as expected: nothing funny there.

But the default branch is the only one considered to "list" workflows that can be triggered from the UI.

.github/workflows/release.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-name: release
+name: release-federation
```
francoisferrand: release.yaml → name should be `release` (for consistency and discoverabiltiy)

.github/workflows/release.yaml
francoisferrand, line 1: this will waterflow into the 8 branches, and can easily cause stuff to be removed from there...
I would prefer to either:
- keep a single release here, but in the `release-federation.yaml` file to avoid "confusing" waterflow
- or move into a single `release.yaml`, backported (or similar) to what we have on latest branches, and which builds both federation and "simple" image

PR #2626, opened 2025-03-11
BackbeatConsumer : ensure ordered processing of messages with the same key

welansari:
Instead of processing all messages concurrently, we now queue messages with the same key and process them one after the other in order. Messages with no key are processed concurrently as before.

The `taskScheduler` was re-used for this task as it allows queuing and sequential processing of tasks based on a key. This class was enhanced to support limiting the concurrency of tasks without a key + helper function to allow flow control in `BackbeatConsumer`.

Issue: BB-645

### DIFF
--- a/extensions/gc/GarbageCollector.js
+++ b/extensions/gc/GarbageCollector.js
@@ -138,6 +138,10 @@ class GarbageCollector extends EventEmitter {
             topic: this._gcConfig.topic,
             groupId: this._gcConfig.consumer.groupId,
             concurrency: this._gcConfig.consumer.concurrency,
+            // respecting the order of messages for the same object
+            // is not required in the garbage collector (messages
+            // sent to the gc don't have a key anyways)
+            orderByFunc: null,
             queueProcessor: this.processKafkaEntry.bind(this),
         });
         this._consumer.on('error', () => {
--- a/extensions/lifecycle/LifecycleConfigValidator.js
+++ b/extensions/lifecycle/LifecycleConfigValidator.js
@@ -8,7 +8,7 @@ const {
     retryParamsJoi,
 } = require('../../lib/config/configItems.joi');
 
-const { supportedLifecycleRules }  = require('../../lib/constants');
+const { supportedLifecycleRules, backbeatConsumer: { MAX_QUEUED_DEFAULT } } = require('../../lib/constants');
 
 const joiSchema = joi.object({
     zookeeperPath: joi.string().required(),
@@ -59,6 +59,7 @@ const joiSchema = joi.object({
         groupId: joi.string().required(),
         retry: retryParamsJoi,
         concurrency: joi.number().greater(0).default(10),
+        maxQueued: joi.number().greater(0).default(MAX_QUEUED_DEFAULT),
         probeServer: probeServerJoi.default(),
         circuitBreaker: joi.object().optional(),
     },
@@ -67,6 +68,7 @@ const joiSchema = joi.object({
         groupId: joi.string().required(),
         retry: retryParamsJoi,
         concurrency: joi.number().greater(0).default(10),
+        maxQueued: joi.number().greater(0).default(MAX_QUEUED_DEFAULT),
         probeServer: probeServerJoi.default(),
         circuitBreaker: joi.object().optional(),
     },
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -9,14 +9,17 @@ const { BreakerState, CircuitBreaker } = require('breakbeat').CircuitBreaker;
 
 const BackbeatProducer = require('./BackbeatProducer');
 const OffsetLedger = require('./OffsetLedger');
+const TaskScheduler = require('./tasks/TaskScheduler');
 const KafkaBacklogMetrics = require('./KafkaBacklogMetrics');
 const {
     startCircuitBreakerMetricsExport,
 } = require('./CircuitBreaker');
 const { observeKafkaStats } = require('./util/probe');
+const {
+    CONCURRENCY_DEFAULT,
+    MAX_QUEUED_DEFAULT,
+} = require('./constants').backbeatConsumer;
 
-// controls the number of messages to process in parallel
-const CONCURRENCY_DEFAULT = 1;
 const CLIENT_ID = 'BackbeatConsumer';
 const { withTopicPrefix } = require('./util/topic');
 
@@ -107,13 +110,33 @@ class BackbeatConsumer extends EventEmitter {
             circuitBreakerMetrics: joi.object({
                 type: joi.string().required(),
             }).optional(),
+            // Function to determine what to order the processing queue by.
+            // Default function orders by the key of the message, if the message
+            // has a key. Messages with the same key are processed in order.
+            // Messages with different keys are processed in parallel.
+            // Setting this to null will disable ordering and will process messages
+            // in parallel.
+            orderByFunc: joi.function()
+                .default(ctx => ctx?.entry?.key?.toString())
+                .allow(null),
+            // When ordering is enabled, one message is processed at a time
+            // for a given key, other messages with the same key are queued.
+            // This parameter controls the maximum number of messages that can
+            // be queued for processing (not per key but globally).
+            // Avoid having a very high value for this parameter as it will increase
+            // the risk of re-processing previously processed messages in case of
+            // a crash or rebalance, as the queued messages will block offset commit
+            // for more recent messages with different keys that might have been processed.
+            maxQueued: joi.number()
+                .greater(0)
+                .default(MAX_QUEUED_DEFAULT),
         });
         const validConfig = joi.attempt(config, configJoi,
                                         'invalid config params');
 
         const { clientId, zookeeper, kafka, topic, groupId, queueProcessor,
-                fromOffset, concurrency, fetchMaxBytes,
-                canary, bootstrap, circuitBreaker, circuitBreakerMetrics } = validConfig;
+                fromOffset, concurrency, maxQueued, fetchMaxBytes,
+                canary, bootstrap, circuitBreaker, circuitBreakerMetrics, orderByFunc } = validConfig;
 
         this._zookeeperEndpoint = zookeeper && zookeeper.connectionString;
         this._kafkaHosts = kafka.hosts;
@@ -128,12 +151,18 @@ class BackbeatConsumer extends EventEmitter {
         this._groupId = groupId;
         this._queueProcessor = queueProcessor;
         this._concurrency = concurrency;
+        this._maxQueued = maxQueued;
         this._fetchMaxBytes = fetchMaxBytes;
         this._canary = canary;
         this._bootstrap = bootstrap;
         this._offsetLedger = new OffsetLedger();
+        this._processingQueue = new TaskScheduler(
+            (ctx, done) => this._processTask(ctx.entry, done),
+            orderByFunc,
+            null,
+            this._concurrency,
+        );
 
-        this._processingQueue = null;
         this._messagesConsumed = 0;
         // this variable represents how many kafka messages have been
         // requested without having been received yet, i.e. still
@@ -334,9 +363,6 @@ class BackbeatConsumer extends EventEmitter {
             });
         }
 
-        this._processingQueue = async.queue(
-            this._queueProcessor, this._concurrency);
-
         this._consumer.on('event.error', error => {
             // This is a bit hacky: the "broker transport failure"
             // error occurs when the kafka broker reaps the idle
@@ -361,19 +387,22 @@ class BackbeatConsumer extends EventEmitter {
 
     /**
      * Get how many messages we may attempt to consume at this time,
-     * considering the maximum concurrency, current processing queue
-     * length and message retrievals still pending.
+     * considering the maximum concurrency, queued messages, and message
+     * retrievals still pending.
      *
      * @return {integer} a strictly positive number representing the
      * number of new messages that the pipeline is ready to process,
      * or zero if the processing pipeline is 100% busy
      */
     _getAvailableSlotsInPipeline() {
-        const nSlots = this._concurrency
+        // do not exceed max queued tasks limit
+        const nQueuedSlots = this._maxQueued
+                - this._processingQueue.length();
+        // do not exceed max concurrency limit
+        const nRunningSlots = this._concurrency
               - this._processingQueue.running()
-              - this._processingQueue.length()
               - this._nConsumePendingRequests;
-        return nSlots > 0 ? nSlots : 0;
+        return Math.max(0, Math.min(nQueuedSlots, nRunningSlots));
     }
 
     _tryConsume() {
@@ -421,8 +450,7 @@ class BackbeatConsumer extends EventEmitter {
                         entry.topic, entry.partition, entry.offset);
                     this._messagesConsumed++;
 
-                    const finishProcessingTask = this._startProcessingTask(entry);
-                    this._processingQueue.push(entry, (err, completionArgs) => {
+                    this._processingQueue.push({ entry }, (err, completionArgs, finishProcessingTask) => {
                         this._onEntryProcessingDone(err, entry, completionArgs);
                         finishProcessingTask(err);
                     });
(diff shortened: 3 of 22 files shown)


lib/tasks/TaskScheduler.js
```diff
@@ -22,18 +31,28 @@ class TaskScheduler {
...
         this._processingFunc = processingFunc;
         this._getQueueKeyFunc = getQueueKeyFunc;
         this._getDedupeKeyFunc = getDedupeKeyFunc;
+        this._drainFunc = defaultDrain;
         this._taskQueues = {};
         this._dedupeCache = {};
+        if (noQueueKeyConcurrency) {
+            this._noQueueKeyTasks = async.queue(this._processingFunc, noQueueKeyConcurrency);
+            this._noQueueKeyTasks.drain = this._tryDrain.bind(this);
+        }
```
francoisferrand: what is the point?
it should not be the responsability of the TaskScheduler to have multiple implementations :
- either it degrades seamlessly (but we go through the same code), e.g. if getDedupeKeyFunc() returns something unique or maxQueueSize == 1;
- or we really need to have multiple implementations, and we should have an abstraction and multiple implementations (and the proper one is created on BackbeatConsumer startup)

lib/tasks/TaskScheduler.js
```diff
@@ -84,8 +103,57 @@ class TaskScheduler {
...
         return process.nextTick(() => this._processingFunc(ctx, onTaskEnd));
     }
+
+    /**
+     * @returns {Number} The number of tasks waiting to be processed
+     */
+    length() {
+        const noQueueKeyTasksLength = this._noQueueKeyTasks ? this._noQueueKeyTasks.length() : 0;
+        const taskQueuesLength = Object.keys(this._taskQueues)
+                .reduce((acc, key) => acc + this._taskQueues[key].length(), 0);
```
francoisferrand: `length()` will be called many times, not sure it is worth doing the computation every time: better to just update the `length` whenever there is a push or a completion

lib/tasks/TaskScheduler.js
```diff
@@ -84,8 +103,57 @@ class TaskScheduler {
...
+        return noQueueKeyTasksLength + taskQueuesLength;
+    }
+
+    /**
+     * @returns {Number} The number of tasks currently being processed
+     */
+    running() {
+        const noQueueKeyTasksRunning = this._noQueueKeyTasks ? this._noQueueKeyTasks.running() : 0;
+        // queues are deleted upon drain so having them in the map means one task is running per queue (no concurrency)
+        const taskQueuesRunning = Object.keys(this._taskQueues).length;
```
francoisferrand: same

lib/tasks/TaskScheduler.js
```diff
@@ -84,8 +103,57 @@ class TaskScheduler {
                 return queue.push(ctx, onTaskEnd);
             }
         }
+        if (this._noQueueKeyTasks) {
```
francoisferrand: what is the point?
is this not eseentially the same as the following line?

lib/BackbeatConsumer.js
```diff
@@ -17,6 +18,16 @@ const { observeKafkaStats } = require('./util/probe');
...
+// controls the max number of messages to queue for processing.
+// messages with the same key are queued to ensure they are
+// processed in order. We should avoid having a high value for
+// this parameter as it may lead to:
+// - Memory exhaustion if the queue is not drained fast enough
+// - Higher risk of processing previously consumed messages in case
+// of a crash or rebalance, as the offsets are committed strictly in
+// order, and the queueing of messages blocks the next offsets for
+// the same partition from being commited.
+const MAX_QUEUED_DEFAULT = 1000;
```
francoisferrand: should this be a "total" value, or a per-queue limit (i.e. total max queued items would be `concurrency x maxQueueLength`) ?
(i don't know, really just wondering)

extensions/gc/GarbageCollector.js
```diff
@@ -138,6 +138,8 @@ class GarbageCollector extends EventEmitter {
             topic: this._gcConfig.topic,
             groupId: this._gcConfig.consumer.groupId,
             concurrency: this._gcConfig.consumer.concurrency,
+            maxQueued: this._gcConfig.consumer.maxQueued,
+            ordered: this._gcConfig.consumer.ordered,
```
francoisferrand: ordered is not a "config" parameter : it is an intrinsic parameter of the implementation of each backbeat consumer/extension

lib/BackbeatConsumer.js
```diff
@@ -107,13 +118,17 @@ class BackbeatConsumer extends EventEmitter {
             circuitBreakerMetrics: joi.object({
                 type: joi.string().required(),
             }).optional(),
+            // if true, ensures that the processing of messages with the same key
+            // is done in order, messages with no key are processed in parallel
+            ordered: joi.boolean().default(true),
```
francoisferrand: i wonder if this should be a boolean, or if we should have a callback: to allow different keying strategies where needed (null --> no order, or callback to get the key, etc...)

lib/BackbeatConsumer.js
```diff
@@ -361,19 +379,22 @@ class BackbeatConsumer extends EventEmitter {
...
-              - this._processingQueue.length()
+        // do not exceed max queued tasks limit
+        const nQueuedSlots = this._maxQueued
+                - this._taskScheduler.length();
+        // do not exceed max concurrency limit
+        const nRunningSlots = this._concurrency
+              - this._taskScheduler.running()
               - this._nConsumePendingRequests;
-        return nSlots > 0 ? nSlots : 0;
+        return Math.max(0, Math.min(nQueuedSlots, nRunningSlots));
```
francoisferrand: would this not allow exceeding the concurrency, not sure that is what we want?

```suggestion
        // do not exceed max queued tasks limit
        const nQueuedSlots = this._maxQueued
                - this._taskScheduler.length();
        // do not exceed max concurrency limit
        const nRunningSlots = this._concurrency
              - this._taskScheduler.running()
              - this._taskScheduler.length()
              - this._nConsumePendingRequests;
        return Math.max(0, Math.min(nQueuedSlots, nRunningSlots));
```

or is that with `async.queues()`, once we 'started' a task it could be in the queue but not yet running : and so we had to compute both running() and length() ?

lib/BackbeatConsumer.js
```diff
@@ -453,19 +476,36 @@ class BackbeatConsumer extends EventEmitter {
...
      */
-    _startProcessingTask(entry) {
+    _startProcessingTask(entry, queuedTime) {
         const { topic, partition, offset } = entry;
         const consumerGroup = this._groupId;
         const taskStartTime = Date.now();
 
+        const queuedDuration = taskStartTime - queuedTime;
+        KafkaBacklogMetrics.onTaskStarted(topic, partition, consumerGroup, queuedDuration);
+
```
francoisferrand: should be in `_processTask`

lib/KafkaBacklogMetrics.js
```diff
@@ -56,6 +56,25 @@ const taskProcessingTime = metrics.ZenkoMetrics.createHistogram({
     buckets: [0.01, 0.1, 1, 10, 50, 100, 200, 300],
 });
 
+const taskQueuedTime = metrics.ZenkoMetrics.createHistogram({
```
francoisferrand: is this metrics useful?
in 99% of cases, queues will be empty, and queue time will be 0... And in the other cases, the time will be equal to the duration of the processing.

lib/KafkaBacklogMetrics.js
```diff
@@ -56,6 +56,25 @@ const taskProcessingTime = metrics.ZenkoMetrics.createHistogram({
...
+    name: promMetricNames.taskQueuedTime,
+    help: 'Time a task waits in the queue',
+    labelNames: ['topic', 'partition', 'group'],
+    buckets: [0.01, 0.1, 1, 10, 50, 100, 200, 300],
+});
+
+const taskQueuedCountGauge = metrics.ZenkoMetrics.createGauge({
```
francoisferrand:
```suggestion
const queuedTasksCountGauge = metrics.ZenkoMetrics.createGauge({
```

lib/KafkaBacklogMetrics.js
```diff
@@ -56,6 +56,25 @@ const taskProcessingTime = metrics.ZenkoMetrics.createHistogram({
...
+    name: promMetricNames.taskQueuedCount,
+    help: 'Number of tasks in the queue',
+    labelNames: ['topic', 'partition', 'group'],
+});
+
+const taskRunningCountGauge = metrics.ZenkoMetrics.createGauge({
```
francoisferrand:
```suggestion
const runningTasksCountGauge = metrics.ZenkoMetrics.createGauge({
```

lib/KafkaBacklogMetrics.js
```diff
@@ -56,6 +56,25 @@ const taskProcessingTime = metrics.ZenkoMetrics.createHistogram({
...
+    name: promMetricNames.taskRunningCount,
+    help: 'Number of tasks running',
```
francoisferrand:
```suggestion
    help: 'Number of tasks currently running',
```

lib/KafkaBacklogMetrics.js
```diff
@@ -56,6 +56,25 @@ const taskProcessingTime = metrics.ZenkoMetrics.createHistogram({
...
+const taskQueuedTime = metrics.ZenkoMetrics.createHistogram({
+    name: promMetricNames.taskQueuedTime,
+    help: 'Time a task waits in the queue',
+    labelNames: ['topic', 'partition', 'group'],
+    buckets: [0.01, 0.1, 1, 10, 50, 100, 200, 300],
+});
+
+const taskQueuedCountGauge = metrics.ZenkoMetrics.createGauge({
+    name: promMetricNames.taskQueuedCount,
+    help: 'Number of tasks in the queue',
```
francoisferrand:
```suggestion
    help: 'Number of tasks currently in the queue',
```

lib/constants.js
```diff
@@ -8,6 +8,9 @@ const constants = {
             rebalanceTotal: 's3_backbeat_queue_rebalance_total',
             slowTasksCount: 's3_backbeat_queue_slowTasks_count',
```
francoisferrand: typo:
```suggestion
            slowTasksCount: 's3_backbeat_queue_slowtasks_count',
```

lib/constants.js
```diff
@@ -8,6 +8,9 @@ const constants = {
...
             taskProcessingTime: 's3_backbeat_queue_task_processing_time_seconds',
+            taskQueuedTime: 's3_backbeat_queue_task_queued_time_seconds',
+            taskQueuedCount: 's3_backbeat_queue_task_queued_count',
+            taskRunningCount: 's3_backbeat_queue_task_running_count',
         },
```
francoisferrand: for consistency with slow tasks:
```suggestion
            queuedTasksCount: 's3_backbeat_queue_queuedtasks_count',
            runningTasksCount: 's3_backbeat_queue_runningtasks_count',
        },
```
or
```suggestion
            queuedTasksCount: 's3_backbeat_queue_tasks_count',
            runningTasksCount: 's3_backbeat_queue_runningtasks_count',
        },
```

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -852,17 +852,17 @@ describe('BackbeatConsumer shutdown tests', () => {
         async.series([
             next => {
                 const interval = setInterval(() => {
-                    if (consumer._processingQueue.idle()) {
+                    if (consumer._taskScheduler.idle()) {
```
francoisferrand: would it make sense to keep the old name `_processingQueue` (which it still is, IMHO 😄) : so we avoid extra changes?

tests/unit/lib/tasks/TaskScheduler.spec.js
```diff
@@ -115,4 +116,94 @@ describe('TaskScheduler', () => {
             }
         }
     });
+
```
francoisferrand, line 124: can you please double-check this class is properly and well tested (including pre-existing code)? i.e. we cover every case and more (timing/sequence of calls...)

4 tests does not seem like a lot, esp. seeing all the branches in the code (and the subtelties that can arise)

lib/BackbeatConsumer.js
welansari: Including `length()` in the calculation of concurrency slots won't really work here, as we may have configured a much larger "queuing factor".

Even if the tasks that need to be run are first "queued" for a brief moment this should still work as we always take the smallest value between concurrent slots and queued slots, so the tasks that we end up pushing fit in both the queue and concurrency slots.

lib/tasks/TaskScheduler.js
welansari: I used the queue to ease the tracking of concurrent tasks without having to keep and update a state.

lib/BackbeatConsumer.js
welansari: I think having a per-queue limit would make the code more complex for little practical benefit in prod. Having that would mean tracking the length of each of the queues, blocking when one gets full or otherwise read what we can. Can be done, but not sure it's worth it.

tests/unit/lib/tasks/TaskScheduler.spec.js
welansari: The test i added goes through most of the paths possible but i added unit tests for each of the methods.

extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -51,6 +51,7 @@ const joiSchema = joi.object({
         // the processing, no need to add more here to avoid
         // overloading the system
         concurrency: joi.number().greater(0).default(1),
+        maxQueued: joi.number().greater(0).default(1000),
```
ghost: Should we make the default consistent?

```suggestion
        maxQueued: joi.number().greater(0).default(MAX_QUEUED_DEFAULT),
```

lib/BackbeatConsumer.js
ghost: On the value 1000, was it chosen arbitrary or do we have a reason not to have higher numbers? (not very high, so we don't risk re processing events in case of a crash)

lib/tasks/TaskScheduler.js
```diff
@@ -81,11 +101,53 @@ class TaskScheduler {
...
+        this._drainFunc = func;
+    }
+
+    /**
+     * Call the drain function if there are no tasks running
+     * @returns {undefined}
+     */
+    _tryDrain() {
+        if (this.idle()) {
+            this._drainFunc();
```
ghost: to avoid creating an empty function you can keep it as null if not defined and do

```suggestion
            this._drainFunc?.();
```

(or check that it's defined)?

lib/tasks/TaskScheduler.js
```diff
@@ -81,11 +101,53 @@ class TaskScheduler {
...
+    idle() {
+        return this._running === 0 && this._inQueue === 0;
+    }
+
+    /**
+     * Set a function to be called after all tasks have been processed
+     * @param {DrainFunction} func drain function
+     * @returns {undefined}
+     */
+    drain(func) {
```
ghost: should this be `setDrain`? drain alone makes me think this function is supposed to actually drain

extensions/gc/GarbageCollectorConfigValidator.js
```diff
@@ -13,6 +13,7 @@ const joiSchema = joi.object({
         groupId: joi.string().required(),
         retry: retryParamsJoi,
         concurrency: joi.number().greater(0).default(10),
+        maxQueued: joi.number().greater(0).default(1000),
```
francoisferrand: regarding the config, I think we should double check where (which extension) this feature should be used?

i.e. we probably don't need it everywhere, so best be explicit and add the `maxQueued` parameter only in the extensions where it makes sense (i.e. there is indeed a `key`), and set the `orderByFunc` to null otherwise.

francoisferrand (approved):

lib/BackbeatConsumer.js
welansari: I avoided setting a high value because of the reason you stated yes, i.e with a high value there would be a higher risk of processing previous messages in case of a restart or rebalance as the queued messages will block the offset to be committed (offset commits are strictly done in order)

In practice, this value should be high enough as it's unlikely that we would be replicating/transitioning/expiring a large number of versions of the same object at a time.

ghost: thanks!

ghost (approved):

extensions/gc/GarbageCollectorConfigValidator.js
welansari: Removed from The Bucket Processor as it's not needed, and the GC as well as the messages it receives don't have a key.

PR #2632, opened 2025-03-24
Support having multiple crr locations configured in the queueProcessor

welansari:
QueueProcessor currently sets the destination host to the host of the last crr site in the bootstrap list, so when having multiple crr locations configured this will not work.

Issue: BB-662

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -257,17 +257,12 @@ class QueueProcessor extends EventEmitter {
             this._setupRedis(redisConfig);
         }
 
-        // FIXME support multiple scality destination sites
-        if (Array.isArray(destConfig.bootstrapList)) {
-            destConfig.bootstrapList.forEach(dest => {
-                if (Array.isArray(dest.servers)) {
-                    this.destHosts =
-                        new RoundRobin(dest.servers, { defaultPort: 80 });
-                    if (dest.echo) {
-                        this._setupEcho();
-                    }
-                }
-            });
+        if (Array.isArray(this.destConfig.replicationEndpoint.servers)) {
+            this.destHosts =
+                new RoundRobin(this.destConfig.replicationEndpoint.servers, { defaultPort: 80 });
+            if (this.destConfig.replicationEndpoint.echo) {
+                this._setupEcho();
+            }
         }
 
         this.taskScheduler = new TaskScheduler(
@@ -889,10 +884,8 @@ class QueueProcessor extends EventEmitter {
                 sourceEntry.getReplicationStorageClass();
             const sites = getLocationsFromStorageClass(replicationStorageClass);
             if (sites.includes(this.site)) {
-                const replicationEndpoint = this.destConfig.bootstrapList
-                    .find(endpoint => endpoint.site === this.site);
-                if (replicationEndpoint &&
-                replicationBackends.includes(replicationEndpoint.type)) {
+                if (this.destConfig.replicationEndpoint &&
+                    replicationBackends.includes(this.destConfig.replicationEndpoint.type)) {
                     task = new MultipleBackendTask(this);
                 } else {
                     task = new ReplicateObject(this);
--- a/extensions/replication/queueProcessor/task.js
+++ b/extensions/replication/queueProcessor/task.js
@@ -185,10 +185,13 @@ function initAndStart(zkClient) {
 
             async.each(allSites, (site, next) => {
                 if (updatedSites.includes(site)) {
+                    const siteConfig = config.getReplicationSiteDestConfig(site);
                     if (!destConfig[site]) {
-                        destConfig[site] = config.getReplicationSiteDestConfig(site);
+                        destConfig[site] = siteConfig;
                     } else {
-                        destConfig[site].bootstrapList = updatedBootstrapList;
+                        // Only updating replicationEndpoint to keep maintaining reference
+                        // in the queue processor instance
+                        destConfig[site].replicationEndpoint = siteConfig.replicationEndpoint;
                     }
                     if (!activeSites.includes(site)) {
                         const qp = new QueueProcessor(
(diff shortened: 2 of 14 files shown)


ghost (approved):
Changes LGTM (codecov still reports that all lines were not covered, yet the comment says the opposite...)

welansari:
The comments point to a previous run of codecov which is weird, but i added tests explicitly for those cases so they should be covered.

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -257,17 +258,16 @@ class QueueProcessor extends EventEmitter {
...
-        // FIXME support multiple scality destination sites
         if (Array.isArray(destConfig.bootstrapList)) {
-            destConfig.bootstrapList.forEach(dest => {
-                if (Array.isArray(dest.servers)) {
-                    this.destHosts =
-                        new RoundRobin(dest.servers, { defaultPort: 80 });
-                    if (dest.echo) {
-                        this._setupEcho();
-                    }
+            const dest = destConfig.bootstrapList.find(d => d.site === this.site);
```
francoisferrand: for this I gather that the QueueProcessor only uses a single site from `dest.bootstrapList` : so should we not instead "select" the appropriate site's bootstrap list/endpoint(s) when building `destConfig`?

(since building every destConfig is done once, when loading the config, this would also allow to do it efficiently, without going searching linearly for every site)

welansari: done

extensions/replication/queueProcessor/task.js
```diff
@@ -187,8 +187,6 @@ function initAndStart(zkClient) {
                 if (updatedSites.includes(site)) {
                     if (!destConfig[site]) {
```
francoisferrand, line 189: is this condition still relevant : should we not update the destConfig in any case?

lib/Config.js
```diff
@@ -313,7 +313,7 @@ class Config extends EventEmitter {
         return {
             transport: destConfig.sites?.[site]?.transport || destConfig.transport,
             auth: destConfig.sites?.[site]?.auth || destConfig.auth,
-            bootstrapList: this.bootstrapList,
+            replicationEndpoint: this.bootstrapList.find(endpoint => endpoint.site === site),
```
francoisferrand, line 316: is there anything more useful in bootstrapList ?
* `servers` may be an array (or sometimes not? then that case should be handled)
* the `site` field is nto relvant once we build a specific sites destination endpoint

```suggestion
            endpoints: this.bootstrapList.find(endpoint => endpoint.site === site).servers,
```
or
```suggestion
            ...this.bootstrapList.find(endpoint => endpoint.site === site),
```

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -889,10 +884,8 @@ class QueueProcessor extends EventEmitter {
                 sourceEntry.getReplicationStorageClass();
             const sites = getLocationsFromStorageClass(replicationStorageClass);
             if (sites.includes(this.site)) {
-                const replicationEndpoint = this.destConfig.bootstrapList
-                    .find(endpoint => endpoint.site === this.site);
-                if (replicationEndpoint &&
-                replicationBackends.includes(replicationEndpoint.type)) {
+                if (this.destConfig.replicationEndpoint &&
+                replicationBackends.includes(this.destConfig.replicationEndpoint.type)) {
```
francoisferrand: bad indent

lib/Config.js
```diff
@@ -313,7 +313,7 @@ class Config extends EventEmitter {
...
         };
```
francoisferrand, line 317: probably not such a big deal today with limited number of locations, but could become one: we are looping through the whole list every time..... even though actually we are actually already looping over every site (looking at task.js)

```
        const bootstrapList = config.getBootstrapList();
        const siteNames = bootstrapList.map(i => i.site);

        // initialize per site destination configs
        const destConfig = {};
        siteNames.forEach(site => {
            destConfig[site] = config.getReplicationSiteDestConfig(site);
        });
```

probably not something we can change in the API, but when loading the config is could do some reprocessing to just linearly build just once both the list of sites and replicationEndpoints.

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -22,21 +22,14 @@ const MPU_GCP_MAX_PARTS = 1024;
 
 class CopyLocationTask extends BackbeatTask {
 
-    _getReplicationEndpointType() {
```
francoisferrand, line 25: nit: to simplify the change, we could have kept this function and (greatly) simplify its implementation


```suggestion
    _getReplicationEndpointType() {
        return this.destConfig.replicationEndpoint.type
    }
```

tests/functional/replication/pauseResumeState.js
```diff
@@ -25,12 +25,16 @@ const repConfig = config.extensions.replication;
...
-    auth: { type: 'skip', vault: 'skip' },
-    bootstrapList: repConfig.destination.bootstrapList,
-};
 const mConfig = config.metrics;
 
+function getDestSiteConfig(site) {
+    return {
+        auth: { type: 'skip', vault: 'skip' },
+        replicationEndpoint: repConfig.destination.bootstrapList
+            .find(endpoint => endpoint.site === site),
```
francoisferrand: probably a simpler way to rework this: `firstSite` and `secondSite` are already build by indexing into the bootstrapList, not sure there is a point to add some lookup code (and possily bug) in the tests

francoisferrand (approved):

lib/Config.js
welansari: There are other useful fields yes. For CRR: `servers`, `default`, `echo`. And for MultipleBackends: `type`

extensions/replication/queueProcessor/task.js
welansari: We should not update the whole config in the other case as the QueueProcessor instance will loose reference to the config, but yes we should update the `replicationEndpoint` field. Added that.

PR #2635, opened 2025-04-03
BB-676: Ignore bucket migration oplog operations

dvasilas:
In the context of handling bucket migration in SUR, we have introduced a new type of operations that appear in the Metadata oplog and are consumed by Backbeat.

The format the of new operation type is (using an example):
```
    {
      "db": "some-bucket",
      "method": 10,
      "entries": [
        {
          "type": "bucket_migration_stage",
          "value": "{\"currentStage\":\"finalize\",\"owner-id\":\"d6633a8b3cfa8ce37c1ed8530f2fbfaa4713f96c416bf3162499483e863757b2\",\"raftSessionType\":\"source\"}"
        }
      ],
      "timestamp": "2025-04-03T06:36:50.401Z"
    },
```

`method: 10` is a new method type, `APPEND_TO_LOG:` https://github.com/scality/MetaData/blob/f2f3bd87b4a47d4d0849fbf497dd55aedd8c052f/lib/protocol.json#L30 

To make sure that Backbeat ignores these operations, and any new operations added in the future (until we explicitly want to add support for  new operations), this change adds a check for operations that are known and supported by Backbeat.

### DIFF
--- a/lib/queuePopulator/LogReader.js
+++ b/lib/queuePopulator/LogReader.js
@@ -355,7 +355,26 @@ class LogReader {
         });
     }
 
+    isSupportedOperation(record) {
+        // Check if the given record’s method is supported by this LogReader.
+        // Currently, only methods with an method ID < 10 are supported.
+        // This is intentional so that any future methods (>= 10) that are
+        // not explicitly supported will be rejected by default.
+        // List of methods defined in Metadata/lib/protocol.json.
+        return !record.method || record.method < 10;
+    }
+
     _processLogEntry(batchState, record, entry, cb) {
+        if (!this.isSupportedOperation(record)) {
+            this.log.debug('skipping unsupported operation',
+                {
+                    method: 'LogReader._processLogEntry',
+                    record,
+                    entry,
+                });
+            return cb();
+        }
+
         function _transformKey(key) {
             if (!key) {
                 return undefined;
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "7.70.18",
+  "version": "7.70.19",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
--- a/tests/unit/lib/queuePopulator/LogReader.spec.js
+++ b/tests/unit/lib/queuePopulator/LogReader.spec.js
@@ -65,21 +65,22 @@ describe('LogReader', () => {
         processedKey: 'v1-version-key\u0000version-id',
     }].forEach(testCase => {
         it(`LogReader::_processLogEntry() should process entry with a ${testCase.desc}`, () => {
-            logReader._processLogEntry(null, { db: 'db' }, {
+            logReader._processLogEntry(null, { db: 'db', method: 8 }, {
                 type: 'put',
                 key: testCase.key,
                 value: '{}',
-            });
-            assert.deepStrictEqual(filteredEntry, {
-                type: 'put',
-                bucket: 'db',
-                key: testCase.processedKey,
-                value: '{}',
-                logReader,
-                overheadFields: {
-                    versionId: undefined,
-                    commitTimestamp: undefined,
-                },
+            }, () => {
+                assert.deepStrictEqual(filteredEntry, {
+                    type: 'put',
+                    bucket: 'db',
+                    key: testCase.processedKey,
+                    value: '{}',
+                    logReader,
+                    overheadFields: {
+                        versionId: undefined,
+                        commitTimestamp: undefined,
+                    },
+                });
             });
         });
     });
@@ -87,6 +88,7 @@ describe('LogReader', () => {
     it('should pass through overhead.versionId if available in record', () => {
         const record = {
             db: 'db',
+            method: 8,
         };
         const versionId = 'v1';
         const entry = {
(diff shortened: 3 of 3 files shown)


lib/queuePopulator/LogReader.js
```diff
@@ -444,6 +444,13 @@ class LogReader {
         if (!record || !record.entries) {
             return done();
         }
+        // Method 10 is APPEND_TO_LOG in Metadata.
+        // For now, operations with this method are used for communication between
+        // metadata-migration and SUR, and between SUR ingestances.
```
jonathan-gramain: "ingestance"? :slightly_smiling_face:

jonathan-gramain (approved):

lib/queuePopulator/LogReader.js
dvasilas: I was probably trying to write "ingest instances" :laughing:

dvasilas:
```suggestion
        // metadata-migration and SUR, and between SUR ingest instances.
```

lib/queuePopulator/LogReader.js
```diff
@@ -444,6 +444,13 @@ class LogReader {
         if (!record || !record.entries) {
             return done();
         }
+        // Method 10 is APPEND_TO_LOG in Metadata.
+        // For now, operations with this method are used for communication between
+        // metadata-migration and SUR, and between SUR ingest instances.
+        // Ignore them for now.
+        if (record.method && record.method === 10) {
```
rachedbenmustapha: Should we instead have a known-supported list of operation types, and ignore those we don't (yet) know about? This would shield the backbeat codebase and oplog clients from future changes unless we explicitly want to support them and add any necessary new code.

dvasilas: That does make sense, and it avoids the need to add code to ignore new operations in the future (and the risk to forget to).

It's not immediately clear to me how to define the list of supported operations (maybe using `op.type`), but I can try.

dvasilas: @rachedbenmustapha @jonathan-gramain following Rached's suggestion, I changed the filter to check for known/supported operations and filter out any operation that does not match.

I did an Integration run with the changes in this PR and it seems to be going well: https://github.com/scality/Integration/actions/runs/14264674778

lib/queuePopulator/LogReader.js
```diff
@@ -355,7 +355,30 @@ class LogReader {
...
 
+    _isSupportedOperation(record, entry) {
+        // List of methods defined in Metadata/lib/protocol.json.
+        if (record.method && record.method < 10) {
+            if (!entry.type ||
+                entry.type === 'put' ||
+                entry.type === 'del' ||
+                entry.type === 'delete') {
+                return true;
+            }
```
jonathan-gramain: I don't think we need this block, as it would be isofunctional without it for all methods < 10 (and we save a bit of processing time as well).

Ideally I'd say it's up to the extension to check what it should consider and what to ignore, but I am aligned that this workaround is the least risky and intrusive considering the existing code.

dvasilas: Removed the checks on `type`.

I also changed the condition to not filter out operations without a `method` (for merging in in `8` and `9` branches, not sure `method` exists in Artesca).

lib/queuePopulator/LogReader.js
```diff
@@ -355,7 +355,25 @@ class LogReader {
         });
     }
 
+    _isSupportedOperation(record) {
+        // List of methods defined in Metadata/lib/protocol.json.
+        if (!record.method || record.method < 10) {
+            return true;
+        }
+        return false;
```
BourgoisMickael:
```suggestion
        return !record.method || record.method < 10;
```

BourgoisMickael (approved):

dvasilas:
@reviewers could you also review the integration PRs?
- `development/8.6`: no change
- `development/9.0`: same changes as here + version bump
- `development/9.1`: same changes as here + version bump

lib/queuePopulator/LogReader.js
```diff
@@ -355,7 +355,22 @@ class LogReader {
         });
     }
 
+    _isSupportedOperation(record) {
+        // List of methods defined in Metadata/lib/protocol.json.
```
nicolas2bert, line 363: Can you specify that we checks if the given record’s method is supported by this LogReader.
And currently, only methods with an ID less than 10 are supported. 
This is intentional so that any future methods (>= 10) that we
do not explicitly support will be rejected by default.

lib/queuePopulator/LogReader.js
```diff
@@ -355,7 +355,22 @@ class LogReader {
         });
     }
 
+    _isSupportedOperation(record) {
```
nicolas2bert: NIT: I don't remember the standard. Since this method will likely be overwritten by the Zenko/Mongo log reader class, should we remove the private prefix: `_` ?

nicolas2bert (approved):

jonathan-gramain (approved):
I reviewed the integration PRs, LGTM, although I'm surprised that the version on 9.1 didn't get bumped to 9.1.x (same version than 9.0). For sure not related to this PR but wondering if we shouldn't fix it on the fly. Technically I believe they should be distinguished to avoid caching issues with yarn, in particular.

dvasilas:
> I reviewed the integration PRs, LGTM, although I'm surprised that the version on 9.1 didn't get bumped to 9.1.x (same version than 9.0). For sure not related to this PR but wondering if we shouldn't fix it on the fly. Technically I believe they should be distinguished to avoid caching issues with yarn, in particular.

I was also surprised by the version in `development/9.1`, I will bump to `9.1.0`.

dvasilas:
ping

PR #2642, opened 2025-04-10
Share S3 client between CRR tasks

welansari:
Share the same S3 clients (and their credentials) between multiple CRR tasks of the same site.

The `ClientsManager` class is an improved version of `ClientManager` that allows handling multiple clients with different auth configs. For now we still keep the `ClientManager` class to limit the scope of this PR, but code that uses it should be updated to use the newer class. This will be done in another PR.

Issue: BB-663

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -14,6 +14,7 @@ const RoundRobin = require('arsenal').network.RoundRobin;
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
 const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
 const VaultClientCache = require('../../../lib/clients/VaultClientCache');
+const ClientsManager = require('../../../lib/clients/ClientsManager');
 const QueueEntry = require('../../../lib/models/QueueEntry');
 const TaskScheduler = require('../../../lib/tasks/TaskScheduler');
 const { getTaskSchedulerQueueKey,
@@ -226,6 +227,7 @@ class QueueProcessor extends EventEmitter {
         this.echoMode = false;
         this.scheduledResume = null;
         this.circuitBreakerConfig = circuitBreakerConfig;
+        this.clientsManager = null;
 
         this.logger = new Logger(
             `Backbeat:Replication:QueueProcessor:${this.site}`);
@@ -253,6 +255,8 @@ class QueueProcessor extends EventEmitter {
         }
 
         this._setupVaultclientCache();
+        this._setupClientsManager();
+        
         if (redisConfig) {
             this._setupRedis(redisConfig);
         }
@@ -356,6 +360,14 @@ class QueueProcessor extends EventEmitter {
                     this.httpsConfig.ca);
             }
         }
+    }
+
+    _setupClientsManager() {
+        if (this.destConfig.auth.type !== libConstants.authTypeAssumeRole) {
+            return;
+        }
+        this.clientsManager = new ClientsManager(`crr-${this.site}`, this.logger);
+        this.clientsManager.initCredentialsManager();
     }
 
     _setupProducer(done) {
@@ -689,6 +701,7 @@ class QueueProcessor extends EventEmitter {
             destHTTPAgent: this.destHTTPAgent,
             vaultclientCache: this.vaultclientCache,
             accountCredsCache: this.accountCredsCache,
+            clientsManager: this.clientsManager,
             replicationStatusProducer: this.replicationStatusProducer,
             mProducer: this._mProducer,
             logger: this.logger,
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -5,7 +5,6 @@ const errors = require('arsenal').errors;
 const jsutil = require('arsenal').jsutil;
 const ObjectMDLocation = require('arsenal').models.ObjectMDLocation;
 
-const ClientManager = require('../../../lib/clients/ClientManager');
 const BackbeatClient = require('../../../lib/clients/BackbeatClient');
 const BackbeatMetadataProxy = require('../../../lib/BackbeatMetadataProxy');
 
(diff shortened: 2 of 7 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -358,6 +362,14 @@ class QueueProcessor extends EventEmitter {
         }
     }
 
+    _setupClientsManager() {
+        if (this.destConfig.auth.type !== libConstants.authTypeAssumeRole) {
```
ghost, line 366: Maybe a comment here explaining why it is limited to assume role type would be great

lib/clients/ClientsManager.js
```diff
@@ -0,0 +1,286 @@
+const CredentialsManager = require('../credentials/CredentialsManager');
+const BackbeatMetadataProxy = require('../BackbeatMetadataProxy');
+const { createBackbeatClient, createS3Client } = require('./utils');
+const { authTypeAssumeRole } = require('../constants');
+const { http: HttpAgent, https: HttpsAgent } = require('httpagent');
+
+const DELETE_INACTIVE_CREDENTIALS_INTERVAL = 1000 * 60 * 30; // 30m
+const MAX_INACTIVE_DURATION = 1000 * 60 * 60 * 2; // 2hr
```
ghost, line 8: If these are role sessions, they will likely expire before this timeout? Shall we link each set of creds with their expiration date (updated each time we refresh them)?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -719,17 +722,17 @@ class ReplicateObject extends BackbeatTask {
...
 
-    _setupDestClients(targetRole, log) {
-        this.destBackbeatHost = this.destHosts.pickHost();
-
-        if (this.destConfig.auth.type === authTypeAssumeRole) {
-            const accountId = _extractAccountIdFromRole(targetRole);
-            const roleName = _extractRoleNameFromRole(targetRole);
-            this.clientManager = new ClientManager({
-                id: accountId,
+    _setupAssumeRoleDestClient(targetRole) {
```
ghost, line 725: Due to how errors are reported as null in case of issue, should we at least add some logs?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -741,10 +744,29 @@ class ReplicateObject extends BackbeatTask {
...
+                    port : this.destBackbeatHost.port,
+                };
+                this.clientsManager.removeClients(clientId);
+            }
+        }
+        this.backbeatDest = this.clientsManager.getBackbeatClient(clientId);
+    }
+
+    _setupDestClients(targetRole, log) {
+        this.destBackbeatHost = this.destHosts.pickHost();
```
francoisferrand, line 766: is `destBackbeatHost` really a class attribute: should it not be just a parameter of `_setupAssumeRoleDestClient` ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -741,10 +744,29 @@ class ReplicateObject extends BackbeatTask {
...
+
+        if (this.destConfig.auth.type === authTypeAssumeRole) {
+            this._setupAssumeRoleDestClient(targetRole);
```
francoisferrand, line 769: since we have an asbtraction (clientManager), can't we just rely on it to create the client in any case?

i.e. maybe introduce an abstract class "ClientFactory", with 2 implementations...

lib/clients/ClientsManager.js
```diff
@@ -0,0 +1,286 @@
...
+        });
+
+        if (credentials === null) {
+            return null;
+        }
+
+        const client = this.s3Clients[clientId];
+
+        if (client) {
+            return client;
```
francoisferrand, line 197: when we go through here, the credentials (`credentials = this.credentialsManager.getCredentials`) are not used --> useless to do it

lib/clients/ClientsManager.js
```diff
@@ -0,0 +1,286 @@
...
+        });
+
+        if (credentials === null) {
+            return null;
+        }
+
+        const client = this.backbeatClients[clientId];
+
+        if (client) {
+            return client;
```
francoisferrand, line 260: same, credentials not used in that case

lib/clients/ClientsManager.js
```diff
@@ -0,0 +1,286 @@
...
+        const client = this.getBackbeatClient(clientId);
+        if (client === null) {
+            return null;
+        }
+
+        const config = this._configs[clientId];
+        if (!config) {
+            return null;
+        }
+
```
francoisferrand, line 226: useless, this would mean client is `null`

extensions/replication/tasks/ReplicateObject.js
francoisferrand: note that it may be worth having 3 implementations:
* CredentialsManager-based "AssumeRole" implementation
* the simpler _createCredentials based implementation used in the 'else' branch of thsi functioj
* `CachedClientFactory` adapter, which could be used on top of either implementation (as needed)

lib/clients/ClientsManager.js
```diff
@@ -0,0 +1,286 @@
...
+    getClientConfig(clientId) {
+        return this._configs[clientId] || null;
+    }
+
+    /**
+     * Return an S3 client instance
+     * @param {String} clientId - The client id.
+     * @return {AWS.S3} The S3 client instance to make requests with
+     */
+    getS3Client(clientId) {
```
francoisferrand, line 177: the structure of getS3Client and getBackbeatClient is very redundant : probably should be refactored with a `_getClient` function, maybe something like

```
    _getClient(clients /* map */, clientId /* string */, createClient /* func */) {
        const client = clients[clientId];
        if (client) {
            return client;
        }

        const config = this._configs[clientId];
        if (!config) {
            return null;
        }

        const credentials = this.credentialsManager.getCredentials({
            id: clientId,
            accountId: config.accountId,
            stsConfig: config.stsConfig,
            authConfig: config.authConfig,
        });
        if (credentials === null) {
            return null;
        }

        const client = createClient(config, credentials);
        this.s3Clients[clientId] = client;
        return client;
   }

PR #2643, opened 2025-04-11
Make sure calls to consume() are sequential

welansari:
Calls to consume() should be done sequentially to make sure messages are consumed in order, as there might be delay in the transport layer causing the responses of the consume() calls to be received out of order in a case of concurrent calls.

Consuming the messages out of order also triggers a behaviour in the OffsetLedger where it considers this a replay scenario (can happen after a rebalance) and proceeds by removing all offsets larger than the most recently consumed message, this causes the offsets to never be committed until new messages with a larger offset are consumed. In Zenko this triggers lag alerts which is problematic.

Issue: BB-666

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -149,6 +149,10 @@ class BackbeatConsumer extends EventEmitter {
         this._consumedEventTimeout = null;
         this._drainProcessQueueTimeout = null;
         this._tryConsumedTimeout = null;
+        // This variable is set to true when tasks have been completed
+        // since the last consume, and is used to trigger a new consume
+        // immediately after the current one
+        this._tasksCompletedSinceLastConsume = false;
 
         /** @type {ConsumerStats} */
         this.consumerStats = { lag: {} };
@@ -391,6 +395,16 @@ class BackbeatConsumer extends EventEmitter {
             // processing pipeline is 100% busy, do not attempt to consume
             return undefined;
         }
+
+        // Calls to consume() should be done sequentially to make sure messages
+        // are consumed in order, as there might be delay in the transport
+        // layer causing the responses of the consume() calls to be received
+        // out of order in a case of concurrent calls.
+        if (this._nConsumePendingRequests > 0) {
+            this._tasksCompletedSinceLastConsume = true;
+            return undefined;
+        }
+
         this._nConsumePendingRequests += nNewConsumeRequests;
         return this._consumer.consume(nNewConsumeRequests, (err, entries) => {
             this._nConsumePendingRequests -= nNewConsumeRequests;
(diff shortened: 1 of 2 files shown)


lib/BackbeatConsumer.js
```diff
@@ -435,6 +449,13 @@ class BackbeatConsumer extends EventEmitter {
                         entry.timestamp / 1000);
                     return undefined;
                 });
+                // Immediately try to consume more messages
+                // if tasks have completed since the last consume
+                if (this._tasksCompletedSinceLastConsume) {
```
francoisferrand, line 454: should this not be merged with the next block (line 460) : i.e. just add a `_tasksCompletedSinceLastConsume` condition to go there? (and cleanup `_tasksCompletedSinceLastConsume` anyway, it does no harm)

francoisferrand: ok, this is done separately to avoid the extra 1 sec delay?

lib/BackbeatConsumer.js
```diff
@@ -435,6 +449,13 @@ class BackbeatConsumer extends EventEmitter {
...
+                    this._tasksCompletedSinceLastConsume = false;
+                    process.nextTick(() => this._tryConsume());
+                    return;
```
francoisferrand, line 457: this (new) branch was not covered : something missing in the tests?

welansari: Tests cover 100% of the changes (see [here](https://github.com/scality/backbeat/pull/2643/checks?check_run_id=40383308128)), not sure why there are still warnings in the diff GUI.

lib/BackbeatConsumer.js
welansari: Yes, we should consume immediately after with no delay. That last block is run when Kafka doesn't have enough messages to consume or when it returns an error, i also didn't want to merge the logics.

ghost, line 454: As we return here we basically ignore the `err` variable that might be set, is there any problem if we actually have an error but this condition is true? won't that conflict?

tests/unit/backbeatConsumer.js
```diff
@@ -33,127 +33,208 @@ describe('backbeatConsumer', () => {
...
+            });
         });
 
-        it('should do nothing when state is unknown', () => {
-            const resumeStub = sinon.stub(consumer, '_resumePausedPartitions');
-            const pauseStub = sinon.stub(consumer, '_pauseAssignments');
-            consumer._onCircuitBreakerStateChanged(-1);
-            assert(resumeStub.notCalled);
-            assert(pauseStub.notCalled);
+        it('should only consume once if no tasks completed since last consume', done => {
```
ghost, line 217: Related to my last comment, should we add two tests when `_tasksCompletedSinceLastConsume` is `true`:
- One with an error from kafka
- One when we have less entries than expected

?

lib/BackbeatConsumer.js
welansari: This `if` statement is within the first one that checks that no errors were returned.

ghost (approved):

francoisferrand (approved):

benzekrimaha (approved):

PR #2644, opened 2025-04-25
Improve BackbeatAPI pause/resume logic

benzekrimaha:
Issue: BB-658


lib/util/LocationStatusManager.js
```diff
@@ -238,9 +238,10 @@ class LocationStatusManager {
      * in mongo based on local values
      * @param {string} location location name
      * @param {callback} cb callack
-     * @returns {undefined}
+     * @returns {undefined}z
      */
     _updateServiceStatusForLocation(location, cb) {
+        const previousState = JSON.parse(JSON.stringify(this._locationStatusStore[location].getValue()));
```
welansari: `this._locationStatusStore[location].getValue()` Already returns a JSON no need to stringify and then parse ?

welansari: Actually, this is not really the previous state, if you look at the Mongo query you'll see that it's the value we use to update the document. The previous state can be acquire in the caller of this function.

lib/util/LocationStatusManager.js
```diff
@@ -252,14 +253,31 @@ class LocationStatusManager {
...
+                        this._locationStatusStore[location].resumeLocation(service, null);
+                    }                            
+                    if (previousState[service].scheduledResume) {
+                        this._locationStatusStore[location].setServiceResumeSchedule(
+                            service, 
+                            new Date(previousState[service].scheduledResume)
+                        );
+                    } else {
+                        this._locationStatusStore[location].setServiceResumeSchedule(service, null);
+                    }
```
welansari: An easier approach would be to initialize the `LocationStatus` class with the value of the older one.

```
// example
const previousState = this._locationStatusStore[location].getValue();
this._locationStatusStore[location] = new LocationStatus(this._supportedServices, previousState);
```

lib/util/LocationStatusManager.js
```diff
@@ -438,7 +434,9 @@ class LocationStatusManager {
                     date.setMinutes(date.getMinutes() + 1);
                     this._scheduledResumeJobs[service][location] = schedule.scheduleJob(date,
                         triggerResume.bind(this));
+                    return;
```
francoisferrand: should add a test for this case

lib/util/LocationStatusManager.js
```diff
@@ -470,9 +468,20 @@ class LocationStatusManager {
             }
             const paused = this._locationStatusStore[location].getServicePauseStatus(service);
             if (paused) {
-                this._deleteResumeJob(location, service);
-                this._locationStatusStore[location].setServiceResumeSchedule(service, null);
-                return this._updateServiceStatusForLocation(location, next);
+                this._deleteResumeJob(location, service);
```
francoisferrand: here `_deleteResumeJob()` will be applied even if we fail to update the state in mongo : leading to inconsistent state...

lib/util/LocationStatusManager.js
```diff
@@ -438,7 +434,9 @@ class LocationStatusManager {
                     date.setMinutes(date.getMinutes() + 1);
                     this._scheduledResumeJobs[service][location] = schedule.scheduleJob(date,
                         triggerResume.bind(this));
+                    return;
                 }
+                this._locationStatusStore[location].resumeLocation(service);
```
francoisferrand: setting the "in-memory" value afterwards allows to handle errors, but does not actually solve everything: there are still many race conditions here...

- First and foremost, since we set the value afterwards, changing the value quickly will not "toggle" but keep trying to do the same transition
- In addition, the state is changed asynchronously: so we may make multiple changes to mongo in parallel, and the results may not be processed in the same order...and so the (in-mem) state may not match the state in mongo (i.e. send "Pause" to mongo ; send "Resume" to mongo ; process the end of the "Resume" call ; process of the "Suspend" call → in mongo we may have state Resume, but store "suspend" in memory)

→ the way to handle this is to actually have a proper automaton, so we can detect the "conflicting" operations and deal with them (reject, retry, ignore/wait for completion...)

lib/util/LocationStatusManager.js
```diff
@@ -470,9 +468,20 @@ class LocationStatusManager {
             }
             const paused = this._locationStatusStore[location].getServicePauseStatus(service);
             if (paused) {
-                this._deleteResumeJob(location, service);
-                this._locationStatusStore[location].setServiceResumeSchedule(service, null);
-                return this._updateServiceStatusForLocation(location, next);
+                this._deleteResumeJob(location, service);                
+                return this._updateServiceStatusForLocation(location, err => {
+                    if (err) {
+                        this._logger.debug('failed to delete scheduled resume', {
```
francoisferrand: not tested

lib/util/LocationStatusManager.js
```diff
@@ -507,11 +516,23 @@ class LocationStatusManager {
                 return next();
             }
             const paused = this._locationStatusStore[location].getServicePauseStatus(service);
-                if (!paused) {
+            if (!paused) {
+                this._updateServiceStatusForLocation(location, err => {
+                    if (err) {
+                        this._logger.debug('failed to pause service', {
```
francoisferrand: maybe this log should be moved to `_updateServiceStatusForLocation`, may be less redundant than having it after each call...

lib/util/LocationStatusManager.js
```diff
@@ -423,74 +1004,49 @@ class LocationStatusManager {
...
         } else {
-            this._scheduledResumeJobs[service][location] = schedule.scheduleJob(date,
-                triggerResume.bind(this));
+                this._scheduledResumeJobs[service][location] = schedule.scheduleJob(date, triggerResume);
             }
-    }
+        }
 
-    /**
+   /**
```
ghost:
```suggestion
    /**
```

lib/util/LocationStatusManager.js
```diff
@@ -343,7 +924,7 @@ class LocationStatusManager {
             return cb(null, statuses);
         });
     }
-
+
```
ghost:
```suggestion

```

lib/util/LocationStatusManager.js
```diff
@@ -12,20 +12,156 @@ const actions = {
     resume: 'resumeService',
 };
 
+// Operation types for the state machine
+const OPERATION_TYPES = {
```
ghost:
```suggestion
const OPERATION_TYPES = Object.freeze({
```

if we want to use enum-like values without TS I suggest using this to ensure it will never be changed in any way

lib/util/LocationStatusManager.js
```diff
@@ -12,20 +12,156 @@ const actions = {
...
+        this.options = options;
+        this.state = OPERATION_STATES.PENDING;
+        this.timestamp = new Date();
+        this.retryCount = 0;
+        this.maxRetries = 3;
+        this.error = null;
+    }
+
+    canTransitionTo(newState) {
+        const validTransitions = {
```
ghost:
```suggestion
        const validTransitions = Object.freeze({
```

lib/util/LocationStatusManager.js
```diff
@@ -12,20 +12,156 @@ const actions = {
...
+const SERVICE_STATES = {
+    RUNNING: 'running',
+    PAUSED: 'paused',
+    SCHEDULED_RESUME: 'scheduled_resume',
+};
+
+/**
+ * Represents a queued operation
+ */
+class QueuedOperation {
```
ghost: This is a generic class name that should:
- Have a more specific name if we can only handle specific operations, here it seems it's related to location, so it should be in the name (maybe `ServiceOperation`?)
- Be fully unit tested
- Be in a dedicated file, as we tend to avoid having multiple classes in the same file

lib/util/LocationStatusManager.js
```diff
@@ -12,20 +12,156 @@ const actions = {
...
+            [OPERATION_STATES.FAILED]: [OPERATION_STATES.PROCESSING, OPERATION_STATES.CANCELLED],
+            [OPERATION_STATES.COMPLETED]: [],
+            [OPERATION_STATES.CANCELLED]: [],
+        };
+        return validTransitions[this.state].includes(newState);
+    }
+
+    setState(newState, error = null) {
+        if (!this.canTransitionTo(newState)) {
+            throw new Error(`Invalid state transition from ${this.state} to ${newState}`);
```
ghost: Maybe include some more info for the debugging as well? (the id, type and location?)

lib/util/LocationStatusManager.js
```diff
@@ -12,20 +12,156 @@ const actions = {
...
+        this.location = location;
+        this.options = options;
+        this.state = OPERATION_STATES.PENDING;
+        this.timestamp = new Date();
+        this.retryCount = 0;
+        this.maxRetries = 3;
+        this.error = null;
+    }
+
+    canTransitionTo(newState) {
```
ghost: As this is used internally only, we can make it private

```suggestion
    #canTransitionTo(newState) {
```

lib/util/LocationStatusManager.js
```diff
@@ -12,20 +12,156 @@ const actions = {
...
 /**
- * Contains methods to incrememt different metrics
- * @typedef {Object} ServiceConfig
- * @property {string} namespace - crr config
- * @property {string} statePath - zookeeper state path
- * @property {string} topic - redis channel name
- * @property {string} isMongo - true if service uses mongo to store status
+ * State machine for service location status
  */
+class ServiceLocationStateMachine {
```
ghost: From what I understand, both `QueuedOperation` and `ServiceLocationStateMachine` are state machines:
- `QueuedOperation` for a single lifecycle task (PENDING -> PROCESSING -> COMPLETED/FAILED)
- `ServiceLocationStateMachine` for the service status given a location (RUNNING -> PAUSED -> SCHEDULED_RESUME).

I would suggest to follow the DRY principle here and create a generic `BaseStateMachine` that both classes can extend. The generic can be as simple as an initial state, a set of valid states (to be overriden) and a `transitionTo` function.

Then the child classes can define their specific states with `static` class variables, so it's globally available as well, and then we simplify the implementation overall.

tests/functional/api/routes.js
```diff
@@ -903,65 +903,149 @@ describe('API routes', () => {
...
+                ], err => {
+                    if (err) {
+                        //eslint-disable-next-line no-console
+                        console.error('Error in setupZkTestState:', err);
+                        return cb(err);
+                    }
+                    return cb();
                 });
-            }
+            }
```
ghost:
```suggestion
            }
```

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+                        const error = new Error(`Operations failed: 
+                            ${failed.map(op => op.error?.message || 'Unknown error').join(', ')}`);
+                        reject(error);
+                    } else {
+                        resolve();
+                    }
+                    
+                    operations.forEach(op => {
+                        this._pendingOperations.delete(op.id);
+                    });
```
ghost: If we call a callback, we should not perform any more operation after: that can lead to inconsistencies. Better to cleanup first, then call the cb

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+                } else {
+                    setTimeout(checkCompletion, 100);
+                }
+            };
+
+            setTimeout(checkCompletion, 10);
```
ghost: The hard coded delays and the missing `return`s even if we go through the above if/else calling a callback is strange, we should either cleanup, return a callback and that's it, or make this delay (if absolutely necessary to have it) configurable

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+            [OPERATION_TYPES.DELETE_SCHEDULE]: actions.deleteSchedule,
+        };
+        return mapping[operationType];
+    }
+
+    /**
+     * Execute pause operation
+     */
+    async _executePauseOperation(location, service, stateMachine) {
+        const paused = this._locationStatusStore[location].getServicePauseStatus(service);
```
ghost: should we add a debug/info log if we try to pause something already paused?

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+                throw error;
+            }
+        }
+    }
+
+    /**
+     * Execute resume operation
+     */
+    async _executeResumeOperation(location, service, stateMachine) {
+        const paused = this._locationStatusStore[location].getServicePauseStatus(service);
```
ghost: Maybe here as well, a log if we try to resume something already running?

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+        return new Promise((resolve, reject) => {
+            this._updateServiceStatusForLocation(location, err => {
+                if (err) {
+                    reject(err);
+                } else {
+                    resolve();
+                }
+            });
+        });
+    }
```
ghost:
```suggestion
    _updateServiceStatusForLocationAsync(location) {
      return promisify(this._updateServiceStatusForLocation.bind(this))(location);
    }
```

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+
+    /**
+     * Queue an operation for processing
+     */
+    _queueOperation(type, service, locations, options = {}) {
```
ghost: I know we don't properly do it at Scality but the standard is to use # to make a method of a class private. `_` is very old style.

```suggestion
    #queueOperation(type, service, locations, options = {}) {
```

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+            default:
+                throw new Error(`Unknown operation type: ${type}`);
+        }
+    }
+
+    /**
+     * Get Redis action from operation type
+     */
+    _getRedisAction(operationType) {
+        const mapping = {
```
ghost: This looks like a candidate for a static member of the class, to avoid recomputing it each time

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+                
+                if (!stillProcessing) {
+                    resolve();
+                } else {
+                    setTimeout(checkCompletion, 100);
+                }
+            };
+            checkCompletion();
+        });
+    }
```
ghost: You coudl do something like this

```suggestion
    async _waitForOperations(operations) {
        const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
        while (operations.some(op => op.state === OPERATION_STATES.PROCESSING)) {
            await delay(100);
        }
    }
```

(with the delay being a util function)
But in any case, I would suggest, for all functions doing this kind of loop, to have some kind of timeout, to avoid infinite loops.

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+                method: 'LocationStatusManager._processOperation',
+                operationId: operation.id,
+                error: error.message,
+                retryCount: operation.retryCount,
+            });
+
+            if (operation.retryCount < operation.maxRetries) {
+                setTimeout(() => {
+                    this._operationQueue.push(operation);
+                }, 1000 * Math.pow(2, operation.retryCount));
```
ghost: Do we have support for exponentional backoff in arsenal?
We should add a comment maybe stating the periods it gives given the number of retries we have right now, maybe it will be too big or too small

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+
+                if (stateMachineInstance.state !== newSmState || 
+                    (stateMachineInstance.scheduledResumeDate ? 
+                        stateMachineInstance.scheduledResumeDate.toISOString() : null) !== 
+                    (newScheduledResumeDate ? newScheduledResumeDate.toISOString() : null)) {
+                    this._logger.info('Updating ZK-based state machine from ZK data due to detected changes', 
+                        { key, oldState: stateMachineInstance.state, newState: newSmState, 
+                          oldScheduledResume: stateMachineInstance.scheduledResumeDate, 
+                          newScheduledResume: newScheduledResumeDate,
+                          zkDataUsed: currentZkData });
```
ghost:
```suggestion
                        {
                            key, oldState: stateMachineInstance.state,
                            newState: newSmState, 
                            oldScheduledResume: stateMachineInstance.scheduledResumeDate, 
                            newScheduledResume: newScheduledResumeDate,
                            zkDataUsed: currentZkData,
                          });
```

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+                        this._logger.error('Invalid scheduledResume date format from ZK during refresh', 
+                            { key, dateValue: currentZkData.scheduledResume, error: dateError.message });
+                        newSmState = SERVICE_STATES.PAUSED;
+                        newScheduledResumeDate = null;
+                    }
+                }
+
+                if (stateMachineInstance.state !== newSmState || 
+                    (stateMachineInstance.scheduledResumeDate ? 
+                        stateMachineInstance.scheduledResumeDate.toISOString() : null) !==
```
ghost:
```suggestion
                    (stateMachineInstance.scheduledResumeDate?.toISOString?.() 
```

Also what if we get two `null` (or `undefined` with my suggestion) -> the condition will pass, is that what we want?

lib/util/LocationStatusManager.js
```diff
@@ -59,6 +195,452 @@ class LocationStatusManager {
...
+        console.log('checking state machine for key', this._stateMachines.has(key));
+        if (!this._stateMachines.has(key)) {
+            let initialState = SERVICE_STATES.RUNNING;
+            let sourceData = null; 
+            const serviceConfig = this._serviceConfig[service];
+            this._logger.debug('Service Config state machine', { serviceConfig });
+            if (this._serviceConfig[service] && !this._serviceConfig[service].isMongo) {
+                this._logger.debug('Initializing state machine for ZK-based service', { key, service, location });
+                try {
+                    sourceData = await this._getZkDataForLocation(service, location);
```
ghost: This calls ZK every time we process an operation, is that OK? I saw we increased the timeouts in tests from 1000ms to 2000ms, is this change fine if we pause/resume at high rate?

lib/util/LocationStatusManager.js
```diff
@@ -12,20 +12,156 @@ const actions = {
...
+            [OPERATION_TYPES.DELETE_SCHEDULE]: () => {
+                this.state = SERVICE_STATES.PAUSED;
+                this.scheduledResumeDate = null;
+            },
+        };
+
+        const transition = transitions[operationType];
+        if (transition) {
+            transition();
+        }
```
ghost:
```suggestion
            transition?.();
```

+ A log in case there is no transition?

ghost (commented):
I tried to review everything but due to some changes I discuss, some of my comments will likely end up irrelevant, I suggest reading the whole review and work on the "big part"s first, so you don't waste time addressing some of my comments I wrote with the current code

lib/util/LocationStatusManager.js
benzekrimaha: I don't think we can have something such as high rate pause /resume ( we need to wait for the resume to occur before pause , if pause is called twice nothing will happen etc... ) so getting the state here from zk enables us to actally keep the consistency between the states saved  @Kerkesni  what do you think?

lib/util/LocationStatusManager.js
benzekrimaha: we have two retries max for the moment so not really that big

lib/util/LocationStatusManager.js
benzekrimaha: For consistency I'd rather stick with how we do it because we'll need to change it everywhere later , we can create a ticket for this if you want

lib/util/LocationStatusManager.js
benzekrimaha: Went with the `_ `

lib/util/LocationStatusManager.js
benzekrimaha: Done we do have now a base classe then a ServiceLocationOperation one in separate files , and then ServiceLocationStateMachine with the locationStatusManager extending the base one

lib/util/BaseStateMachine.js
```diff
@@ -0,0 +1,68 @@
...
+        throw new Error('getTransitions() must be implemented by child class');
+    }
+
+    /**
+     * Check if transition to new state is valid
+     * @param {string} newState - The state to transition to
+     * @returns {boolean} True if transition is valid
+     */
+    _canTransitionTo(newState) {
+        const validTransitions = Object.freeze(this.constructor.getTransitions());
```
francoisferrand: why double freeze (both here and in LocationStatusManager) ?

lib/util/LocationStatusManager.js
```diff
@@ -1,31 +1,175 @@
...
+            ],
+            [SERVICE_STATES.SCHEDULED_RESUME]: [
+                SERVICE_STATES.RUNNING,
+                SERVICE_STATES.PAUSED,
+                SERVICE_STATES.SCHEDULED_RESUME // Allow rescheduling
+            ],
+        });
+    }
+
+    static getValidOperations() {
```
francoisferrand: I don't understand this method : the valid operations should be the same as the transitions, i.e. in a FSM an operation is valid if there is a transition from the current state using that operation...

lib/util/LocationStatusManager.js
```diff
@@ -1,31 +1,175 @@
...
+
+    static getStates() {
+        return SERVICE_STATES;
+    }
+
+    static getOperationTypes() {
+        return OPERATION_TYPES;
+    }
+
+    static getTransitions() {
```
francoisferrand: a transition is what happens when an operation is performed on a state.

so if you want to model with array it would more likely be something like: `[state][operation] → { nextState, action}`

lib/util/LocationStatusManager.js
```diff
@@ -1,31 +1,175 @@
...
+            if (operation.isProcessing()) {
+                if (this._areOperationsConflicting(operation.type, newOperation.type)) {
+                    conflicts.push(operation);
+                }
+            }
+        }
+        return conflicts;
+    }
+
+    _areOperationsConflicting(operation1, operation2) {
```
francoisferrand: conflicting operations should not be computed : this should be just a corrolary of the FSM definition (a conflict happens when the operation is not a valid transition from the current state)

lib/util/LocationStatusManager.js
```diff
@@ -1,31 +1,175 @@
...
+
+    addPendingOperation(operation) {
+        return this.pendingOperations.set(operation.id, operation);
+    }
+
+    removePendingOperation(operationId) {
+        return this.pendingOperations.delete(operationId);
+    }
+
+    executeStateTransition(operationType, options = {}) {
```
francoisferrand: since you introduced a `BaseStateMachine`, the "execution" should be part of that class.

and the action to perform should be part of the "data" defining the FSM (i.e. the map of transitions) : having it here makes the code very hard to follow (and splitting the definition of the FSM), even though we end up hard-coding the FSM in this function....

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+                        {
+                            key, oldState: stateMachineInstance.state,
+                            newState: newSmState, 
+                            oldScheduledResume: stateMachineInstance.scheduledResumeDate, 
+                            newScheduledResume: newScheduledResumeDate,
+                            zkDataUsed: currentZkData,
+                          });
+                }
+                stateMachineInstance.state = newSmState;
+                stateMachineInstance.scheduledResumeDate = newScheduledResumeDate;
```
francoisferrand: this updates the state, so must be done through a transition (otherwise no point having a FSM)

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+            this._logger.debug('Service Config state machine', { serviceConfig });
+            if (this._serviceConfig[service] && !this._serviceConfig[service].isMongo) {
+                this._logger.debug('Initializing state machine for ZK-based service', { key, service, location });
+                try {
+                    sourceData = await this._getZkDataForLocation(service, location);
+                    if (sourceData.paused) {
+                        initialState = sourceData.scheduledResume ?
+                            SERVICE_STATES.SCHEDULED_RESUME : SERVICE_STATES.PAUSED;
+                    } else {
+                        initialState = SERVICE_STATES.RUNNING;
```
francoisferrand: this is repeated throughout : either you set a default at the beginning (286), or you set it in every branch, but not both.

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+                    stateMachineInstance.scheduledResumeDate = null;
+                }
+            }
+            this._stateMachines.set(key, stateMachineInstance);
+        }
+
+        const stateMachineInstance = this._stateMachines.get(key);
+
+        if (this._serviceConfig[service] && !this._serviceConfig[service].isMongo) {
+            this._logger.debug('State machine is ZK-based. Refreshing/validating its state from ZK.',
```
francoisferrand: refresh looks the same as init (for ZK backend), should not create so much duplication

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
             ingestion: {},
             lifecycle: {},
         };
+
+        this._stateMachines = new Map();
```
francoisferrand: name should be more precise : the purpose of this variables is not to store an arbitrary collection of state machines, but to keep track of the state of each location (and the state is managed through a state machine, but the "data" of a state machine is really just the state of the machine...)

→ the name of the variable should reflect that, so we easily understand what this is about...

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+            if (initialState === SERVICE_STATES.SCHEDULED_RESUME && sourceData.scheduledResume) {
+                try {
+                    stateMachineInstance.scheduledResumeDate = new Date(sourceData.scheduledResume);
+                } catch (dateError) {
+                    this._logger.error('Invalid scheduledResume date format, cannot set on state machine',
+                        { key, service, location, dateValue: sourceData.scheduledResume, error: dateError.message });
+                    stateMachineInstance.state = SERVICE_STATES.PAUSED; 
+                    stateMachineInstance.scheduledResumeDate = null;
+                }
+            }
```
francoisferrand: should be a transition.
the purpose of the StateMachine is to manage the state, so you must _never_ set the state directly

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+                this._logger.debug('Operation no longer valid, cancelling', {
+                    method: 'LocationStatusManager._processOperation',
+                    operationId: operation.id,
+                    currentState: stateMachine.state,
+                    operationType: operation.type,
+                });
+                operation.setState(ServiceLocationOperation.STATES.COMPLETED);
+                stateMachine.removePendingOperation(operation.id);
+                return;
+            }
```
francoisferrand: all this should be part of the state machine...

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+
+    /**
+     * Process a single operation from the queue
+     */
+    async _processOperation(operation) {
+        this._logger.debug('Processing operation', { id: operation.id, type: operation.type, state: operation.state });
+        const stateMachine = await this._getStateMachine(operation.service, operation.location);
+        
+        try {
+            operation.setState(ServiceLocationOperation.STATES.PROCESSING);
```
francoisferrand: what is the point of having a "state" on the operation?
an operation is the "event" that triggers a transition in the FSM, so should either it is handled (valid transition & state updated) or rejected (conflict / invalid state) but there is not further processing or state...
(and looking at the PR, I don't see where we use the state of operation)

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
             ingestion: {},
             lifecycle: {},
         };
+
+        this._stateMachines = new Map();
+
+        this._operationQueue = async.queue((operation, callback) => {
```
francoisferrand: why do we need a queue, we already have an FSM and node.js is already single-threaded : so in node.js, we can safely make transitions in the FSM wherever we need...
(while a queue opens the risk of queueing similar events which "cancel" each other, but which we will process sequentially...)

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+            };
+
+            operations.forEach(operation => {
+                this._operationQueue.push(operation);
+            });
+            const checkCompletion = () => {
+                const allCompleted = operations.every(op => 
+                    op.state === ServiceLocationOperation.STATES.COMPLETED || 
+                    op.state === ServiceLocationOperation.STATES.CANCELLED
+                );
```
francoisferrand: this is reinventing callbacks (async.forEach) or promises (Promise.all) : code will be much simpler and efficient (no timer) if we just trigger each location's transition in parallel...

lib/util/LocationStatusManager.js
francoisferrand: should the log be generated automatically from the FSM, to avoid "polluting" the logic here?

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+    static getRedisAction(operationType) {
+       return REDIS_ACTION_MAPPING[operationType] || null;
+    }
+
+    /**
+     * Execute pause operation
+     */
+    async _executePauseOperation(location, service, stateMachine) {
+        const paused = this._locationStatusStore[location].getServicePauseStatus(service);   
+        if (!paused) {
```
francoisferrand: is this not the whole purpose of the FSM to handle the state?

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+    async _executeOperation(operation, stateMachine) {
+        const { type, service, location, options } = operation;
+    
+        if (this._serviceConfig[service] && !this._serviceConfig[service].isMongo) {
+            this._pushActionToRedis(service, location, 
+                LocationStatusManager.getRedisAction(type), options.scheduleDate);
+            stateMachine.executeStateTransition(type, options);
+            return;
+        }
+        switch (type) {
```
francoisferrand: one of the purpose of FSM is to avoid these kind of switch: if there is some operation to performed, it should be part of the FSM definition (linked to a transition or entry into the state, but there anyway)

lib/util/LocationStatusManager.js
```diff
@@ -59,11 +203,476 @@ class LocationStatusManager {
...
+    async _waitForOperations(operations, timeoutMs = OPERATION_WAIT_TIMEOUT_MS) {
+        const startTime = Date.now();
+        
+        while (operations.some(op => op.state === ServiceLocationOperation.STATES.PROCESSING)) {
+            if (Date.now() - startTime > timeoutMs) {
+                throw new Error('Timeout waiting for operations to complete');
+            }
+            await delay(COMPLETION_POLL_INTERVAL_MS);
+        }
+    }
```
francoisferrand: should not poll, but use callabcks/promises

benzekrimaha:
Note for reviewers : changes on going please do not review yet

lib/util/BaseStateMachine.js
```diff
@@ -0,0 +1,82 @@
...
+        this.metadata = metadata;
+    }
+
+    /**
+     * Get valid states - to be overridden by child classes
+     * @returns {Object} Object containing state constants
+     */
+    static getStates() {
+        throw new Error('getStates() must be implemented by child class');
+    }
```
francoisferrand: not needed

lib/util/BaseStateMachine.js
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Get valid transitions - to be overridden by child classes
+     * @returns {Object} Object mapping [state][operation] → { nextState, action }
+     */
+    static getTransitions() {
+        throw new Error('getTransitions() must be implemented by child class');
+    }
```
francoisferrand: no reason to use inheritance here : the state machine only needs "data", which can be provided when initializing the state machine (in the constructor).

this will also avoid this weird trick to have inheritance on a static method (using `this.constructor.getTransitions()`)

lib/util/LocationStatusManager.js
```diff
@@ -52,18 +131,262 @@ class LocationStatusManager {
...
+
+    /**
+     * Execute operation directly on the state machine
+     */
+    async _executeOperation(service, location, operationType, options = {}) {
+        const stateMachine = await this._getStateMachine(service, location);
+        
+        // Check if operation is valid in current state
+        if (!stateMachine.canExecuteOperation(operationType)) {
```
francoisferrand: no reason to do this : this is handled already in `executeOperation` (it will return an error if it fails, and we must then deal with it), and create _opportunities_ for race conditions (i.e. check that we can execute the transition, do something, than perform the transition: but state may have change in between...)

lib/util/BaseStateMachine.js
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Check if currently in any of the specified states
+     * @param {string[]} states - States to check
+     * @returns {boolean} True if in any of the specified states
+     */
+    isInAnyState(states) {
+        return states.includes(this.state);
+    }
```
francoisferrand: does not seem to be used, and best avoid this: it can easily lead to anti pattern when the code looks at the state, instead of reacting to automaton callback

lib/util/BaseStateMachine.js
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Check if an operation is valid in the current state
+     * @param {string} operation - The operation to check
+     * @returns {boolean} True if operation is valid
+     */
+    canExecuteOperation(operation) {
+        const transitions = this.constructor.getTransitions();
+        return transitions[this.state]?.[operation] !== undefined;
+    }
```
francoisferrand: should be removed, the goal of the automaton is to perform kind of "atomic" transitions : which is not possible with such a function... The code using the automaton should simply attempt the transition(s), and it will get an error if the transition fails.

lib/util/BaseStateMachine.js
```diff
@@ -0,0 +1,82 @@
...
+        throw new Error('getTransitions() must be implemented by child class');
+    }
+
+    /**
+     * Execute an operation and transition to the resulting state
+     * @param {string} operation - The operation to execute
+     * @param {Object} context - Additional context for the operation
+     * @returns {Object} The transition result { nextState, action }
+     */
+    executeOperation(operation, context = {}) {
```
francoisferrand: nit:
```suggestion
    transition(operation, context = {}) {
```

lib/util/BaseStateMachine.js
```diff
@@ -0,0 +1,82 @@
...
+        
+        if (!stateTransitions || !stateTransitions[operation]) {
+            throw new Error(`Invalid operation ${operation} for state ${this.state}`);
+        }
+        
+        const transition = stateTransitions[operation];
+        
+        // Execute the action if it's a function
+        if (typeof transition.action === 'function') {
+            transition.action.call(this, context);
```
francoisferrand: the callback may be asynchronous : so it must be called `after` updating the state

lib/util/LocationStatusManager.js
```diff
@@ -52,18 +131,262 @@ class LocationStatusManager {
...
+                newState: result.nextState,
+            });
+            return result;
+        }
+    }
+
+    /**
+     * Update the in-memory store based on operation
+     */
+    async _updateServiceInStore(service, location, operationType, options = {}) {
```
francoisferrand: this is the kind of thing that should be in the automaton callbacks...
it seems you were adding too many abstraction layers, and wrapped everything: on various events, you know exactly which operation you are dealing with, so it may somewhat look like this:

```
if (state.transition(OPERATION.PAUSE)) {
    locationStore.pauseLocation(service);
}
```

or alternatively we simply "issue" transitions, and keep all the code in the callbacks, to be executed when the transition actually happens.

lib/util/LocationStatusManager.js
```diff
@@ -52,18 +131,262 @@ class LocationStatusManager {
...
+        const locationStore = this._locationStatusStore[location];
+        
+        switch (operationType) {
+            case OPERATION_TYPES.PAUSE:
+                if (!locationStore.getServicePauseStatus(service)) {
```
francoisferrand: why do we check the status in the store? Isn't this why we introduced the automaton?

lib/util/BaseStateMachine.js
```diff
@@ -0,0 +1,82 @@
...
+     * @returns {Object} Object mapping [state][operation] → { nextState, action }
+     */
+    static getTransitions() {
+        throw new Error('getTransitions() must be implemented by child class');
+    }
+
+    /**
+     * Execute an operation and transition to the resulting state
+     * @param {string} operation - The operation to execute
+     * @param {Object} context - Additional context for the operation
```
francoisferrand: kind of mixed feelings regarding the `context` passed to callback: the point of having "callbacks" in the automaton is to have the same code executed on transition, and so it should not take `params.... and if we really need params, maybe a better design would be to just "proceed" depending on the success or failure of the transition:

```
if (stateMachine.executeOperation(op)) {
   // do what needs to be done!
} else {
   // failed to transition, maybe return an error?
}
```

maybe this is appropriate, but should double check if this is really the way to go.

lib/util/LocationStatusManager.js
```diff
@@ -1,31 +1,110 @@
...
+    static getStates() {
+        return SERVICE_STATES;
+    }
+
+    static getOperationTypes() {
+        return OPERATION_TYPES;
+    }
+
+    static getTransitions() {
+        return Object.freeze({
```
francoisferrand: this automaton looks way too simple: the goal of the PR was to handle intermediate states (i.e. pause requested and in-memory state updated, but not applied everywhere), so I would imagine solving it would involve having these intermediate states in here...
(and this is not what fixes the issue, is there really a point refactoring with the state machine?)

lib/util/LocationStatusManager.js
```diff
@@ -1,31 +1,110 @@
...
+                    }
+                },
+                [OPERATION_TYPES.SCHEDULE_RESUME]: {
+                    nextState: SERVICE_STATES.SCHEDULED_RESUME,
+                    action(context) {
+                        this.scheduledResumeDate = context.scheduleDate;
+                        this.metadata.scheduledResumeDate = context.scheduleDate;
+                    }
+                },
+                [OPERATION_TYPES.DELETE_SCHEDULE]: {
```
francoisferrand: no difference with the PAUSE operation, is operation really needed?

lib/util/LocationStatusManager.js
```diff
@@ -52,18 +131,262 @@ class LocationStatusManager {
...
+                    return pausedStateMachine;
+                }
+            }
+            
+            this._serviceLocationStates.set(key, stateMachine);
+        }
+    
+        const stateMachine = this._serviceLocationStates.get(key);
+    
+        // For ZK services, sync with latest state
```
francoisferrand: what is the point of _syncing_ the state?
- we are using an automaton so that we can actually manage the state, and there should thus not be anything to "sync" : from the initial state, we should just "listen" to state changes from Zookeeper to update the automaton
- if we reset the state based on zookeeper, then the automaton is not so useful : it should be used to trigger other transition, and thus allow handling the intermediate states (RUNNING → PAUSE_IN_PROGRESS → PAUSED, etc...)

lib/util/LocationStatusManager.js
```diff
@@ -52,18 +131,262 @@ class LocationStatusManager {
...
+                    if (expectedDate) {
+                        zkSyncedStateMachine.scheduledResumeDate = expectedDate;
+                        zkSyncedStateMachine.metadata.scheduledResumeDate = expectedDate;
+                    }
+                    this._serviceLocationStates.set(key, zkSyncedStateMachine);
+                    return zkSyncedStateMachine;
+                }
+            } catch (err) {
+                this._logger.error('Failed to sync ZK state', { key, error: err.message });
+            }
```
francoisferrand: same code as in the beginning of the function (when `!this._serviceLocationStates.has(key)`), should be de-duplicated...

lib/util/LocationStatusManager.js
```diff
@@ -52,18 +131,262 @@ class LocationStatusManager {
...
         this._locationStatusColl = null;
-
         this._locationStatusStore = {};
         this._scheduledResumeJobs = {
             crr: {},
             ingestion: {},
             lifecycle: {},
         };
+
+        this._serviceLocationStates = new Map();
```
francoisferrand: we already store the state (for each workflow/location) in `LocationStatus` instances, in `_locationStatusStore` : so the state is kind of redundant....

francoisferrand (commented):
Overall, still seems very complex to me, mixing different layers and trying to introduce functions which "abstract everything" and make it very complex to understand what happens.

- We need to find a way to keep the "abstraction" (mongo vs zookeeper) somewhat cleanly separated, and not intermingle it with state management ;
- The state is already managed in memory (that is the starting point of the bug), in LocationStatus, so this  is where the state management would need to be updated ;
- Since we want to fix inconsistent state, it means we need intermediary states (like PAUSE_IN_PROGRESS) in the FSM, and transition both before initiating the backend operation and once it has succeeded (or failed)

---

Since we have multiple processes (or restart), I wonder however if this approach really works in every case : we may write the state to the backend, but then another process could overwrite this state in between, and we would still be inconsistent?
(i.e. maybe the safer -and simpler- way is to rely on the backend information and using more "transactional" operations instead of in-memory cache)

lib/util/LocationStatusManager.js
```diff
@@ -136,7 +140,8 @@ class LocationStatusManager {
      * @returns {undefined}
      */
     _deleteInvalidLocations(locations, cb) {
-        const validLocationNames = Object.keys(validLocations);
+        const validLocationDetails = locations
+            .filter(loc => validLocationNames.includes(loc._id));
```
francoisferrand: instead of linear search, should `validLocationNames` be transformed into some sort of object/map, to allow immediately accessing the location?

lib/util/LocationStatusManager.js
```diff
@@ -136,7 +140,8 @@ class LocationStatusManager {
...
         const invalidLocations = locations
             .filter(loc => !validLocationNames.includes(loc._id))
```
francoisferrand: we traverse and filter the list of locations twice : is there a somewhat "standard" function to "split" the list according to a criteria, in a single call? (maybe split or partition, something like that)

lib/util/LocationStatusManager.js
```diff
@@ -263,6 +289,37 @@ class LocationStatusManager {
...
+     */
+    _getMongoStateDetails(service, locations, cb) {
+        this._locationStatusColl.find(
+            { _id: { $in: locations } },
+            { projection: { _id: 1, [`value.${service}`]: 1 } }
+        )
+        .toArray()
+        .then(docs => {
+            const states = docs.reduce((acc, doc) => {
+                if (doc.value && doc.value[service]) {
```
francoisferrand: why do filtering in javascript? should we do this directly in `find()` ?

lib/util/LocationStatusManager.js
```diff
@@ -424,8 +477,20 @@ class LocationStatusManager {
...
+                [`value.${service}.paused`]: true,
+                [`value.${service}.scheduledResume`]: { $eq: date }
+            };
+            const update = {
+                $set: {
+                    [`value.${service}`]: {
+                        paused: false,
+                        scheduledResume: null,
+                    }
+                }
```
francoisferrand: seems a lot to pass on each call to `_updateServiceStatusForLocation`, would be better to abstract the "detail" of mongo implementation/mapping: i.e. passing at the very most the relevant "filters" (`{ paused: true, scheduledResume: date }`) and values (`{ paused: false, scheduledResume: null }`)...

maybe we could make the query something like `{ _id: location, ``[value.${service}]`` : { paused: true, scheduledResume: {$eq: date} } }` (inside _updateServiceStatusForLocation), so everything is consistent?

lib/util/LocationStatusManager.js
```diff
@@ -424,8 +477,20 @@ class LocationStatusManager {
...
+            };
+            this._updateServiceStatusForLocation(query, update, err => {
                 if (err) {
                     this._logger.error('error resuming scheduled job, retrying in 1min', {
                         method: 'LocationStatusManager._scheduleResumeJob',
```
francoisferrand, line 504: What happens if the entry is not found?
in that case, it means we basically hit some kind of race condition, and we should ignore the "resume"

--> if it is an error, should not be retried
--> if it is not an error, should not display that location was resumed

francoisferrand: (same for other functions as well)

lib/util/LocationStatusManager.js
```diff
@@ -192,24 +214,31 @@ class LocationStatusManager {
...
             if (err) {
-                this._logger.error('Could not setup location statuses in mongo', {
-                    method: 'LocationStatusManager._setupLocationStatusStore',
-                    error: err.message,
-                });
                 return cb(err);
             }
-            this._logger.info('Locations status setup complete', {
-                method: 'LocationStatusManager._setupLocationStatusStore',
+            if (cluster.isWorker) {
```
francoisferrand, line 221: not sure why we use worker mode, esp. with only one worker: please create a followup ticket so we can look at it.

welansari: [BB-717](https://scality.atlassian.net/browse/BB-717)

lib/util/LocationStatusManager.js
welansari: It's not filtering i'm transforming the array into a map

lib/util/LocationStatusManager.js
welansari: replaced the two filter functions with a single foreach that builds the two structures

lib/util/LocationStatusManager.js
```diff
@@ -234,35 +263,74 @@ class LocationStatusManager {
...
-            {
-                $set: {
-                    _id: location,
-                    value: this._locationStatusStore[location].getValue(),
-                }
+    _updateServiceStatusForLocation(location, service, conditions, newStatus, cb) {
+        const query = { _id: location };
+        Object.keys(conditions).forEach(key => {
+            query[`value.${service}.${key}`] = conditions[key];
+        });
```
francoisferrand, line 278: nit: seems like a `map` call?

lib/util/LocationStatusManager.js
francoisferrand, line 507: we already log error inside `_updateServiceStatusForLocation` : is it worth logging it again here?

lib/util/LocationStatusManager.js
```diff
@@ -234,35 +263,74 @@ class LocationStatusManager {
...
-            .catch(err => {
-                if (err) {
-                    this._logger.error('Could not update location service status in MongoDB', {
-                        method: 'LocationStatusManager._updateServiceStatusForLocation',
-                        error: err.message,
-                    });
-                    return cb(err);
+            .then(res => {
+                if (res.modifiedCount === 0) {
+                    return cb(invalidStateTransitionError);
```
francoisferrand, line 295: should log the error here; but maybe not as an "error", just as warning ?
(same for every API)

lib/util/LocationStatusManager.js
```diff
@@ -563,15 +631,17 @@ class LocationStatusManager {
                 this._pushActionToRedis(service, location, actions.resume, schedule);
                 return next();
             }
-            const paused = this._locationStatusStore[location].getServicePauseStatus(service);
-            if (paused) {
-                this._locationStatusStore[location].resumeLocation(service, schedule);
+            const conditions = { paused: true };
+            const newStatus = { paused: !!schedule, scheduledResume: schedule || null };
```
francoisferrand: nit: maybe more "readable" with ternary?

```suggestion
            const newStatus = schedule ? { paused: true, scheduledResume: schedule }
                                       : { paused: false, scheduledResume: null };
```

francoisferrand (approved):

francoisferrand (commented):
also please rebase 🙏

lib/util/LocationStatusManager.js
welansari: not really

benzekrimaha (commented):
can't approve as I opened the PR walid worked on later but lgtm

DarkIsDude (approved):

PR #2645, opened 2025-04-25
enableIngestionUpdates value parametrize

benzekrimaha:
Issue: BB-677


welansari (approved):
LGTM, the only impact this should have is the ingestion related routes of Backbeat API not working, as the server won't know the ingestion location names.

ghost (approved):
LGTM 
Just to ensure no future regression, maybe good to test this (the test should just fail if we later change the behavior, later...)

francoisferrand (approved):

PR #2646, opened 2025-05-15
use auth v4 in backbeat client

jonathan-gramain:
Modify the backbeat client definition file to force it to use authv4, so that Cloudserver contacts Vault with POST requests, avoiding errors due to limitations on token header size.

I had to disable signing of the POST body, which is not supported in streaming mode by the AWS v2 client.

Issue: BB-682

### DIFF
--- a/lib/clients/backbeat-2017-07-01.api.json
+++ b/lib/clients/backbeat-2017-07-01.api.json
@@ -8,7 +8,7 @@
         "protocol": "rest-json",
         "serviceAbbreviation": "Backbeat",
         "serviceFullName": "Backbeat Internal Routes",
-        "signatureVersion": "s3",
+        "signatureVersion": "v4",
         "timestampFormat": "rfc822",
         "uid": "backbeat-2017-07-01"
     },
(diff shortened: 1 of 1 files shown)


francoisferrand (approved):
Looks ok, but not sure about the impact/risk of not signing anymore (or maybe it wasn't signed already, with old auth scheme?)

jonathan-gramain:
> Looks ok, but not sure about the impact/risk of not signing anymore (or maybe it wasn't signed already, with old auth scheme?)

I did a quick test using the old V2 signing and the new V4 with `v4-unsigned-body` auth type for a streamed request.

They both seem equivalent wrt. signature of the request body:

- in V2, the string to sign includes the provided `Content-MD5` header

- in V4, the string to sign both includes the provided `Content-MD5` header and a `X-Amz-Content-Sha256: UNSIGNED-PAYLOAD` header, which is equivalent to just including the `Content-MD5` header.

Likewise, in cases where the `Content-MD5` is **not** provided, neither of the methods signs the request body.

welansari (approved):

jonathan-gramain:
ping

PR #2649, opened 2025-05-22
Support CRR Location

welansari:
Issue: BB-680

### DIFF
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -26,7 +26,7 @@ const destinationAuthJoi = joi.object({
         port: joi.number().greater(0).optional(),
         adminPort: joi.number().greater(0).optional(),
         adminCredentialsFile: joi.string().optional(),
-    }).when('type', { is: 'role', then: joi.required() }),
+    }),
     sts: hostPortJoi
         .when('type', { is: 'assumeRole', then: joi.required() }),
 });
--- a/extensions/replication/ReplicationConfigValidator.js
+++ b/extensions/replication/ReplicationConfigValidator.js
@@ -1,7 +1,8 @@
 const fs = require('fs');
 const joi = require('joi');
 const { hostPortJoi, transportJoi, bootstrapListJoi, adminCredsJoi,
-        retryParamsJoi, probeServerJoi, probeServerPerSite } =
+        retryParamsJoi, probeServerJoi, probeServerPerSite, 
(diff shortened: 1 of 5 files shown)


extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -243,9 +243,9 @@ class QueueProcessor extends EventEmitter {
         }
         if (destConfig.transport === 'https') {
             this.destHTTPAgent = new HttpsAgent.Agent({
-                key: httpsConfig.key,
-                cert: httpsConfig.cert,
-                ca: httpsConfig.ca,
+                key: httpsConfig?.key,
+                cert: httpsConfig?.cert,
+                ca: httpsConfig?.ca,
```
francoisferrand, line 248: should we allow adding these parameters in the location definition?
(for now we would not support these in other type of location, but it would pave the way...)

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -243,9 +243,9 @@ class QueueProcessor extends EventEmitter {
         }
         if (destConfig.transport === 'https') {
             this.destHTTPAgent = new HttpsAgent.Agent({
```
francoisferrand, line 245: should we do this instead?
```suggestion
this.destHTTPAgent = new HttpsAgent.Agent(httpsConfig);
```

(requires that httpsAgent has been validated earlier)

francoisferrand (approved):

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -1,7 +1,8 @@
 const fs = require('fs');
 const joi = require('joi');
 const { hostPortJoi, transportJoi, bootstrapListJoi, adminCredsJoi,
-        retryParamsJoi, probeServerJoi, probeServerPerSite } =
+        retryParamsJoi, probeServerJoi, probeServerPerSite, 
+        stsConfigJoi} =
```
ghost:
```suggestion
        retryParamsJoi, probeServerJoi, probeServerPerSite, 
        stsConfigJoi } =
```

ghost (approved):

extensions/replication/queueProcessor/QueueProcessor.js
welansari: As discussed offline, we'll keep things simple for now and won't be adding this for now.

PR #2651, opened 2025-06-06
Fix NonCurrentVersionTransition not processed

francoisferrand:
- Fix filtering of supported rules
- Release backbeat 9.0.10

Issue: BB-691

### DIFF
--- a/extensions/lifecycle/util/rules.js
+++ b/extensions/lifecycle/util/rules.js
@@ -199,14 +199,15 @@ function rulesToParams(versioningStatus, currentDate, bucketLCRules, bucketData,
  * they correspond to how they are written in the lifecycle configuration.
  * @param {Object} lcConfig - Lifecycle configuration object
  * @param {string[]} lcConfig.supportedLifecycleRules - List of supported lifecycle rules
- * @returns {string[]} formatted supported lifecycle rules
+ * @returns {string[]} formatted supported lifecycle rules. These must match the field
+ * names in arsenal's LifecycleRuleData type
  */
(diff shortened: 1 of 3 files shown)


jonathan-gramain (approved):

jonathan-gramain:
@francoisferrand do we need this one for S3C as well?

FYI, I saw that Federation install failed when using backbeat 9.0.8 (current version in Federation `development/9` is 9.0.7), backbeat wasn't listening on its port, but I couldn't find logs showing what happened. At first I suspect that my change to use V4 signatures may not be compatible with cloudserver 7.70, although it's just an assumption. In any case, if we need to bump backbeat in Federation, we'll probably have to dig into this one as well, and maybe revert the incompatible change.

jonathan-gramain:
@francoisferrand sorry I didn't notice it was about transitions which are not supported in S3C, so we should be good :crossed_fingers:

francoisferrand:
> @francoisferrand do we need this one for S3C as well?

no, transitions are only supported in Zenko (S3C only support Expiration, which is not affected)

> FYI, I saw that Federation install failed when using backbeat 9.0.8 (current version in Federation `development/9` is 9.0.7), backbeat wasn't listening on its port, but I couldn't find logs showing what happened. At first I suspect that my change to use V4 signatures may not be compatible with cloudserver 7.70, although it's just an assumption. In any case, if we need to bump backbeat in Federation, we'll probably have to dig into this one as well, and maybe revert the incompatible change.

yes, @Kerkesni saw issues as well, and is working on it : https://scality.atlassian.net/browse/BB-689

PR #2652, opened 2025-06-08
use `{Key+}` in DeleteObjectFromExpiration URI to prevent signature mismatch

welansari:
The DeleteObjectFromExpiration operation was previously using `{Key}` in its requestUri path. This caused issues when object keys contained slashes, as `{Key}` only matches a single path segment. As a result, requests for objects with multi-segment keys would not match the intended route, leading to incorrect canonical request construction and AWS Signature V4 verification failures (signature mismatch errors).

This fix updates the requestUri to use `{Key+}` instead, ensuring that the full object key—including any slashes—is correctly captured and signed. This change aligns the route with S3 API conventions and prevents signature mismatch errors for keys with slashes.

Issue: BB-689

### DIFF
--- a/lib/clients/backbeat-2017-07-01.api.json
+++ b/lib/clients/backbeat-2017-07-01.api.json
@@ -194,7 +194,7 @@
         "DeleteObjectFromExpiration": {
             "http": {
                 "method": "DELETE",
-                "requestUri": "/_/backbeat/expiration/{Bucket}/{Key}"
+                "requestUri": "/_/backbeat/expiration/{Bucket}/{Key+}"
             },
             "input": {
                 "type": "structure",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "9.0.10",
+  "version": "9.0.11",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {


lib/clients/backbeat-2017-07-01.api.json
```diff
@@ -194,7 +194,7 @@
         "DeleteObjectFromExpiration": {
             "http": {
                 "method": "DELETE",
-                "requestUri": "/_/backbeat/expiration/{Bucket}/{Key}"
+                "requestUri": "/_/backbeat/expiration/{Bucket}/{Key+}"
```
ghost, line 197: (note: I don't have the full picture here, just asking to fully understand)
Should we exclude the `?` in case we have query parameters (like the versionID, or anything actually, so we are robust)?

benzekrimaha (approved):

francoisferrand (approved):

BourgoisMickael (approved):

lib/clients/backbeat-2017-07-01.api.json
welansari: For context, this is the definition of the BackbeatClient which is a "custom" S3 client that can target internal routes in Cloudserver.

The final URI as well as the signature are constructed by this custom client, so we control what we send. Dynamic query parameters (such as the versionId with this API) are specified in the "inputs" section and are automatically appended to the URI and taken into account when calculating the signature.

PR #2657, opened 2025-07-08
logReader: manually commit offset when using kafka logConsumer

welansari:
Kafka log consumer currently auto-commits messages after they are consumed. This is an issue as it means we loose messages after restart or crash even if their processing didn't finish.

Like we do with other logConsumer implementations, we now manually store offsets after the end of each batch. The auto-commit mechanism commits offsets stored locally after each 5 seconds.

Contrary to the BackbeatConsumer, there is no risk of offsets being committed in disorder as only one batch is processed at a time, and a partition is only assigned to a single instance of the QueuePopulator.

Issue: BB-698

### DIFF
--- a/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
+++ b/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
@@ -348,8 +348,8 @@ class LogConsumer extends EventEmitter {
                         // to initiate the rebalance process
                         // allowing the message processing to finish
                         // and offsets to be committed
-                        this._consumer.unsubscribe();
                         this.once(EVENTS.UNASSIGNED, () => next());
+                        this._consumer.unsubscribe();
                         return null;
                     }
                 }
@@ -357,8 +357,8 @@ class LogConsumer extends EventEmitter {
             },
             next => {
                 if (this._consumer?.isConnected()) {
-                    this._consumer.disconnect();
                     this._consumer.once('disconnected', () => next());
+                    this._consumer.disconnect();
                     return null;
                 }
                 return next();
--- a/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
+++ b/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
@@ -146,7 +146,7 @@ class LogConsumer extends EventEmitter {
                     return;
                 }
                 // wait for consumed messages to finish being processed
-                if (this._topicPartition && this._topicPartition.length > 0) {
+                if (this._hasUnprocessedMessages()) {
                     return;
                 }
                 // everything processed but not yet committed
@@ -201,11 +201,14 @@ class LogConsumer extends EventEmitter {
      */
     _consumeKafkaMessages(limit, cb) {
         this._resetRecordStream();
-        if (!this._consumer.isConnected()) {
-            this._log.info('Consumer is not ready, skipping message consumption', {
+        const isConnected = this._consumer.isConnected();
+        if (!isConnected || this._hasUnprocessedMessages()) {
+            this._log.info('Skipping message consumption', {
                 method: 'LogConsumer._consumeKafkaMessages',
                 topic: this._topic,
                 consumerGroupId: this._consumerGroupId,
+                reason: !isConnected ? 'consumer not connected' :
+                    'awaiting processing of previous batch',
             });
             return cb();
         }
@@ -264,7 +267,7 @@ class LogConsumer extends EventEmitter {
      * @returns {undefined}
      */
     storeOffsets() {
-        if (!this._topicPartition || this._topicPartition.length === 0) {
+        if (!this._hasUnprocessedMessages()) {
             this._log.debug('No offsets to store', {
                 method: 'LogConsumer.storeOffsets',
                 topic: this._topic,
@@ -280,6 +283,10 @@ class LogConsumer extends EventEmitter {
         });
         this._pendingCommit = true;
         this._topicPartition = null;
+    }
+
+    _hasUnprocessedMessages() {
+        return this._topicPartition?.length > 0;
     }
 
     /**
--- a/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
+++ b/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
@@ -2,9 +2,6 @@ const kafka = require('node-rdkafka');
 const async = require('async');
 const ListRecordStream = require('./ListRecordStream');
 
-// maximum time to wait for consumer group to rebalance (ms)
-const maxDelayRebalanceMs = 120000;
-
 class LogConsumer {
 
     /**
@@ -16,17 +13,20 @@ class LogConsumer {
      * @param {Logger} logger logger
      */
     constructor(kafkaConfig, logger) {
-        const { hosts, topic, consumerGroupId } = kafkaConfig;
+        const { hosts, topic, consumerGroupId, maxPollIntervalMs } = kafkaConfig;
         this._kafkaHosts = hosts;
+        this._maxPollIntervalMs = maxPollIntervalMs || 300000; // default to 5 minutes
         this._topic = topic;
         this._consumerGroupId = consumerGroupId;
         this._topicPartition = null;
+        this._pendingCommit = false;
         this._log = logger;
+        this._consumer = null;
     }
 
     /**
      * Get partition offsets
-     * Offsets are stored in kafka are not managed
+     * Offsets are stored in kafka and not managed
      * by the logReader, only keeping this function
      * to not alter the logic path in the logReader
      * @returns {null}
@@ -50,6 +50,7 @@ class LogConsumer {
             // Default auto-commit interval is 5 seconds
             'enable.auto.commit': true,
             'offset_commit_cb': this._onOffsetCommit.bind(this),
+            'rebalance_cb': this._onRebalance.bind(this),
             'metadata.broker.list': this._kafkaHosts,
             'group.id': this._consumerGroupId,
             'max.poll.interval.ms': this._maxPollIntervalMs,
@@ -63,33 +64,6 @@ class LogConsumer {
             this._consumer.subscribe([this._topic]);
             done();
         });
-    }
-
-    /**
-     * Waits for consumer group to rebalance
-     * @param {number} wait amount of time to wait already waited for, in milliseconds
-     * @param {Function} cb callback
-     * @returns {undefined}
-     */
-    _waitForAssignment(wait, cb) {
-        setTimeout(() => {
-            // assignements contain the partitions
-            // assigned to this consumer, they are only
-            // set once consumer group is balanced
-            const assignments = this._consumer.assignments();
-            if (assignments.length === 0) {
-                if (wait > maxDelayRebalanceMs) {
-                    this._log.error('Timeout waiting for consumer to be assigned to partitions', {
-                        method: 'LogConsumer._waitForAssignment',
-                        topic: this._topic,
-                        consumerGroupId: this._consumerGroupId,
-                    });
-                    return cb();
-                }
-                return this._waitForAssignment(wait + 2000, cb);
-            }
-            return cb();
-        }, 2000);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -20,17 +20,19 @@ class LogConsumer {
...
         this._consumerGroupId = consumerGroupId;
+        this._topicPartition = null;
         this._log = logger;
-        this._offsets = {};
     }
 
     /**
      * Get partition offsets
-     * @returns {string} stored partition offsets
+     * Offsets are stored in kafka are not managed
```
francoisferrand: typo?
```suggestion
     * Offsets are stored in kafka and not managed
```

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -42,13 +44,17 @@ class LogConsumer {
     setup(done) {
         // partition offsets will be managed by kafka
         const consumerParams = {
-            'enable.auto.offset.store': true,
+            // Manually manage storing offsets to ensure they are only stored
+            // after the batch processing is fully completed.
+            'enable.auto.offset.store': false,
+            // Default auto-commit interval is 5 seconds
+            'enable.auto.commit': true,
```
francoisferrand, line 61: with `auto-commit`, we will still keep a window (5 seconds) of uncertainty:
- is it the same with backbeatconsumer? I don't see the flag there...
- in backbeatConsume, I think we have code to ensure we "flush" the commited offset when needed : should we do the same here?

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -42,13 +44,17 @@ class LogConsumer {
     setup(done) {
         // partition offsets will be managed by kafka
         const consumerParams = {
-            'enable.auto.offset.store': true,
+            // Manually manage storing offsets to ensure they are only stored
+            // after the batch processing is fully completed.
```
francoisferrand, line 58: bonus issue: this will raise the issues we had (and fixed) in BackbeatConsumer around handling of rebalance, i.e. the need to delay the rebalance until the batch has been completed and offsets commited accordingly, and the need to detect/handle "slow tasks" which could block all further processing... It may be done in another PR (or maybe we could reuse BackbeatConsumer for KafkaLogConsumer?), but we probably can't go to prod without that kind of things...

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -155,16 +180,39 @@ class LogConsumer {
...
-                return next(null, { log: this._listRecordStream, tailable: false });
+        ], err =>{
+            if (err) {
+                return cb(err);
             }
-        ], (err, res) => cb(err, res[3]));
+            // ending and returning the stream
+            this._listRecordStream.end();
+            return cb(null, { log: this._listRecordStream, tailable: false });
+        });
```
francoisferrand: nit: i think the original flow was better, as it managed the callback (`cb`) in a single place, with no added condition (`if`) to handle error... or was there an issue? (we can also do `res?.[3]` if res could be null)

lib/queuePopulator/LogReader.js
```diff
@@ -712,25 +712,50 @@ class LogReader {
...
-                this._metricsHandler.logReadOffset(this.getMetricLabels(), this.logOffset);
-                this._metricsHandler.logSize(this.getMetricLabels(), logSize);
+                this._updateMetrics(logRes.info);
             }
 
             return this._writeLogOffset(logger, done);
         }
+
+        // Handle offset managed externally (e.g., Kafka)
+        if (!this.isOffsetManaged() && this.logConsumer.storeOffsets) {
```
francoisferrand: When `!this.isOffsetManaged()`, then `storeOffsets` should always be defined, right?
If this is the case, then this option is redundant (an assertion in constructor is enough)

This would allow a much nicer flow:
```
    if (!this.isOffsetManaged()) {
        this.logConsumer.storeOffsets();
    } else if (this._shouldUpdateLogOffset(logRes, nextLogOffset)) {
        ...
    }
```

welansari: > When !this.isOffsetManaged(), then storeOffsets should always be defined, right?

Currently yes, the only logReader for which we don't manage offsets is the Kafka logReader.

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
welansari: > is it the same with backbeatconsumer? I don't see the flag there...

Yes, we have the same in `BackbeatConsumer`, the option is enabled by default i just decided to included it here for clarity.

> in backbeatConsume, I think we have code to ensure we "flush" the commited offset when needed : should we do the same here?

Yes, i should handle rebalance to avoid replay cases

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -136,8 +219,20 @@ class LogConsumer {
...
+                topicPartition.set(`${message.topic}-${message.partition}`, {
+                    topic: message.topic,
+                    partition: message.partition,
+                    offset: message.offset + 1, // next offset to commit
+                });
+                // writing consumed messages to the stream
+                this._listRecordStream.write(message);
+            });
+            // format offsets for commit
+            this._topicPartition = Array.from(topicPartition.values());
```
francoisferrand, line 273: Should we not `merge` this map with the existing `this._topicPartition` ?
e.g. in `_consumeKafkaMessages` is called multiple times before `storeOffsets()` is called

(maybe this does not happen, but it is not clear from the prototype of LogConsumer that it MUST work this way...)

In that case, may be better to keep a map in LogConsumer, and build the array only as a temporary varialbe in storeOffsets

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -42,70 +46,141 @@ class LogConsumer {
...
+     * @param {kafka.KafkaError} err Rebalance event
+     * @param {TopicPartition[]} assignment List of (un)assigned partitions
+     * @returns {void}
      */
-    _storeCurrentOffsets(cb) {
-        this._consumer.committed(5000, (err, topicPartitions) => {
-            if (err) {
-                this._log.error('Error while getting offsets', {
-                    method: 'LogConsumer._storeCurrentOffsets',
+    _onRebalance(err, assignment) {
```
francoisferrand, line 119: should this update metrics: `KafkaBacklogMetrics.onRebalance(this._topic, this._groupId, status)` ?
(or is that metrics specific to BackbeatConsumer)

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -42,70 +46,141 @@ class LogConsumer {
...
+                        method: 'LogConsumer._onRebalance',
+                        topic: this._topic,
+                        consumerGroupId: this._consumerGroupId,
+                    });
+                    clearInterval(rebalanceTimeoutInterval);
+                    this._consumer.disconnect();
+                    return;
+                }
+                // wait for consumed messages to finish being processed
+                if (this._topicPartition && this._topicPartition.length > 0) {
```
francoisferrand: instead of polling, can't we trigger this callback when the topic gets empty?
similar to BackbeatConsumer, we could create a `drain` callback which is called when `_topicPartition` is reset ; and set it when needed in rebalance.

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -153,18 +248,80 @@ class LogConsumer {
...
+                        return null;
+                    }
+                }
+                return next();
+            },
+            next => {
+                if (this._consumer?.isConnected()) {
+                    this._consumer.disconnect();
+                    this._consumer.once('disconnected', () => next());
+                    return null;
```
francoisferrand: should we not do it the other way, to avoid race conditions?
(node is thread safe, so probably fine as is... but could change with futurechanges...)
```suggestion                  
                    this._consumer.once('disconnected', () => next());
                    this._consumer.disconnect();
                    return null;
```

(unless the semantics is that setting `once('disconnected')` callback will immediately invoke the callback if the consumer is alreadyd disconnected)

lib/queuePopulator/LogReader.js
```diff
@@ -111,10 +111,18 @@ class LogReader {
      * @return {undefined}
      */
     close(cb) {
-        async.each(
-            Object.values(this._producers),
-            (producer, done) => producer.close(done),
-            cb);
+        async.parallel([
```
francoisferrand: should we not first (in series!) stop the consumer -waiting for all 'current' messages to have been consumer- then stop the producers?

tests/unit/lib/queuePopulator/LogReader.spec.js
```diff
@@ -599,30 +599,34 @@ describe('LogReader', () => {
...
                 shouldSave: false,
             },
             {
                 desc: 'should not save the offset if the offset did not change',
                 managed: true,
+                hasStoreOffsetsFn: false,
                 currentOffset: 0,
                 nextLogOffset: 0,
                 shouldSave: false,
             },
```
francoisferrand, line 622: should we really remove that test:
- that would mean we changed the behavior, which seems weird?
- the other tests seem to have similar conditions ("offsets not managed"), so this not a corner case we dropped
- if we remove it from here, we should probably create a new one (or multiple, if offset are/aren't managed)

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
welansari: The queuePopulator doesn't support parallel processing of batches, only a single batch of messages is consumed, processed and then committed at a time. I don't see a need to support that currently.

I enforced this by not allowing consumption of new messages when a batch is in progress.

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
welansari: No the metric is not specific to BackbeatConsumer, i added it

tests/unit/lib/queuePopulator/LogReader.spec.js
welansari: The behaviour did change for the non managed logConsumer (Kafka) and it's not weird, the whole point of this PR is to now handle how/when the offsets get committed. Before, we used to do nothing and the offsets would be committed as soon as messages were consumed, now we do it manually after each batch of messages finished processing.

francoisferrand (approved):

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -42,70 +49,168 @@ class LogConsumer {
...
+            if (err.code === kafka.CODES.ERRORS.ERR__NO_OFFSET) {
+                return undefined;
             }
-            return cb();
-        }, 2000);
+            this._log.error('Error committing offsets to kafka', {
+                method: 'LogConsumer._onOffsetCommit',
+                errorCode: err,
+                topicPartitions,
+                groupId: this._groupId
```
ghost: I'm not sure this is defined in this class?

```suggestion
                groupId: this._consumerGroupId
```

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -42,70 +49,168 @@ class LogConsumer {
...
+            return undefined;
+        }
+        this._pendingCommit = false;
+        if (!this._hasUnprocessedMessages()) {
+            this.emit('drained');
+        }
+        this._log.debug('Offsets committed', {
+            method: 'LogConsumer._onOffsetCommit',
+            topicPartitions,
+            groupId: this._groupId
```
ghost:
```suggestion
            groupId: this._consumerGroupId
```

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -42,70 +49,168 @@ class LogConsumer {
...
+                method: 'LogConsumer._onOffsetCommit',
+                errorCode: err,
+                topicPartitions,
+                groupId: this._groupId
+            });
+            return undefined;
+        }
+        this._pendingCommit = false;
+        if (!this._hasUnprocessedMessages()) {
+            this.emit('drained');
```
ghost:
```suggestion
            this.emit(constants.unassignStatus.DRAINED);
```

(same at other places)

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -153,18 +281,84 @@ class LogConsumer {
...
+            method: 'LogConsumer.storeOffsets',
+            consumerGroupId: this._consumerGroupId,
+            topicPartition: this._topicPartition,
+        });
+        this._pendingCommit = true;
+        this._topicPartition = null;
+    }
+
+    _hasUnprocessedMessages() {
+        return this._topicPartition && this._topicPartition.length > 0;
```
ghost:
```suggestion
        return this._topicPartition?.length > 0;
```

lib/queuePopulator/LogReader.js
```diff
@@ -777,6 +807,13 @@ class LogReader {
         return this._batchTimedOut;
     }
 
+    isLogConsumerReady() {
+        if (this.logConsumer?.isReady) {
+            return this.logConsumer.isReady();
+        }
+        return true;
```
ghost:
```suggestion
        return this.logConsumer?.isReady?.() ?? true;
```

(just a personal preference, not blocking)

ghost (approved):
Nothing blocking to me, the first comment should probably be addressed though, as I think it will always log undefined otherwise

PR #2658, opened 2025-07-09
fix notification cached buckets count

SylvainSenechal:
Issue : BB-692

### DIFF
--- a/extensions/notification/configManager/MongoConfigManager.js
+++ b/extensions/notification/configManager/MongoConfigManager.js
@@ -46,15 +46,11 @@ const cachedBuckets = ZenkoMetrics.createGauge({
     help: 'Total number of cached buckets in the notification config manager',
 });
 
-function onConfigManagerCacheUpdate(op) {
+function onConfigManagerCacheUpdate(op, nbCachedBuckets) {
     cacheUpdates.inc({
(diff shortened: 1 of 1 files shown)


extensions/notification/configManager/MongoConfigManager.js
```diff
@@ -46,15 +46,11 @@ const cachedBuckets = ZenkoMetrics.createGauge({
...
+function onConfigManagerCacheUpdate(op, cacheInstance) {
     cacheUpdates.inc({
         [CONFIG_MANAGER_OPERATION_TYPE]: op,
     });
-    if (op === 'add') {
-        cachedBuckets.inc({});
-    } else if (op === 'delete') {
-        cachedBuckets.dec({});
-    }
+    cachedBuckets.set({}, cacheInstance.count());
```
welansari: Maybe better to pass the count directly for better separation of concern ? `onConfigManagerCacheUpdate` is just a helper and isn't supposed to know about the classes we use internally.

SylvainSenechal: yeah let's do this

welansari (approved):

benzekrimaha (approved):

PR #2659, opened 2025-07-15
BB-694: Update auth config schema

tmacro:
Extends the existing schema to support basic auth and stricter checking of kerberos.

### DIFF
--- a/extensions/notification/NotificationConfigValidator.js
+++ b/extensions/notification/NotificationConfigValidator.js
@@ -1,18 +1,36 @@
 const joi = require('joi');
 const { probeServerJoi } = require('../../lib/config/configItems.joi');
 
-const authSchema = joi.object({
-    type: joi.string(),
-    ssl: joi.boolean(),
-    protocol: joi.string(),
+const sslSchema = joi.object({
+    ssl: joi.boolean().default(false),
     ca: joi.string(),
     client: joi.string(),
     key: joi.string(),
     keyPassword: joi.string(),
-    keytab: joi.string(),
-    principal: joi.string(),
-    serviceName: joi.string(),
 });
+
+const saslAuthSchema = sslSchema.append({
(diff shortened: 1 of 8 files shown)


tcarmet (approved):

tests/unit/notification/NotificationConfigValidator.js
```diff
@@ -1,81 +1,236 @@
...
-            monitorNotificationFailures: true,
-            notificationFailedTopic: 'failed-topic',
-            queueProcessor: {
-                groupId: 'groupId',
-                concurrency: 1000,
-            },
-            destinations: [{
+    const testCases = [
+        {
+            valid:false,
```
jonathan-gramain:
```suggestion
            valid: false,
```

extensions/notification/NotificationConfigValidator.js
```diff
@@ -1,27 +1,45 @@
 const joi = require('joi');
 const { probeServerJoi } = require('../../lib/config/configItems.joi');
 
-const authSchema = joi.object({
-    type: joi.string(),
-    ssl: joi.boolean(),
-    protocol: joi.string(),
+const sslSchema = joi.object({
+    ssl: joi.boolean().default(false),
```
jonathan-gramain, line 5: Would having `type: "ssl"` be more consistent with the other auth types?

jonathan-gramain (approved):

BourgoisMickael (approved):

nicolas2bert (approved):

extensions/notification/NotificationConfigValidator.js
tmacro: The ssl auth isn't really a distinct type, it can be used with kerberos and basic auth as well. My feeling is most deploys using kerberos and basic auth will use at least `ca` and `key` so the credentials aren't sent over the wire unencrypted.

PR #2663, opened 2025-07-29
add old location var to be able to clear location MD

benzekrimaha:
Issue: BB-699

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
+++ b/extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
@@ -185,6 +185,8 @@ class LifecycleUpdateExpirationTask extends BackbeatTask {
                     coldLocation, startTime - restoreExpirationDate);
 
                 // Reset archive flags to no longer show it as restored
+                const oldLocation = objMD.getLocation();
+                objMD.setLocation();
                 objMD.setArchive({
                     archiveInfo: archive.archiveInfo,
                 });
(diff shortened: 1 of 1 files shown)


extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
```diff
@@ -181,6 +182,8 @@ class LifecycleUpdateExpirationTask extends BackbeatTask {
                     coldLocation, startTime - restoreExpirationDate);
 
                 // Reset archive flags to no longer show it as restored
+                oldLocation = objMD.getLocation();
```
welansari:
```suggestion
                const oldLocation = objMD.getLocation();
```
`oldLocation` is a constant, we can just define it here

welansari (approved):

francoisferrand (approved):

extensions/lifecycle/tasks/LifecycleUpdateExpirationTask.js
francoisferrand: the variable is not defined here, but at the beginning of the function : so that it can be set in this first step of the waterflow, and used in the next one...

The way to use a `const` would be to make the variable local indeed, remove the early definition (line 149), and change the `next` call line 198:

```
                    next(err, objMD, oldLocation);
                });
            },
            (objMD, oldLocation, next) => this._garbageCollectLocation(
                entry, oldLocation, log, next,
            ),
```

PR #2667, opened 2025-08-01
BB-705: Bump rdkafka to fix DEP0048 util.isError()

BourgoisMickael:
This deprecation warning spams stderr logs

Fix done in this release: https://github.com/Blizzard/node-rdkafka/releases/tag/v3.3.1

Going from 2.18.0 to 3.4.1 updates librdkafka from 2.3.0 to 2.10.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "minimatch": "^10.0.1",
     "mongodb": "^6.11.0",
     "node-forge": "^1.3.1",
-    "node-rdkafka": "^2.12.0",
+    "node-rdkafka": "^3.6.0",
     "node-rdkafka-prometheus": "^1.0.0",
     "node-schedule": "^2.1.1",
     "node-zookeeper-client": "^1.1.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "minimatch": "^10.0.1",
     "mongodb": "^6.11.0",
     "node-forge": "^1.3.1",
-    "node-rdkafka": "^2.12.0",
+    "node-rdkafka": "^3.4.1",
```
ghost: Could we test a zenko build with this change early, as this is a major version bump?

BourgoisMickael: The major is only because they don't support old nodejs version.
But right now tests here are not passing so it seems to break something

BourgoisMickael:
For some reason it spams these debug messages:

```json
{"name":"BackbeatConsumer","time":1754069124682,"topic":"backbeat-func-test-repstatus",
 "groupId":"backbeat-func-test-group-id","level":"debug",
 "message":"not enough message available yet, retry in 1s: 0",
 "hostname":"pkrvmjbmru5nbw0","pid":8418}

{"name":"BackbeatConsumer","time":1754069124682,"topic":"backbeat-func-test-repstatus",
 "groupId":"backbeat-func-test-group-id","level":"debug",
 "message":"error, retry in 1s: KafkaConsumer is not connected",
 "hostname":"pkrvmjbmru5nbw0","pid":8418}
```

francoisferrand:
bumping librdkafka (especially here, with a jump to a new major version of the node package and 7 minors of the native lib) is risky, we should only do it in a newer Backbeat branch → c.f. https://github.com/scality/backbeat/pull/2728

Can we close this PR?

BourgoisMickael:
> bumping librdkafka (especially here, with a jump to a new major version of the node package and 7 minors of the native lib) is risky, we should only do it in a newer Backbeat branch → c.f. #2728
> 
> Can we close this PR?

Yes we can close

PR #2668, opened 2025-08-12
Fix conductor always triggering v1 listing when using bucketd

welansari:
bucketd and zookeeper listing don't include the property `isLifecycled` in the task object.

Issue: BB-709


tests/functional/lifecycle/LifecycleConductor.spec.js
francoisferrand, line 1: Many (all?) tests updated here : are we covering all nominal cases (v1, v2....) in functional tests?

francoisferrand (approved):

tests/functional/lifecycle/LifecycleConductor.spec.js
welansari: added tests for v1 listing

nicolas2bert (approved):

PR #2672, opened 2025-08-26
add replicateObject policy to the bucket processor

welansari:
The bucket processor needs extra permissions to use backbeat routes such as GetMetadata

Issue: BB-712

### DIFF
--- a/extensions/lifecycle/bucketProcessor/policy.json
+++ b/extensions/lifecycle/bucketProcessor/policy.json
@@ -13,7 +13,8 @@
                 "s3:GetObjectTagging",
                 "s3:GetObjectVersionTagging",
                 "s3:GetObject",
-                "s3:GetObjectVersion"
+                "s3:GetObjectVersion",
+                "s3:ReplicateObject"
             ],
             "Resource": [
                 "*"


DarkIsDude (approved):

SylvainSenechal (approved):

extensions/lifecycle/bucketProcessor/policy.json
```diff
@@ -13,7 +13,8 @@
                 "s3:GetObjectTagging",
                 "s3:GetObjectVersionTagging",
                 "s3:GetObject",
-                "s3:GetObjectVersion"
+                "s3:GetObjectVersion",
+                "s3:ReplicateObject"
```
francoisferrand, line 17: It seems weird that `ReplicateObject` permission is needed to update the metadata of the object, where all that bucketProcessor needs is to update a few states...
→ this should probably be reviewed eventually, to have separate routes and/or permissions for different fields (e.g. bucket processor should not be able to create a new version or modify the 'location' of the data, though it may modify some fields : cold storage status, ...) : can you please create a followup ticket about making this permissions cleaner & more granular?

PR #2674, opened 2025-09-01
pass proper params to the GC

welansari:
The deleteData API expects the source attribute to contain the following fields:
- bucket
- objectKey
- storageClass
- lastModified (when deleting conditionally)

Some of these fields are never passed and cause a
SignatureDoesNotMatch error, leading to data never getting removed from disk.

Issue: BB-713

### DIFF
--- a/tests/unit/lifecycle/LifecycleTask.spec.js
+++ b/tests/unit/lifecycle/LifecycleTask.spec.js
@@ -1686,12 +1686,12 @@ describe('lifecycle task helper methods', () => {
             lct2.reset();
         });
 
-        it('should popuplate and return null if heap has space', () => {
+        it('should populate and return null if heap has space', () => {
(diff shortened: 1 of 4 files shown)


tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -2190,4 +2190,156 @@ describe('lifecycle task helper methods', () => {
             });
         });
     });
+
+    describe('_getTransitionActionEntry', () => {
+        let lct2;
```
DarkIsDude: Can we find better than `2` as a description ? Also this variable is used a lot, doing acronym is not a good practice to help people to read your code. `lifecycleTaskMockForTransitionActionEntry` seems better ?

DarkIsDude (approved):

francoisferrand (approved):

PR #2678, opened 2025-10-01
Create backbeat client with smithy in preparation of aws sdk v3 migration 

SylvainSenechal:
This one only creates the new client, but not in use yet in backbeat
Issue: BB-715


lib/clients/smithy/models/multipleBackendDeleteObjectTagging.smithy
```diff
@@ -0,0 +1,52 @@
...
+    SourceBucket: String,
+    
+    @httpHeader("X-Scal-Source-Version-Id")
+    SourceVersionId: String,
+    
+    @httpHeader("X-Scal-Replication-Endpoint-Site")
+    ReplicationEndpointSite: String,
+    
+    /// TODO : REVIEW THIS FIELD
+    @httpPayload
```
SylvainSenechal, line 43: Not sure I wanna keep this field : I have to add 
@suppress(["HttpMethodSemantics.UnexpectedPayload"]) 
or else smithy doesn't want to build (not expecting this kind of payload for a delete operation)

And when I look at the way we use this api in the code, I see that the field is unused

lib/clients/smithy/localTests/initBucketForTests.ts
SylvainSenechal, line 1: I understand we may not want to keep this test folder, but it's been quite helpful and I think it could be useful for the next who will have to modify the client

PR #2680, opened 2025-10-09
Prevent crash when processing a paused location

welansari:
A race condition could occur in `IngestionPopulator` if a location is paused while its `logReader` is being processed. This pause action removes the `logReader` from the active list.

If the asynchronous processing completes and attempts a recursive call to `_processLogReaderEntries`, it could do so with a `logReader` that is now `undefined`, causing a crash.

Issue: BB-716

### DIFF
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -770,6 +770,9 @@ class IngestionPopulator {
     }
 
     _processLogReaderEntries(logReader, params, cb) {
+        if (!logReader) {
+            return process.nextTick(cb);
+        }
         // skipping if circuit breaker tripped
         if (this._circuitBreaker.state !== BreakerState.Nominal) {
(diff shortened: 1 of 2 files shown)


ghost (approved):

lib/queuePopulator/IngestionPopulator.js
SylvainSenechal, line 1: Is the logic line 792 "skipping if location is paused" likely to never be run because of this check ?

SylvainSenechal (approved):

lib/queuePopulator/IngestionPopulator.js
welansari: When the `raftlog` has more entries than what one pass can handle (this is especially the case when performing the initial listing) we immediately continue the listing without waiting for the cron, you can see that being done at the end of the `_processLogReaderEntries` method, in this case the next call receives the still alive reference of the logReader used in the initial processing, so we do still need to check if it's paused.

PR #2681, opened 2025-10-17
Fix alert when expiring delete-markers

francoisferrand:
EODM happens when a DeleteMarker is the only remaining revision of an object.

When expiring it, either because ExpiredObjectDeleteMarker is set in the rule or because it matches the age criteria, the transitionTime is thus not intrinsic to the DeleteMarker but really depends on the time the other revisions were removed.

We don't have this information, so being precise is not possible here; however we must not use the DeleteMarker's LastModified date: it creates seemingly very slow expiration latency metrics (since the DeleteMarker may be arbitrary old, even though expiration is possible only for a very short time) which triggers alerts about slow lifecycle, even though it is working as expected.

In addition, depending on the exact rule, there may not be an age limit in that specific rule, which causes getTransitionTimestamp() to return null, which prevents collecting the metrics properly.

Consequently, we now just use the current time in that case, to avoid making metrics suspicious and triggering alert.

Issue: BB-719

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -1425,19 +1425,14 @@ class LifecycleTask extends BackbeatTask {
                     new Date(deleteMarker.LastModified)
                 );
 
-                const eodm = rules.Expiration &&
-                    rules.Expiration.ExpiredObjectDeleteMarker;
-
                 // Backbeat performs automatic ExpiredObjectDeleteMarker cleanup
                 // for compatibility with Amazon S3,
                 // - either when the delete markers meet the age criteria
                 // - or when the ExpiredObjectDeleteMarker tag is set to true.
                 const applicableExpRule = rules.Expiration && (
-                    (rules.Expiration.Days !== undefined
-                     && daysSinceInitiated >= rules.Expiration.Days)
-                    || (rules.Expiration.Date !== undefined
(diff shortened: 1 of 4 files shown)


extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1520,6 +1516,7 @@ class LifecycleTask extends BackbeatTask {
                 .setAttribute('target.key', verToExpire.Key)
                 .setAttribute('target.version', verToExpire.VersionId)
                 .setAttribute('details.dataStoreName', verToExpire.StorageClass || '')
+                // details.lastModified is not set for NCVE...
```
DarkIsDude, line 1519: What is `NCVE` ? Non current version expiration ?

DarkIsDude (approved):
Same do we have tests ?

extensions/lifecycle/tasks/LifecycleTask.js
SylvainSenechal: Guess so, but is this comment a TODO ?

SylvainSenechal (approved):

extensions/lifecycle/tasks/LifecycleTask.js
francoisferrand: no, it just (tries to) make it explicit we do not set `details.lastModified` in this specific case, contrary to all other similar places where we send a message to object processor

PR #2682, opened 2025-10-17
Use special location when expiring delete marker

francoisferrand:
When expiring a delete marker, the value in the metrics used to be `undefined`. Use a special marker instead when expiring a delete marker, to make it easier to analyze.

Issue: BB-720


DarkIsDude (approved):
No tests ?

SylvainSenechal (approved):

PR #2683, opened 2025-10-17
Ensure we don't try to transition deleteMarkers

francoisferrand:
Issue: BB-722

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -1653,7 +1653,7 @@ class LifecycleTask extends BackbeatTask {
             return done();
         }
 
-        if (rules.NoncurrentVersionTransition) {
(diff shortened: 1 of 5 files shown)


DarkIsDude (approved):

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1653,7 +1653,7 @@ class LifecycleTask extends BackbeatTask {
             return done();
         }
 
-        if (rules.NoncurrentVersionTransition) {
+        if (rules.NoncurrentVersionTransition && !this._isDeleteMarker(version)) {
```
SylvainSenechal, line 1656: Should this be placed at the top of the _checkAndApplyNCVTransitionRule function ?

SylvainSenechal: Same comment for the other modifications

SylvainSenechal (approved):

PR #2684, opened 2025-10-21
BB-724: Use originOp from overheadFields

BourgoisMickael:
For lifecycle expiration on metadata

### DIFF
--- a/extensions/notification/NotificationQueuePopulator.js
+++ b/extensions/notification/NotificationQueuePopulator.js
@@ -223,7 +223,9 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
                 return undefined;
             }
             const { eventMessageProperty, deleteEvent } = notifConstants;
-            let eventType = value[eventMessageProperty.eventType];
+            // Delete can have an overheadField originOp that should be used for lifecycle expiration (on metadata)
+            // This takes precedence over the originOp from the value object
+            let eventType = overheadFields?.[eventMessageProperty.eventType] || value[eventMessageProperty.eventType];
             if (eventType === undefined && type === 'del') {
                 eventType = deleteEvent;
             }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "9.0.17",
+  "version": "9.0.18",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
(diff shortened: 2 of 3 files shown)


tmacro (approved):

extensions/notification/NotificationQueuePopulator.js
```diff
@@ -227,6 +227,10 @@ class NotificationQueuePopulator extends QueuePopulatorExtension {
             if (eventType === undefined && type === 'del') {
                 eventType = deleteEvent;
             }
+            // Delete can have an overheadField originOp that should be used for lifecycle expiration (on metadata)
+            if (type === 'del' && overheadFields?.[eventMessageProperty.eventType]) {
```
jonathan-gramain: Do we need the `type === 'del'` check? It may not hurt to just use the `eventType` field when set, whatever the request type.

Also, on the style: it can be easier to read if we order the setters of the `eventType` by decreasing priority, rather than "last write wins" style:

```
if (<present in overhead>) {
    eventType = <overhead>
} else if (type === 'del') {
    eventType = deleteEvent
} else {
    eventType = <message property>
}
```

BourgoisMickael: I don't know the format of overhead for other kind of operation and in other backend than metadata, like mongoDB. `originOp` could be present for other reason, here I want to specifically target delete operations without impacting other operations.

I've changed the condition a little

ghost (approved):

extensions/notification/NotificationQueuePopulator.js
jonathan-gramain: I may be wrong but I think the `originOp` was introduced for bucket notifications purpose in the first place, in which case it can make sense to just forward it to the notification message, without extra filtering. It makes the code easier to maintain and avoids having to track down where to change things when we want to add new event types to handle.

Even in the case where `originOp` wasn't added specifically for BN, it can still be worth checking if it can be blindly forwarded, as it can help having a cleaner design.

BourgoisMickael: I reworked the condition to use overheadFIelds without condtion on delete. [`f55a33c` (#2684)](https://github.com/scality/backbeat/pull/2684/commits/f55a33cd2db37fd10a6acdf2dd7a9fc9be240c35)

jonathan-gramain (approved):

PR #2687, opened 2025-11-14
BB-727: Fix ZookeeperConfigManager event listener leaks

BourgoisMickael:
- improve some logs
- warn early for event listener leaks
- avoid send a remove to zk when a watcher notify the removed bucket config
- separate list and get configs

Below a sequence diagram showing in red a few problems fixed here

bucket-notif-zookeeper drawio


nicolas2bert (approved):

BourgoisMickael:
@bert-e reset

PR #2689, opened 2025-12-01
Improvement/bb 728/filter out internal buckets

fredmnl:
We have added a new `internal..backupIndex` internal bucket for repd internal information, it is being tracked by the MD oplog. Backbeat doesn't need to track these buckets since they are not related to user traffic. 

We extend the internal bucket concept to all buckets containing the `..` pattern so we can re-use this pattern in the future.

### DIFF
--- a/extensions/replication/ReplicationQueuePopulator.js
+++ b/extensions/replication/ReplicationQueuePopulator.js
@@ -18,8 +18,13 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
             // bucket updates have no key in raft log
             return undefined;
         }
+        // users..bucket has a special role for "echo mode"
         if (entry.bucket === usersBucket) {
             return this._filterBucketOp(entry);
+        }
+        // internal buckets (other than users..bucket) are ignored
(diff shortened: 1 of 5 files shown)


extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -21,6 +21,9 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
         if (entry.bucket === usersBucket) {
             return this._filterBucketOp(entry);
         }
+        if (entry.bucket.includes('..')) {
```
nicolas2bert, line 26: same comment as for notification can be helpful:
`// ignore internal buckets`

nicolas2bert: NIT: if we don’t want the comment and want to avoid duplicating logic, we can introduce a clear central function: isInternalBucket(bucketName)

nicolas2bert (approved):

dvasilas (approved):

fredmnl:
hello

PR #2691, opened 2025-12-04
BB-731 upgrade kafka version used on tests

tcarmet:
Since we are upgrading kafka at the product level, doing some housekeeping to keep the version used in the tests up to date, a bit more synced with the product as this one was very outdated.

Downloading the kafka package directly instead of using the kafka image the 3.9.0 version of it does not support zookeeper anymore.

And since spotify/kafka used to come with zookeeper + kafka, we are now executing both separately.

### DIFF
--- a/.github/dockerfiles/ft/docker-compose.yaml
+++ b/.github/dockerfiles/ft/docker-compose.yaml
@@ -51,12 +51,24 @@ services:
     image: ${REDIS_IMAGE}
     ports:
       - 6379:6379
+  zookeeper:
+    profiles: ['s3c']
+    image: zookeeper:3.9.4
+    ports:
+      - 2181:2181
+    environment:
+      ZOOKEEPER_CLIENT_PORT: 2181
+      ZOOKEEPER_TICK_TIME: 2000
   kafka:
     profiles: ['s3c']
     image: ${KAFKA_IMAGE}
+    depends_on:
+      - zookeeper
     ports:
       - 9092:9092
-      - 2181:2181
     environment:
-      ADVERTISED_HOST: "localhost"
-      ADVERTISED_PORT: 9092
+      KAFKA_BROKER_ID: 1
+      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:9092
+      KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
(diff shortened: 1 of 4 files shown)


jonathan-gramain (approved):

.github/dockerfiles/ft/docker-compose.yaml
```diff
@@ -51,12 +51,24 @@ services:
     image: ${REDIS_IMAGE}
     ports:
       - 6379:6379
+  zookeeper:
+    profiles: ['s3c']
+    image: zookeeper:3.9.3
```
anurag4DSB: Does Kafka 3.9 support ZK migration in Zookeeper 3.9.x?
I am thinking if there are features not supported in 3.9 ZK which were there in 3.8 zk and needed for ZK migration by Kafka 3.9

[Reference](https://kafka.apache.org/blog#apache_kafka_390_release_announcement)

> Upgrade to Kafka 3.4 (a newer version will not work due to ZK incompatibilites)
Upgrade to ZooKeeper 3.8
Upgrade to Kafka 3.9.
Perform ZK migration.
Upgrade to Kafka 4.0.

tcarmet: Good catch indeed, I think we should technically be picking up the 3.8 cause that's what officially stated and tested. But I went to double check on federation side, we actually use `3.9.4` there (and shipped it).

So now I wonder, what should be the strategy here? Cause runtime wise, I think we are fine. The issue come with the zk migration indeed, which may actually work (it's just not officially supported, but I don't think they dropped and broke everything in a single digit minor bump).

Here's what I'm thinking, we keep the 3.9 of zookeeper because that's what federation is using, so that it forces us to check with that version when the times come to play around zk migration. I want to avoid a downgrade because looking at the [commits](https://github.com/scality/Federation/commit/06fe928ecce1a24186a228805572e2d74ea25e74), we bumped zookeeper for CVE fixes, it wouldn't send the right message if they come back and downgrading a component may come with bigger risks than performing the migration in a higher version.

I'm also thinking about spending some time next week setting up a env with kraft and zookeeper 3.9.4 in dual-write mode to see how kraft behaves, if it's happy or not.

Let me know what you think, if it's worth spending some effort this way, or if we should rather downgrade zookeeper for this kafka upgrade.

tcarmet: Took kraft for a spin with the same version of zookeeper we use in federation `3.9.4` and was able to play around the migration steps required to remove zookeeper from the picture.

I think it's best to set zookeeper to `3.9.4` here, downgrading sounds more risky to me, and when the time comes to take the migration more seriously, we force ourselves to go against the same version we use in federation.

anurag4DSB (approved):

PR #2692, opened 2025-12-18

delthas:

Issue: BB-732


delthas:
Moved to #2693

PR #2693, opened 2025-12-18

delthas:

Issue: BB-732

### DIFF
--- a/extensions/notification/NotificationConfigValidator.js
+++ b/extensions/notification/NotificationConfigValidator.js
@@ -21,10 +21,19 @@ const kerberosAuthSchema = saslAuthSchema.append({
     serviceName: joi.string().required(),
 });
 
-const basicAuthSchema = saslAuthSchema.append({
+const basicAuthBaseSchema = saslAuthSchema.append({
     type: joi.string().valid('basic').required(),
-    credentialsFile: joi.string().required(),
 });
+
+const basicAuthSchema = joi.alternatives().try(
+    basicAuthBaseSchema.append({
+        credentialsFile: joi.string().required(),
+    }),
+    basicAuthBaseSchema.append({
+        username: joi.string().required(),
+        password: joi.string().required(),
+    }),
+);
 
 const credentialsFileSchema = joi.object({
     username: joi.string().required(),
--- a/extensions/notification/queueProcessor/task.js
+++ b/extensions/notification/queueProcessor/task.js
@@ -38,7 +38,11 @@ let destinationAuth = {
     keytab: process.env.KEYTAB,
     principal: process.env.PRINCIPAL,
     serviceName: process.env.SERVICE_NAME,
+    username: process.env.BASIC_USERNAME,
+    password: process.env.BASIC_PASSWORD,
 };
+// Drop undefined environment variables to prevent hitting unknown fields when validating the auth schema.
+Object.keys(destinationAuth).forEach(key => destinationAuth[key] === undefined ? delete destinationAuth[key] : {});
 const isDestinationAuthEmpty = Object.values(destinationAuth)
     .every(x => !x);
 if (isDestinationAuthEmpty) {
--- a/extensions/notification/utils/auth.js
+++ b/extensions/notification/utils/auth.js
@@ -85,19 +85,26 @@ function generateKafkaAuthObject(auth) {
             break;
         }
         case 'basic': {
-            const { protocol, credentialsFile } = auth;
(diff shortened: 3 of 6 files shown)


extensions/notification/NotificationConfigValidator.js
```diff
@@ -25,8 +25,10 @@ const kerberosAuthSchema = saslAuthSchema.append({
 
 const basicAuthSchema = saslAuthSchema.append({
     type: joi.string().valid('basic').required(),
-    credentialsFile: joi.string().required(),
-});
+    credentialsFile: joi.string(),
+    username: joi.string(),
+    password: joi.string(),
+}).or('credentialsFile', 'username').with('username', 'password');
```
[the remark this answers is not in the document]
delthas: No, I don't care about a standalone password.

extensions/notification/utils/auth.js
```diff
@@ -85,19 +85,24 @@ function generateKafkaAuthObject(auth) {
...
-            authObject['sasl.password'] = credentials.password;
+            if (credentialsFile) {
+                const credsFilePath = getAuthFilePath(credentialsFile);
+                if (credsFilePath === null) {
+                    throw new Error(`Credentials file ${credentialsFile} not found in ${authFilesFolder}`);
+                }
+                const credentials = readCredentialsFile(credsFilePath);
+                authObject['sasl.username'] = credentials.username;
+                authObject['sasl.password'] = credentials.password;
+            } else {
```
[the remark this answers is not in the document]
delthas: No, we use joi validation for that.

extensions/notification/utils/auth.js
```diff
@@ -85,19 +85,24 @@ function generateKafkaAuthObject(auth) {
...
+                authObject['sasl.username'] = username;
+                authObject['sasl.password'] = password;
+            }
```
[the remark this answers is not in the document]
delthas: There is a test now.

extensions/notification/queueProcessor/task.js
```diff
@@ -38,7 +38,11 @@ let destinationAuth = {
     keytab: process.env.KEYTAB,
     principal: process.env.PRINCIPAL,
     serviceName: process.env.SERVICE_NAME,
+    username: process.env.BASIC_USERNAME,
+    password: process.env.BASIC_PASSWORD,
```
francoisferrand, line 42: "basic" is for http auth, in SASL this is the "plain" mechanism I think:
```suggestion
    username: process.env.PLAIN_USERNAME,
    password: process.env.PLAIN_PASSWORD,
```

francoisferrand: but then I see the auth type is (incorrectly) called `basic`, so maybe best to keep it consistent at least...

extensions/notification/NotificationConfigValidator.js
francoisferrand: kind of difficult to read, can it be written more clearly with `Joi.alternatives()` or something similar?

something like:
```
const basicAuthSchema = saslAuthSchema.append({
    type: joi.string().valid('basic').required(),
	joi.alternatives().try([
		joi.object({
	     credentialsFile: joi.string().required(),
       },
	   joi.object({
		 username: joi.string().required(),
         password: joi.string().required(),
       })
    ]),
})
```

extensions/notification/NotificationConfigValidator.js
```diff
@@ -25,8 +25,10 @@ const kerberosAuthSchema = saslAuthSchema.append({
 
 const basicAuthSchema = saslAuthSchema.append({
```
francoisferrand: please add tests for the validation/rejection of different `saslAuthSchema` values : `joi` can be subtle, best to "unit-test" the rules...
(unless every case is covered already by the tests in `tests/unit/notification/utils/auth.js` ?)

extensions/notification/NotificationConfigValidator.js
delthas: OK, could not use it exactly as is because this joi.alternatives must be assigned to a field, but I simplified it.

extensions/notification/NotificationConfigValidator.js
delthas: OK, added some more tests

delthas:
Replaced by #2699

francoisferrand:
> Replaced by #2699

Note that there is no need to close a PR to change the target branch : you can just click "edit" (on top-right to the page) to change the target branch, and force-push the updated/rebased branch...

PR #2694, opened 2025-12-19
BB-731 downgrade to kafka 3.8.1 for lifecycle compat

tcarmet:
As we are most likely heading towards using kafka 3.8.1, since there are some incompatibilities on backbeat-lifecycle testsuite suggesting to downgrade a bit to keep things in sync. But we can also chose not to merge and keep ourselves more up to date.


tmacro (approved):

tcarmet:
turns out the reason why the tests were failing was unrelated to kafka's version, we are good to run with 3.9 and there's no reason to downgrade. Closing this PR.

PR #2696, opened 2025-12-22
🐛 (BB-733) Make sts port optional

DarkIsDude:
Issue: BB-733


francoisferrand:
Why close this? Is there a reason not-to handle the default in backbeat?

PR #2698, opened 2025-12-29
Make STS destination port optional for CRR destinations 

francoisferrand:
We can use the transport (http or https) to compute a default port, so we don't need to patch the code in many places.

Issue: BB-733


delthas (approved):
LGTM

SylvainSenechal (approved):

PR #2699, opened 2026-01-07

delthas:

Issue: BB-732

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,7 @@ FROM node:${NODE_VERSION}
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
+        ca-certificates \
         jq \
         openssl \
         tini \
--- a/extensions/notification/utils/auth.js
+++ b/extensions/notification/utils/auth.js
@@ -85,19 +85,24 @@ function generateKafkaAuthObject(auth) {
             break;
         }
         case 'basic': {
-            const { protocol, credentialsFile } = auth;
+            const { protocol, credentialsFile, username, password } = auth;
             if (!supportedSaslProtocols.includes(protocol)) {
                 throw new Error(`Unsupported security.protocol: ${protocol}`);
             }
-            const credsFilePath = getAuthFilePath(credentialsFile);
-            if (credsFilePath === null) {
-                throw new Error(`Credentials file ${credentialsFile} not found in ${authFilesFolder}`);
-            }
-            const credentials = readCredentialsFile(credsFilePath);
             authObject['sasl.mechanisms'] = 'PLAIN';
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -39,6 +39,7 @@ FROM node:${NODE_VERSION}
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
+        ca-certificates \
```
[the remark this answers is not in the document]
SylvainSenechal: Nit but agree with the bot : could be a separate commit with a message to explain why its added

SylvainSenechal (approved):
Looks good, for some reason I really struggle to read Joi validation code

extensions/notification/utils/auth.js
```diff
@@ -85,19 +85,26 @@ function generateKafkaAuthObject(auth) {
...
+                    throw new Error(`Credentials file ${credentialsFile} not found in ${authFilesFolder}`);
+                }
+                const credentials = readCredentialsFile(credsFilePath);
+                authObject['sasl.username'] = credentials.username;
+                authObject['sasl.password'] = credentials.password;
+            } else if (username && password) {
+                authObject['sasl.username'] = username;
+                authObject['sasl.password'] = password;
+            } else {
+                throw new Error('Missing credentials');
```
francoisferrand: can this really happen?
- if `credentialsFile` is set, we call `readCredentialsFile()` which validates the field are set indeed
- otherwise `username` & `password` should have been validated when validating the config with joi


```suggestion
			} else {
                authObject['sasl.username'] = username;
                authObject['sasl.password'] = password;
            }
```

tests/unit/notification/utils/auth.js
```diff
@@ -144,6 +160,14 @@ describe('generateKafkaAuthObject', () => {
...
             },
         },
+        {
+            description: 'Basic authentication with missing credentials',
+            valid: false,
+            input: {
+                type: 'basic',
+                protocol: 'SASL_PLAINTEXT',
+            },
+        },
```
francoisferrand: for completeness, should also have test case when only username of password is missing or empty, when `credentialsFile` is empty, and when both username/password and credentialsFile are specified

francoisferrand (approved):

tests/unit/notification/utils/auth.js
delthas: Added

extensions/notification/utils/auth.js
delthas: Hm indeed. I got confused because I wrongly created my unit tests against generateKafkaAuthObject, meaning after the validation. I moved the unit tests to target the validation and now they properly pass without this added else case.

PR #2703, opened 2026-01-21
sdk v3 migration : follow up on a few details

SylvainSenechal:
ISSUE: BB-730

### DIFF
--- a/bin/ensureServiceUser
+++ b/bin/ensureServiceUser
@@ -166,6 +166,7 @@ class PolicyHandler extends BaseHandler {
             } catch (err) {
                 // Workaround a Vault issue on 8.3 branch
                 // https://scality.atlassian.net/browse/VAULT-238
+                // Update : Should be possible to use now, but got TLS errors in Zenko CI, needs further testing
                 accountId = '000000000000';
             }
         }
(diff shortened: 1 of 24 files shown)


bin/ensureServiceUser
```diff
@@ -164,9 +164,10 @@ class PolicyHandler extends BaseHandler {
                 const res = await this.stsClient.send(command);
                 accountId = res.Account;
             } catch (err) {
-                // Workaround a Vault issue on 8.3 branch
```
SylvainSenechal, line 167: https://scality.atlassian.net/browse/VAULT-238
Get caller identity is available in vault
image

conf/config.json
SylvainSenechal, line 1: i'll remove this

SylvainSenechal:
ci not passing but you can review, i think it's a small thing because of the test mock returning err.name or something like that

bin/ensureServiceUser
francoisferrand: is it available in both Vault 1 and Vault 2 ?

francoisferrand (approved):

bin/ensureServiceUser
SylvainSenechal: Implemented in vault 2 here https://scality.atlassian.net/browse/VAULT-291

SylvainSenechal: and here for vault 1 https://scality.atlassian.net/browse/VAULT-214

benzekrimaha (approved):
LGTM

PR #2705, opened 2026-01-29
replication: Always refresh location data

delthas:
This is an intermediate step for later removing location data from the Kafka topic (in order to save space).

Issue: BB-491


delthas:
Replaced by #2708

PR #2711, opened 2026-02-03
Optimize oplog size by stripping locations of large objects

delthas:
Issue: [BB-491](https://scality.atlassian.net/browse/BB-491)

[BB-491]: https://scality.atlassian.net/browse/BB-491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -304,6 +304,7 @@
             "kafkaConnectHost": "127.0.0.1",
             "kafkaConnectPort": 8083,
             "numberOfConnectors": 1,
+            "locationStrippingThreshold": 5,
             "probeServer": {
                 "bindAddress": "0.0.0.0",
                 "port": 8556
--- a/extensions/oplogPopulator/pipeline/MultipleBucketsPipelineFactory.js
+++ b/extensions/oplogPopulator/pipeline/MultipleBucketsPipelineFactory.js
@@ -9,6 +9,16 @@ const PipelineFactory = require('./PipelineFactory');
  */
 class MultipleBucketsPipelineFactory extends PipelineFactory {
     /**
+     * @constructor
+     * @param {number} locationStrippingThreshold threshold for stripping location data
+     */
+    constructor(locationStrippingThreshold) {
+        super(locationStrippingThreshold);
+        // getPipeline is used standalone later, make sure its `this` reference binds to us.
+        this.getPipeline = this.getPipeline.bind(this);
+    }
+
+    /**
      * Checks if an existing pipeline is valid against the current
      * factory.
      * @param {string[]} bucketList pipeline
@@ -22,24 +32,21 @@ class MultipleBucketsPipelineFactory extends PipelineFactory {
     }
 
     /**
-     * Makes new connector pipeline that includes
-     * buckets assigned to this connector.
+     * Makes connector pipeline stage, that includes buckets assigned to this connector.
      * @param {string[] | undefined} buckets buckets assigned to this connector
-     * @returns {string} new connector pipeline
+     * @returns {object} connector pipeline stage
      */
-    getPipeline(buckets) {
+    getPipelineStage(buckets) {
         if (!buckets || !buckets.length) {
-            return JSON.stringify([]);
+            return null;
         }
-        return JSON.stringify([
-            {
-                $match: {
-                    'ns.coll': {
-                        $in: buckets,
-                    }
+        return {
+            $match: {
+                'ns.coll': {
+                    $in: buckets,
                 }
             }
-        ]);
+        };
     }
 }
 
--- a/extensions/oplogPopulator/pipeline/PipelineFactory.js
+++ b/extensions/oplogPopulator/pipeline/PipelineFactory.js
@@ -8,6 +8,14 @@ const { wildCardForAllBuckets } = require('../constants');
  */
 class PipelineFactory {
     /**
+     * @constructor
+     * @param {number} locationStrippingThreshold threshold for stripping location data
+     */
+    constructor(locationStrippingThreshold) {
+        this._locationStrippingThreshold = locationStrippingThreshold;
+    }
+
+    /**
      * Checks if an existing pipeline is valid against the current
      * factory.
      * @param {string} pipeline pipeline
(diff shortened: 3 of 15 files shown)


extensions/oplogPopulator/pipeline/PipelineFactory.js
```diff
@@ -58,6 +66,31 @@ class PipelineFactory {
...
         throw errors.NotImplemented;
     }
+
+    /**
+     * Builds a conditional location stripping stage.
+     * Excludes location[] from Kafka if there are too many items (to save space).
+     * Keeps location[] for small objects (to avoid extra S3 fetches).
+     * @returns {Object} MongoDB $set aggregation stage
+     */
+    _getLocationStrippingStage() {
```
francoisferrand: instead of calling this function (and possibly forgetting the call) in each implementation, could we tweak the interface and `getPipeline()` to keep this part in a single place?

maybe something like:

```
getPipelineAbstract(buckets) { // eslint-disable-line no-unused-vars
    throw errors.NotImplemented;
}

getPipeline(bucket) {
    return JSON.stringify([
		getPipelineAbstract(buckets), // return a single object (not a string)
        _getLocationStrippingStage(),
    ])
}
```

conf/config.json
```diff
@@ -304,6 +304,7 @@
             "kafkaConnectHost": "127.0.0.1",
             "kafkaConnectPort": 8083,
             "numberOfConnectors": 1,
+            "locationStrippingThreshold": 100,
```
francoisferrand: 100 seems quite a log already? (maybe around ~500kBytes or more for the locations)

anything specific for choosing this specific value? How can be define what a "good" default would be?

extensions/oplogPopulator/pipeline/MultipleBucketsPipelineFactory.js
```diff
@@ -8,6 +8,16 @@ const PipelineFactory = require('./PipelineFactory');
  * given a list of buckets.
  */
 class MultipleBucketsPipelineFactory extends PipelineFactory {
+    /**
+     * @constructor
+     * @param {number} locationStrippingThreshold threshold for stripping location data
+     */
+    constructor(locationStrippingThreshold = 100) {
```
francoisferrand: should not have default in these constructors, otherwise we just duplicate this (default) value

extensions/oplogPopulator/pipeline/PipelineFactory.js
```diff
@@ -58,6 +66,31 @@ class PipelineFactory {
...
+        return {
```
francoisferrand: should we allow disabling location stripping?
e.g. if `_locationStrippingThreshold` is undefined or null, skip all filtering.

this would allow us to easily revert to the default behavior in case of issue (and esp. ensure no extra load on mongo, if this ends up being an issue)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -864,12 +865,34 @@ class ReplicateObject extends BackbeatTask {
...
+                            log.info('replication already completed, skipping', {
+                                entry: sourceEntry.getLogInfo(),
+                            });
+                            return next(errorAlreadyCompleted);
+                        }
+                    }
+                    if (isLocationStripped) {
+                        // Reassign sourceEntry to use fresh metadata
+                        sourceEntry = refreshedEntry;
+                    }
```
francoisferrand, line 876: do we really need the conditionals?

we may systematically reassign `sourceEntry` I guess, and since we update the object, we can and should check if already replicated to skip immediately if needed?

francoisferrand (approved):

francoisferrand (approved):

extensions/oplogPopulator/pipeline/PipelineFactory.js
delthas: Agree it would be nicer :+1:

extensions/replication/tasks/ReplicateObject.js
delthas: I agree, I was considering doing this but was not sure about the impact of that change. Will add this. :+1:

extensions/oplogPopulator/pipeline/PipelineFactory.js
delthas: Would be useful. Thinking about translating this to the Zenko CR, maybe we could disable the feature when locationStrippingThreshold < 0 ?

So, for example:
- -1: never strip
- 0: always strip
- unset: use the default value
- 123: strip if >= 123 parts

In both the Zenko CR and the backbeat config.

conf/config.json
delthas: Indeed, it was an arbitrary choice but it is quite large looking at it again.

Maybe a way to define a "good" value is how large the location array gets relative to the rest of the object? Maybe we should allow a location array up to x10 times the rest of the object? So if a "base" object is 1~2 kB, and a location is ~5kB, we should have a default threshold of 5 items?

I guess it would also depend on whether there would be a performance bottleneck on the replication caused by having to read the object.

delthas:
Wondering if there might be an issue where we would refresh items that just had an empty location array (that was not stripped, but was just empty to start with). Investigating.

delthas:
> Wondering if there might be an issue where we would refresh items that just had an empty location array (that was not stripped, but was just empty to start with). Investigating.

I added a check on content length > 0 (if content length is 0, no need to fetch the location, there were not stripped, but rather empty in the first place).

extensions/oplogPopulator/pipeline/PipelineFactory.js
delthas: I just implemented that ^

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -864,12 +865,30 @@ class ReplicateObject extends BackbeatTask {
                 this._setTargetAccountMd(destEntry, targetRole, log, next);
             },
             next => {
-                if (!mdOnly &&
-                    sourceEntry.getContentLength() / 1000000 >=
-                    this.repConfig.queueProcessor.sourceCheckIfSizeGreaterThanMB) {
-                    return this._checkSourceReplication(sourceEntry, log, next);
```
francoisferrand, line 872: is this function (`_checkSourceReplication`) still used?
if not we shoud remove it, and/or maybe change its continuation callback to "return" the updated entry (but this may create lots of changes, if used in multiple places)

francoisferrand (approved):

conf/config.json
delthas: OK, went for 5 items.

extensions/oplogPopulator/pipeline/MultipleBucketsPipelineFactory.js
```diff
@@ -8,6 +8,16 @@ const PipelineFactory = require('./PipelineFactory');
  * given a list of buckets.
  */
 class MultipleBucketsPipelineFactory extends PipelineFactory {
+    /**
+     * @constructor
+     * @param {number} locationStrippingThreshold threshold for stripping location data
+     */
+    constructor(locationStrippingThreshold) {
+        super(locationStrippingThreshold);
+        // getPipeline is used standalone later, make sure its this binds to us.
```
SylvainSenechal: this doesnt look like english

SylvainSenechal (approved):

extensions/oplogPopulator/pipeline/MultipleBucketsPipelineFactory.js
delthas: It is!! Make sure that its `this` [variable] binds to us. But I guess I should wrap the this with backticks ;P

francoisferrand:
Should we merge this in a 9.3 branch? Or are we confident enough that we are happy to deliver this in a patch release (maybe even 9.2) ?

delthas:
> Should we merge this in a 9.3 branch? Or are we confident enough that we are happy to deliver this in a patch release (maybe even 9.2) ?

Let's merge in: 9.3.

delthas:
Moved to #2715

PR #2713, opened 2026-02-09
Support SASL/SCRAM bucket notification destination

delthas:
Issue: BB-742

### DIFF
--- a/extensions/notification/utils/auth.js
+++ b/extensions/notification/utils/auth.js
@@ -84,36 +84,20 @@ function generateKafkaAuthObject(auth) {
             }
             break;
         }
-        case 'basic': {
-            const { protocol, credentialsFile, username, password } = auth;
-            if (!supportedSaslProtocols.includes(protocol)) {
-                throw new Error(`Unsupported security.protocol: ${protocol}`);
-            }
-            authObject['sasl.mechanisms'] = 'PLAIN';
-            authObject['security.protocol'] = protocol;
-            if (credentialsFile) {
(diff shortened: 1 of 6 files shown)


extensions/notification/utils/auth.js
```diff
@@ -105,6 +105,30 @@ function generateKafkaAuthObject(auth) {
             }
             break;
         }
+        case 'scram': {
```
francoisferrand, line 88: looks like this whole block is duplicated with `basic`, the only difference is

```javascript
            authObject['sasl.mechanisms'] = 'PLAIN';
```
vs
```javascript
            if (!supportedScramMechanisms.includes(mechanism)) {
                throw new Error(`Unsupported SCRAM mechanism: ${mechanism}`);
            }
            authObject['sasl.mechanisms'] = `SCRAM-${mechanism}`;
```

so seems better refactor and avoid the duplication:
```javascript
case 'basic', 'scram':
    [...]
    if (type == 'basic') {
        authObject['sasl.mechanisms'] = 'PLAIN';
    } else if (!supportedScramMechanisms.includes(mechanism)) {
        throw new Error(`Unsupported SCRAM mechanism: ${mechanism}`);
    }
    authObject['sasl.mechanisms'] = `SCRAM-${mechanism}`;

    authObject['security.protocol'] = protocol;
    if (credentialsFile) {
    [...]
```

delthas:
Tested on a live AWS MSK cluster with SASL/SCRAM, works fine!

maeldonn (approved):
LGTM

francoisferrand (approved):

PR #2715, opened 2026-02-12
Optimize oplog size by stripping locations of large objects

delthas:
Issue: [BB-491](https://scality.atlassian.net/browse/BB-491)

[BB-491]: https://scality.atlassian.net/browse/BB-491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -304,7 +304,6 @@
             "kafkaConnectHost": "127.0.0.1",
             "kafkaConnectPort": 8083,
             "numberOfConnectors": 1,
-            "locationStrippingThreshold": 500,
             "probeServer": {
                 "bindAddress": "0.0.0.0",
                 "port": 8556
--- a/conf/config.json
+++ b/conf/config.json
@@ -304,6 +304,7 @@
             "kafkaConnectHost": "127.0.0.1",
             "kafkaConnectPort": 8083,
             "numberOfConnectors": 1,
+            "locationStrippingThreshold": 500,
             "probeServer": {
                 "bindAddress": "0.0.0.0",
                 "port": 8556
--- a/extensions/oplogPopulator/OplogPopulatorConfigValidator.js
+++ b/extensions/oplogPopulator/OplogPopulatorConfigValidator.js
@@ -6,7 +6,7 @@ const joiSchema = joi.object({
     kafkaConnectHost: joi.string().required(),
     kafkaConnectPort: joi.number().required(),
     numberOfConnectors: joi.number().required().min(0),
-    locationStrippingThreshold: joi.number().default(5),
+    locationStrippingBytesThreshold: joi.number().min(0).default(0),
     prefix: joi.string().optional(),
     probeServer: probeServerJoi.default(),
     connectorsUpdateCronRule: joi.string().default('*/1 * * * * *'),
--- a/extensions/oplogPopulator/OplogPopulatorConfigValidator.js
+++ b/extensions/oplogPopulator/OplogPopulatorConfigValidator.js
@@ -6,6 +6,7 @@ const joiSchema = joi.object({
     kafkaConnectHost: joi.string().required(),
     kafkaConnectPort: joi.number().required(),
     numberOfConnectors: joi.number().required().min(0),
+    locationStrippingThreshold: joi.number().default(5),
     prefix: joi.string().optional(),
     probeServer: probeServerJoi.default(),
     connectorsUpdateCronRule: joi.string().default('*/1 * * * * *'),
--- a/extensions/oplogPopulator/modules/Connector.js
+++ b/extensions/oplogPopulator/modules/Connector.js
@@ -314,10 +314,14 @@ class Connector extends EventEmitter {
      * That is why we use updateConnectorConfig() instead of
      * updateConnectorPipeline()
      * @param {boolean} [doUpdate=false] updates connector if true
+     * @param {boolean} [updateSupported=true] whether full pipeline updates
+     * (including match stage changes) are supported. When false, the update
+     * is only applied if the match stage has not changed, to protect resume
+     * tokens on MongoDB >= 6.0.
      * @returns {Promise|boolean} connector did update
      * @throws {InternalError}
      */
-    async updatePipeline(doUpdate = false) {
+    async updatePipeline(doUpdate = false, updateSupported = true) {
         // Only update when buckets changed and when not already updating
         if (!this._state.bucketsGotModified || this._state.isUpdating) {
             return false;
@@ -325,6 +329,21 @@ class Connector extends EventEmitter {
         this._config.pipeline = this._getPipeline([...this._buckets]);
         try {
             if (doUpdate && this._isRunning) {
+                if (!updateSupported) {
+                    const currentConfig = await this._kafkaConnect.getConnectorConfig(this._name);
+                    const currentPipeline = currentConfig.pipeline
+                        ? JSON.parse(currentConfig.pipeline) : [];
+                    const newPipeline = JSON.parse(this._config.pipeline);
+                    const currentMatch = JSON.stringify(currentPipeline[0]?.$match);
+                    const newMatch = JSON.stringify(newPipeline[0]?.$match);
+                    if (currentMatch !== newMatch) {
+                        this._logger.info('Skipping pipeline update: match stage differs', {
+                            method: 'Connector.updatePipeline',
+                            connector: this._name,
+                        });
+                        return false;
+                    }
+                }
                 const timeBeforeUpdate = Date.now();
                 this._state.isUpdating = true;
                 this.emit(constants.connectorUpdatedEvent, this);
--- a/extensions/oplogPopulator/modules/Connector.js
+++ b/extensions/oplogPopulator/modules/Connector.js
@@ -45,6 +45,7 @@ class Connector extends EventEmitter {
         this._name = params.name;
         this._config = params.config;
         this._buckets = new Set(params.buckets);
+        this._liveBuckets = new Set(params.buckets);
         this._isRunning = params.isRunning;
         this._getPipeline = params.getPipeline;
         this._state = {
@@ -166,6 +167,7 @@ class Connector extends EventEmitter {
                 config: this._config,
             });
             this._isRunning = true;
+            this._liveBuckets = new Set(this._buckets);
         } catch (err) {
             this._logger.error('Error while spawning connector', {
                 method: 'Connector.spawn',
@@ -193,6 +195,7 @@ class Connector extends EventEmitter {
             this.emit(constants.connectorUpdatedEvent, this);
             await this._kafkaConnect.deleteConnector(this._name);
             this._isRunning = false;
+            this._liveBuckets = new Set();
             // resetting the resume point to set a new one on creation of the connector
             delete this._config['startup.mode.timestamp.start.at.operation.time'];
         } catch (err) {
@@ -330,13 +333,9 @@ class Connector extends EventEmitter {
         try {
             if (doUpdate && this._isRunning) {
                 if (!updateSupported) {
-                    const currentConfig = await this._kafkaConnect.getConnectorConfig(this._name);
-                    const currentPipeline = currentConfig.pipeline
-                        ? JSON.parse(currentConfig.pipeline) : [];
-                    const newPipeline = JSON.parse(this._config.pipeline);
-                    const currentMatch = JSON.stringify(currentPipeline[0]?.$match);
-                    const newMatch = JSON.stringify(newPipeline[0]?.$match);
-                    if (currentMatch !== newMatch) {
+                    const sameBuckets = this._liveBuckets.size === this._buckets.size &&
+                        this._liveBuckets.isSubsetOf(this._buckets);
+                    if (!sameBuckets) {
                         this._logger.info('Skipping pipeline update: match stage differs', {
                             method: 'Connector.updatePipeline',
                             connector: this._name,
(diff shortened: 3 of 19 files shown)


delthas:
Manually tested with varying values and single-part, empty, and multi-part objects, works.

Also manually tested changing the configuration on the fly, properly applies, and properly resumes the pipeline with the token.

conf/config.json
```diff
@@ -304,6 +304,7 @@
             "kafkaConnectHost": "127.0.0.1",
             "kafkaConnectPort": 8083,
             "numberOfConnectors": 1,
+            "locationStrippingThreshold": 5,
```
francoisferrand: should use "proper" default (as discussed in ZKOP PR)

extensions/oplogPopulator/pipeline/PipelineFactory.js
```diff
@@ -55,7 +63,53 @@ class PipelineFactory {
...
+                        $cond: {
+                            if: {
+                                $gte: [
+                                    { $size: { $ifNull: ['$fullDocument.value.location', []] } },
+                                    this._locationStrippingThreshold
+                                ]
+                            },
+                            then: '$$REMOVE',
+                            else: { $ifNull: ['$fullDocument.value.location', '$$REMOVE'] },
+                        }
```
francoisferrand: can't we do `$ifNull` before `$cond`, to simplify the pipeline a bit (reduce redundancy) ?

extensions/oplogPopulator/pipeline/PipelineFactory.js
```diff
@@ -55,7 +63,53 @@ class PipelineFactory {
...
+                            if: {
+                                $gte: [
+                                    { $size: { $ifNull: ['$updateDescription.updatedFields.value.location', []] } },
+                                    this._locationStrippingThreshold
+                                ]
+                            },
+                            then: '$$REMOVE',
+                            else: { $ifNull: ['$updateDescription.updatedFields.value.location', '$$REMOVE'] },
+                        }
+                    },
```
francoisferrand, line 96: this is the same pipeline as above, applied to a different field.
should we introduce a helper function taking the field name in argument (maybe even just an local anonymous function), to deduplicate these?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -314,17 +314,35 @@ class Connector extends EventEmitter {
...
      * updateConnectorPipeline()
      * @param {boolean} [doUpdate=false] updates connector if true
+     * @param {boolean} [updateMatchStage=true] when false, only update
+     * if the $match stage has not changed (to protect resume tokens
+     * on MongoDB >= 6.0)
      * @returns {Promise|boolean} connector did update
      * @throws {InternalError}
      */
-    async updatePipeline(doUpdate = false) {
+    async updatePipeline(doUpdate = false, updateMatchStage = true) {
```
francoisferrand: maybe the `updateSupported` name would be more helpfull ?

this way we can document in here why we may still update when update is not supported: it just becomes an implementation detail.

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -314,17 +314,35 @@ class Connector extends EventEmitter {
...
         try {
             if (doUpdate && this._isRunning) {
+                if (!updateMatchStage) {
+                    const currentConfig = await this._kafkaConnect.getConnectorConfig(this._name);
+                    const currentPipeline = currentConfig.pipeline
+                        ? JSON.parse(currentConfig.pipeline) : [];
+                    const newPipeline = JSON.parse(this._config.pipeline);
+                    const currentMatch = JSON.stringify(currentPipeline[0]?.$match);
+                    const newMatch = JSON.stringify(newPipeline[0]?.$match);
+                    if (currentMatch !== newMatch) {
```
francoisferrand: * does this case actually happen? changing the "match" (i.e. list of buckets associated with a pipeline) should be handled in the strategy.... if update not supported, the strategy will create a new pipeline.
* maybe it happens when "removing" a bucket (i.e. we don't need its oplog anyway) : in that case, we used to do nothing indeed.... but we should probably still do the update to change the "config" of the pipeline → should generate the pipeline BUT keep the existing list of buckets (and maybe report that we did not fully update...)

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -314,17 +314,35 @@ class Connector extends EventEmitter {
...
+    async updatePipeline(doUpdate = false, updateMatchStage = true) {
         // Only update when buckets changed and when not already updating
         if (!this._state.bucketsGotModified || this._state.isUpdating) {
             return false;
         }
         this._config.pipeline = this._getPipeline([...this._buckets]);
         try {
             if (doUpdate && this._isRunning) {
+                if (!updateMatchStage) {
+                    const currentConfig = await this._kafkaConnect.getConnectorConfig(this._name);
```
francoisferrand: retrieving the pipeline is already done before, so it is redondant & inefficient to make yet another call here.
Basically, what we want here is to keep the list of buckets almost immutable (i.e. not modifying it, but still knowing when no buckets are needed anymore) : can we instead check this list of buckets directly? and possibly tweak handling/tracking there, so we can always regenerate the pipeline with the "current" list of buckets?

tests/unit/oplogPopulator/pipeline/MultipleBucketsPipelineFactory.js
SylvainSenechal, line 1: I will review everything later but general question : Would it be hard to setup a functional test around the connector and oplog logic ? Is it possible to only instantiate a Connector, and have a running mongodb, to do functional testing ? I'm asking because the changes, and the overall logic looks complicated and functional tests would help here

edit: i guess also need kafka so it starts to become a bit complex

delthas: We would need Kafka + Kafka Connect , so likely not (or would warrant its own ticket)

extensions/oplogPopulator/pipeline/PipelineFactory.js
delthas: We can't simply do `$ifNull` before `$cond` (e.g. with `$let`) because it would change behavior for null/missing location fields.

If we do:
```js
$let: {
    vars: { loc: { $ifNull: [`$${fieldPath}`, []] } },
    in: {
        $cond: {
            if: { $gte: [{ $size: '$$loc' }, threshold] },
            then: '$$REMOVE',
            else: '$$loc'
        }
    }
}
```

When the field is null/missing, `$$loc` becomes `[]`, size is 0 (below threshold), so the `else` branch returns `[]` — this would create an empty `location: []` on documents that never had a location field. We want `$$REMOVE` in that case to preserve the original document shape.

Instead, we'll use `$switch` with an `$isArray` guard, which eliminates both `$ifNull` usages entirely:
```js
$switch: {
    branches: [
        { case: { $not: [{ $isArray: `$${fieldPath}` }] }, then: '$$REMOVE' },
        { case: { $gte: [{ $size: `$${fieldPath}` }, threshold] }, then: '$$REMOVE' },
    ],
    default: `$${fieldPath}`
}
```

`$isArray` returns false for null/missing, so the first branch handles that. `$switch` short-circuits, so `$size` in the second branch only evaluates when the field is an array.

extensions/oplogPopulator/modules/Connector.js
delthas: Added a _liveBuckets variable that tracks what is actually being run by Kafka Connect, removing the need for the fetch

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -330,13 +339,10 @@ class Connector extends EventEmitter {
...
-                    const currentConfig = await this._kafkaConnect.getConnectorConfig(this._name);
-                    const currentPipeline = currentConfig.pipeline
-                        ? JSON.parse(currentConfig.pipeline) : [];
-                    const newPipeline = JSON.parse(this._config.pipeline);
-                    const currentMatch = JSON.stringify(currentPipeline[0]?.$match);
-                    const newMatch = JSON.stringify(newPipeline[0]?.$match);
-                    if (currentMatch !== newMatch) {
+                    const sameSize = this._liveBuckets.size === this._buckets.size;
+                    const sameContent = sameSize &&
+                        [...this._liveBuckets].every(b => this._buckets.has(b));
```
francoisferrand: `[...this._liveBuckets]` is a costly operation (esp. for many entries), as it creates a new list/array

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -79,6 +80,12 @@ class Connector extends EventEmitter {
      */
     get buckets() { return [...this._buckets]; }
 
+    /**
+     * Getter for connector live buckets (currently configured on Kafka Connect)
+     * @returns {string[]} live buckets
+     */
+    get liveBuckets() { return [...this._liveBuckets]; }
```
francoisferrand: this is quite costly (allocates and copies the whole list), can we avoid this?

extensions/oplogPopulator/modules/Connector.js
```diff
@@ -45,6 +45,7 @@ class Connector extends EventEmitter {
         this._name = params.name;
         this._config = params.config;
         this._buckets = new Set(params.buckets);
+        this._liveBuckets = new Set(params.buckets);
```
francoisferrand, line 48: we already have a Set of buckets (`_buckets`), which create another one? (esp. to convert it to a list whenever it is used...)

francoisferrand: also not sure, but I seem to remember we already "track" which buckets are actually deployed (maybe through the state of each connector? or indirectly through _buckets, due to the way they are updated? ...) Can't we reuse that info (if indeed present), instead of re-computing it?

delthas: Buckets stores the buckets we want, while live buckets stores the buckets currently used by Kafka Connect. For example, say we start on an upgraded cluster that has buckets "A" and "B" on a conncetor (previously having multiple buckets).

Next we stop listening to "B". We're only interested in "A". `_buckets` would contain `[A]`, whereas live buckets would contain `[A, B]`. This is how we would see `[A, B] != [A]` and prevent the update if not allowed by the pipeline factory.

extensions/oplogPopulator/OplogPopulatorConfigValidator.js
```diff
@@ -6,6 +6,7 @@ const joiSchema = joi.object({
     kafkaConnectHost: joi.string().required(),
     kafkaConnectPort: joi.number().required(),
     numberOfConnectors: joi.number().required().min(0),
+    locationStrippingThreshold: joi.number().default(5),
```
francoisferrand: should have default consistent with `sourceCheckIfSizeGreaterThanMB` : since parts are expected to be 5MB at least (according to AWS spec), should be 20 by default?

extensions/oplogPopulator/modules/Connector.js
delthas: Dropped (FWIW there is exactly the same getter above for buckets)

extensions/oplogPopulator/modules/Connector.js
delthas: Dropped (FWIW there is the same expresison for buckets just above)

delthas:
Tested on a live cluster, working.

extensions/oplogPopulator/pipeline/PipelineFactory.js
```diff
@@ -55,7 +63,49 @@ class PipelineFactory {
...
+        const pipeline = [
+            stage,
+        ];
+        if (this._locationStrippingBytesThreshold > 0) {
+            pipeline.push({
+                $set: {
+                    'fullDocument.value.location': {
+                        $cond: {
+                            if: { $gte: [
+                                '$fullDocument.value.content-length',
```
francoisferrand, line 80: how does `$gte` behave if there is no `$fullDocument.value.content-length` field : does it go into the `then` or `else` case? (i.e. how does it compare `undefined` vs a number)

may be good to add a test for this specific case (not sure it really happen, but maybe in some corner case like pushing empty objet...), to ensure we are not adding any risk

francoisferrand (approved):

extensions/oplogPopulator/pipeline/PipelineFactory.js
delthas: In MongoDB's BSON comparison order, null/missing values are less than any number (MinKey < Null < Numbers < ...). So when content-length is missing, $gte: [null, threshold] evaluates to false, and the else branch is taken, preserving the location field as-is.                                                                                                                                                             

This is the correct behavior — if we don't know the object size, we keep the location.                                                                                                                            

Working on a test.

delthas: Actually we do not have functional tests using Mongo + Kafka + Kafka Connect here? So not really testable as is here. Maybe in Zenko?

francoisferrand: I was thinking more of a "unit" test, i.e. just trying to use the pipeline returned by `getPipeline` with mongo : create object in mongo, then make a query with the pipeline and check the result (and repeat with different "kinds" of objects)

SylvainSenechal: if later you can add functional tests in zenko i think it would be nice

SylvainSenechal (approved):

extensions/oplogPopulator/pipeline/PipelineFactory.js
delthas: Unit test added.

PR #2718, opened 2026-02-17
Manage CRR endpoints without explicit port

maeldonn:
Issue: ARTESCA-16783

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -258,8 +258,9 @@ class QueueProcessor extends EventEmitter {
         }
 
         if (Array.isArray(this.destConfig.replicationEndpoint.servers)) {
+            const defaultPort = this.destConfig.transport === 'https' ? 443 : 80;
             this.destHosts =
-                new RoundRobin(this.destConfig.replicationEndpoint.servers, { defaultPort: 80 });
+                new RoundRobin(this.destConfig.replicationEndpoint.servers, { defaultPort });
             if (this.destConfig.replicationEndpoint.echo) {
                 this._setupEcho();
             }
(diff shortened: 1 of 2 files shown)


SylvainSenechal (approved):
why is getQueueProcessorConfig returning an array instead of an object 😆

benzekrimaha (approved):
lgtm

francoisferrand:
> Issue: ARTESCA-16783

Backbeat PR must reference a Backbeat ticket (`BB-xxx`), not an Artesca ticket, c.f. message from Bert-e:

> The Jira issue ARTESCA-16783 specified in the source
> branch name, does not belong to project BB.

so you will need to reopen this PR with a different branch name

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -258,8 +258,9 @@ class QueueProcessor extends EventEmitter {
         }
 
         if (Array.isArray(this.destConfig.replicationEndpoint.servers)) {
+            const defaultPort = this.destConfig.transport === 'https' ? 443 : 80;
```
francoisferrand, line 261: can't the transport be specified in each replication endpoint?

francoisferrand (commented):
is this really the only place where the server port is added?

if in doubt, "validating" input (and handling default) may be done when loading the configuration, to ensure it is available later in every case

extensions/replication/queueProcessor/QueueProcessor.js
maeldonn: Transport IS already resolved per-site (config.getReplicationSiteDestConfig()). The varible destConfig is constructed per-site, so the transport is not a global value, but the current site transport.

maeldonn:
Moved to https://github.com/scality/backbeat/pull/2721

PR #2721, opened 2026-02-23
Manage CRR endpoints without explicit port

maeldonn:
Issue: BB-746

This is the same PR as #2718 but with a different branch name.

For fixing ARTESCA-16783

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -90,14 +90,19 @@ class Config extends EventEmitter {
             throw new Error('dataMoverTopic is required when lifecycle transitions is supported');
         }
 
-        if (parsedConfig.extensions && parsedConfig.extensions.replication
-            && parsedConfig.extensions.replication.destination
-            && parsedConfig.extensions.replication.destination.bootstrapList) {
-            this.bootstrapList = parsedConfig.extensions.replication
-            .destination.bootstrapList;
-        } else {
-            this.bootstrapList = [];
-        }
+        const destination = parsedConfig.extensions?.replication?.destination;
+        this.bootstrapList = destination?.bootstrapList?.map(endpoint => {
+            if (!endpoint.servers) {
+                return endpoint;
+            }
+
+            const transport = destination.sites?.[endpoint.site]?.transport || destination.transport || 'http';
+            const defaultPort = transport === 'https' ? 443 : 80;
+            const servers = endpoint.servers.map(server => server.includes(':') ? server : `${server}:${defaultPort}`);
+
+            return { ...endpoint, servers };
+        }) ?? [];
+
         const shouldRestrictToSite = process.env.BOOTSTRAP_SITE_NAME;
         if (shouldRestrictToSite) {
             this.bootstrapList = this.bootstrapList.filter(item => item.site === shouldRestrictToSite);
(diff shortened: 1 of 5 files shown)


bin/replication.js
```diff
@@ -49,9 +52,8 @@ function _createSetupReplication(command, options, log) {
         target: {
             bucket: targetBucket,
             credentials: targetCredentials,
-            hosts: targetIsExternal ?
-                undefined : new RoundRobin(destinationEndpoint.servers),
-            transport: destination.transport,
+            hosts: targetIsExternal ? undefined : new RoundRobin(destSiteConfig.replicationEndpoint.servers),
+            transport: destSiteConfig?.transport || destination.transport,
```
francoisferrand: nit: harder to read with a conditional (line 43) followed here by a "default"...

maybe better to either inline this new `destSiteConfig`:
```suggestion
            transport: destinationEndpoint
            		? config.getReplicationSiteDestConfig(destinationEndpoint.site)
            		:  destination.transport,
```

or to directly compute the transport there:
```suggestion
const destTransport = destinationEndpoint
        ? config.getReplicationSiteDestConfig(destinationEndpoint.site).transport
        : destination.transport;
```

francoisferrand: why do we need this change here: `transport` was always set by config, to ensure data is always in the same "form"... We should continue doing the same, and fixing the port in Config (as you do) should be enough without these extra changes?

maeldonn: I initially implemented normalization within getReplicationSiteDestConfig. However, after further review of the codebase, I determined that this was not the optimal location. I have refactored the logic to handle normalization during config initialization. This simplifies the implementation and prevents potential downstream errors.

SylvainSenechal (approved):

tests/unit/conf/configs/replicationServersConfig.json
```diff
@@ -0,0 +1,204 @@
...
+                    },
+                    "http-site": {
+                        "transport": "http"
+                    }
+                },
+                "bootstrapList": [
+                    { "site": "https-site", "servers": ["s3.example.com"] },
+                    { "site": "http-site", "servers": ["s3.example.com"] },
+                    { "site": "explicit-port-site", "servers": ["s3.example.com:8443"] },
+                    { "site": "aws-site", "type": "aws_s3" }
```
francoisferrand, line 86: missing case(s) where `servers` is not an array

lib/Config.js
```diff
@@ -90,14 +90,19 @@ class Config extends EventEmitter {
...
-            && parsedConfig.extensions.replication.destination.bootstrapList) {
-            this.bootstrapList = parsedConfig.extensions.replication
-            .destination.bootstrapList;
-        } else {
-            this.bootstrapList = [];
-        }
+        const destination = parsedConfig.extensions?.replication?.destination;
+        this.bootstrapList = destination?.bootstrapList?.map(endpoint => {
+            if (!Array.isArray(endpoint.servers)) {
+                return endpoint;
```
francoisferrand, line 96: don't we need to add the default port in this case (e.g. single `server`) as well ?

tests/unit/conf/configs/replicationServersConfig.json
maeldonn: Servers cannot be a string. There is a validator for this: https://github.com/scality/backbeat/blob/development/9.3/lib/config/configItems.joi.js#L22

lib/Config.js
maeldonn: This was to handle different entry types without servers field. Update the condition.

francoisferrand (approved):

PR #2726, opened 2026-03-20
Fix close() functions

SylvainSenechal:
Issue: BB-752

### DIFF
--- a/extensions/gc/GarbageCollector.js
+++ b/extensions/gc/GarbageCollector.js
@@ -167,13 +167,17 @@ class GarbageCollector extends EventEmitter {
     }
 
     /**
-     * Close the lifecycle consumer
+     * Close the garbage collector consumer
      * @param {function} cb - callback function
      * @return {undefined}
      */
     close(cb) {
         this._logger.debug('closing garbage collector consumer');
-        this._consumer.close(cb);
+        if (this._consumer) {
+            this._consumer.close(cb);
+        } else {
+            cb();
+        }
     }
 
     processKafkaEntry(kafkaEntry, done) {
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
@@ -145,7 +145,11 @@ class LifecycleObjectProcessor extends EventEmitter {
             clearInterval(this._deleteInactiveCredentialsInterval);
         }
 
-        this._consumer.close(cb);
+        if (this._consumers) {
+            this._consumers.close(cb);
+        } else {
+            cb();
+        }
     }
 
     /**
--- a/extensions/gc/service.js
+++ b/extensions/gc/service.js
@@ -100,7 +100,7 @@ initAndStart();
 
 process.on('SIGTERM', () => {
     logger.info('received SIGTERM, exiting');
-    garbageCollector.stop(() => {
+    garbageCollector.close(() => {
         process.exit(0);
     });
 });
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -823,6 +823,7 @@ class BackbeatConsumer extends EventEmitter {
         });
         // ensure consumer is active when calling offsetsStore() on
         // it, to avoid raising an exception (invalid state)
+        // TODO : potential issue here, see BB-758
         if (committableOffset !== null && !this.isPaused()) {
             this._consumer.offsetsStore([{ topic, partition,
                                            offset: committableOffset }]);
--- a/tests/unit/gc/GarbageCollector.spec.js
+++ b/tests/unit/gc/GarbageCollector.spec.js
@@ -92,4 +92,47 @@ describe('garbage collector', function garbageCollector() {
               .setAttribute('target.owner', ownerId);
         gcTask.processActionEntry(action, done);
     });
+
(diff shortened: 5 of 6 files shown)


extensions/gc/GarbageCollector.js
```diff
@@ -167,13 +167,17 @@ class GarbageCollector extends EventEmitter {
     }
 
     /**
-     * Close the lifecycle consumer
+     * Stop the garbage collector consumer
```
SylvainSenechal: You can tell a code was hard copy/pasted when the comments of the functions belongs to another file 🌚

extensions/gc/GarbageCollector.js
```diff
@@ -167,13 +167,17 @@ class GarbageCollector extends EventEmitter {
...
      * @param {function} cb - callback function
      * @return {undefined}
      */
-    close(cb) {
+    stop(cb) {
         this._logger.debug('closing garbage collector consumer');
-        this._consumer.close(cb);
+        if (this._consumer) {
```
SylvainSenechal, line 176: This is defensive and following similar pattern as the other files. The most important fix is that on sigterm, we are calling a function called stop and not close

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -145,7 +145,11 @@ class LifecycleObjectProcessor extends EventEmitter {
             clearInterval(this._deleteInactiveCredentialsInterval);
         }
 
-        this._consumer.close(cb);
+        if (this._consumers) {
```
SylvainSenechal, line 148: - misspelled _consumer with an S, was giving js errors as close function doesn't exists on a property _consumer that doesnt exists

extensions/gc/GarbageCollector.js
SylvainSenechal, line 1: I don't necessarily expect theses changes to be that impactful, as these functions are called on sigterm, so no matter what, the server will be shutdown, but still nice to fix

extensions/gc/GarbageCollector.js
```diff
@@ -167,13 +167,17 @@ class GarbageCollector extends EventEmitter {
...
+            this._consumer.close(cb);
+        } else {
+            cb();
```
francoisferrand, line 179: should we take the chance to migrate this function to async ?

extensions/gc/GarbageCollector.js
```diff
@@ -167,13 +167,17 @@ class GarbageCollector extends EventEmitter {
     }
 
     /**
-     * Close the lifecycle consumer
+     * Stop the garbage collector consumer
      * @param {function} cb - callback function
      * @return {undefined}
      */
-    close(cb) {
+    stop(cb) {
```
francoisferrand: looking at the various extensions, it seems there is no clear convention between close() and stop():

```
Populator : close
queuePopulator: close
gc : close implemented, but stop used
bucketProcessor : close
objectProcessor : close
mongoProcessor: stop
queueProcessor: stop
replicationStatusProcessor: stop
---
logReader: close
circuitBreaker: stop
backbearConsumer: close
BackbeatConsumerManager: close
```

so I wonder how we should fix this one: should we modify the GarbageCollector to `stop()` as you did, or instead the gc startup script to use `close()` which a small majority?

Or is there an argument to be made that stop is more appropriate than close? (as far as naming conventions go, `close` is usually terminal, while we could restart after a `stop`)

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -145,7 +145,11 @@ class LifecycleObjectProcessor extends EventEmitter {
...
+            this._consumers.close(cb);
+        } else {
+            cb();
```
francoisferrand, line 151: also here, should we make it async?

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
francoisferrand, line 148: this function was probably not tested, should add a test

extensions/gc/GarbageCollector.js
francoisferrand, line 176: should add a test to cover the new case at least (and possibly the nominal one as well, if not already covered)

francoisferrand (approved):

francoisferrand:
Does this bug exist in 9.1 as well?
Does it have an impact, e.g. preventing proper shutdown may not stop BackbeatConsumer, and we could end up replaying the same message? Do you estimate this would this be an issue for GC process?

DarkIsDude (approved):

maeldonn (approved):
LGTM, are these functions tested ?

SylvainSenechal:
> Does this bug exist in 9.1 as well? Does it have an impact, e.g. preventing proper shutdown may not stop BackbeatConsumer, and we could end up replaying the same message? Do you estimate this would this be an issue for GC process?

Yeah it's in previous versions too, and I think it can have an impact on kafka, we should probably merge on 9.1 and 9.2

extensions/gc/GarbageCollector.js
SylvainSenechal: Yes actually I'm keeping the name close

SylvainSenechal:
Added tests

PR #2727, opened 2026-03-25
Fix ingestion from raft session >=10

francoisferrand:
The JSON was causing the raftId string (which is the http body of the response) to be returned to
the code as an array of string : [ '1', '3' ]. Which was worked-around by taking the first element.
That fix worked for raft sesion < 10, but breaks when raft session cluster grows.

The fix is to properly define the model, with the `payload` attribute allowing to indicate that the
body of the request should map to a specified field of the model.

Impact is limited since we only make a call to this API in a single place.

Issue: BB-759

### DIFF
--- a/lib/clients/backbeat-2017-07-01.api.json
+++ b/lib/clients/backbeat-2017-07-01.api.json
@@ -1142,7 +1142,13 @@
                 }
             },
             "output": {
-                "type": "string"
+                "type": "structure",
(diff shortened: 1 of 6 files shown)


maeldonn (approved):

SylvainSenechal (approved):
probably gonna have some nice conflict on waterfall

benzekrimaha (approved):

PR #2728, opened 2026-03-26
fix(deps): bump node-rdkafka to ^3.6.0 to fix cooperative-sticky rebalance bug

delthas:
## Summary

Bumps `node-rdkafka` from `^2.12.0` (librdkafka 2.3.0) to `^3.6.0` (currently resolving to 3.6.1 / librdkafka 2.12.0) as a maintenance upgrade, picking up bugfixes accumulated across librdkafka 2.3.0→2.12.0.

Also fixes a latent race condition in `BackbeatConsumer._bootstrapConsumer()` exposed by librdkafka 2.10.0+.

## Context

This was initially motivated by investigating a flaky CI failure in the **"Kafka Cleaner"** test ([Zenko CI run](https://github.com/scality/Zenko/actions/runs/23431247112/job/68489839492?pr=2360)), where the `backbeat-metrics` topic had unconsumed messages after rapid pod replacement. While the original hypothesis (cooperative-sticky rebalance bug [confluentinc/librdkafka#4908](https://github.com/confluentinc/librdkafka/pull/4908)) turned out not to apply — BackbeatConsumer uses eager rebalance (`range,roundrobin`), not cooperative-sticky — the upgrade is still worthwhile for the accumulated bugfixes, and exposed a real bug in the bootstrap code that needed fixing.

## Bootstrap consumer fix

The librdkafka upgrade exposed a latent race condition in `_bootstrapConsumer()`. The bootstrap previously used `setInterval(200ms)` to call `consume(1, consumeCb)`, dispatching C++ async workers to the libuv thread pool (each with a 1000ms timeout). Up to 5 workers could be in flight concurrently.

When the bootstrap match was found, the old code called `clearInterval` then immediately `unsubscribe()`. But `clearInterval` only prevents *new* consume calls — workers already in the C++ thread pool continue running. With librdkafka < 2.10.0, `unsubscribe()` effectively invalidated these stale workers (they'd return empty or error). With librdkafka >= 2.10.0 ("Enhanced handling for subscribe/unsubscribe edge cases"), these workers survive across the unsubscribe→subscribe transition and dequeue messages from the next subscription, delivering them to the bootstrap's `consumeCb` which silently ignores non-bootstrap messages.

This was confirmed by local reproduction and bisection:

| node-rdkafka | librdkafka | `_bootstrapConsumer` test |
|---|---|---|
| 3.3.1 | 2.8.0 | PASS |
| 3.4.0 | 2.10.0 | FAIL |
| 3.6.1 | 2.12.0 | FAIL |

The fix replaces `setInterval` with chained `setTimeout`: each `consume(1)` is only scheduled after the previous one completes, guaranteeing at most one C++ async worker is in flight at a time. This makes `unsubscribe()` safe to call directly from the callback with no drain/polling needed.

## Why ^3.6.0

| node-rdkafka | librdkafka |
|---|---|
| 2.18.0 (latest 2.x) | 2.3.0 |
| 3.0.0 | 2.3.0 |
| 3.4.0 | 2.10.0 |
| **3.6.1** | **2.12.0** |

We set the floor to `^3.6.0` (librdkafka 2.12.0). KIP-848 (new consumer group protocol) is opt-in (`group.protocol=consumer` must be explicitly set) and does not affect the default `classic` protocol.

## Upgrade safety

**librdkafka 2.3.0 → 2.12.0**: The [librdkafka CHANGELOG](https://github.com/confluentinc/librdkafka/blob/master/CHANGELOG.md) shows no consumer-breaking changes in this range. The only notable breaking change is in v2.4.0 where `INVALID_RECORD` producer errors became non-retriable — this does not affect consumers. The metadata recovery behavior change in 2.10.0 (brokers not in metadata responses are removed and clients re-bootstrap) is a minor behavioral difference that should not impact normal operation.

**node-rdkafka 2.x → 3.x**: The [3.0.0 release](https://github.com/Blizzard/node-rdkafka/releases/tag/v3.0.0) only dropped support for EOL Node.js versions — no API changes. Backbeat requires Node >= 20 and runs on Node 22.14.0.

Issue: BB-760

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -876,51 +876,46 @@ class BackbeatConsumer extends EventEmitter {
         let lastBootstrapId;
         let producer; // eslint-disable-line prefer-const
         let producerTimer;
-        let bootstrapDone = false;
         // Use chained setTimeout instead of setInterval to ensure
         // only one consume() worker is in flight at a time. With
         // setInterval, multiple C++ async workers can overlap, and
         // since librdkafka 2.10.0 they survive an
         // unsubscribe→subscribe transition, stealing messages from
         // the next subscription.
-        function consumeNext() {
+        function consume() {
             self._consumer.consume(1, (err, messages) => {
-                if (err || bootstrapDone) {
-                    if (!bootstrapDone) {
-                        setTimeout(consumeNext, 200);
-                    }
+                if (err) {
+                    setTimeout(consume, 200);
                     return undefined;
                 }
-                let matched = false;
-                messages.forEach(message => {
+                const match = messages.find(message => {
                     const bootstrapId = JSON.parse(message.value).bootstrapId;
-                    if (bootstrapId) {
-                        self._log.info('bootstraping backbeat consumer: ' +
-                            'received bootstrap message',
-                            { bootstrapId, topic: self._topic, groupId: self._groupId });
-                        if (bootstrapId === lastBootstrapId) {
-                            self._log.info('backbeat consumer is bootstrapped',
-                                { topic: self._topic, groupId: self._groupId });
-                            matched = true;
-                            bootstrapDone = true;
-                            clearInterval(producerTimer);
-                            self._consumer.offsetsStore([{
-                                topic: self._topic,
-                                partition: message.partition,
-                                offset: message.offset + 1,
-                            }]);
-                            self._consumer.commit();
-                            self._consumer.unsubscribe();
-                            producer.close(() => {
-                                self._bootstrapping = false;
-                                self._onReady();
-                            });
-                        }
+                    if (!bootstrapId) {
+                        return false;
                     }
+                    self._log.info('bootstraping backbeat consumer: ' +
+                        'received bootstrap message',
+                        { bootstrapId, topic: self._topic, groupId: self._groupId });
+                    return bootstrapId === lastBootstrapId;
                 });
-                if (!matched) {
-                    setTimeout(consumeNext, 200);
+                if (!match) {
+                    setTimeout(consume, 200);
+                    return undefined;
                 }
+                self._log.info('backbeat consumer is bootstrapped',
+                    { topic: self._topic, groupId: self._groupId });
+                clearInterval(producerTimer);
+                self._consumer.offsetsStore([{
+                    topic: self._topic,
+                    partition: match.partition,
+                    offset: match.offset + 1,
+                }]);
+                self._consumer.commit();
+                self._consumer.unsubscribe();
+                producer.close(() => {
+                    self._bootstrapping = false;
+                    self._onReady();
+                });
                 return undefined;
             });
         }
@@ -947,7 +942,7 @@ class BackbeatConsumer extends EventEmitter {
                         this._initConsumer();
                         this._consumer.on('ready', () => {
                             this._consumer.subscribe([this._topic]);
-                            consumeNext();
+                            consume();
                         });
                     }, 500);
                 }
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -219,6 +219,10 @@ class BackbeatConsumer extends EventEmitter {
         const consumerParams = {
             'metadata.broker.list': this._kafkaHosts,
             'group.id': this._groupId,
+            // This is the default in our current librdkafka version, but we
+            // pin it explicitly because we depend on eager rebalancing and
+            // don't want it changed implicitly by future version updates.
+            'partition.assignment.strategy': 'range,roundrobin',
             // we manage stored offsets based on the highest
             // contiguous offset fully processed by a worker, so
             // disabling automatic offset store is needed
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -876,52 +876,53 @@ class BackbeatConsumer extends EventEmitter {
         let lastBootstrapId;
         let producer; // eslint-disable-line prefer-const
         let producerTimer;
-        let consumerTimer;
-        let inFlightConsumes = 0;
-        // Wait for all in-flight consume(1, consumeCb) workers to complete
-        // before calling unsubscribe(). The setInterval(200ms) pattern can
-        // have multiple consume workers running concurrently in the C++
-        // thread pool (each with a 1000ms timeout). Since librdkafka 2.10.0,
-        // these workers survive an unsubscribe→subscribe transition and can
-        // steal messages from the next subscription.
-        function _finishBootstrap(partition, offset) {
-            if (inFlightConsumes > 0) {
-                setTimeout(() => _finishBootstrap(partition, offset), 50);
-                return;
-            }
-            self._consumer.offsetsStore([{
-                topic: self._topic,
-                partition,
-                offset,
-            }]);
-            self._consumer.commit();
-            self._consumer.unsubscribe();
-            producer.close(() => {
-                self._bootstrapping = false;
-                self._onReady();
-            });
-        }
-        function consumeCb(err, messages) {
-            inFlightConsumes--;
-            if (err) {
-                return undefined;
-            }
-            messages.forEach(message => {
-                const bootstrapId = JSON.parse(message.value).bootstrapId;
-                if (bootstrapId) {
-                    self._log.info('bootstraping backbeat consumer: ' +
-                        'received bootstrap message',
-                        { bootstrapId, topic: self._topic, groupId: self._groupId });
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -67,7 +67,7 @@
     "minimatch": "^10.0.1",
     "mongodb": "^6.11.0",
     "node-forge": "^1.3.1",
-    "node-rdkafka": "^2.12.0",
+    "node-rdkafka": "^3.5.0",
```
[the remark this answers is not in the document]
delthas: Oudated

package.json
```diff
@@ -67,7 +67,7 @@
     "minimatch": "^10.0.1",
     "mongodb": "^6.11.0",
     "node-forge": "^1.3.1",
-    "node-rdkafka": "^2.12.0",
+    "node-rdkafka": "^3.6.0",
```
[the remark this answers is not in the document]
delthas: Outdated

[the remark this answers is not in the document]
delthas: Outdated

delthas:
On hold because it is not known yet whether this PR actually helps with the CI.

lib/BackbeatConsumer.js
```diff
@@ -873,7 +877,32 @@ class BackbeatConsumer extends EventEmitter {
...
         let consumerTimer;
+        let inFlightConsumes = 0;
+        // Wait for all in-flight consume(1, consumeCb) workers to complete
+        // before calling unsubscribe(). The setInterval(200ms) pattern can
+        // have multiple consume workers running concurrently in the C++
+        // thread pool (each with a 1000ms timeout). Since librdkafka 2.10.0,
+        // these workers survive an unsubscribe→subscribe transition and can
+        // steal messages from the next subscription.
+        function _finishBootstrap(partition, offset) {
+            if (inFlightConsumes > 0) {
```
francoisferrand: if I understand correctly, the issue would be that the timer is called while we are processing the last message (`bootstrapId === lastBootstrapId`) - and initiate a new consume?

instead of waiting here (which allows the timer to hit again, with no guarantee of converging) and since there is no message afterwards: should we not fix the timer handling instead?

i.e. either
- use `setTimeout()` (i.e. schedule a single consume), and call it again on consumeCb if needed (e.g. if it did not get the initial message)
- introduce some flag to ensure we don't initiate another consume call if we did not yet consume (e.g. the same `inFlightConsumes` you have) - though practically would actually become a boolean...

delthas: OK, I wasn't sure if there was a particular reason for the setInterval but it seems not. Moved to a chained `setTimeout`.

lib/BackbeatConsumer.js
```diff
@@ -872,37 +876,53 @@ class BackbeatConsumer extends EventEmitter {
...
+        // unsubscribe→subscribe transition, stealing messages from
+        // the next subscription.
+        function consumeNext() {
+            self._consumer.consume(1, (err, messages) => {
+                if (err || bootstrapDone) {
+                    if (!bootstrapDone) {
+                        setTimeout(consumeNext, 200);
+                    }
+                    return undefined;
+                }
```
francoisferrand:
```suggestion
if (bootstrapDone) {
   return undefined;
}

if (err) {
   setTimeout(consumeNext, 200);
   return undefined;
}
```

lib/BackbeatConsumer.js
```diff
@@ -872,37 +876,53 @@ class BackbeatConsumer extends EventEmitter {
...
+                let matched = false;
+                messages.forEach(message => {
```
francoisferrand: instead of forEach and a variable, best to use `find()`:
- so we don't even try to process the message beyond the bootstrap
- make core more readable

```suggestion
cont receivedBootstrap = messages.find(message => {
   const bootstrapId = JSON.parse(message.value).bootstrapId;
   self._log.info('bootstraping backbeat consumer: received bootstrap message', {
      bootstrapId, topic: self._topic, groupId: self._groupId
   });
   return bootstrapId === lastBootstrapId;
});

if (!receivedBootstrap) {
   return setTimeout(consumeNext, 200);
}

self._log.info('backbeat consumer is bootstrapped', { topic: self._topic, groupId: self._groupId });

bootstrapDone = true;
clearInterval(producerTimer);
self._consumer.offsetsStore([{ topic: self._topic, partition: message.partition, offset: message.offset + 1 }]);
self._consumer.commit();
self._consumer.unsubscribe();
producer.close(() => {
   self._bootstrapping = false;
   self._onReady();
});
```

lib/BackbeatConsumer.js
```diff
@@ -872,37 +876,53 @@ class BackbeatConsumer extends EventEmitter {
...
+                    const bootstrapId = JSON.parse(message.value).bootstrapId;
+                    if (bootstrapId) {
+                        self._log.info('bootstraping backbeat consumer: ' +
+                            'received bootstrap message',
+                            { bootstrapId, topic: self._topic, groupId: self._groupId });
+                        if (bootstrapId === lastBootstrapId) {
+                            self._log.info('backbeat consumer is bootstrapped',
+                                { topic: self._topic, groupId: self._groupId });
+                            matched = true;
+                            bootstrapDone = true;
```
francoisferrand: is this variable still needed?

- we update it when processing the "result" of `consume()`
- since we are now guaranteed to have a single call to `consume()`, it is the current one
- thus no guard needed

lib/BackbeatConsumer.js
```diff
@@ -872,37 +876,53 @@ class BackbeatConsumer extends EventEmitter {
...
+                            clearInterval(producerTimer);
```
francoisferrand: `producerTimer` is a local variable, and does not seem to be initialized?

lib/BackbeatConsumer.js
```diff
@@ -872,37 +876,53 @@ class BackbeatConsumer extends EventEmitter {
...
-                            self._onReady();
-                        });
+        let bootstrapDone = false;
+        // Use chained setTimeout instead of setInterval to ensure
+        // only one consume() worker is in flight at a time. With
+        // setInterval, multiple C++ async workers can overlap, and
+        // since librdkafka 2.10.0 they survive an
+        // unsubscribe→subscribe transition, stealing messages from
+        // the next subscription.
+        function consumeNext() {
```
francoisferrand: since it runs the timer repeatedly until bootstrap received, we can't really call this function consume _next_... so maybe just `consume` ?

lib/BackbeatConsumer.js
delthas: Done — separated the checks and removed `bootstrapDone` entirely.

lib/BackbeatConsumer.js
delthas: Done — switched to `find()` and flattened the match handling out of the callback.

lib/BackbeatConsumer.js
delthas: Confirmed not needed — removed. With chained setTimeout there is exactly one `consume(1, cb)` in flight at a time. When the match is found, we do not schedule another `setTimeout(consume, 200)`, so the chain stops. There is no path where the callback fires again after the match.

lib/BackbeatConsumer.js
delthas: `producerTimer` is declared with `let` at the top of `_bootstrapConsumer()` and assigned inside the `producer.on("ready", ...)` callback (`producerTimer = setInterval(...)`). The `clearInterval(producerTimer)` call only runs from the consume callback, which can only fire after the consumer is ready, which happens inside the same setInterval callback — so `producerTimer` is always assigned before `clearInterval` is called. It is a closure over a shared local variable, not uninitialized access.

lib/BackbeatConsumer.js
delthas: Good point — renamed to just `consume`.

francoisferrand (approved):

lib/BackbeatConsumer.js
```diff
@@ -876,51 +876,46 @@ class BackbeatConsumer extends EventEmitter {
...
         // the next subscription.
-        function consumeNext() {
+        function consume() {
             self._consumer.consume(1, (err, messages) => {
-                if (err || bootstrapDone) {
-                    if (!bootstrapDone) {
-                        setTimeout(consumeNext, 200);
-                    }
+                if (err) {
+                    setTimeout(consume, 200);
```
benzekrimaha, line 893: Shouldn't we have a debug log here to see how many times we actually got an error ?

benzekrimaha (approved):

lib/BackbeatConsumer.js
delthas: Done — added a debug log with the error message, topic, and groupId.

PR #2729, opened 2026-04-01
Skip lifecycle index creation when disk space is insufficient

delthas:
## Summary

- Before attempting to create lifecycle v2 indexes, check available disk space and estimate the index creation cost.
- If `fsFreeSize < 3 * _id_ index size`, skip index creation and fall back to v1 listing.
- On error checking disk space, also fall back to v1 (consistent with existing error handling in `_indexesGetOrCreate`).

## Context

When a large bucket fills the MongoDB PV, operators need lifecycle to empty it. The conductor tries to create v2 indexes for the bucket, which either fails (ENOSPC) or — worse — **crashes the MongoDB shard**. We confirmed this on a test cluster: attempting index creation with 50MB free on a 10GB PV caused "Connection closed by peer" and the shard went into a crash loop (560+ restarts). Removing the filler files was required to recover.

This change adds a proactive disk space check using two MongoDB metadata calls run sequentially (~11ms total, no locks):
- `getDiskUsage()` (`dbStats`) — filesystem free space
- `getCollectionStats()` (`collStats`) — per-collection index sizes

## Why `3 * _id_` index size?

After running `compact` on a test collection with 1.37M objects, all three indexes (`_id_`, both lifecycle) are the same size (~60MB each). However, the `_id_` index is typically bloated from incremental inserts (101MB before compact vs 58MB after). Since each lifecycle index is freshly built at its compact size, `3 * _id_` provides a safe margin:
- When `_id_` is bloated (101MB): threshold = 303MB, actual cost ~120MB — conservative, safe
- When `_id_` is compact (58MB): threshold = 174MB, actual cost ~120MB — still safe

## Design decisions

- **Calls run sequentially** (`async.series`) to avoid creating extra MongoDB connections/sessions. Latency is not critical here.
- **Check happens per-bucket in `_indexesGetOrCreate`**, only for buckets that actually need index creation (indexes don't exist, auto-create enabled, not at concurrent limit). Most buckets already have indexes and return early.
- **Falls back to v1 on any error** — consistent with every other error path in this method.
- **Depends on Arsenal PR** scality/Arsenal#2605 for the fixed `getDiskUsage` and new `getCollectionStats` methods.

### DIFF
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -285,26 +285,57 @@ class LifecycleConductor {
                 return cb(null, lifecycleTaskVersions.v1);
             }
 
-            LifecycleMetrics.onLegacyTask(log, 'putBucketIndexes');
-
-            this.activeIndexingJobs.push({
-                bucket: task.bucketName,
-                indexes: indexesForFeature.lifecycle.v2,
-            });
-
-            return backbeatMetadataProxy.putBucketIndexes(
-                task.bucketName,
-                indexesForFeature.lifecycle.v2,
-                log,
-                err => {
-                    if (err) {
-                        log.warn('unable to create lifecycle indexes', {
-                            bucket: task.bucketName,
-                            error: err,
-                        });
-                    }
+            return async.series({
+                diskUsage: done => this._mongodbClient.getDiskUsage(done),
+                collStats: done => this._mongodbClient.getCollectionStats(
+                    task.bucketName, log, done),
+            }, (err, results) => {
+                if (err) {
+                    log.warn('unable to check disk space, skipping index creation', {
+                        bucket: task.bucketName,
+                        error: err,
+                    });
+                    LifecycleMetrics.onLegacyTask(log, 'diskSpaceCheckFailed');
                     return cb(null, lifecycleTaskVersions.v1);
+                }
+
+                const fsFreeSize = results.diskUsage.free;
+                // Each lifecycle index is roughly the same size as the
+                // _id_ index. We create two, so 3x gives a safe margin
+                // accounting for _id_ index bloat from incremental inserts.
+                const idIndexSize = results.collStats.indexSizes?._id_ || 0;
+
+                if (fsFreeSize < 3 * idIndexSize) {
+                    log.warn('insufficient disk space for index creation', {
+                        bucket: task.bucketName,
+                        fsFreeSize,
+                        threshold: 3 * idIndexSize,
+                    });
+                    LifecycleMetrics.onLegacyTask(log, 'insufficientDiskSpace');
+                    return cb(null, lifecycleTaskVersions.v1);
+                }
+
+                LifecycleMetrics.onLegacyTask(log, 'putBucketIndexes');
+
+                this.activeIndexingJobs.push({
+                    bucket: task.bucketName,
+                    indexes: indexesForFeature.lifecycle.v2,
                 });
+
+                return backbeatMetadataProxy.putBucketIndexes(
+                    task.bucketName,
+                    indexesForFeature.lifecycle.v2,
+                    log,
+                    err => {
+                        if (err) {
+                            log.warn('unable to create lifecycle indexes', {
+                                bucket: task.bucketName,
+                                error: err,
+                            });
+                        }
+                        return cb(null, lifecycleTaskVersions.v1);
+                    });
+            });
         });
     }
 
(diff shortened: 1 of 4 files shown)


extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -285,26 +285,55 @@ class LifecycleConductor {
...
+                    return cb(null, lifecycleTaskVersions.v1);
+                }
 
-            this.activeIndexingJobs.push({
-                bucket: task.bucketName,
-                indexes: indexesForFeature.lifecycle.v2,
-            });
+                const fsFreeSize = results.diskUsage.free;
+                const idIndexSize =
+                    (results.collStats.indexSizes && results.collStats.indexSizes._id_) || 0;
```
delthas: Confirmed in practice that the indexes created are about the same size as the default _id_ index.

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -285,26 +285,55 @@ class LifecycleConductor {
                 return cb(null, lifecycleTaskVersions.v1);
             }
 
-            LifecycleMetrics.onLegacyTask(log, 'putBucketIndexes');
+            return async.parallel({
```
francoisferrand: is it better to do in parallel, or sequentially?
- parallel may require creating 2 connections/sessions - which could be slower than just doing requests sequentially on the same connection/session
- latency is not an issue here

extensions/lifecycle/conductor/LifecycleConductor.js
francoisferrand:
```suggestion
                const idIndexSize = results.collStats.indexSizes?._id_ || 0;
```

francoisferrand: this is logical: since the index cover "all objects", the size of the index should be roughly sum of the length of every key (+/- compression)

- should this be added in a comment?
- can we really go in the `|| 0` case? when the collection is empty maybe?

francoisferrand (approved):

francoisferrand:
should https://github.com/scality/Arsenal/pull/2605 be bumped here first?
(will need to land before this actually works... and until then, it would probably crash...)

SylvainSenechal (approved):
Did this error ever happened in production ? What's the context that lead to this work, I don't see the linked ticket in the pr

I'm a bit surprised as you say : 
"The _id_ index size is used as an estimate for the lifecycle index cost. On a test cluster with 1M objects: _id_ = 76MB, each lifecycle index ≈ 60MB — so 2 * _id_ is a conservative estimate for both indexes combined." 

It means a bucket with 100 million objects would only take ~7gb, which is.. fine ? Or maybe I misunderstood something and we are talking about ram instead of disk storage. Maybe the bigger problem is not the storage, but simply adding index on millions of objects is hammering the db for tens of minutes which also affects other apis calls

extensions/lifecycle/conductor/LifecycleConductor.js
delthas: `_id_` is always set, but given that we have a `?` it's better to have `0` than a check against `null` below imo.

delthas:
> Did this error ever happened in production ? What's the context that lead to this work, I don't see the linked ticket in the pr
> 
> I'm a bit surprised as you say : "The _id_ index size is used as an estimate for the lifecycle index cost. On a test cluster with 1M objects: _id_ = 76MB, each lifecycle index ≈ 60MB — so 2 * _id_ is a conservative estimate for both indexes combined."
> 
> It means a bucket with 100 million objects would only take ~7gb, which is.. fine ? Or maybe I misunderstood something and we are talking about ram instead of disk storage. Maybe the bigger problem is not the storage, but simply adding index on millions of objects is hammering the db for tens of minutes which also affects other apis calls

The use case is for when your disk is (almost) full and you want to create a workflow to empty the bucket, but you can't because there isn't enough space to create the workflow index.

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -285,26 +285,57 @@ class LifecycleConductor {
...
-                indexesForFeature.lifecycle.v2,
-                log,
-                err => {
-                    if (err) {
-                        log.warn('unable to create lifecycle indexes', {
-                            bucket: task.bucketName,
-                            error: err,
-                        });
-                    }
+                const fsFreeSize = results.diskUsage.free;
```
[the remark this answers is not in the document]
delthas: `free` == `available` in Arsenal.

delthas:
> when `indexSizes._id_` is missing or zero

Guaranteed not to happen by Mongo.

PR #2731, opened 2026-04-08
Harden onEntryCommittable against ERR__STATE

SylvainSenechal:
**TL;DR** — defense-in-depth around `onEntryCommittable`. The known sync and async races that caused the BB-758 crash are closed structurally by BB-776 (#2739, merged) and BB-777 (#2740, merged). This PR is the small safety net for residual edge cases: late callbacks after the outer-timeout disconnect, the BB-772 poison-pill paths, and any future race we haven't enumerated. Log and move on instead of crashing.

## Background

BB-758 was a worker crash with `ERR__STATE` (-172) when `onEntryCommittable` called `offsetsStore` after the partition had been revoked. Two structural fixes have since landed:

- **BB-776 (#2739)** — closes the synchronous tail: `TaskScheduler` now invokes the per-task `done` callback before triggering the drain check, so the rebalance handler can't observe an "idle" queue while the last `offsetsStore` call is still in flight on its caller's stack.
- **BB-777 (#2740)** — closes the asynchronous tail: the rebalance handler now waits for the `OffsetLedger` to drain (in addition to the processing queue) before unassigning. Deferred `onEntryCommittable` callbacks (the `committable: false` flow used by replication) land safely before the partition is released.

## What this PR adds

In `BackbeatConsumer.onEntryCommittable`:

1. An additional `this._consumer.isConnected()` guard alongside the existing `!this.isPaused()` check.
2. A `try/catch` around `offsetsStore()` that logs at error level instead of crashing.

The offset will be re-committed after the partition is re-assigned, so swallowing the error costs at most a duplicate delivery.

## Why keep it after BB-776 + BB-777

A few cases still slip past the structural fixes:

- **Late callback after the outer-timeout disconnect.** If a deferred ack (e.g. a Kafka status producer's delivery callback) lands more than `max.poll.interval - 1s` after the rebalance starts, BB-777 has already disconnected the consumer. The next `onEntryCommittable` would crash.
- **BB-772 poison-pill paths.** `GarbageCollectorTask` and `LifecycleColdStatusArchiveTask` use `committable: false` without ever calling `onEntryCommittable`. They trigger the outer timeout today. Until BB-772 lands, the residual `onEntryCommittable` calls from other tasks in the same process can race the disconnect.
- **Future races we haven't enumerated.** The `try/catch` is ~5 lines and costs nothing in the happy path.

## Tests

5 unit tests on `onEntryCommittable` cover the new guards and catch:
- calls `offsetsStore` when active + connected
- does not call when paused (unsubscribed)
- does not call when not connected
- does not throw + logs error on `ERR__STATE`
- does not throw + logs error on unexpected error

The original functional test was dropped after the rebase: with BB-777 in place, it would deadlock on `consumer1.once('unassign', ...)` (the unassign now waits for the ledger to drain, and the test only drains the ledger inside the `unassign` callback) and "pass" only after the 5-minute outer-timeout fires. The unit tests cover the change directly and cheaply.

Issue: BB-758

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -882,10 +882,24 @@ class BackbeatConsumer extends EventEmitter {
         });
         // ensure consumer is active when calling offsetsStore() on
         // it, to avoid raising an exception (invalid state)
-        // TODO : potential issue here, see BB-758
-        if (committableOffset !== null && !this.isPaused()) {
-            this._consumer.offsetsStore([{ topic, partition,
-                                           offset: committableOffset }]);
+        if (committableOffset !== null && this._consumer.isConnected() && !this.isPaused()) {
+            try {
+                this._consumer.offsetsStore([{ topic, partition, offset: committableOffset }]);
+            } catch (e) {
+                // offsetsStore() should not throw given the guards above;
+                // if it does (e.g. ERR__STATE on a still-connected
+                // consumer whose partition was revoked), log as error
+                // but do not crash — the message will be re-delivered
+                // and committed by whichever consumer holds the
+                // partition next.
+                this._log.error('offsetsStore failed', {
+                    error: e.toString(),
+                    topic,
+                    partition,
+                    offset: committableOffset,
+                    groupId: this._groupId,
+                });
+            }
         }
     }
 
--- a/tests/unit/backbeatConsumer.js
+++ b/tests/unit/backbeatConsumer.js
@@ -2,6 +2,7 @@ const assert = require('assert');
 const sinon = require('sinon');
 
 const BackbeatConsumer = require('../../lib/BackbeatConsumer');
+const { CODES } = require('node-rdkafka');
 
 const { kafka } = require('../config.json');
 const { BreakerState } = require('breakbeat').CircuitBreaker;
@@ -235,6 +236,80 @@ describe('backbeatConsumer', () => {
                 assert.strictEqual(consumer._tasksCompletedSinceLastConsume, false);
                 done();
             }, 500);
+        });
+    });
+
+    describe('onEntryCommittable', () => {
+        let consumer;
+        let mockConsumer;
+
+        const entry = {
+            topic: 'my-test-topic',
+            partition: 2,
+            offset: 280,
+            key: null,
+            timestamp: Date.now(),
+        };
+
+        beforeEach(() => {
+            consumer = new BackbeatConsumerMock({
+                kafka,
+                groupId: 'unittest-group',
+                topic: 'my-test-topic',
+            });
+
+            mockConsumer = {
+                offsetsStore: sinon.stub(),
+                subscription: sinon.stub().returns(['my-test-topic']),
+                isConnected: sinon.stub().returns(true),
+            };
+            consumer._consumer = mockConsumer;
+
+            // pre-register the offset as consumed so onOffsetProcessed returns a value
+            consumer._offsetLedger.onOffsetConsumed(entry.topic, entry.partition, entry.offset);
+        });
+
+        afterEach(() => {
+            sinon.restore();
+        });
+
+        it('should call offsetsStore when consumer is active and connected', () => {
+            consumer.onEntryCommittable(entry);
+            assert(mockConsumer.offsetsStore.calledOnce);
+        });
+
+        it('should not call offsetsStore when consumer is paused (unsubscribed)', () => {
+            mockConsumer.subscription.returns([]);
+            consumer.onEntryCommittable(entry);
+            assert(mockConsumer.offsetsStore.notCalled);
+        });
+
+        it('should not call offsetsStore when consumer is not connected', () => {
+            mockConsumer.isConnected.returns(false);
+            consumer.onEntryCommittable(entry);
+            assert(mockConsumer.offsetsStore.notCalled);
+        });
+
+        it('should not throw and always log at error level when offsetsStore throws', () => {
+            const errState = new Error('Local: Erroneous state');
+            errState.code = CODES.ERRORS.ERR__STATE;
+            mockConsumer.offsetsStore.throws(errState);
+
+            const errorSpy = sinon.spy(consumer._log, 'error');
+
+            assert.doesNotThrow(() => consumer.onEntryCommittable(entry));
+            assert(errorSpy.calledOnce);
+        });
+
+        it('should not throw and log at error level when offsetsStore throws an unexpected error', () => {
+            const unexpectedErr = new Error('unexpected kafka error');
+            unexpectedErr.code = -1;
+            mockConsumer.offsetsStore.throws(unexpectedErr);
+
+            const errorSpy = sinon.spy(consumer._log, 'error');
+
+            assert.doesNotThrow(() => consumer.onEntryCommittable(entry));
+            assert(errorSpy.calledOnce);
         });
     });
 


lib/BackbeatConsumer.js
```diff
@@ -823,10 +823,23 @@ class BackbeatConsumer extends EventEmitter {
...
         // it, to avoid raising an exception (invalid state)
-        // TODO : potential issue here, see BB-758
-        if (committableOffset !== null && !this.isPaused()) {
-            this._consumer.offsetsStore([{ topic, partition,
-                                           offset: committableOffset }]);
+        if (committableOffset !== null && !this.isPaused() && this._consumer.isConnected()) {
+            try {
+                this._consumer.offsetsStore([{ topic, partition, offset: committableOffset }]);
+            } catch (e) {
+                // ERR__STATE (-172) means the consumer is not in a valid state
```
SylvainSenechal: So this stuff seems to happen on rebalance or weird kafka status. 

- Before, the pod would crash, the offset wouldn't be commited, and another one pod would reconsume and re process it
- Now : the pod doesn't crash, but the offset is still not commited. and reconsumed later by another pod, or maybe the same pod since now it doesn't crash anymore

francoisferrand: do we know when this happens?

in order to properly handle rebalance, we actually _need_ to commit before the rebalance (hence the logic implemented here). If there are some cases where it does not work, we need to understand why/how, and not only mask the problem....

(there will always be corner cases where we cannot commit and will re-play the message, that is fine... but we need to understand why/when it happens, to make sure this is really the unavoidable corner case and not a deeper issue)

lib/BackbeatConsumer.js
```diff
@@ -823,10 +823,23 @@ class BackbeatConsumer extends EventEmitter {
...
+                // to store offsets (e.g. during a rebalance, shutdown, or other
+                // state transition) — treat as non-fatal and log at debug level.
+                const logger = this._consumer.isConnected() &&
+                    e.code !== kafka.CODES.ERRORS.ERR__STATE
+                    ? this._log.error : this._log.debug;
```
francoisferrand: this is never a debug log : we don't expect this to happen at all... So always ouptutting an error log may be appropriate

lib/BackbeatConsumer.js
francoisferrand: could the root cause be the same as https://github.com/scality/backbeat/pull/2728/changes/11fe6eee361cd509a3002dc2f87c2a56fb368c95 ?

lib/BackbeatConsumer.js
```diff
@@ -823,10 +823,23 @@ class BackbeatConsumer extends EventEmitter {
         });
         // ensure consumer is active when calling offsetsStore() on
         // it, to avoid raising an exception (invalid state)
-        // TODO : potential issue here, see BB-758
-        if (committableOffset !== null && !this.isPaused()) {
-            this._consumer.offsetsStore([{ topic, partition,
-                                           offset: committableOffset }]);
+        if (committableOffset !== null && !this.isPaused() && this._consumer.isConnected()) {
+            try {
```
francoisferrand, line 886: since you added the `isConnected()` guard, do we also need to catch the exception?

lib/BackbeatConsumer.js
SylvainSenechal: I don't really know when, and more importantly why this happen :/ 
Maybe it could be interesting to wait for the pr from Thomas to be merged and run in zenko ci to see if this issue still exists

lib/BackbeatConsumer.js
SylvainSenechal: I think we could still have cases where is connected returns true, but then inside offsetStore something wrong happens 🤔

lib/BackbeatConsumer.js
SylvainSenechal: Okay, I repulled the zenko ci logs from a successful run and asked claude to analyze and pay attention to the timing of the logs. 

What was found was that this is happening during a rebalance, we first have one pod handling all the partitions, then the operator deployed another pod, and and when that new pod joined the Kafka consumer group, a rebalance was triggered. 
The old pod had 5 partitions revoked and was re-assigned only 2 of them. The problem is: some tasks that were already in-flight finished processing after the revoke happened. When their callback tried to call offsetsStore() to mark the offset as ready to commit, librdkafka threw ERR__STATE because those partitions were no longer assigned to this pod so it crashed

tests/unit/backbeatConsumer.js
```diff
@@ -238,6 +238,83 @@ describe('backbeatConsumer', () => {
...
+
+        it('should not call offsetsStore when consumer is not connected', () => {
+            mockConsumer.isConnected.returns(false);
+            consumer.onEntryCommittable(entry);
+            assert(mockConsumer.offsetsStore.notCalled);
+        });
+
+        it('should not throw and log at debug level when offsetsStore throws ERR__STATE', () => {
+            const errState = new Error('Local: Erroneous state');
+            errState.code = -172; // ERR__STATE
```
maeldonn: Why not using the same constant used in production code ?

maeldonn (approved):

lib/BackbeatConsumer.js
francoisferrand: so the problem is actually much deeper: the whole logic here is done to delay the _whole rebalance_ until we commit :
- which we normally have the time (not sure about the numbers, but rebalance can last a significant time - unless some kafka params changed)
- unless we have so-called "slow tasks", i.e. messages which take more than the time limit to process. So there is an issue in the processing: either timeouts do not match our target (and esp. on a possibly overloaded CI), or we really hit some corner case where processing is stuck

Good thing is it happens on CI: so whatever the reason, we have a way to reproduce, somehow...

lib/BackbeatConsumer.js
francoisferrand: do we ? this is javascript, so single threaded...

francoisferrand (commented):
seems like there is a deeper issue, we should investigate more to make sure we don't just hide another bug

tests/unit/backbeatConsumer.js
SylvainSenechal: can
i switched to CODES.ERRORS.ERR__STATE

SylvainSenechal:
> seems like there is a deeper issue, we should investigate more to make sure we don't just hide another bug

Ok considering you other comments, I will try to setup a functional test to see if we can first reproduce the errors, if we can reproduce it, we will be able to work on a fix and really verify that it works

SylvainSenechal:
@francoisferrand 
I added a test that setup conditions to reproduces this err -172 that is now catched with the try catch. I also run this test locally with and without the try catch that I added and confirmed that this error is reproductible when a rebalance happens

francoisferrand (commented):
> I added a test that setup conditions to reproduces this err -172 that is now catched with the try catch. I also run this test locally with and without the try catch that I added and confirmed that this error is reproductible when a rebalance happens

Nice! But still quite a few questions:

- No clear in your message: do we actually need the try/catch? (if we do, then probably we don't need the isConnected/... guard?)
- So what is the scenario?
- There may be a deeper issue, we need to make sure we don't just "handle" another bug: i.e. lile some message taking too long to process and triggering the bug, or another bug in our rebalance handling which must be fixed to ensure we commit on time...

lib/BackbeatConsumer.js
```diff
@@ -823,10 +823,22 @@ class BackbeatConsumer extends EventEmitter {
...
-            this._consumer.offsetsStore([{ topic, partition,
-                                           offset: committableOffset }]);
+        if (committableOffset !== null && this._consumer.isConnected() && !this.isPaused()) {
+            try {
+                this._consumer.offsetsStore([{ topic, partition, offset: committableOffset }]);
+            } catch (e) {
+                // offsetsStore() should not throw given the guards above;
+                // if it does (e.g. ERR__STATE race during a rebalance or
+                // shutdown), log as error but do not crash — the offset will
+                // be re-committed after the partition is re-assigned.
```
delthas: Message is a bit unclear: the offset is not committed, the other consumer will reprocess the same messages, and then commit the offset once it's done. There's no hidden "re-commit after re-assignment" thing.

delthas (approved):

delthas (changes requested):
Now that https://github.com/scality/backbeat/pull/2739 and https://github.com/scality/backbeat/pull/2740 are merged, I suggest we keep only the lib/BackbeatConsumer.js part, and merge. Useful to have this try/catch just in case. (Tests already added in other PRs)

lib/BackbeatConsumer.js
delthas: The two guards protect different failure modes: `isConnected()` catches calls into a disconnected consumer (e.g. after the BB-777 outer-timeout disconnect fires), while the `try/catch` catches `ERR__STATE` on a still-connected consumer whose partition was revoked — which is the original BB-758 scenario. Single-threaded JS doesn't help here because the state lives on the librdkafka side and changes via rebalance callbacks between our guard and the `offsetsStore()` call. With BB-776 (#2739) and BB-777 (#2740) now merged, the structural races are closed and this catch becomes the residual safety net (see the updated PR description).

lib/BackbeatConsumer.js
delthas: Fixed in [`c52fbccb`](https://github.com/scality/backbeat/pull/2731/commits/c52fbccb). The comment now says the message will be re-delivered and committed by whichever consumer holds the partition next, with no implication of a same-consumer re-commit.

lib/BackbeatConsumer.js
```diff
@@ -882,10 +882,22 @@ class BackbeatConsumer extends EventEmitter {
         });
         // ensure consumer is active when calling offsetsStore() on
         // it, to avoid raising an exception (invalid state)
-        // TODO : potential issue here, see BB-758
-        if (committableOffset !== null && !this.isPaused()) {
-            this._consumer.offsetsStore([{ topic, partition,
-                                           offset: committableOffset }]);
+        if (committableOffset !== null && this._consumer.isConnected() && !this.isPaused()) {
```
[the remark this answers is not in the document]
delthas: The short-circuit ordering (`isConnected()` before `isPaused()`) is intentional. Even if `subscription()` did throw on a disconnected consumer, the surrounding `try/catch` would catch it. Resolving as a nit.

delthas (approved):

PR #2736, opened 2026-04-30
Properly get location in lifecycle metrics

benzekrimaha:
Issue: BB-721

## Context analysis

Lifecycle execution metrics (`s3_lifecycle_started_total` / `s3_lifecycle_completed_total`, emitted via `LifecycleMetrics.onLifecycleStarted` / `onLifecycleCompleted`) carry a `location` label used by Grafana panels and alerts to break lifecycle activity down by storage backend.

For **expiration of a restored cold object**, that label was wrong: the completion metric reported the object's *warm* location instead of the cold location the archive actually lives on. On platforms with cold storage this made per-location panels and alerts unusable for expiration of restored data.

## Root cause

`LifecycleDeleteObjectTask` computed the metric location as:

```js
const location = this.objectMD?.dataStoreName
    || entry.getAttribute('details.dataStoreName');
```

For a restored cold object, CloudServer's restore-completion path (`overwritingVersioning` in `cloudserver:lib/api/apiUtils/object/versioning.js`) deliberately:

- **preserves** `x-amz-storage-class` as the cold location name (e.g. `location-dmf-v1`), and
- **resets** `dataStoreName` to the warm location where the restored data was actually written.

So `objectMD.dataStoreName` is the warm location, while the queued action entry's `details.dataStoreName` (set at queue time in `LifecycleTask` from the listing's `StorageClass`, which mirrors `x-amz-storage-class`) is the cold location. The `objectMD || entry` precedence therefore picked the wrong one and mislabelled the metric.

`LifecycleUpdateExpirationTask` was not affected — it already resolves the metric location from `target.location` (the cold location).

## Fix implemented in this PR

- **`LifecycleDeleteObjectTask`**: drop the `objectMD.dataStoreName` override and use `entry.getAttribute('details.dataStoreName')` directly, so the completion metric reports the same location the action was queued against.

- **`LifecycleTask._compareObject`**: make the queued `details.dataStoreName` an explicit, authoritative source instead of reusing `obj.StorageClass` from the listing verbatim. On the `HeadObject` call already issued per candidate, attach `x-amz-scal-archive-info: true` (via SDK middleware) so the response always carries an explicit storage class — including the plain `STANDARD` case, which CloudServer's default `HeadObject` path strips (`collectResponseHeaders`) but `setArchiveInfoHeaders` always sets. Propagate the resulting `StorageClass` onto the candidate object as `data.StorageClass || obj.StorageClass`, so downstream queuing sees a consistent, non-empty value regardless of whether the object is on the default location, a warm non-default location, cold, or restored-from-cold.

- **`extensions/lifecycle/bucketProcessor/policy.json`**: grant `scality:GetObjectArchiveInfo` so the bucket processor is allowed to send the archive-info request header.

- Propagate the resulting StorageClass onto the candidate object (data.StorageClass — always set by CloudServer when the archive-info header is present, never STANDARD)
## Tests

New / updated unit tests:

- `LifecycleTask.spec.js`: `HeadObject` is sent with `x-amz-scal-archive-info: true`, the returned `StorageClass` is propagated onto the candidate, and the queued entry's `details.dataStoreName` reflects it (including for restored / cold objects).
- `LifecycleDeleteObjectTask.spec.js`: expiration completion metric uses the queued `details.dataStoreName`, no longer influenced by `objectMD.dataStoreName`; covers the restored-cold-object case.
- `LifecycleUpdateExpirationTask.spec.js`: regression assertion that `onLifecycleStarted` / `onLifecycleCompleted` are called with the cold location for restored-object expiration. `LifecycleBucketProcessorPolicy.spec.js`: policy exposes the new `scality:GetObjectArchiveInfo` action.
=> the _compareObject test is now a four-scenario matrix (hot default / hot non-default / cold / restored)

### DIFF
--- a/extensions/lifecycle/bucketProcessor/policy.json
+++ b/extensions/lifecycle/bucketProcessor/policy.json
@@ -14,7 +14,8 @@
                 "s3:GetObjectVersionTagging",
                 "s3:GetObject",
                 "s3:GetObjectVersion",
-                "s3:ReplicateObject"
+                "s3:ReplicateObject",
+                "scality:GetObjectArchiveInfo"
             ],
             "Resource": [
                 "*"
--- a/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
+++ b/extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
@@ -195,7 +195,12 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
 
         const actionType = entry.getActionType();
         const transitionTime = entry.getAttribute('transitionTime');
-        const location = this.objectMD?.dataStoreName || entry.getAttribute('details.dataStoreName');
+        // The entry attribute carries the location resolved at queue time
+        // (the cold class for cold/restored objects). Fall back to the object
+        // metadata for entries missing it, e.g. expired object delete markers
+        // queued by LifecycleTaskV2, which do not set it. objectMD is only
+        // fetched for versioned deleteObject actions, hence the optional call.
+        const location = entry.getAttribute('details.dataStoreName') || this.objectMD?.getDataStoreName();
 
         let reqMethod = 'deleteObject';
         let actionMethod = this._deleteObject.bind(this);
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -59,6 +59,19 @@ const MAX_RETRIES = 4;
 // parallel tasks, so the total delay of retries should about 1m30s.
 const MAX_RETRIES_TOTAL = CONCURRENCY_DEFAULT * MAX_RETRIES * 10;
 
+function attachArchiveInfoHeader(command) {
+    command.middlewareStack.add(next => async args => {
+        if (args.request && args.request.headers) {
+            // eslint-disable-next-line no-param-reassign
+            args.request.headers['x-amz-scal-archive-info'] = 'true';
+        }
+        return next(args);
+    }, {
+        step: 'build',
+        name: 'attachArchiveInfoHeader',
+    });
+}
+
 /**
  * compare 2 version by their stale dates returning:
  * - LT (-1) if v1 is less than v2
@@ -1588,12 +1601,20 @@ class LifecycleTask extends BackbeatTask {
             params.VersionId = obj.VersionId;
         }
         const command = new HeadObjectCommand(params);
+        attachArchiveInfoHeader(command);
         attachReqUids(command, log.getSerializedUids());
         return this.s3target.send(command)
             .then(data => {
                 LifecycleMetrics.onS3Request(log, 'headObject', 'bucket', null);
                 const object = Object.assign({}, obj,
-                    { LastModified: data.LastModified });
+                    {
+                        LastModified: data.LastModified,
+                        // With the archive-info header, CloudServer always
+                        // returns the real storage class (never STANDARD):
+                        // the preserved cold class for cold/restored objects,
+                        // the data-store name otherwise.
+                        StorageClass: data.StorageClass,
+                    });
 
                 // There is an order of importance in cases of conflicts
                 // Expiration and NoncurrentVersionExpiration should be priority
--- a/tests/unit/lifecycle/LifecycleDeleteObjectTask.spec.js
+++ b/tests/unit/lifecycle/LifecycleDeleteObjectTask.spec.js
@@ -7,6 +7,7 @@ const { ObjectMD } = require('arsenal').models;
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
 const LifecycleDeleteObjectTask = require(
     '../../../extensions/lifecycle/tasks/LifecycleDeleteObjectTask');
+const { LifecycleMetrics } = require('../../../extensions/lifecycle/LifecycleMetrics');
 
 const day = 1000 * 60 * 60 * 24;
 
@@ -51,6 +52,7 @@ describe('LifecycleDeleteObjectTask', () => {
     });
 
     afterEach(() => {
+        sinon.restore();
         backbeatMdProxyClient.setError(null);
     });
 
@@ -275,6 +277,55 @@ describe('LifecycleDeleteObjectTask', () => {
             assert.ifError(err);
             assert.strictEqual(backbeatClient.times.deleteObjectFromExpiration, 1);
             assert.strictEqual(s3Client.calls.deleteObject, 0);
+            done();
+        });
+    });
+
+
+    it('should emit expiration metrics with the cold location carried by the entry', done => {
+        const startedMetric = sinon.stub(LifecycleMetrics, 'onLifecycleStarted');
+        const completedMetric = sinon.stub(LifecycleMetrics, 'onLifecycleCompleted');
+        // Simulate a restored cold object: the object metadata carries the
+        // warm location the restored copy was written to, while the entry
+        // carries the cold location resolved at queue time. The metric must
+        // use the entry location, not the metadata one.
+        objMd.setDataStoreName('warm-restored-location');
+        const entry = ActionQueueEntry.create('deleteObject')
+            .setAttribute('target.owner', 'testowner')
+            .setAttribute('target.bucket', 'testbucket')
+            .setAttribute('target.accountId', 'testid')
+            .setAttribute('target.key', 'testkey')
+            .setAttribute('target.version', 'testversion')
+            .setAttribute('details.dataStoreName', 'cold-location')
+            .setAttribute('transitionTime', Date.now() - day);
+        backbeatClient.setResponse(null, {});
+        task.processActionEntry(entry, err => {
+            assert.ifError(err);
+            assert.strictEqual(startedMetric.firstCall.args[2], 'cold-location');
+            assert.strictEqual(completedMetric.firstCall.args[2], 'cold-location');
+            done();
+        });
+    });
+
+    it('should fall back to the metadata location when the entry does not carry one', done => {
+        const startedMetric = sinon.stub(LifecycleMetrics, 'onLifecycleStarted');
+        const completedMetric = sinon.stub(LifecycleMetrics, 'onLifecycleCompleted');
+        // Entries queued without details.dataStoreName (e.g. expired object
+        // delete markers from LifecycleTaskV2): the metric falls back to the
+        // location from the object metadata fetched during processing.
+        objMd.setDataStoreName('md-location');
+        const entry = ActionQueueEntry.create('deleteObject')
+            .setAttribute('target.owner', 'testowner')
+            .setAttribute('target.bucket', 'testbucket')
+            .setAttribute('target.accountId', 'testid')
+            .setAttribute('target.key', 'testkey')
+            .setAttribute('target.version', 'testversion')
+            .setAttribute('transitionTime', Date.now() - day);
+        backbeatClient.setResponse(null, {});
+        task.processActionEntry(entry, err => {
+            assert.ifError(err);
+            assert.strictEqual(startedMetric.firstCall.args[2], 'md-location');
+            assert.strictEqual(completedMetric.firstCall.args[2], 'md-location');
             done();
         });
     });
--- a/tests/unit/lifecycle/LifecycleTask.spec.js
+++ b/tests/unit/lifecycle/LifecycleTask.spec.js
@@ -10,6 +10,8 @@ const LifecycleTask = require(
     '../../../extensions/lifecycle/tasks/LifecycleTask');
 const LifecycleTaskV2 = require(
     '../../../extensions/lifecycle/tasks/LifecycleTaskV2');
+const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
+const { LifecycleMetrics } = require('../../../extensions/lifecycle/LifecycleMetrics');
 const fakeLogger = require('../../utils/fakeLogger');
 const { withActiveSpan } = require('../../utils/withActiveSpan');
 const { timeOptions } = require('../../functional/lifecycle/configObjects');
@@ -2470,6 +2472,121 @@ describe('lifecycle task helper methods', () => {
             });
         });
     });
+
+    describe('_sendObjectAction', () => {
+        it('should emit trigger metrics with the entry location', done => {
+            const lifecycleTask = new LifecycleTask(lp);
+            const sentEntries = [];
+            lifecycleTask.objectTasksTopic = 'object-topic';
+            lifecycleTask.circuitBreakers = {
+                tripped: sinon.stub().returns(false),
+            };
+            lifecycleTask.producer = {
+                sendToTopic: (topic, entries, cb) => {
+                    sentEntries.push({ topic, entries });
+                    cb();
+                },
+            };
+            const triggeredMetric = sinon.stub(LifecycleMetrics, 'onLifecycleTriggered');
+
+            const entry = ActionQueueEntry.create('deleteObject')
+                .setAttribute('target.owner', 'test-owner')
+                .setAttribute('target.bucket', 'test-bucket')
+                .setAttribute('target.accountId', 'test-account')
+                .setAttribute('target.key', 'test-key')
+                .setAttribute('details.dataStoreName', 'us-east-1')
+                .setAttribute('transitionTime', Date.now() - HOUR);
+            lifecycleTask._sendObjectAction(entry, err => {
+                assert.ifError(err);
+                assert.strictEqual(entry.getAttribute('details.dataStoreName'), 'us-east-1');
+                assert.strictEqual(lifecycleTask.circuitBreakers.tripped.firstCall.args[1], 'us-east-1');
+                assert.strictEqual(triggeredMetric.firstCall.args[3], 'us-east-1');
+                assert.strictEqual(sentEntries.length, 1);
+                done();
+            });
+        });
+    });
+
+    describe('_compareObject location metrics', () => {
+        // With the x-amz-scal-archive-info request header, CloudServer's
+        // HeadObject always returns the real storage class: the preserved
+        // cold class for cold and restored objects, the data-store name
+        // otherwise (never the STANDARD placeholder the listing may carry).
+        [
+            {
+                desc: 'hot object on the default location',
+                listedStorageClass: 'STANDARD',
+                headStorageClass: 'us-east-1',
+            },
+            {
+                desc: 'hot object on a non-default location',
+                listedStorageClass: 'site-azure',
+                headStorageClass: 'site-azure',
+            },
+            {
+                desc: 'cold object',
+                listedStorageClass: 'location-dmf-v1',
+                headStorageClass: 'location-dmf-v1',
+            },
+            {
+                // restored objects keep the cold class in
+                // x-amz-storage-class even though the restored copy lives
+                // on a warm location
+                desc: 'restored cold object',
+                listedStorageClass: 'location-dmf-v1',
+                headStorageClass: 'location-dmf-v1',
+            },
+        ].forEach(({ desc, listedStorageClass, headStorageClass }) => {
+            it(`should queue the expiration of a ${desc} with the ` +
+            'archive-info storage class', done => {
+                class LifecycleTaskMock extends LifecycleTask {
+                    _sendObjectAction(entry, cb) {
+                        this.latestEntry = entry;
+                        return cb();
+                    }
+                }
+
+                const lifecycleTask = new LifecycleTaskMock(lp);
+                lifecycleTask.s3target = {
+                    send: sinon.stub().resolves({
+                        LastModified: new Date().toISOString(),
+                        StorageClass: headStorageClass,
+                    }),
+                };
+
+                const bucketData = {
+                    target: {
+                        owner: 'test-owner',
+                        bucket: 'test-bucket',
+                        accountId: 'test-account',
+                    },
+                    details: {},
+                };
+                const rules = {
+                    Expiration: {
+                        Date: new Date(Date.now() - DAY),
+                    },
+                };
+                const listedObject = Object.assign({}, OBJECT, {
+                    StorageClass: listedStorageClass,
+                });
+
+                lifecycleTask._compareObject(bucketData, listedObject, rules, fakeLogger, err => {
+                    assert.ifError(err);
+                    assert.strictEqual(lifecycleTask.s3target.send.calledOnce, true);
+                    const command = lifecycleTask.s3target.send.firstCall.args[0];
+                    assert(command.middlewareStack.identify()
+                        .includes('attachArchiveInfoHeader - build'));
+                    assert.strictEqual(
+                        lifecycleTask.latestEntry.getAttribute('details.dataStoreName'),
+                        headStorageClass
+                    );
+                    done();
+                });
+            });
+        });
+    });
+
 });
 
 describe('LifecycleTask trace-context propagation', () => {
--- a/tests/unit/lifecycle/LifecycleBucketProcessorPolicy.spec.js
+++ b/tests/unit/lifecycle/LifecycleBucketProcessorPolicy.spec.js
@@ -1,0 +1,14 @@
+const assert = require('assert');
+
+const bucketProcessorPolicy = require('../../../extensions/lifecycle/bucketProcessor/policy.json');
+
+describe('LifecycleBucketProcessor policy', () => {
+    it('should allow archive info reads for lifecycle metric location resolution', () => {
+        const actions = bucketProcessorPolicy.Statement
+            .find(statement => statement.Sid === 'LifecycleExpirationBucketProcessor')
+            .Action;
+
+        assert(actions.includes('scality:GetObjectArchiveInfo'));
+        assert(!actions.includes('s3:GetBucketLocation'));
+    });
+});
--- a/tests/unit/lifecycle/LifecycleUpdateExpirationTask.spec.js
+++ b/tests/unit/lifecycle/LifecycleUpdateExpirationTask.spec.js
@@ -1,10 +1,12 @@
 const assert = require('assert');
+const sinon = require('sinon');
 const werelogs = require('werelogs');
 
 const { ObjectMD } = require('arsenal').models;
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
 const LifecycleUpdateExpirationTask = require(
     '../../../extensions/lifecycle/tasks/LifecycleUpdateExpirationTask');
+const { LifecycleMetrics } = require('../../../extensions/lifecycle/LifecycleMetrics');
 
 const {
     GarbageCollectorProducerMock,
@@ -60,7 +62,13 @@ describe('LifecycleUpdateExpirationTask', () => {
         task = new LifecycleUpdateExpirationTask(objectProcessor);
     });
 
+    afterEach(() => {
+        sinon.restore();
+    });
+
     it('should expire restored object', done => {
+        const startedMetric = sinon.stub(LifecycleMetrics, 'onLifecycleStarted');
+        const completedMetric = sinon.stub(LifecycleMetrics, 'onLifecycleCompleted');
         const requestedAt = new Date();
         const restoreCompletedAt = new Date();
         const expireDate = new Date();
@@ -94,6 +102,8 @@ describe('LifecycleUpdateExpirationTask', () => {
             const receivedGcEntry = gcProducer.getReceivedEntry();
             assert.strictEqual(receivedGcEntry.getActionType(), 'deleteData');
             assert.deepStrictEqual(receivedGcEntry.getAttribute('target.locations'), oldLocation);
+            assert.strictEqual(startedMetric.firstCall.args[2], 'location-dmf-v1');
+            assert.strictEqual(completedMetric.firstCall.args[2], 'location-dmf-v1');
             done();
         });
     });


extensions/lifecycle/util/lifecycleLocation.js
```diff
@@ -0,0 +1,50 @@
...
+    if (!objectMD) {
+        return undefined;
+    }
+    if (typeof objectMD[getterName] === 'function') {
+        return objectMD[getterName]();
+    }
+    return objectMD[fieldName];
+}
+
+function isRealLocation(location) {
```
SylvainSenechal: Still reading the pr, but I dont fully understand the link between the function name and what its doing.

Basically checking that a given location is non null and not equal to 'STANDARD'. So why not name it "isStandardLocation" or "isNotStandardLocation" ?

Reading a bit further, I think the standard location is some kind of default value, and we try to resolve some locations, so maybe the intent is "isResolvedLocations" ? I think some of this can be documented by code comments too

extensions/lifecycle/util/lifecycleLocation.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const STANDARD_LOCATION = 'STANDARD';
+
+function _getObjectMDValue(objectMD, getterName, fieldName) {
```
SylvainSenechal: Haven't digged too much into this, but I'm surprised we have to create a helper function for this, since there must already be some existing code trying to access object metadatas, I wonder how it's being done ?

Don't we have getters functions in Arsenal for this already ?

Looking a bit more into it : The function is only called locally from resolveLifecycleMetricObjectLocation, we can check that objectMD is non nil directly at the beginning of resolveLifecycleMetricObjectLocation, 
then when it's not nil, I think we don't need this helper and can directly call
objectMD.getArchive()
objectMD..getDatastoreName()
etc

extensions/lifecycle/util/lifecycleLocation.js
```diff
@@ -0,0 +1,50 @@
...
+
+function isRealLocation(location) {
+    return !!location && location !== STANDARD_LOCATION;
+}
+
+function shouldResolveLifecycleMetricLocation(location) {
+    return !isRealLocation(location);
+}
+
+function resolveLifecycleMetricObjectLocation(objectMD, fallbackLocation) {
```
SylvainSenechal: I find it a bit weird that in this function, we are calling isRealLocation on different object metadata attributes. Maybe its fine but it probably deserve some code comments to explain a bit more what its doing

SylvainSenechal: Can also explain why we only look at the first locations[0], of course future developers can always ask an llm to know (claude said its because mpu object have an array of locations, but all locations are the same so its ok to look at the first only) but I think its nicer to have a human written comment for things that can be confusing. Maybe here it does deserves a helper function like "getMpuObjectLocation" or more generic "getObjectLocation" that handle both mpu and non mpu objects

extensions/lifecycle/util/lifecycleLocation.js
SylvainSenechal, line 1: Can discuss with other reviewers : its a bit annoying to do one more pr but, should all these functions belongs to Arsenal ?

extensions/lifecycle/util/lifecycleLocation.js
SylvainSenechal: update : looking at the jira ticket, it gives more context : "In some cases, we have STANDARD as location in lifecycle metrics : as we get object information through regular S3 API."

I think this is the kind of context that could be added as code comment

extensions/lifecycle/util/lifecycleLocation.js
```diff
@@ -0,0 +1,50 @@
...
+        return objectMD[getterName]();
+    }
+    return objectMD[fieldName];
+}
+
+function isRealLocation(location) {
+    return !!location && location !== STANDARD_LOCATION;
+}
+
+function shouldResolveLifecycleMetricLocation(location) {
```
SylvainSenechal: after reading the pr some more, I think this is more a 
needsResolveLifecycleMetricLocation
than
shouldResolveLifecycleMetricLocation

It makes the function calling it easier to understand what they are doing

extensions/lifecycle/util/lifecycleLocation.js
DarkIsDude: I agree with Sylvain on this one, something is not clear

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
...
+
+            const metadataLocation = resolveLifecycleMetricObjectLocation(objectMD, fallbackLocation);
+            if (!shouldResolveLifecycleMetricLocation(metadataLocation)) {
+                return cb(null, metadataLocation);
+            }
+            return this._resolveLifecycleMetricLocationFromArchiveInfo(entry, metadataLocation, log, cb);
+        }, false);
+    }
+
+    _resolveLifecycleMetricLocation(entry, fallbackLocation, log, cb) {
```
SylvainSenechal: i feel like something is off around here :
We have 
_resolveLifecycleMetricLocation calling 

-  _resolveLifecycleMetricLocationFromBucket
-  _resolveLifecycleMetricLocationFromMetadata which calls these 2
  - _resolveLifecycleMetricLocationFromArchiveInfo
  - _resolveLifecycleMetricLocationFromBucket

So Frombucket can be called twice, and we have something else smelly : the if entry action type == deleteMpu is checked in this function, but also in _resolveLifecycleMetricLocationFromMetadata


I have a feeling it would be better to have a single function, even if larger, that can be read more naturally with less indirection.

Discussed this with Claude and providing his answer here : 

Image

extensions/lifecycle/util/lifecycleLocation.js
DarkIsDude: I'm not sure we need a new function to just call an another one ?

extensions/lifecycle/util/lifecycleLocation.js
DarkIsDude: I think the idea is to be able to manage when the value are wrapped in a class (Arsenal) or when it's a raw data. But we should try to always manipulate class ?

DarkIsDude (commented):
Can you also tell me more about `In some cases, we have STANDARD as location in lifecycle metrics : as we get object information through regular S3 API.` which cases and when it's an issue to use regular s3 api ?

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
...
-            Date.now() - entry.getAttribute('transitionTime'));
+        if (!this.s3target) {
+            return process.nextTick(cb);
+        }
 
-        const entries = [{ message: entry.toKafkaMessage() }];
-        this.producer.sendToTopic(this.objectTasksTopic, entries,  err => {
-            LifecycleMetrics.onKafkaPublish(null, 'ObjectTopic', 'bucket', err, 1);
-            return cb(err);
+        const command = new GetBucketLocationCommand({ Bucket: bucket });
```
francoisferrand: `getBucketLocation()` returns the default location of a bucket, which could change (it happens in S3C, in some cases):
- so would be a race condition here...
- the cache makes this issue worse, as it may get poisoned by the "old" value of the location

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
...
+        attachReqUids(command, log.getSerializedUids());
+        return this.s3target.send(command)
+            .then(data => {
+                LifecycleMetrics.onS3Request(log, 'getBucketLocation', 'bucket', null);
+                const location = data && data.LocationConstraint;
+                this._bucketLocationCache.set(bucket, location);
```
francoisferrand: there may be **many** buckets, not sure we should allow unbounded cache

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
...
 
-        const shouldBreak = this.circuitBreakers.tripped(
-            'expiration',
-            location,
-            this.objectTasksTopic,
-        );
-        if (shouldBreak) {
-            process.nextTick(() => cb(errorCircuitBreakerTripped));
-            return;
+        this._resolveLifecycleMetricLocation(entry, initialLocation, this.log, (err, location) => {
```
francoisferrand: instead of the extra metadata call, could we change one (existing) call?
i.e. if we already call HeadObject, we could instead pass the appropriate flag (x-amz-scal-archive-info) to get more details? Or use getObjectAttribute/listObjectV2 with attributes to get the required info more efficiently?

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
      * @return {undefined}
      */
     _sendObjectAction(entry, cb) {
-        const location = entry.getAttribute('details.dataStoreName');
+        const initialLocation = entry.getAttribute('details.dataStoreName');
```
francoisferrand: this is not the "initialLocation" : if the location is already set (e.g. object has been transitioned already or written directly to a non-default location), this value should be correct already : and it must be used.
→ this is really the "real" location, except sometimes it is "STANDARD" (i.e. default bucket location) and needs to be resolved

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
...
+                log.debug('failed to get bucket location for lifecycle metrics', {
+                    method: 'LifecycleTask._getBucketLocationConstraint',
+                    bucket,
+                    error: err.message,
+                });
+                return cb();
+            });
+    }
+
+    _resolveLifecycleMetricLocationFromBucket(entry, fallbackLocation, log, cb) {
```
francoisferrand: * this so-called "fallbackLocation" is NOT a fallback, this is the actual location
* parameter is redundant : this is part of entry already

extensions/lifecycle/util/lifecycleLocation.js
francoisferrand: it is already a "real" location : shoud be `isStandardLocation` or `isDefaultLocation`

extensions/lifecycle/util/lifecycleLocation.js
francoisferrand: "should resolve lifecycle metric location" is really just confusing, as their is no real notion behind : the code (in lifecycle) would be much easier to read without this abstraction

```
const location = entry.getAttribute('details.dataStoreName');
if (isDefaultLocation(location) ) {
    location = getDefaultBucketLocation(bucket);
}
[...]
```

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
...
+
+        return this._getArchiveInfoLocation(entry, log, (err, archiveInfoLocation) => {
+            if (err || !isRealLocation(archiveInfoLocation)) {
+                return this._resolveLifecycleMetricLocationFromBucket(entry, fallbackLocation, log, cb);
+            }
+            return cb(null, archiveInfoLocation);
         });
     }
 
+    _resolveLifecycleMetricLocationFromMetadata(entry, fallbackLocation, log, cb) {
```
francoisferrand: not a "fallback" location...

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
...
+        if (!this.backbeatMetadataProxy) {
```
francoisferrand: why do we need 3 cases?
- `backbeatMetadataProxy` is always available AFAICS (always initialized in [Bucket Processor](https://github.com/scality/backbeat/blob/8f38358b263711adfb42c84ead0f8f756786499d/extensions/lifecycle/bucketProcessor/LifecycleBucketProcessor.js))
- resolving from ObjectMD or Bucket should work in all cases : but at the cost of extra IO (read ObjectMD) or race condition (read default bucket location)

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -197,30 +217,176 @@ class LifecycleTask extends BackbeatTask {
...
+            LifecycleMetrics.onS3Request(log, 'getMetadata', 'bucket', err);
+            if (err) {
+                log.debug('failed to get object metadata for lifecycle metric location', {
+                    method: 'LifecycleTask._resolveLifecycleMetricLocationFromMetadata',
+                    bucket: params.bucket,
+                    objectKey: params.objectKey,
+                    versionId: params.versionId,
+                    error: err.message,
+                });
+                return this._resolveLifecycleMetricLocationFromArchiveInfo(entry, fallbackLocation, log, cb);
```
francoisferrand: why make the call to getObjectMD if we already had the data in `entry` ?

extensions/lifecycle/util/lifecycleLocation.js
francoisferrand: not possible. When you call a function, you MUST know what the parameters are : and then you implement the way which is needed, instead of trying to cope with various cases dynamically...

extensions/lifecycle/tasks/LifecycleTask.js
francoisferrand: also sometimes needs to be resolved because it is a "cold" location, according to the ticket.

Though practically, not sure what the problem really is - of if there one : sendObjectAction would be called on expiration, and seems ok if we report the expiration of a cold object as happening on the "cold" location itself?

tests/unit/lifecycle/LifecycleDeleteObjectTask.spec.js
```diff
@@ -279,6 +281,40 @@ describe('LifecycleDeleteObjectTask', () => {
         });
     });
 
+    it('should emit expiration metrics with restored object cold location', done => {
```
francoisferrand: expiration metrics should be made with cold location: since this is where the data is "really" stored...
Or are there some cases where this creates inconsistencies?

extensions/lifecycle/bucketProcessor/policy.json
```diff
@@ -6,6 +6,7 @@
             "Effect": "Allow",
             "Action": [
                 "s3:GetLifecycleConfiguration",
+                "s3:GetBucketLocation",
```
francoisferrand: also need extra permission to get "archive info" in HeadObject

francoisferrand (commented):
- no clear which scenario exactly creates which problem - and hence if the fix is relevant
- by trying to be very generic, the code creates lots of complexity, with a very complex algorithm being implemented to retrieve the location in different cases - which also makes it very difficult to be sure "what" will actually end up in the metrics
- this is a "small" improvement: instead of STANDARD we get in some case, we want to get the actual field which is already in metadata; so don't want to pay for this (if it needs so much complexity and IOs, may be better to just live with it)

With the advances in Cloudserver api (headObject, list object v2, ...), is there no way to just tweak the existing approach, to get to this information efficiently and keep this simple?

extensions/lifecycle/util/lifecycleLocation.js
benzekrimaha: I removed the ambiguous resolution helpers from the bucket task path. The object action now uses the location already set on the entry.

benzekrimaha: Done differently: the STANDARD resolution is no longer hidden in this helper. For non-versioned objects, the existing HeadObject call with archive info sets the real StorageClass before the action is queued.

extensions/lifecycle/util/lifecycleLocation.js
benzekrimaha: Removed the generic helper usage from LifecycleTask; no more mixed raw/ObjectMD handling in that path. The resolution is now explicit at the existing HeadObject call site. I also removed the dynamic raw/class abstraction from the bucket task path.

extensions/lifecycle/util/lifecycleLocation.js
benzekrimaha: Removed this logic from the bucket task path, so the extra comment is no longer needed there.We no longer inspect ObjectMD/location arrays while queuing object actions

extensions/lifecycle/util/lifecycleLocation.js
benzekrimaha: Not needed anymore for this PR scope: most of the helper logic was removed. The remaining object-processor helper only handles an ObjectMD already fetched for deletion.

extensions/lifecycle/util/lifecycleLocation.js
benzekrimaha: The function was removed

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: Removed the multi-step resolution chain. _sendObjectAction no longer performs metadata/head/bucket lookups

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: Fixed by removing GetBucketLocation from this path entirely, so there is no bucket-default-location race anymore.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: I removed the bucket location cache.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: I reused the existing HeadObject call in _compareObject and kept the archive-info header there instead of adding another metadata/HeadObject call later.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: We no longer call this initial/fallback. The action entry carries the metric location directly. I also kept expiration metrics on the real cold/archive location when Cloudserver returns it through archive info; and removed the bucket fallback logic.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: Fixed by removing the fallback param/path from LifecycleTask.

extensions/lifecycle/util/lifecycleLocation.js
benzekrimaha: Adjusted: the remaining util now uses default-location wording instead of “real location”.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: Fixed by removing the fallback-location flow.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: We no longer do extra ObjectMD or bucket-location lookup while queuing the action.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: No extra getObjectMD call is made in _sendObjectAction anymore.

extensions/lifecycle/util/lifecycleLocation.js
benzekrimaha: Agreed. I removed the dynamic parameter handling.

tests/unit/lifecycle/LifecycleDeleteObjectTask.spec.js
benzekrimaha: I kept this behavior: expiration execution metrics still resolve restored/cold ObjectMD to the cold location in LifecycleDeleteObjectTask.

extensions/lifecycle/bucketProcessor/policy.json
benzekrimaha: Added `scality:GetObjectArchiveInfo` to the lifecycle bucket processor policy and updated the policy test.

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -14,6 +14,32 @@ const {
 class ObjectLockedError extends Error {}
 class PendingReplicationError extends Error {}
 
+function _isDefaultLocation(location) {
+    return !location || location === 'STANDARD';
+}
+
+function _getLifecycleMetricLocation(objectMD, currentLocation) {
+    const archiveLocation = objectMD?.getArchive() && objectMD.getAmzStorageClass();
```
francoisferrand: will never be STANDARD in that case. getAmzStorageClass is set exactly to have a cold class, which cannot be the default location.

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -14,6 +14,32 @@ const {
...
+    const dataStoreName = objectMD?.getDataStoreName();
+    if (!_isDefaultLocation(dataStoreName)) {
+        return dataStoreName;
+    }
+
+    // MPU metadata keeps part locations in an array; the metric only needs the
+    // storage site, which is shared by the listed parts.
+    const locations = objectMD?.getLocation();
+    const partLocation = Array.isArray(locations) && locations[0]?.dataStoreName;
+    if (!_isDefaultLocation(partLocation)) {
```
francoisferrand: this cannot happen : location's dataStoreName is always a "real" location

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1588,12 +1601,16 @@ class LifecycleTask extends BackbeatTask {
...
+        attachArchiveInfoHeader(command);
         attachReqUids(command, log.getSerializedUids());
         return this.s3target.send(command)
             .then(data => {
                 LifecycleMetrics.onS3Request(log, 'headObject', 'bucket', null);
                 const object = Object.assign({}, obj,
-                    { LastModified: data.LastModified });
+                    {
+                        LastModified: data.LastModified,
+                        StorageClass: data.StorageClass || obj.StorageClass,
```
francoisferrand: why the "or" ?
- storageClass will always be set
- it will the "cold" class (object.x-amz-storage-class) for cold object, even if restored
- it will never be STANDARD (contrary to obj.StorageClass), i.e. always the "real" storage even if not default

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -195,7 +221,10 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
 
         const actionType = entry.getActionType();
         const transitionTime = entry.getAttribute('transitionTime');
-        const location = this.objectMD?.dataStoreName || entry.getAttribute('details.dataStoreName');
+        const location = _getLifecycleMetricLocation(
+            this.objectMD,
+            entry.getAttribute('details.dataStoreName')
```
francoisferrand: with the headObject change, the storage class is "initially" correct.
would it be enough to just trust `entry.getAttribute('details.dataStoreName')` now ?

francoisferrand (commented):
very difficult to review this PR, missing the analysis of the ticket : what problems _exactly_ is it solving? in which situation do we have an issue, and why?

→ please document this in the ticket and PR

DarkIsDude (approved):
LGTM but I'm still newbie to backbeat so I suggest to have at least one review from someone who worked on it 🚀

francoisferrand:
> very difficult to review this PR, missing the analysis of the ticket : what problems _exactly_ is it solving? in which situation do we have an issue, and why?
> 
> → please document this in the ticket and PR

@benzekrimaha

francoisferrand (changes requested):
very difficult to review this PR, missing the analysis of the ticket : what problems exactly is it solving? in which situation do we have an issue, and why?

→ please document this in the ticket and PR

benzekrimaha:
@francoisferrand I've updated the PR description with a full problem analysis (root cause + the two concrete scenarios where the location label is wrong: cold-storage expiration and restored-object cleanup).

benzekrimaha:
Added problem analysis to the PR description and [BB-721](https://scality.atlassian.net/browse/BB-721) ticket per review feedback.

**TL;DR for reviewers:** Expiration metrics can report the wrong location because lifecycle used listing/STANDARD semantics and, for restored cold objects, completion metrics preferred warm objectMD.dataStoreName over the cold location on the action entry. The fix resolves the real storage class at queue time via archive-info HeadObject, then uses details.dataStoreName downstream.

[BB-721]: https://scality.atlassian.net/browse/BB-721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -205,7 +218,7 @@ class LifecycleTask extends BackbeatTask {
      * @return {undefined}
      */
     _sendObjectAction(entry, cb) {
-        const location = entry.getAttribute('details.dataStoreName');
+        const location = entry.getAttribute('details.dataStoreName') || '';
```
francoisferrand: * when would this happen?
* how would this "preserve the existing behavior" ? Both undefined and '' are falsy, so code will behave the same...

extensions/lifecycle/tasks/LifecycleTask.js
francoisferrand: @benzekrimaha ?

tests/unit/lifecycle/LifecycleTask.spec.js
```diff
@@ -2469,4 +2471,82 @@ describe('lifecycle task helper methods', () => {
             });
         });
     });
+
```
francoisferrand, line 2475: I wonder if these unit tests are enough: the issue was a correlation between different values at different layers (STANDARD or cold class in ListObject/HeadObject, hot class in GetMetadata, the way the kafka message's dataStoreName was computed...):
- should we have some sort higher level test where we exercise or simulate the whole sequence?
- do the existing tests cover every kind of lifecycle op (expiration of hot / cold / restored objects, transition from hot to cold, transition from STANDARD to hot, transition from hot to STANDARD...) and verify the metrics are appropriate eventually?

benzekrimaha: I extended the coverage into a scenario matrix at the queue level: expiration of a hot object on the default location (listing says STANDARD, HeadObject returns the real location), a hot object on a non-default location, a cold object, and a restored cold object (cold class preserved in x-amz-storage-class). Each case asserts the archive-info middleware is attached to the HeadObject and that the queued entry's details.dataStoreName is the head-resolved storage class. The execution side (metric emitted from the entry location) is covered in LifecycleDeleteObjectTask.spec.js and LifecycleUpdateExpirationTask.spec.js, including the restored-object case.
Transitions are not affected by this change: transition trigger metrics use the rule's target site, not the listed storage class.
A full end-to-end test (listing → HeadObject → Kafka entry → processor → metric) would need the functional suite with a real CloudServer; I'd propose tracking that as a follow-up ticket rather than blocking this fix WDYT?

tests/unit/lifecycle/LifecycleDeleteObjectTask.spec.js
```diff
@@ -279,6 +281,30 @@ describe('LifecycleDeleteObjectTask', () => {
...
+
+    it('should emit expiration metrics with the cold location carried by the entry', done => {
+        const startedMetric = sinon.stub(LifecycleMetrics, 'onLifecycleStarted');
+        const completedMetric = sinon.stub(LifecycleMetrics, 'onLifecycleCompleted');
+        // The bucket processor resolves the real (cold) storage class via the
+        // archive-info HeadObject and stores it on the entry, so the metric
+        // location is taken directly from details.dataStoreName.
+        const entry = ActionQueueEntry.create('deleteObject')
+            .setAttribute('target.owner', 'testowner')
+            .setAttribute('target.bucket', 'testbucket')
```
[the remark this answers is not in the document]
benzekrimaha: the mock metadata now carries a warm dataStoreName so the test fails with the pre-fix code.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: Indeed. With the `x-amz-scal-archive-info` header, CloudServer always sets x-amz-storage-class in the response (setArchiveInfoHeaders returns objMD['x-amz-storage-class'] || objMD.dataStoreName), so data.StorageClass is always the real class, the preserved cold class for cold/restored objects, the data-store name otherwise , and never STANDARD. The fallback was dead code and reintroduced the listing ambiguity we aiming to remove: dropped it, and I documented the contract in a comment at the call site.

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
```diff
@@ -195,7 +195,7 @@ class LifecycleDeleteObjectTask extends BackbeatTask {
 
         const actionType = entry.getActionType();
         const transitionTime = entry.getAttribute('transitionTime');
-        const location = this.objectMD?.dataStoreName || entry.getAttribute('details.dataStoreName');
+        const location = entry.getAttribute('details.dataStoreName');
```
francoisferrand: `details.dataStoreName` was already added to entries _before_ this PR right?

→ If not, we should keep a fallback (`entry.getAttribute('details.dataStoreName') || this.objectMD?.dataStoreName`) in case the new backbeat processes messages generated by previous version

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -59,6 +59,19 @@ const MAX_RETRIES = 4;
 // parallel tasks, so the total delay of retries should about 1m30s.
 const MAX_RETRIES_TOTAL = CONCURRENCY_DEFAULT * MAX_RETRIES * 10;
 
+function attachArchiveInfoHeader(command) {
```
francoisferrand, line 62: nit: we should use cloudserverclient instead of (manually) adding such header → please create followup

francoisferrand (approved):

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
benzekrimaha: Yes, details.dataStoreName has been set by the bucket processor since 7814584d , so messages from previous versions do carry it. But it can legitimately be empty (obj.StorageClass || '') and the EODM path in LifecycleTaskV2 never sets it, so I added the fallback as suggested: entry.getAttribute('details.dataStoreName') || this.objectMD?.getDataStoreName() (getter, since objectMD is an Arsenal ObjectMD instance). Covered by a new unit test.

extensions/lifecycle/tasks/LifecycleTask.js
benzekrimaha: Follow-up ticket created: BB-800

extensions/lifecycle/tasks/LifecycleDeleteObjectTask.js
benzekrimaha: that's what the current version does; per your later comment I kept a metadata fallback only for entries that don't carry the attribute

PR #2739, opened 2026-05-19
fix: store offset before drain on rebalance tail

delthas:
## TL;DR

A two-line reorder in `TaskScheduler.onTaskEnd` so the per-task callback runs before the drain callback. This closes a synchronous race during Kafka consumer rebalances that prevents the last task's offset from being committed and produces the `LibrdKafkaError: Local: Erroneous state` (`ERR__STATE`) crashes reported in BB-758. The crash was previously contained by a `try/catch` in `BackbeatConsumer.onEntryCommittable`; this PR fixes the underlying ordering bug.

## What this changes

`lib/tasks/TaskScheduler.js`, inside the `onTaskEnd` function created in `push`:

\`\`\`diff
 this._running--;
-this._tryDrain();
 done(...args);
+this._tryDrain();
\`\`\`

That's the whole behavior change. Plus a unit test that asserts the new ordering, and a functional smoke test at the BackbeatConsumer integration level.

## What the bug is

`BackbeatConsumer` uses `TaskScheduler` to run incoming Kafka messages through a user-supplied `queueProcessor`. When a task finishes, `onTaskEnd` runs three things in sequence:

1. Decrement the running-task counter.
2. Maybe fire the drain callback (if the queue is now idle).
3. Call the per-task callback (`done(...args)`), which is `BackbeatConsumer`'s `_onEntryProcessingDone` → `onEntryCommittable` → librdkafka `offsetsStore`.

Before this PR, the order is `_tryDrain()` then `done(...args)`. During a rebalance, the drain callback is the one that runs `_consumer.commit()` and `_consumer.unassign()`. So when the **last** task of a batch completes while a rebalance is pending:

1. The drain callback runs, queueing `commit()` and `unassign()` ops to librdkafka's main thread.
2. `done(...args)` runs, calling `offsetsStore` for that task's offset.
3. By the time step 2 reaches librdkafka, the main thread may have already processed the `unassign` op and cleared the partition's internal `RD_KAFKA_TOPPAR_F_ASSIGNED` flag.
4. `offsetsStore` returns `RD_KAFKA_RESP_ERR__STATE` ("Partition is not assigned"). The error throws into JS and bubbles up out of `onTaskEnd`.

That's the production stack trace recorded in BB-758 (visible in CI artifacts from `lifecycle-bucket-processor`, `queue-populator`, `notification-processor`, etc.).

The race only affects the **last** task of a batch — the one that flips the running counter to zero. Earlier tasks don't trigger the drain callback, so their `offsetsStore` always runs while the partition is still assigned.

## Why the swap is the right fix

`offsetsStore` checks exactly one piece of state inside librdkafka: the `RD_KAFKA_TOPPAR_F_ASSIGNED` flag on the topic partition. That flag is cleared in `rd_kafka_assignment_serve_removals` (in `librdkafka/src/rdkafka_assignment.c`), which only runs **in response to the application's `unassign()` call**. Until the application calls `unassign()`, the flag stays set, and `offsetsStore` succeeds.

With the new order:

1. `done(...args)` runs. `offsetsStore` is invoked while the partition is still assigned — succeeds, the offset is locally stored in librdkafka.
2. `_tryDrain()` runs. The drain callback calls `_consumer.commit()` (synchronous in JS, queued as an op for the main thread). `commit()` picks up the just-stored offset and flushes it to the broker.
3. The drain callback then calls `_consumer.unassign()`. The main thread eventually clears `F_ASSIGNED`.

The offset is committed before the partition is released. The race is closed by construction, not by timing luck.

## Why this is safe

- `setDrain` is only called by `BackbeatConsumer`'s rebalance handler (`lib/BackbeatConsumer.js:725` and `:768`). Audited all other `TaskScheduler` callers (`extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js`, `extensions/replication/queueProcessor/QueueProcessor.js`) — both only use `push()`, no `setDrain`. Ingestion goes through `BackbeatConsumer` indirectly. No production caller depends on drain firing before per-task callbacks.
- All 25 pre-existing `TaskScheduler` unit tests pass unchanged.
- The new unit test deterministically demonstrates the bug on the pre-PR code (`expected ['done', 'drain'] but got ['drain']`) and passes after the swap. Runs in ~5ms without any Kafka dependency.

## Background

BB-441 introduced rebalance drain handling in October 2023 with the stated goal "ensure the offsets are stored (and committed) on rebalance." The implementation waited for `_processingQueue.idle()`, but the synchronous tail of `onTaskEnd` left this race intact. This PR finishes that work for the `committable: true` path.

## Tests

- **`tests/unit/lib/tasks/TaskScheduler.spec.js`** — new `it()` "should run the per-task callback before drain when the last task completes". Pinpoint discriminator for the ordering invariant: passes with the swap, fails with a clear diff without it. Deterministic, no Kafka, ~5ms.
- **`tests/functional/lib/BackbeatConsumer.js`** — new "BackbeatConsumer rebalance during batch processing > commits all task offsets when rebalance fires during a batch". Integration smoke test: sends 10 messages, triggers consumer2 mid-batch, asserts all offsets committed. Doesn't pinpoint the sync-tail race (timing makes the original bug hard to reproduce deterministically at the integration layer with `backlogMetrics` enabled and `concurrency=1` + `nextTick` queueProcessor), but catches end-to-end regressions of the rebalance-during-processing path.

## What this PR does NOT fix

- The async `committable: false` race (deferred `onEntryCommittable` called from a Kafka producer's delivery callback after `unassign` has already run). Addressed in #2740 (BB-777).
- The "poison pill" `committable: false` on error paths in GC and lifecycle (tracked separately in BB-772).

Both are unrelated to the ordering bug fixed here.

Issue: BB-776

### DIFF
--- a/tests/functional/lib/BackbeatConsumer.js
+++ b/tests/functional/lib/BackbeatConsumer.js
@@ -380,6 +380,132 @@ describe('BackbeatConsumer rebalance tests', () => {
     }).timeout(60000);
 });
 
+describe('BackbeatConsumer rebalance during batch processing', () => {
+    // Smoke test that exercises the rebalance handler while a batch of
+    // messages is being processed. Catches regressions where offsets
+    // can be dropped at the boundary between processing and rebalance.
+    // The pinpoint sync-tail race is covered by the TaskScheduler unit
+    // test (`should run the per-task callback before drain when the
+    // last task completes`).
+    const topic = 'backbeat-consumer-spec-sync-tail';
+    const groupId = `replication-group-sync-tail-${Math.random()}`;
+    let producer;
+    let consumer1;
+    let consumer2;
+
+    before(function before(done) {
+        this.timeout(60000);
+        producer = new BackbeatProducer({
+            kafka: producerKafkaConf, topic,
+            pollIntervalMs: 100,
+            compressionType: 'none',
+        });
+        consumer1 = new BackbeatConsumer({
+            clientId: 'BackbeatConsumer-SYNC-TAIL-1',
+            zookeeper: zookeeperConf,
+            kafka: { ...consumerKafkaConf, compressionType: 'none' },
+            groupId, topic,
+            queueProcessor: (_msg, cb) => cb(),
+            bootstrap: true,
+        });
+        async.parallel([
+            innerDone => producer.on('ready', innerDone),
+            innerDone => consumer1.on('ready', innerDone),
+        ], err => {
+            if (err) {return done(err);}
+            consumer2 = new BackbeatConsumer({
+                clientId: 'BackbeatConsumer-SYNC-TAIL-2',
+                zookeeper: zookeeperConf,
+                kafka: { ...consumerKafkaConf, compressionType: 'none' },
+                groupId, topic,
+                queueProcessor: (_msg, cb) => cb(),
+            });
+            return consumer2.on('ready', done);
+        });
+    });
+
+    after(function after(done) {
+        this.timeout(10000);
+        async.parallel([
+            innerDone => producer.close(innerDone),
+            innerDone => consumer1.close(innerDone),
+            innerDone => (consumer2 ? consumer2.close(innerDone) : innerDone()),
+        ], done);
+    });
+
+    it('commits all task offsets when rebalance fires during a batch', done => {
+        const N = 10;
+        let processed = 0;
+        consumer1._queueProcessor = (_message, cb) => {
+            processed++;
+            process.nextTick(cb);
+        };
+        consumer2._queueProcessor = (_msg, cb) => process.nextTick(cb);
+
+        consumer2._consumer.committed(
+            [{ topic, partition: 0 }], 5000, (e1, base) => {
+                assert.ifError(e1);
+                const baseline = base[0].offset;
+
+                consumer1.subscribe();
+                const messages = [];
+                for (let i = 0; i < N; i++) {
+                    messages.push({ key: `k${i}`, message: `{"i":${i}}` });
+                }
+                producer.send(messages, err => {
+                    assert.ifError(err);
+                });
+
+                // Trigger consumer2.subscribe() once consumer1 has
+                // started processing the batch — that way the
+                // rebalance fires while the batch is in flight (some
+                // tasks already done, some still running). The
+                // synchronous-tail ordering matters for whichever task
+                // happens to be the last to flip the queue's running
+                // counter to zero.
+                let triggered = false;
+                consumer1.on('consumed', () => {
+                    if (triggered) {return;}
+                    triggered = true;
+                    consumer2.subscribe();
+                });
+
+                // Wait for processing + rebalance to settle, then
+                // verify all N offsets reached the broker. Use consumer1
+                // (still connected) and retry on transient "Local:
+                // Timed out" errors that can occur while the rebalance
+                // is mid-flight.
+                const queryCommitted = (attempts, finalCb) => {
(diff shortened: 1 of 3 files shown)


tests/unit/lib/tasks/TaskScheduler.spec.js
```diff
@@ -233,6 +233,22 @@ describe('TaskScheduler', () => {
             sinon.stub(taskScheduler, 'idle').returns(false);
             assert.doesNotThrow(() => taskScheduler._tryDrain());
         });
+        it('should run the per-task callback before drain when the last ' +
```
francoisferrand, line 237:
```suggestion

        it('should run the per-task callback before drain when the last ' +
```

francoisferrand (commented):
LGTM

- The changes in TaskScheduler were introduced esp. for Ingestion I think, can you double check there is no impact? (i.e. is drain callback used only for BackbeatConsumer, or are there other cases)
- This issue should have been caught by `BackbeatConsumer` tests (which already exercise rebalance) : is it possible to add a test in there?

delthas:
Thanks for the review @francoisferrand. Addressed both points in [`094bac62`](https://github.com/scality/backbeat/pull/2739/commits/094bac62):

**1. TaskScheduler / setDrain caller audit**

`grep -rn 'new TaskScheduler\|setDrain\b' --include='*.js'` outside `tests/` and `node_modules/`:

- `lib/BackbeatConsumer.js` — only place that calls `setDrain` (in the rebalance handler at lines 729 and 772).
- `extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js:243` — creates a TaskScheduler, uses `push()` only, no `setDrain`.
- `extensions/replication/queueProcessor/QueueProcessor.js:268` — same: `push()` only, no `setDrain`.
- `extensions/ingestion/` — does not use TaskScheduler directly. Ingestion goes through `BackbeatConsumer`, so the only impact path is via the BackbeatConsumer rebalance handler (which is what this PR fixes).

No production caller relies on drain firing before the per-task callback. The change is contained to the path the BB-758 stack trace actually exercises.

**2. BackbeatConsumer-level test**

Added `BackbeatConsumer rebalance during batch processing > commits all task offsets when rebalance fires during a batch` in `tests/functional/lib/BackbeatConsumer.js`. It sends 10 messages, triggers consumer2.subscribe() once consumer1 has begun processing, waits for the rebalance to settle, and asserts the broker's committed offset advanced to `baseline + N`.

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -380,6 +380,133 @@ describe('BackbeatConsumer rebalance tests', () => {
...
+        this.timeout(10000);
+        async.parallel([
+            innerDone => producer.close(innerDone),
+            innerDone => consumer1.close(innerDone),
+            innerDone => (consumer2 ? consumer2.close(innerDone) : innerDone()),
+        ], done);
+    });
+
+    it('commits all task offsets when rebalance fires during a batch',
+    done => {
```
francoisferrand:
```suggestion
    it('commits all task offsets when rebalance fires during a batch', done => {
```

francoisferrand (approved):

SylvainSenechal (approved):

tests/functional/lib/BackbeatConsumer.js
delthas: Applied in [`37841a3e`](https://github.com/scality/backbeat/pull/2739/commits/37841a3e).

PR #2740, opened 2026-05-19
improvement: wait for OffsetLedger to drain on rebalance

delthas:
## TL;DR

Extend `BackbeatConsumer`'s rebalance wait to also drain the `OffsetLedger` (in-flight deferred commits), not just the processing queue. Closes the async-path race that has been quietly producing `LibrdKafkaError: Local: Erroneous state` log noise and forcing Kafka to redeliver messages whose `onEntryCommittable` callbacks land after `unassign()`. Finishes the work BB-441 started for the deferred-commit (`committable: false`) flow. BB-776 (#2739) fixed the synchronous-tail race separately.

## What this changes

`lib/BackbeatConsumer.js`, in `_onRebalance`'s `ERR__REVOKE_PARTITIONS` branch:

\`\`\`diff
-if (this._processingQueue.idle()) {
+const isFullyDrained = () =>
+    this._processingQueue.idle() &&
+    this._offsetLedger.getProcessingCount(this._topic) === 0;
+
+if (isFullyDrained()) {
     unassign(unassignStatus.IDLE);
     return;
 }
-this._processingQueue.setDrain(() => unassign(unassignStatus.DRAINED));
+const checkFullyDrained = () => {
+    if (isFullyDrained()) {
+        unassign(unassignStatus.DRAINED);
+    }
+};
+this._processingQueue.setDrain(checkFullyDrained);
+this._ledgerEmptyListener = ledgerTopic => {
+    if (ledgerTopic === this._topic) checkFullyDrained();
+};
+this._offsetLedger.on('empty', this._ledgerEmptyListener);
\`\`\`

`lib/OffsetLedger.js` extends `EventEmitter` and emits `'empty'` (with the topic name) from `onOffsetProcessed` when the in-flight count for that topic transitions to zero. The emit is deferred via `process.nextTick(...)` so callers of `onOffsetProcessed` (chiefly `BackbeatConsumer.onEntryCommittable`) can finish their synchronous `offsetsStore` call before any listener triggers the rebalance unassign.

Plus listener cleanup in the existing `unassign` closure, and three functional tests covering the wait predicate's transition points.

The existing outer timeout (`max.poll.interval - 1s`) is unchanged behaviorally — still fires `unassign(TIMEOUT) + disconnect()`. The only difference is *what it's waiting for*: queue idle **and** ledger empty, rather than just queue idle.

## What the bug is

`BackbeatConsumer` supports two ways for a task to complete:

- **`committable: true`** (the default): the task's `done(...)` callback synchronously triggers `onEntryCommittable` → `offsetsStore` inside `BackbeatConsumer._onEntryProcessingDone`. BB-776 made sure this runs *before* the rebalance drain handler fires.
- **`committable: false`** (introduced in ZENKO-860 for replication): the task returns immediately, and the application is responsible for calling `onEntryCommittable(entry)` later — typically from a Kafka status producer's delivery callback. Used today by `ReplicateObject._publishReplicationStatus` and `CopyLocationTask._publishCopyLocationStatus`.

For the second pattern, the queue is "idle" from the worker's perspective even though the offset hasn't been stored yet. At rebalance time, the handler today only checks `_processingQueue.idle()`, so it unassigns immediately. The later `onEntryCommittable` then hits `ERR__STATE` because `RD_KAFKA_TOPPAR_F_ASSIGNED` has been cleared by librdkafka's main thread in response to the JS `unassign()` call (`librdkafka/src/rdkafka_assignment.c:361`). The BB-758 `try/catch` swallows the throw and Kafka redelivers — at-least-once but wasteful.

## Why the OffsetLedger is the right signal

`OffsetLedger.getProcessingCount(topic)` tracks consumed-but-not-yet-marked-processed offsets per topic. An entry enters `processing` in `onOffsetConsumed` (when the consumer dispatches the message) and leaves in `onOffsetProcessed` (which is the same code path that calls `offsetsStore`). When the count reaches 0, every consumed message has had its offset stored locally in librdkafka. The rebalance handler's existing `_consumer.commit()` (just before `unassign`) then flushes those stored offsets to the broker.

So "ledger empty" == "all offsets that should be committed are now stored," which is exactly what we need before releasing the partition.

## Why an event (and why the deferral)

Earlier iteration used `setInterval(checkFullyDrained, 100)` for ledger drain detection. Per review feedback this is now event-based: `OffsetLedger` extends `EventEmitter` and emits `'empty'` from `onOffsetProcessed` when the count transitions to zero for a topic. No busy loop, deterministic firing.

The emit is wrapped in `process.nextTick(...)` because callers of `onOffsetProcessed` (i.e. `BackbeatConsumer.onEntryCommittable`) still need to call `_consumer.offsetsStore()` after `onOffsetProcessed` returns. A synchronous listener triggering `unassign()` would clear `F_ASSIGNED` before `offsetsStore` runs, reintroducing the BB-758 race. Deferring to the next tick lets the caller finish its synchronous work first. Safe because librdkafka pauses fetching during rebalance (`rd_kafka_assignment_pause` at the C level), so no new offsets enter the ledger between the schedule and the emit.

## Why no JS-side fetch gate during the wait

A previous attempt (October 2023, `_unassignInProgress` flag) gated `_tryConsume` during rebalance, then was removed without a documented rationale. Reading librdkafka's source, `rd_kafka_assignment_pause` is called inside the cgrp before queuing the rebalance op for the application (`rdkafka_cgrp.c:1102-1108`), which pauses fetching at the C level. A JS-side gate would be redundant.

## Tests

Three `it()` blocks in `tests/functional/lib/BackbeatConsumer.js > BackbeatConsumer ledger drain on rebalance`, one per transition point of the wait predicate (`queue.idle && ledger.empty`):

- **`takes the IDLE path when both queue and ledger are already drained at revoke time`** — initial check, no work pending, IDLE path immediate.
- **`completes the wait when the queue drains and ledger is empty`** — committable: true tasks finishing during the rebalance. Sends N=5 messages, triggers consumer2 mid-batch, asserts all offsets committed.
- **`completes the wait when the ledger drains via the empty event`** — the Pattern A scenario (committable: false + deferred ack), DRAINED via the new `'empty'` event.

All three pass in ~15s combined. The Pattern A test would fail with ERR__STATE without the fix (verified by stashing the fix locally).

## What this does NOT fix

- BB-772 (poison-pill `committable: false` in GC and lifecycle that never calls `onEntryCommittable`). Those entries never drain from the ledger. After this change, a rebalance catching such an entry will wait until the outer timeout (~5 min) and disconnect — same as a genuinely stuck worker today. Acceptable; BB-772 will fix it separately.

## Related

- BB-441 (October 2023): originally introduced rebalance drain handling; only covered the queue, not the ledger.
- BB-758: the ERR__STATE crash whose `try/catch` mitigation is no longer needed for this specific case once the wait is in place.
- BB-776 (#2739): synchronous-tail race fix; ordering swap in `TaskScheduler.onTaskEnd`.

Issue: BB-777


lib/BackbeatConsumer.js
```diff
@@ -764,12 +769,32 @@ class BackbeatConsumer extends EventEmitter {
...
             }
 
-            this._processingQueue.setDrain(() => unassign(unassignStatus.DRAINED));
+            const checkFullyDrained = () => {
+                if (isFullyDrained()) {
+                    unassign(unassignStatus.DRAINED);
+                }
+            };
+            this._processingQueue.setDrain(checkFullyDrained);
+            // The OffsetLedger doesn't expose an "empty" event, so poll. The
```
francoisferrand: shouldn't we introduce one instead of polling?

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -380,6 +380,114 @@ describe('BackbeatConsumer rebalance tests', () => {
     }).timeout(60000);
 });
 
+describe('BackbeatConsumer ledger drain on rebalance', () => {
```
francoisferrand, line 509: seems to be missing some cases: there are 2 conditions being checked, at 3 time points (initially, processing queue drained, ledger drained)

→ should have at least 3 tests for ledger drained ? (on top of the existing tests)

lib/BackbeatConsumer.js
delthas: Done in [`3c38b1a5`](https://github.com/scality/backbeat/pull/2740/commits/3c38b1a5): `OffsetLedger` now extends `EventEmitter` and emits `'empty'` (with the topic name) from `onOffsetProcessed` when the in-flight count for that topic transitions to zero. The rebalance handler registers a listener at the start of the wait and removes it in the `unassign` closure. No more `setInterval`.

Followup [`392b48fd`](https://github.com/scality/backbeat/pull/2740/commits/392b48fd) defers the emit via `process.nextTick` so callers of `onOffsetProcessed` (e.g. `onEntryCommittable`) can finish their synchronous `offsetsStore` call before a listener triggers the rebalance unassign — otherwise we'd reintroduce the BB-758 race we're trying to close.

tests/functional/lib/BackbeatConsumer.js
delthas: Done in [`3c38b1a5`](https://github.com/scality/backbeat/pull/2740/commits/3c38b1a5). Three `it()` blocks, one per transition point of the rebalance wait predicate (`queue.idle && ledger.empty`):

- `takes the IDLE path when both queue and ledger are already drained at revoke time` — initial check, no work pending.
- `completes the wait when the queue drains and ledger is empty` — committable: true tasks finishing during the rebalance, DRAINED via queue drain. Sends N=5 messages, triggers consumer2 mid-batch, asserts all offsets committed.
- `completes the wait when the ledger drains via the empty event` — the original Pattern A scenario (committable: false + deferred ack), DRAINED via the new `'empty'` event.

All three pass on the updated code (~15s total).

lib/OffsetLedger.js
```diff
@@ -88,7 +96,12 @@ class OffsetLedger {
         const partitionOffsets = this._getPartitionOffsets(topic, partition);
         partitionOffsets.processing =
             partitionOffsets.processing.filter(pOff => pOff !== offset);
-        return this._getPartitionCommittableOffset(partitionOffsets);
+        const committableOffset =
+            this._getPartitionCommittableOffset(partitionOffsets);
+        if (this.getProcessingCount(topic) === 0) {
+            this.emit('empty', topic);
```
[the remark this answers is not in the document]
delthas: Good catch — fixed in [`392b48fd`](https://github.com/scality/backbeat/pull/2740/commits/392b48fd). Wrapped the emit in `process.nextTick(...)` exactly as suggested. The reasoning checks out: librdkafka pauses fetching during rebalance (via `rd_kafka_assignment_pause` at the C level), so the ledger can't grow between the schedule and the emit, and the caller gets to finish its synchronous work first.

The functional test was passing despite this because the test config has `backlogMetrics` enabled, which makes the rebalance handler's `doUnassign` go through `_publishOffsetsCron` (async). The actual `unassign()` therefore happens on a later tick than the emit-triggered `unassign(DRAINED)` call, so `offsetsStore` runs against a still-assigned partition. Without backlogMetrics — or in production timing under load — the synchronous path would have caught fire.

lib/BackbeatConsumer.js
```diff
@@ -715,22 +716,29 @@ class BackbeatConsumer extends EventEmitter {
                 assignment,
                 queueLen: this._processingQueue.length(),
                 running: this._processingQueue.running(),
+                ledgerProcessing: this._offsetLedger.getProcessingCount(this._topic),
```
francoisferrand: maybe just "ledger" ?
```suggestion
                ledger: this._offsetLedger.getProcessingCount(this._topic),
```

lib/BackbeatConsumer.js
```diff
@@ -715,22 +716,29 @@ class BackbeatConsumer extends EventEmitter {
...
             });
 
             const unassign = jsutil.once(status => {
                 this._log.info(`processing queue ${status}, un-assigning`, {
                     queueLen: this._processingQueue.length(),
                     running: this._processingQueue.running(),
+                    ledgerProcessing: this._offsetLedger.getProcessingCount(this._topic),
```
francoisferrand: same

```suggestion
                    ledger: this._offsetLedger.getProcessingCount(this._topic),
```

lib/OffsetLedger.js
```diff
@@ -88,7 +96,19 @@ class OffsetLedger {
...
+            this._getPartitionCommittableOffset(partitionOffsets);
+        if (this.getProcessingCount(topic) === 0) {
+            // Defer the emit: callers of onOffsetProcessed (e.g.
+            // BackbeatConsumer.onEntryCommittable) still need to call
+            // _consumer.offsetsStore() with committableOffset after we
+            // return. If a 'empty' listener triggers the rebalance
+            // unassign synchronously, offsetsStore would hit ERR__STATE
+            // on the now-unassigned partition. Deferring to the next
+            // tick lets the caller finish its synchronous work first.
+            process.nextTick(() => this.emit('empty', topic));
```
francoisferrand: could the caller "add" entry to the ledger before emit is called?
(and should we thus add a guard before editing)


```suggestion
            process.nextTick(() => {
            	if (this.getProcessingCount(topic) === 0)
		            this.emit('empty', topic));
		        }
		    });
```

lib/OffsetLedger.js
```diff
@@ -88,7 +96,19 @@ class OffsetLedger {
         const partitionOffsets = this._getPartitionOffsets(topic, partition);
         partitionOffsets.processing =
             partitionOffsets.processing.filter(pOff => pOff !== offset);
-        return this._getPartitionCommittableOffset(partitionOffsets);
+        const committableOffset =
+            this._getPartitionCommittableOffset(partitionOffsets);
+        if (this.getProcessingCount(topic) === 0) {
```
francoisferrand: should update ledger test to verify the signal is indeed emitted

lib/BackbeatConsumer.js
```diff
@@ -764,12 +772,40 @@ class BackbeatConsumer extends EventEmitter {
...
+            // land). Either may fire first depending on the path that
+            // had work pending. Both call checkFullyDrained, which
+            // only triggers unassign once both conditions hold.
+            this._processingQueue.setDrain(checkFullyDrained);
+            this._ledgerEmptyListener = ledgerTopic => {
+                if (ledgerTopic === this._topic) {
+                    checkFullyDrained();
+                }
+            };
+            this._offsetLedger.on('empty', this._ledgerEmptyListener);
```
francoisferrand: do we need to reset something on "drain timeout" (e..g line ~816), or is disconnect on line 819 enough?

lib/BackbeatConsumer.js
```diff
@@ -715,22 +716,29 @@ class BackbeatConsumer extends EventEmitter {
...
 
-                // Reset the timer
+                // Reset the timer and ledger drain listener
                 clearTimeout(this._drainProcessQueueTimeout);
                 this._drainProcessQueueTimeout = null;
+                if (this._ledgerEmptyListener) {
+                    this._offsetLedger.removeListener(
+                        'empty', this._ledgerEmptyListener);
+                    this._ledgerEmptyListener = null;
+                }
```
francoisferrand: nit: more closely related to line 732

lib/BackbeatConsumer.js
francoisferrand: both processingQueue and ledger signal should
- be set/cleared at the same time (i.e. line 802 and 732)
- execute the same callback

→ would be worth introducing a `BackbeatConsumer.setDrain(cb)` function which covers both ; cb would be either `checkFullyDrained` or empty callback ; and we would just replace `this._processingQueue.setDrain` with `this.setDrain`

francoisferrand: should we really call `on()` repeatedly? Or would it maybe be safer -less moving parts- to just call it once, to setup up this callback:

```
			ledgerTopic => {
                if (ledgerTopic === this._topic && this._ledgerEmptyListener) {
                    this._ledgerEmptyListener();
                }
            };
```

(kind of fear corner cases where we would not remove the event handler, but add another one later...)

lib/BackbeatConsumer.js
delthas: Done in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f). Renamed to `ledger` in both the revoke log (line 719) and the un-assigning log (line 726).

lib/BackbeatConsumer.js
delthas: Same — applied in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f).

lib/OffsetLedger.js
delthas: Good catch — applied the guard in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f). The deferred callback now re-checks `getProcessingCount(topic) === 0` before emitting; if a caller pushed a new entry between scheduling and firing, the stale emit is suppressed. Also added a unit test for this exact case ("should not emit if a new entry is consumed between scheduling and firing the deferred emit").

lib/OffsetLedger.js
delthas: Done in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f). Added 6 tests in `tests/unit/OffsetLedger.spec.js` under a new `'empty' event` describe block: emit on transition to zero, no emit when count stays non-zero, per-topic isolation, `nextTick` deferral, suppression when a new entry is consumed before the deferred emit fires, and no-op when no listeners (steady-state hot path).

lib/BackbeatConsumer.js
delthas: Added defensive cleanup in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f): the outer-timeout handler now calls `this._setDrain(null)` before `disconnect()`. The `unassign` closure also resets it via the same path, but the explicit clear guards against a late ledger `'empty'` emit firing into a half-torn-down consumer.

lib/BackbeatConsumer.js
delthas: Done in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f). With the new `_setDrain(null)` wrapper, both the timer clearing and the drain reset live on adjacent lines in the unassign closure.

lib/BackbeatConsumer.js
delthas: Done in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f). Introduced `BackbeatConsumer._setDrain(cb)` (around line ~656). The rebalance handler now calls `this._setDrain(checkFullyDrained)` to arm both signals and `this._setDrain(null)` to clear both. Single source of truth for the drain state.

delthas: Good call — fixed in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f). The `'empty'` listener is now installed once at construction (in `BackbeatConsumer`'s constructor, right after `new OffsetLedger()`) and delegates through a `this._drainCallback` flag. The rebalance handler toggles the flag via `_setDrain(cb)` / `_setDrain(null)` — no more `on()/removeListener()` per rebalance, no leaked listener risk.

lib/OffsetLedger.js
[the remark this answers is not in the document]
delthas: Good suggestion — applied in [`63e3499f`](https://github.com/scality/backbeat/pull/2740/commits/63e3499f). The condition is now `getProcessingCount(topic) === 0 && this.listenerCount('empty') > 0` — skips the `nextTick` allocation when no rebalance is active. Cheap and clean.

That said: with the BackbeatConsumer refactor, the persistent ledger listener is installed once at consumer construction, so `listenerCount('empty')` is essentially always `>= 1` for any consumer using this ledger. So in practice the gate optimizes for ledgers created without a BackbeatConsumer (e.g. unit tests, or future callers). Still cheap to keep.

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -506,6 +506,158 @@ describe('BackbeatConsumer rebalance during batch processing', () => {
...
+        ], done);
+    });
+
+    // Three timepoints exercise the rebalance wait predicate
+    // (queue.idle AND ledger.empty):
+    //   (1) initial check — both already drained, IDLE path
+    //   (2) queue drains while ledger was already empty — completes
+    //       via the synchronous tail of the last task processed
+    //   (3) ledger drains via 'empty' event while queue was idle —
+    //       the Pattern A (committable: false + deferred ack) case
```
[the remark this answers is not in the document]
delthas: Right — simplified in [`8e2f11a4`](https://github.com/scality/backbeat/pull/2740/commits/8e2f11a4). The header now notes that two of the three timepoints are exercised here and points to the `BackbeatConsumer rebalance during batch processing` suite (BB-776) for the queue-drain case.

SylvainSenechal (approved):
when this is integrated into zenko, take a look at the artifact to see if we still have the erroneous state errors we used to have

lib/BackbeatConsumer.js
```diff
@@ -764,19 +795,43 @@ class BackbeatConsumer extends EventEmitter {
...
+            // Both the processing queue and the offset ledger must be drained
+            // before unassigning. The queue covers in-flight worker invocations
+            // (committable: true path); the ledger covers entries still waiting
+            // for their deferred onEntryCommittable to land (committable: false
+            // path — typically a Kafka status producer's delivery callback).
+            // Releasing the partition before the ledger drains would cause
+            // offsetsStore to hit ERR__STATE on the deferred callbacks (BB-758).
+            const isFullyDrained = () =>
+                this._processingQueue.idle() &&
+                this._offsetLedger.getProcessingCount(this._topic) === 0;
```
francoisferrand, line 807: could this cause an issue when "paused" (`isPaused()`)

when that happens, we don't commit offset, but also don't have subscription if I remember correctly: so should be ok, but please double check 🙏

francoisferrand (approved):

lib/BackbeatConsumer.js
delthas: Checked — safe in both pause paths. Public `pause()` (line 1018) goes via `unsubscribe()`, which triggers a revoke rebalance; the drain wait engages, `onEntryCommittable` still decrements the ledger (so it drains), and the `!isPaused()` guard on `offsetsStore` (line 886, predating this PR — see the BB-758 TODO) means the redeliver-on-resume semantics are unchanged. Circuit-breaker `_pauseAssignments()` uses librdkafka's native pause without unsubscribing, so `isPaused()` returns false and a rebalance there behaves identically to a normal one.

PR #2741, opened 2026-05-21
oplogPopulator: route oplog key through TransformObjectKey SMT when available

delthas:
## Summary

- Adds an `oplogPopulator.transformObjectKey` config flag (default `false`). When enabled, oplog source connectors are configured with the `com.scality.kafka.connect.transforms.TransformObjectKey` SMT (shipped by Zenko via [ZENKO-5274](https://scality.atlassian.net/browse/ZENKO-5274)), keying messages by the raw S3 object key.
- `extensions/oplogPopulator/constants.js`: `defaultConnectorConfig` (legacy `{ns, fullDocument.value.key}` key schema) + `smtKeyConfig` (projects `documentKey._id`, adds the SMT + `key.converter=StringConverter`), applied on top when the flag is set.
- `ConnectorsManager._getDefaultConnectorConfiguration` picks the SMT key config from the flag. `_processOldConnectors` scrubs SMT-only keys from a stale connector config when the flag is off, so a connector that previously ran with the SMT doesn't keep referencing a class that isn't there.
- `OplogPopulatorConfigValidator`: validates the new boolean flag.

## Context

[BB-768](https://scality.atlassian.net/browse/BB-768): the oplog Kafka topic keys messages from `fullDocument.value.key`, which is `null` on `update`/`delete` events (BB-355 removed `change.stream.full.document=updateLookup`). Those op types collapse onto `hash({ns, null})` while inserts spread across partitions, breaking per-object ordering across op types.

The fix lives in the Zenko-side SMT (ZENKO-5274): it derives the key from `documentKey._id` (always populated) and strips the arsenal master/version encoding, so master and all versions of an S3 object hash to the same partition. This PR is the Backbeat side that configures the connector to use it.

## Enablement model

Rather than Backbeat probing the Kafka Connect plugin path at runtime, the SMT is gated by an explicit config flag owned by the operator (ZKOP / zenkoversion CR). The flag is flipped on once the Connect image ships the `TransformObjectKey` plugin **and** the supporting Backbeat version is deployed. This keeps Backbeat free of auto-detection/runtime-flip machinery and gives a controlled rollout. A config change goes through the existing reconciliation: the connector is updated in place via `PUT /connectors/{name}/config` (no recreate, no resume-token loss — the change touches only the key schema + transforms, not the pipeline match stage).

## Follow-up

- ZKOP / zenko-operator must surface `transformObjectKey` into the generated Backbeat config (tracked separately).

Issue: BB-768

[ZENKO-5274]: https://scality.atlassian.net/browse/ZENKO-5274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BB-768]: https://scality.atlassian.net/browse/BB-768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extensions/oplogPopulator/constants.js
+++ b/extensions/oplogPopulator/constants.js
@@ -1,3 +1,94 @@
+const transformObjectKeyClass =
+    'com.scality.kafka.connect.transforms.TransformObjectKey';
+
+const defaultConnectorConfig = {
+    'connector.class': 'com.mongodb.kafka.connect.MongoSourceConnector',
+    'pipeline': '[]',
+    'collection': '',
+    // If no timestamp is provided, the startup mode will be equivalent
+    // to 'latest' which will pick up from the latest event in the oplog
+    'startup.mode': 'timestamp',
+    // JSON output converter config
+    // Using a string converter to avoid getting an over-stringified
+    // JSON that is returned by default
+    'output.format.value': 'json',
+    'value.converter.schemas.enable': false,
+    'value.converter': 'org.apache.kafka.connect.storage.StringConverter',
+    // Kafka message key config (legacy).
+    // The key schema projects {ns.coll, fullDocument.value.key}. fullDocument
+    // is null on update/delete events (BB-355 removed updateLookup), so for
+    // those op types the key collapses to hash({ns, null}) and lands on a
+    // different partition than insert events for the same S3 object. See
+    // BB-768. This is overridden by smtKeyConfig when the oplogPopulator
+    // 'transformObjectKey' flag is enabled (and the TransformObjectKey SMT
+    // is available in the Kafka Connect plugin path).
+    'output.format.key': 'schema',
+    'output.schema.key': JSON.stringify({
+        type: 'record',
+        name: 'keySchema',
+        fields: [{
+            name: 'ns',
+            type: [{
+                    name: 'ns',
+                    type: 'record',
+                    fields: [{
+                        name: 'coll',
+                        type: ['string', 'null'],
+                    }],
+                }, 'null'],
+        }, {
+            name: 'fullDocument',
+            type: [{
+               type: 'record',
+               name: 'fullDocumentRecord',
+               fields: [{
+                    name: 'value',
+                    type: [{
+                        type: 'record',
+                        name: 'valueRecord',
+                        fields: [{
+                            name: 'key',
+                            type: ['string', 'null'],
+                        }],
+                    }, 'null'],
+               }],
+            }, 'null'],
+        }],
+    }),
+};
+
+// Key-config overrides applied on top of defaultConnectorConfig when the
+// 'transformObjectKey' flag is set. Projects documentKey._id (always
+// populated on every change-stream event) and adds the TransformObjectKey
+// SMT, which strips the arsenal master/version encoding so master and all
+// versions of the same S3 object hash to the same partition. The SMT
+// rewrites the projected Struct into a plain string, serialised by
+// key.converter=StringConverter.
+//
+// The keys here that are absent from defaultConnectorConfig (transforms*,
+// key.converter) are the ones to drop when reverting a connector to the
+// legacy schema; output.schema.key is restored from defaultConnectorConfig.
+const smtKeyConfig = {
+    'output.schema.key': JSON.stringify({
+        type: 'record',
+        name: 'keySchema',
+        fields: [{
+            name: 'documentKey',
+            type: [{
+                type: 'record',
+                name: 'documentKeyRecord',
+                fields: [{
+                    name: '_id',
+                    type: ['string', 'null'],
+                }],
+            }, 'null'],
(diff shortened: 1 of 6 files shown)


francoisferrand (commented):
- Do we need to make this dynamic, or should we just let ZKOP enable the plugin depending on backbeat/zenko version (could be a "feature flag" in zenkoversion cr, so it can be enabled when Zenko has both the the plugin, the supporting backbeat version, and we decide it is time) ?
- How critical/dangerous is this whole fix (SMT...) ? Should it land in a patch release, or should we be more cautious?

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -207,6 +225,69 @@ class ConnectorsManager extends EventEmitter {
...
+     * @returns {undefined}
+     */
+    _applyKeyConfig(connector) {
+        const base = this._smtAvailable
+            ? constants.defaultConnectorConfigSMT
+            : constants.defaultConnectorConfig;
+        connector.config['output.schema.key'] = base['output.schema.key'];
+        if (this._smtAvailable) {
+            for (const k of constants.smtOnlyConfigKeys) {
+                connector.config[k] = base[k];
```
francoisferrand:
```suggestion
			Object.assign(connector.config, base);
```
or
```suggestion
			connector.config = { ...base, ...connector.config };
```

francoisferrand: `base` variable is always `constants.defaultConnectorConfigSMT` here, and not used in "else"
→ both `base` and `constants.defaultConnectorConfig` seem mostly useless (or over-engineered)

francoisferrand: ok, used somewhere else, but still looks kind of weird here : seems that it may be simpler to have just "base config" on one side, and "smp additions" added/removed conditionally.... (and consequently drop smtOnlyConfigKeys altogether I guess)

delthas:
Pivoted per your first point: dropped the runtime `/connector-plugins` probe and the auto-flip machinery entirely. The SMT is now gated by an `oplogPopulator.transformObjectKey` config flag (default off), which ZKOP/zenkoversion flips on once the Connect image ships the plugin and the supporting Backbeat is deployed. This also dissolves the `_applyKeyConfig` / `smtOnlyConfigKeys` complexity flagged inline — `constants.js` is now just `defaultConnectorConfig` + `smtKeyConfig` (the base + additions shape you suggested).

Per your second point: retargeted to `development/9.5` only (dropped the 9.4.1 patch); fix version updated to 9.5.0.

Follow-up: the zenkoversion CR → Backbeat config is tracked in ZKOP-553.

extensions/oplogPopulator/modules/ConnectorsManager.js
```diff
@@ -81,6 +82,11 @@ class ConnectorsManager extends EventEmitter {
         this._oldConnectors = [];
         this._allocationStrategy = params.allocationStrategy;
         this._pipelineFactory = params.pipelineFactory;
+        // When true, connectors are configured with the TransformObjectKey
+        // SMT so the oplog message key is the raw S3 object key. Enabled via
+        // the oplogPopulator 'transformObjectKey' config flag, set by the
+        // operator once Kafka Connect ships the TransformObjectKey plugin.
+        this._transformObjectKey = params.transformObjectKey || false;
```
francoisferrand, line 89: defaulting is not needed here, already handling by `joi` ?

extensions/oplogPopulator/constants.js
```diff
@@ -1,3 +1,94 @@
...
+    'startup.mode': 'timestamp',
+    // JSON output converter config
+    // Using a string converter to avoid getting an over-stringified
+    // JSON that is returned by default
+    'output.format.value': 'json',
+    'value.converter.schemas.enable': false,
+    'value.converter': 'org.apache.kafka.connect.storage.StringConverter',
+    // Kafka message key config (legacy).
+    // The key schema projects {ns.coll, fullDocument.value.key}. fullDocument
+    // is null on update/delete events (BB-355 removed updateLookup), so for
```
francoisferrand, line 19: c.f. https://scality.atlassian.net/browse/BB-768?focusedCommentId=477005 :
- we always pass the full document (AFAIK) on `update`, so the key should be available (but maybe not in "fullDocument")
- the `delete` event is not used AFAIK (we generate an update with the previous document right before it), and may be dropped instead

→ is there a simpler (as in "less maitenance") fix?

delthas:
Replaced by https://github.com/scality/backbeat/pull/2744

PR #2744, opened 2026-05-29
oplogPopulator: synthesise oplog key via $addFields in connector pipeline

delthas:
## Summary

Two commits on this branch:

1. **`oplogPopulator: synthesise oplog key via $addFields in connector pipeline`**
    - `extensions/oplogPopulator/pipeline/PipelineFactory.js`: insert a `$addFields` stage into the connector's change-stream pipeline (right after `$match`, before the conditional location-strip `$set`) that synthesises a top-level `key` field via `$ifNull` coalescing `$fullDocument.value.key` and `$updateDescription.updatedFields.value.key`.
    - `extensions/oplogPopulator/constants.js`: replace the broken `fullDocument` nested record in `output.schema.key` with a top-level `key: [string, null]` field. The existing `ns: {coll}` projection is preserved, so the Kafka message key remains `{ns: {coll: <bucket>}, key: <object-key>}` — bucket-level isolation is unchanged.
    - Tests updated in both pipeline-factory subclasses, the functional pipeline test (real-mongo aggregation against the new stage), and the `connectorConfig` fixture used by `ConnectorsManager` tests.
2. **`queuePopulator: counter for oplog events processed without a synthesised key`**
    - New `lib/queuePopulator/KafkaLogConsumer/KafkaLogConsumerMetrics.js` module (mirrors `LifecycleMetrics`: static class, per-method `try`/`catch` + `handleError`).
    - Counter `s3_oplog_event_missing_key_total` labelled by `opType`, incremented from `ListRecordStream._transform` when a consumed event passes the `objectMd` check but lacks the top-level `key`. Defensive observability — added per review.

Pure Backbeat change. No SMT, no Zenko image change, no operator flag. Always on.

## Context

[BB-768](https://scality.atlassian.net/browse/BB-768): the current key schema projects `fullDocument.value.key`, which is `null` on `update` events (since BB-355 removed `change.stream.full.document=updateLookup`). Every update for a given bucket therefore serialises to the same key Struct (`{ns:{coll:<bucket>}, fullDocument:null}`) and lands on **one** partition — both an ordering problem (insert → update on different partitions for the same object) and a throughput problem (a hot bucket's update traffic can't scale with partition count, blocking [BB-756](https://scality.atlassian.net/browse/BB-756)).

After ticket discussion ([Jira comment 477122](https://scality.atlassian.net/browse/BB-768?focusedCommentId=477122)) we picked the pipeline-`$addFields` route over a Kafka Connect SMT: it's a single-repo fix, and the measured server-side cost is ~600–900 ns/event ≈ ~1–2% of one core at 20k ops/s — not a throughput concern.

For the same logical S3 object — insert (key in `fullDocument.value.key`), master PUT update (key in `updatedFields.value.key`), replication-status update, delete-marker update — all yield the same `key` value → same partition. Master and version documents both store the same `value.key`, so master/version events also collapse to the same partition without prefix-stripping.

## Coupling caveat

The `$ifNull` variant relies on metadata always writing the **whole** `value` subdocument (full `$set`). Confirmed today against arsenal — there's no partial dotted-`$set` path for object MD. A future partial update (`$set: {"value.x": …}`) would not populate `updatedFields.value.key` and the resulting event would mis-partition. Accepted risk per the ticket discussion. The `s3_oplog_event_missing_key_total` counter exists to detect exactly this regression.

## Migration

The change is propagated to existing connectors via the existing in-place `PUT /connectors/{name}/config` reconciliation (no recreate, no resume-token loss — the change touches the key schema + the added `$addFields` stage only, not the `$match` stage). Downstream oplog consumers don't read the Kafka message key, so the new key shape is consumer-transparent. See [this comment](https://github.com/scality/backbeat/pull/2744#issuecomment-4575409482) for the full migration trace.

Issue: BB-768

[BB-768]: https://scality.atlassian.net/browse/BB-768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extensions/oplogPopulator/pipeline/PipelineFactory.js
+++ b/extensions/oplogPopulator/pipeline/PipelineFactory.js
@@ -70,6 +70,21 @@ class PipelineFactory {
         }
         const pipeline = [
             stage,
+            // Synthesise a top-level 'key' from fullDocument.value.key
+            // (insert/replace) or updateDescription.updatedFields.value.key
+            // (update). Relies on object MD writes always $set-ing the whole
+            // 'value' subdocument; a partial dotted $set would mis-partition.
+            // See BB-768 (superseded SMT alternative in PR #2741).
+            {
+                $addFields: {
+                    key: {
+                        $ifNull: [
+                            '$fullDocument.value.key',
+                            '$updateDescription.updatedFields.value.key',
+                        ],
+                    },
+                },
+            },
         ];
         if (this._locationStrippingBytesThreshold > 0) {
             pipeline.push({
--- a/tests/functional/oplogPopulator/PipelineFactory.js
+++ b/tests/functional/oplogPopulator/PipelineFactory.js
@@ -19,6 +19,7 @@ describe('PipelineFactory', function () {
     const collectionName = 'test-pipeline-stripping';
     let collection;
     let setStage;
+    let addFieldsStage;
 
     before(async () => {
         await client.connect();
@@ -26,7 +27,8 @@ describe('PipelineFactory', function () {
 
         const factory = new MultipleBucketsPipelineFactory(THRESHOLD);
         const pipeline = JSON.parse(factory.getPipeline(['test-bucket']));
-        setStage = pipeline[1];
+        addFieldsStage = pipeline[1];
+        setStage = pipeline[2];
     });
 
     afterEach(async () => {
@@ -145,6 +147,47 @@ describe('PipelineFactory', function () {
                 results[0].updateDescription.updatedFields.value.location,
                 undefined
             );
+        });
+    });
+
+    describe('key synthesis ($addFields)', () => {
+        it('should populate key from fullDocument.value.key on insert-shaped docs', async () => {
+            const doc = { fullDocument: { value: { key: 'my/object' } } };
+            await collection.insertOne(doc);
+            const results = await collection.aggregate([addFieldsStage]).toArray();
+            assert.strictEqual(results.length, 1);
+            assert.strictEqual(results[0].key, 'my/object');
+        });
+
+        it('should populate key from updateDescription.updatedFields.value.key on update-shaped docs', async () => {
+            const doc = { updateDescription: { updatedFields: { value: { key: 'my/object' } } } };
+            await collection.insertOne(doc);
+            const results = await collection.aggregate([addFieldsStage]).toArray();
+            assert.strictEqual(results.length, 1);
+            assert.strictEqual(results[0].key, 'my/object');
+        });
+
+        it('should prefer fullDocument.value.key when both are present', async () => {
+            const doc = {
+                fullDocument: { value: { key: 'from-full' } },
+                updateDescription: { updatedFields: { value: { key: 'from-update' } } },
+            };
+            await collection.insertOne(doc);
+            const results = await collection.aggregate([addFieldsStage]).toArray();
+            assert.strictEqual(results[0].key, 'from-full');
+        });
+
+        it('should leave key absent when neither path is populated', async () => {
+            // $ifNull returns missing (not null) when all inputs are missing.
+            // The connector's nullable Avro key schema emits this as null on
+            // the wire, so the partition outcome is the same as an explicit
+            // null. In production deletes are ignored, so this case doesn't
+            // occur for consumed events.
+            const doc = { fullDocument: null };
+            await collection.insertOne(doc);
+            const results = await collection.aggregate([addFieldsStage]).toArray();
+            assert.strictEqual(results.length, 1);
+            assert.strictEqual(results[0].key, undefined);
         });
     });
 });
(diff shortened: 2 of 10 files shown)


extensions/oplogPopulator/pipeline/PipelineFactory.js
```diff
@@ -97,6 +97,16 @@ class PipelineFactory {
                 }
             });
         }
+        pipeline.push({
```
francoisferrand: should be directly in the pipeline, at line 71?
or does it really _need_ to come after the conditional stripping?

```suggestion
        const pipeline = [
            stage,
            $addFields: {
                ...
            },
        ];
```

extensions/oplogPopulator/pipeline/PipelineFactory.js
```diff
@@ -97,6 +97,16 @@ class PipelineFactory {
...
+            $addFields: {
```
francoisferrand: should we add a comment here, explaining why we need workaround (update have the key, but in different place since we always do "full updates" [...]).

The comment could point to the closed PR to handle it with Kafka-connect plugin, if we do partial updates? (and/or mention the limit that partial updates MUST include the key as well)

francoisferrand (commented):
will the recently updated logic correctly _update_ the pipeline in all cases?
or do we have some situations were we may still stay with an old pipeline?

delthas:
> will the recently updated logic correctly _update_ the pipeline in all cases?

Yes, on the next Backbeat restart, no existing connector stays on the old pipeline:

- `OplogPopulator.initializeConnectors` → `ConnectorsManager._processOldConnectors` retrieves each existing connector's config from Kafka Connect, extracts buckets from `pipeline[0].$match` (works on both the old 1–2-stage shape and the new 2–3-stage shape — the bucket stage stays at index 0), then constructs a new `Connector` with `{...oldConfig, ...config}` where `config = _getDefaultConnectorConfiguration()`. The spread overwrites `output.schema.key` with the new single-`key`-field Avro, and the new `Connector` starts with `bucketsGotModified: true`.
- On the **first reconciliation tick** after restart, `Connector.updatePipeline(doUpdate=true)` rebuilds `_config.pipeline` via the new `PipelineFactory.getPipeline()` and PUTs the full `_config` (new schema + new pipeline) via `_kafkaConnect.updateConnectorConfig`. That's an in-place `PUT /connectors/{name}/config` — no recreate, no resume-token loss (the change doesn't touch `$match`).
- If that first PUT fails (network blip / Connect not ready), `bucketsGotModified` stays `true` and it retries on each subsequent tick until success.
- Newly-spawned connectors (from `addConnector`) go through `_getDefaultConnectorConfiguration` directly, so they start with the new schema + new pipeline from the first POST.

The one path that doesn't flip in-process: an already-running Backbeat with old code in memory — the in-process `Connector._config` was constructed before this code was deployed, so it keeps the old schema until restart. Backbeat restart is required as part of the rollout (which is normal for any Backbeat upgrade).

Also addressed in `e0ce22c9`:
- moved `$addFields` directly into the initial pipeline array (alongside `$match`); the location-strip `$set` remains the conditional append. The two stages are commutative (different field paths) so behavior is unchanged.
- added a short inline comment explaining the workaround + the partial-`$set` caveat, pointing at BB-768 and the superseded SMT PR #2741.

tests/functional/oplogPopulator/PipelineFactory.js
```diff
@@ -147,4 +149,45 @@ describe('PipelineFactory', function () {
...
+            const doc = {
+                fullDocument: { value: { key: 'from-full' } },
+                updateDescription: { updatedFields: { value: { key: 'from-update' } } },
+            };
+            await collection.insertOne(doc);
+            const results = await collection.aggregate([addFieldsStage]).toArray();
+            assert.strictEqual(results[0].key, 'from-full');
+        });
+
+        it('should leave key absent when neither path is populated', async () => {
```
francoisferrand, line 180: when this happens, we would write the message in "any" topic...
Is there a way to get some metrics when it happens?

francoisferrand (approved):

tests/functional/oplogPopulator/PipelineFactory.js
delthas: Added a Prometheus counter `s3_oplog_event_missing_key_total` (labelled by `opType`) — see commit f53649e8 (split into its own commit on the branch for clarity). It lives in a new `lib/queuePopulator/KafkaLogConsumer/KafkaLogConsumerMetrics.js` module mirroring the `LifecycleMetrics` pattern (static class, per-method `try`/`catch` + `handleError`). `ListRecordStream._transform` calls `KafkaLogConsumerMetrics.onMissingKey(this._logger, op)` when a consumed event passes the `objectMd` check but has no top-level `key`. Should stay at zero in steady-state; any non-zero rate signals the partitioning fix has been bypassed (e.g. a future partial dotted `$set` on object MD that doesn't carry `value.key`).

delthas:
@francoisferrand : moving this back to 9.4 since 9.4 == 9.5 atm and not sure this warrants a different branch.

maeldonn (approved):

PR #2745, opened 2026-06-01
feat(lifecycle): restrict v2 lifecycle indexes to master entries

delthas:
## Summary

Add `partialFilterExpression: { _id: { $gte: '\x7fM', $lt: '\x7fN' } }` to both v2 lifecycle indexes (`V2LifecycleLastModifiedPrefixed`, `V2LifecycleDataStoreNamePrefixed`) so they only contain master entries.

v2 Current listings only ever query masters; including versions in these indexes was inflating their size and forcing the planner to skip past version entries at scan time.

## Migration

Existing lifecycled buckets are unaffected: the conductor matches indexes by name in `_indexesGetOrCreate`, so non-partial indexes stay in place. Only newly-lifecycled buckets get the partial spec. Lazy migration, no script.

## v0 buckets

The filter matches no v0 entries; for any v0 bucket somehow running v2 lifecycle, the partial index would be empty and MongoDB transparently falls back to `_id_` — same as today's no-index path, no regression.

Full rationale, empirical measurements (3–58× index size reduction), and discussion of trade-offs: [BB-773](https://scality.atlassian.net/browse/BB-773).

Issue: BB-773

[BB-773]: https://scality.atlassian.net/browse/BB-773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,17 +1,13 @@
 const { versioning, algorithms } = require('arsenal');
 
-// Bounds of the v1 master key range in the metadata collection.
-// MASTER_KEY_PREFIX is arsenal's DbPrefixes.Master (\x7fM); MASTER_KEY_PREFIX_END
-// is the inclusive-exclusive upper bound on master keys, derived the same way
-// as arsenal's listingParamsMasterKeysV0ToV1.
-const MASTER_KEY_PREFIX = versioning.VersioningConstants.DbPrefixes.Master;
-const MASTER_KEY_PREFIX_END = algorithms.listTools.DelimiterTools.inc(MASTER_KEY_PREFIX);
+// Bounds of the v1 master key range in the metadata collection. Derived the
+// same way as arsenal's listingParamsMasterKeysV0ToV1.
+const masterKeyPrefix = versioning.VersioningConstants.DbPrefixes.Master;
+const masterKeyPrefixEnd = algorithms.listTools.DelimiterTools.inc(masterKeyPrefix);
 
 const constants = {
-    keyPrefixes: {
-        master: MASTER_KEY_PREFIX,
-        masterEnd: MASTER_KEY_PREFIX_END,
-    },
+    masterKeyPrefix,
+    masterKeyPrefixEnd,
     kafkaBacklogMetrics: {
         promMetricNames: {
             latestPublishedMessageTimestamp: 's3_backbeat_queue_latest_published_message_timestamp',
@@ -69,7 +65,7 @@ const constants = {
                     ],
                     name: 'V2LifecycleLastModifiedPrefixed',
                     partialFilterExpression: {
-                        _id: { $gte: MASTER_KEY_PREFIX, $lt: MASTER_KEY_PREFIX_END },
+                        _id: { $gte: masterKeyPrefix, $lt: masterKeyPrefixEnd },
                     },
                 },
                 {
@@ -80,7 +76,7 @@ const constants = {
                     ],
                     name: 'V2LifecycleDataStoreNamePrefixed',
                     partialFilterExpression: {
-                        _id: { $gte: MASTER_KEY_PREFIX, $lt: MASTER_KEY_PREFIX_END },
+                        _id: { $gte: masterKeyPrefix, $lt: masterKeyPrefixEnd },
                     },
                 },
             ],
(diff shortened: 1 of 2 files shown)


lib/constants.js
```diff
@@ -45,13 +45,17 @@ const constants = {
...
+            // listings only query masters, so excluding versions shrinks
+            // the index and skips them at scan time.
             v2: [
                 {
                     keys: [
                         { key: 'value.last-modified', order: 1 },
                         { key: '_id', order: 1 },
                     ],
                     name: 'V2LifecycleLastModifiedPrefixed',
+                    partialFilterExpression: { _id: { $gte: '\x7fM', $lt: '\x7fN' } },
```
francoisferrand: instead of hard-coding the values, can we use existing constants (from arsenal maybe) ?
or at least introduce them....

francoisferrand (approved):
* Does this change "requirements", i.e. require a minimum version of MongoDB ? (likely not an issue, but today we should still be compatible with 4.0 I think: so worth noting it -and ideally finding a place to document- if this is not the case)
* We don't do a "comparison" on existing indexes, so those would not be affected, right? (i.e. we would just stick with a sub-optimal and too large index)
* Should we create a procedure to "upgrade" the index, or just live with the possible discrepancy in prod? (not in this PRs though)

delthas:
> * Does this change "requirements", i.e. require a minimum version of MongoDB ? (likely not an issue, but today we should still be compatible with 4.0 I think: so worth noting it -and ideally finding a place to document- if this is not the case)
> 
>     * We don't do a "comparison" on existing indexes, so those would not be affected, right? (i.e. we would just stick with a sub-optimal and too large index)
> 
>     * Should we create a procedure to "upgrade" the index, or just live with the possible discrepancy in prod? (not in this PRs though)

- Partial indexes are GA since Mongo 3.2 so we should be good
- Right, we only try to create them if there are no indexes with this name, so there should be no upgrades of existing indexes
- I'd say live with the discrepancy in production for now, since we know creating indexes is super risky now

lib/constants.js
delthas: Done in 112ce5e2 — both bounds now derive from arsenal: `MASTER_KEY_PREFIX = versioning.VersioningConstants.DbPrefixes.Master` and `MASTER_KEY_PREFIX_END = algorithms.listTools.DelimiterTools.inc(MASTER_KEY_PREFIX)`, mirroring the pattern in arsenal's `listingParamsMasterKeysV0ToV1`. Exposed via `constants.keyPrefixes` so future master/version range queries can reuse them.

benzekrimaha (approved):

francoisferrand:
> I'd say live with the discrepancy in production for now, since we know creating indexes is super risky now

my question was about a _procedure_, i.e. for human operator: not automated, not I the case where the index is useless or to large to build, but for long-running platform where we may want to optimize the index size....

lib/constants.js
```diff
@@ -1,4 +1,17 @@
...
+// is the inclusive-exclusive upper bound on master keys, derived the same way
+// as arsenal's listingParamsMasterKeysV0ToV1.
+const MASTER_KEY_PREFIX = versioning.VersioningConstants.DbPrefixes.Master;
+const MASTER_KEY_PREFIX_END = algorithms.listTools.DelimiterTools.inc(MASTER_KEY_PREFIX);
+
 const constants = {
+    keyPrefixes: {
+        master: MASTER_KEY_PREFIX,
+        masterEnd: MASTER_KEY_PREFIX_END,
+    },
```
francoisferrand: why the duplicate: can't we just export `MASTER_KEY_PREFIX` and `MASTER_KEY_PREFIX_END` directly ?

francoisferrand (approved):

lib/constants.js
delthas: Good catch — done in b7bbe65c. Dropped the `keyPrefixes` wrapper; the two values now live directly on `constants` as `masterKeyPrefix` / `masterKeyPrefixEnd` (camelCase to match `lifecycleTaskVersions`, `indexesForFeature`, etc.).

PR #2746, opened 2026-06-01
feat(lifecycle): skip index creation on Veeam SosAPI buckets

delthas:
## Summary

If a bucket has Veeam's SosAPI capability enabled (`bucketInfo.capabilities.VeeamSOSApi`), skip auto-creating the v2 lifecycle indexes and stick with v1.

Veeam manages object cleanup via its own S3 calls rather than S3 lifecycle policies, so the indexes are dead weight on these buckets. Avoiding the build saves disk space and conductor cycles.

## Behavior

- New: SosAPI bucket without indexes → conductor returns v1, no auto-create.
- Existing: SosAPI bucket that already has the v2 indexes → still uses v2 (the indexes are there, no reason to ignore them).
- Non-SosAPI buckets: unchanged.
- New `sosApi` metric tag on `LifecycleMetrics.onLegacyTask` for visibility.

## Implementation

- `listMongodbBuckets` now projects `value.capabilities.VeeamSOSApi` from the metastore and plumbs `hasSosApi` through the task object — zero extra round trips.
- `_indexesGetOrCreate` checks `task.hasSosApi` after the existing-indexes check, returning v1 before the auto-create path.

Full rationale in [BB-778](https://scality.atlassian.net/browse/BB-778).

Issue: BB-778

[BB-778]: https://scality.atlassian.net/browse/BB-778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -265,7 +265,7 @@ class LifecycleConductor {
                 // Veeam buckets (SosAPI enabled) manage cleanup via their own
                 // S3 calls, not lifecycle policies. Skip the expensive index
                 // build and stick with v1.
-                log.trace('skipping index creation: SosAPI bucket');
+                log.trace('skipping index creation: SosAPI/Veeam has its own cleanup management');
                 LifecycleMetrics.onLegacyTask(log, 'sosApi');
                 return cb(null, lifecycleTaskVersions.v1);
             }
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -261,6 +261,15 @@ class LifecycleConductor {
                 return cb(null, lifecycleTaskVersions.v2);
             }
 
+            if (task.hasSosApi) {
+                // Veeam buckets (SosAPI enabled) manage cleanup via their own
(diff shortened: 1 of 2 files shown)


extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -261,6 +261,15 @@ class LifecycleConductor {
                 return cb(null, lifecycleTaskVersions.v2);
             }
 
+            if (task.hasSosApi) {
+                // Veeam buckets (SosAPI enabled) manage cleanup via their own
+                // S3 calls, not lifecycle policies. Skip the expensive index
+                // build and stick with v1.
+                log.trace('skipping index creation: SosAPI bucket');
```
DarkIsDude:
```suggestion
                log.trace('skipping index creation as SosAPI have it\'s own management');
```

DarkIsDude (approved):

extensions/lifecycle/conductor/LifecycleConductor.js
delthas: Addressed in f92d68c4 with slight wording tweak: "skipping index creation: SosAPI/Veeam has its own cleanup management" — fixes the grammar and makes the why clearer.

francoisferrand (approved):

PR #2747, opened 2026-06-03
Add Crr Cascade capabilities to backbeat crr replication

SylvainSenechal:
Issue: BB-767

Related PRs :
Arsenal : https://github.com/scality/Arsenal/pull/2628
Cloudserver : https://github.com/scality/cloudserver/pull/6179
CloudserverClient : https://github.com/scality/cloudserverclient/pull/24
S3utils : https://github.com/scality/s3utils/pull/395

### DIFF
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -1,14 +1,18 @@
 const async = require('async');
 const { S3Client, GetBucketReplicationCommand, GetObjectCommand } = require('@aws-sdk/client-s3');
 
-const errors = require('arsenal').errors;
-const jsutil = require('arsenal').jsutil;
-const ObjectMDLocation = require('arsenal').models.ObjectMDLocation;
-const ReplicationConfiguration = require('arsenal').models.ReplicationConfiguration;
+const { errors, jsutil, versioning } = require('arsenal');
+const { ObjectMDLocation, ReplicationConfiguration } = require('arsenal').models;
+const {
+    encode: encodeMicroVersionId,
+    decode: decodeMicroVersionId,
+    compare: compareMicroVersionId,
+    Ordering,
+} = versioning.VersionID;
 
 const ClientManager = require('../../../lib/clients/ClientManager');
 const BackbeatMetadataProxy = require('../../../lib/BackbeatMetadataProxy');
-const { 
+const {
     BackbeatRoutesClient,
     PutDataCommand,
     BatchDeleteCommand,
@@ -16,6 +20,10 @@ const {
     GetMetadataCommand,
     addContentLengthMiddleware,
     attachReqUids,
+    attachExpectContinueMiddleware,
+    VersionIdCollisionException,
+    StaleMicroVersionIdException,
+    MicroVersionIdAlreadyStoredException,
 } = require('@scality/cloudserverclient');
 
 const mapLimitWaitPendingIfError = require('../../../lib/util/mapLimitWaitPendingIfError');
@@ -159,11 +167,11 @@ class ReplicateObject extends BackbeatTask {
         }, cb);
     }
 
-    _putMetadata(entry, mdOnly, log, cb) {
+    _putMetadata(entry, mdOnly, conflict, log, cb) {
         this.retry({
             actionDesc: 'update metadata on target',
             logFields: { entry: entry.getLogInfo() },
-            actionFunc: done => this._putMetadataOnce(entry, mdOnly,
+            actionFunc: done => this._putMetadataOnce(entry, mdOnly, conflict,
                 log, done),
             shouldRetryFunc: err => err.retryable,
             onRetryFunc: err => {
@@ -452,11 +460,30 @@ class ReplicateObject extends BackbeatTask {
         const mpuConcLimit = this.repConfig.queueProcessor.mpuPartsConcurrency;
         return mapLimitWaitPendingIfError(locations, mpuConcLimit, (part, done) => {
             this._getAndPutPart(sourceEntry, destEntry, part, log, done);
-        }, (err, destLocations) => {
-            if (err) {
-                return this._deleteOrphans(destEntry, destLocations, log, () => cb(err));
+        }, (err, partResults) => {
+            let collisionResult;
+            const uploadedParts = [];
+            for (const result of (partResults || [])) {
+                if (!result) {
+                    continue;
+                }
+                if (result.isCollision) {
+                    collisionResult = collisionResult || result;
+                } else {
+                    uploadedParts.push(result);
+                }
             }
-            return cb(null, destLocations);
+            const hasPutDataConflict = collisionResult !== undefined;
+            if (err || hasPutDataConflict) {
+                // On error or conflict, drop all parts written
+                return this._deleteOrphans(destEntry, uploadedParts, log, () => {
+                    if (hasPutDataConflict) {
+                        return cb(null, [], collisionResult);
+                    }
+                    return cb(err);
+                });
+            }
+            return cb(null, partResults, undefined);
         });
     }
 
@@ -570,7 +597,9 @@ class ReplicateObject extends BackbeatTask {
                     // destination bucket has to be versioning enabled.
                     VersioningRequired: true,
                     RequestUids: log.getSerializedUids(),
+                    VersionId: sourceEntry.getEncodedVersionId(),
                 });
+                attachExpectContinueMiddleware(putCommand, this.backbeatDest.config?.requestHandler);
                 addContentLengthMiddleware(
                     putCommand,
                     response.ContentLength,
@@ -599,6 +628,17 @@ class ReplicateObject extends BackbeatTask {
                             if (incomingMsg.destroy) {
                                 incomingMsg.destroy();
                             }
+                        }
+
+                        if (err instanceof VersionIdCollisionException) {
+                            log.info('cascade putData: data already at destination', {
+                                method: 'ReplicateObject._getAndPutPartOnce',
+                                entry: destEntry.getLogInfo(),
+                            });
+                            return doneOnce(null, {
+                                isCollision: true,
+                                microVersionId: err.microVersionId,
+                            });
                         }
                         // eslint-disable-next-line no-param-reassign
                         err.origin = 'target';
@@ -642,7 +682,13 @@ class ReplicateObject extends BackbeatTask {
             });
     }
 
-    _putMetadataOnce(entry, mdOnly, log, cb) {
+    _putMetadataOnce(entry, mdOnly, conflict, log, cb) {
+        if (this._shouldSkipMetadata(entry.getMicroVersionId(), conflict, log)) {
+            log.info('skipping putMetadata: destination already has same or newer revision', {
+                entry: entry.getLogInfo(),
+            });
+            return cb();
+        }
         log.debug('putting metadata', {
             where: 'target', entry: entry.getLogInfo(),
             replicationStatus: entry.getReplicationSiteStatus(entry.getReplicationBackend()),
@@ -657,9 +703,10 @@ class ReplicateObject extends BackbeatTask {
             accountId = _extractAccountIdFromRole(this.targetRole);
         }
 
-        // sends extra header x-scal-replication-content to the target
-        // if it's a metadata operation only
-        const replicationContent = (mdOnly ? 'METADATA' : undefined);
+        // METADATA: update existing document (preserve stored location).
+        // DATA,METADATA: create a new document.
+        const localMdOnly = mdOnly || !!conflict;
+        const replicationContent = (localMdOnly ? 'METADATA' : 'DATA,METADATA');
         const mdBlob = entry.getSerialized();
         const command = new PutMetadataCommand({
             Bucket: entry.getBucket(),
@@ -671,6 +718,8 @@ class ReplicateObject extends BackbeatTask {
             // destination bucket has to be versioning enabled.
             VersioningRequired: true,
             RequestUids: log.getSerializedUids(),
+            MicroVersionId: entry.getMicroVersionId()
+                ? encodeMicroVersionId(entry.getMicroVersionId()) : '',
         });
         const writeStartTime = Date.now();
         return this.backbeatDest.send(command)
@@ -681,7 +730,9 @@ class ReplicateObject extends BackbeatTask {
             .catch(err => {
                 // eslint-disable-next-line no-param-reassign
                 err.origin = 'target';
-                if (err.ObjNotFound || err.name === 'ObjNotFound') {
+                if (err.ObjNotFound || err.name === 'ObjNotFound' ||
+                    err instanceof MicroVersionIdAlreadyStoredException ||
+                    err instanceof StaleMicroVersionIdException) {
                     return cbOnce(err);
                 }
                 log.error('an error occurred when putting metadata to S3',
@@ -868,7 +919,7 @@ class ReplicateObject extends BackbeatTask {
                 // put metadata in target bucket
                 next => {
                     // TODO check that bucket role matches role in metadata
-                    this._putMetadata(destEntry, false, log, next);
+                    this._putMetadata(destEntry, false, null, log, next);
                 },
             ], err => this._handleReplicationOutcome(
                 err, sourceEntry, destEntry, kafkaEntry, log, done));
@@ -920,16 +971,15 @@ class ReplicateObject extends BackbeatTask {
                     return this._getAndPutData(sourceEntry, destEntry, log,
                                                next);
                 }
-                return next(null, []);
+                return next(null, [], undefined);
             },
             // update location, replication status and put metadata in
             // target bucket
-            (destLocations, next) => {
+            (destLocations, conflict, next) => {
                 destEntry.setLocation(destLocations);
-                this._putMetadata(destEntry, mdOnly, log, err => {
+                return this._putMetadata(destEntry, mdOnly, conflict, log, err => {
                     if (err) {
-                        return this._deleteOrphans(
-                            destEntry, destLocations, log, () => next(err));
+                        return this._deleteOrphans(destEntry, destLocations, log, () => next(err));
                     }
                     return next();
                 });
@@ -938,17 +988,37 @@ class ReplicateObject extends BackbeatTask {
             err, sourceEntry, destEntry, kafkaEntry, log, done));
     }
 
-    _processQueueEntryRetryFull(sourceEntry, destEntry, kafkaEntry, log, done) {
-        log.debug('reprocessing entry as full replication',
-            { entry: sourceEntry.getLogInfo() });
+    // Returns true if putMetadata can be skipped because the destination already
+    // holds this revision or a newer one. Returns false when there is no conflict,
+    // when the destination microVersionId is absent or can't be parsed (proceed
+    // conservatively), or when the source holds a newer revision.
+    _shouldSkipMetadata(sourceMicroVersionId, conflict, log) {
+        if (!conflict) {
+            return false;
+        }
+        let destMvId = null;
+        if (conflict.microVersionId) {
+            const decoded = decodeMicroVersionId(conflict.microVersionId);
+            if (decoded instanceof Error) {
+                log.warn('could not decode microVersionId from putData 409, ' +
+                    'proceeding to putMetadata without skip optimisation', {
+                    error: decoded.message,
+                });
+            } else {
+                destMvId = decoded;
+            }
+        }
+        const comparison = compareMicroVersionId(sourceMicroVersionId, destMvId);
+        return destMvId !== null &&
+            (comparison === Ordering.OLDER || comparison === Ordering.EQUAL);
+    }
 
+    _processQueueEntryRetryFull(sourceEntry, destEntry, kafkaEntry, log, done) {
         return async.waterfall([
             next => this._getAndPutData(sourceEntry, destEntry, log, next),
-            // update location, replication status and put metadata in
-            // target bucket
-            (location, next) => {
-                destEntry.setLocation(location);
-                this._putMetadata(destEntry, false, log, next);
+            (destLocations, conflict, next) => {
+                destEntry.setLocation(destLocations);
+                return this._putMetadata(destEntry, false, conflict, log, next);
             },
         ], err => this._handleReplicationOutcome(
             err, sourceEntry, destEntry, kafkaEntry, log, done));
@@ -956,12 +1026,18 @@ class ReplicateObject extends BackbeatTask {
 
     _handleReplicationOutcome(err, sourceEntry, destEntry, kafkaEntry,
         log, done) {
+        if (err instanceof MicroVersionIdAlreadyStoredException ||
+            err instanceof StaleMicroVersionIdException) {
+            log.info('replication completed: metadata revision already at destination',
+                { entry: sourceEntry.getLogInfo(), reason: err.name });
+            this._publishReplicationStatus(sourceEntry, 'COMPLETED', { kafkaEntry, log });
+            return done(null, { committable: false });
+        }
         if (!err) {
             log.debug('replication succeeded for object, publishing ' +
                 'replication status as COMPLETED',
                 { entry: sourceEntry.getLogInfo() });
-            this._publishReplicationStatus(
-                sourceEntry, 'COMPLETED', { kafkaEntry, log });
+            this._publishReplicationStatus(sourceEntry, 'COMPLETED', { kafkaEntry, log });
             return done(null, { committable: false });
         }
         if (err.BadRole || err.name === 'BadRole' ||
@@ -991,8 +1067,7 @@ class ReplicateObject extends BackbeatTask {
                     { entry: sourceEntry.getLogInfo() });
                 return done();
             }
-            log.info('target object version does not exist, retrying ' +
-                'a full replication',
+            log.info('replication target object not found, retrying with full data write',
                 { entry: sourceEntry.getLogInfo() });
             // TODO: Is this the right place to capture retry metrics?
             return this._processQueueEntryRetryFull(
--- a/tests/unit/replication/ReplicateObject.spec.js
+++ b/tests/unit/replication/ReplicateObject.spec.js
@@ -1,10 +1,18 @@
 const assert = require('assert');
 const sinon = require('sinon');
+const { Readable } = require('stream');
 
 const QueueEntry = require('../../../lib/models/QueueEntry');
 const ReplicateObject = require('../../../extensions/replication/tasks/ReplicateObject');
 const ClientManager = require('../../../lib/clients/ClientManager');
 const locations = require('../../../conf/locationConfig.json');
+const { versioning } = require('arsenal');
+const { generateVersionId, encode } = versioning.VersionID;
+const {
+    VersionIdCollisionException,
+    StaleMicroVersionIdException,
+    MicroVersionIdAlreadyStoredException,
+} = require('@scality/cloudserverclient');
 
 const { replicationEntry } = require('../../utils/kafkaEntries');
 const fakeLogger = require('../../utils/fakeLogger');
@@ -72,6 +80,211 @@ describe('ReplicateObject', () => {
         sinon.restore();
     });
 
+    function makeMicroVersionIds() {
+        const a = generateVersionId('test', 'RG001');
+        const b = generateVersionId('test', 'RG001');
+        const [older, newer] = a > b ? [a, b] : [b, a]; // In case they are generated at the same millisecond
+        return { older, newer, olderEncoded: encode(older), newerEncoded: encode(newer) };
+    }
+
+    function makeBodyStream() {
+        const stream = new Readable({ read() {} });
+        process.nextTick(() => stream.push(null));
+        return stream;
+    }
+
+    function makeSourceEntry(microVersionId) {
+        return {
+            getBucket: () => 'src-bucket',
+            getObjectKey: () => 'key',
+            getEncodedVersionId: () => encode(generateVersionId('test', 'RG001')),
+            getMicroVersionId: () => microVersionId || null,
+            getOwnerId: () => 'canonical-id',
+            getLogInfo: () => ({}),
+            getLocation: () => [{
+                key: 'data-key', size: 10, start: 0,
+                dataStoreName: 'file', dataStoreETag: '1:abc',
+            }],
+            getContentLength: () => 10,
+        };
+    }
+
+    function makeDestEntry() {
+        return {
+            getBucket: () => 'dest-bucket',
+            getObjectKey: () => 'key',
+            getOwnerId: () => 'canonical-id',
+            getLogInfo: () => ({}),
+            setAmzServerSideEncryption: () => {},
+            setAmzEncryptionCustomerAlgorithm: () => {},
+            setAmzEncryptionKeyId: () => {},
+            setLocation: () => {},
+        };
+    }
+
+    describe('putData : cascade VersionIdCollisionException handling', () => {
+        let part;
+
+        beforeEach(() => {
+            part = { key: 'data-key', size: 10, start: 0,
+                dataStoreName: 'file', dataStoreETag: '1:abc' };
+            sinon.stub(task, '_publishReadMetrics').returns();
+            sinon.stub(task, '_publishDataWriteMetrics').returns();
+        });
+
+        function mockDataTransfer(destinationMicroVersionId) {
+            task.S3source = {
+                send: sinon.stub().resolves({
+                    Body: makeBodyStream(),
+                    ContentLength: 10,
+                }),
+            };
+            const err = new VersionIdCollisionException({
+                message: 'version already at destination',
+                microVersionId: destinationMicroVersionId,
+            });
+            task.backbeatDest = {
+                send: sinon.stub().rejects(err),
+            };
+        }
+
+        it('should return collision info on VersionIdCollisionException', done => {
+            const { older, olderEncoded } = makeMicroVersionIds();
+            mockDataTransfer(olderEncoded);
+            task._getAndPutPartOnce(makeSourceEntry(older), makeDestEntry(), part, fakeLogger, (err, result) => {
+                assert.ifError(err);
+                assert.ok(result && result.isCollision, 'should return collision info object');
+                assert.ok('microVersionId' in result,
+                    'collision info should include microVersionId');
+                sinon.assert.notCalled(task._publishDataWriteMetrics);
+                done();
+            });
+        });
+
+        it('should set data location and publish metrics when no collision', done => {
+            task.S3source = {
+                send: sinon.stub().resolves({ Body: makeBodyStream(), ContentLength: 10 }),
+            };
+            task.backbeatDest = {
+                send: sinon.stub().resolves({
+                    Location: [{ key: 'new-key', dataStoreName: 'file' }],
+                }),
+            };
+            task._getAndPutPartOnce(makeSourceEntry(), makeDestEntry(), part, fakeLogger, (err, result) => {
+                assert.ifError(err);
+                assert.deepStrictEqual(result, {
+                    key: 'new-key',
+                    start: 0,
+                    size: 10,
+                    dataStoreName: 'file',
+                    dataStoreETag: '1:abc',
+                    blockId: undefined,
+                });
+                sinon.assert.calledOnce(task._publishDataWriteMetrics);
+                done();
+            });
+        });
+    });
+
+    describe('putMetadata : cascade handling', () => {
+        let entry;
+
+        beforeEach(() => {
+            entry = QueueEntry.createFromKafkaEntry(replicationEntry);
+            task.targetRole = 'arn:aws:iam::123456789012:role/crr-role';
+        });
+
+        it('should pass through MicroVersionIdAlreadyStoredException and skip metrics', done => {
+            const metricsStub = sinon.stub(task, '_publishMetadataWriteMetrics').returns();
+            const loopErr = new MicroVersionIdAlreadyStoredException({
+                message: 'incoming microVersionId already at destination',
+            });
+            task.backbeatDest = { send: sinon.stub().rejects(loopErr) };
+            task._putMetadataOnce(entry, false, null, fakeLogger, err => {
+                assert.ok(err instanceof MicroVersionIdAlreadyStoredException);
+                sinon.assert.notCalled(metricsStub);
+                done();
+            });
+        });
+
+        it('should pass through StaleMicroVersionIdException', done => {
+            sinon.stub(task, '_publishMetadataWriteMetrics').returns();
+            const staleErr = new StaleMicroVersionIdException({
+                message: 'incoming revision is older than destination',
+            });
+            task.backbeatDest = { send: sinon.stub().rejects(staleErr) };
+            task._putMetadataOnce(entry, false, null, fakeLogger, err => {
+                assert.ok(err instanceof StaleMicroVersionIdException);
+                done();
+            });
+        });
+
+        it('should publish metrics and succeed on normal response', done => {
+            const metricsStub = sinon.stub(task, '_publishMetadataWriteMetrics').returns();
+            task.backbeatDest = { send: sinon.stub().resolves({}) };
+            task._putMetadataOnce(entry, false, null, fakeLogger, err => {
+                assert.ifError(err);
+                sinon.assert.calledOnce(metricsStub);
+                done();
+            });
+        });
+    });
+
+    describe('_handleReplicationOutcome : cascade outcomes', () => {
+        let sourceEntry, destEntry, kafkaEntry;
+
+        beforeEach(() => {
+            sourceEntry = QueueEntry.createFromKafkaEntry(replicationEntry);
+            destEntry = makeDestEntry();
+            kafkaEntry = {};
+            sinon.stub(task, '_publishReplicationStatus').returns();
+
+            sinon.stub(sourceEntry, 'toCompletedEntry').returns(sourceEntry);
+            sinon.stub(sourceEntry, 'toFailedEntry').returns(sourceEntry);
+            sinon.stub(sourceEntry, 'setReplicationSiteDataStoreVersionId').returns(sourceEntry);
+            sinon.stub(sourceEntry, 'getReplicationSiteDataStoreVersionId').returns('v1');
+        });
+
+        it('should mark COMPLETED for MicroVersionIdAlreadyStoredException', done => {
+            task._handleReplicationOutcome(
+                new MicroVersionIdAlreadyStoredException({}),
+                sourceEntry, destEntry, kafkaEntry, fakeLogger, () => {
+                    sinon.assert.calledWith(task._publishReplicationStatus,
+                        sourceEntry, 'COMPLETED', sinon.match.any);
+                    done();
+                });
+        });
+
+        it('should mark COMPLETED for StaleMicroVersionIdException', done => {
+            task._handleReplicationOutcome(
+                new StaleMicroVersionIdException({}),
+                sourceEntry, destEntry, kafkaEntry, fakeLogger, () => {
+                    sinon.assert.calledWith(task._publishReplicationStatus,
+                        sourceEntry, 'COMPLETED', sinon.match.any);
+                    done();
+                });
+        });
+
+        it('should mark COMPLETED on successful replication', done => {
+            task._handleReplicationOutcome(
+                null, sourceEntry, destEntry, kafkaEntry, fakeLogger, () => {
+                    sinon.assert.calledWith(task._publishReplicationStatus,
+                        sourceEntry, 'COMPLETED', sinon.match.any);
+                    done();
+                });
+        });
+
+        it('should mark FAILED for real errors', done => {
+            const realErr = Object.assign(new Error('network failure'), { origin: 'target' });
+            task._handleReplicationOutcome(
+                realErr, sourceEntry, destEntry, kafkaEntry, fakeLogger, () => {
+                    sinon.assert.calledWith(task._publishReplicationStatus,
+                        sourceEntry, 'FAILED', sinon.match.any);
+                    done();
+                });
+        });
+    });
+
     describe('_setTargetAccountMd', () => {
         it('should skip gettin target account info when auth type is assumeRole', done => {
             sinon.stub(task, '_setupDestClients').returns();
@@ -107,7 +320,7 @@ describe('ReplicateObject', () => {
                 send: sendStub,
             };
             task.targetRole = 'arn:aws:iam::123456789012:role/crr-role';
-            task._putMetadataOnce(entry, true, fakeLogger, err => {
+            task._putMetadataOnce(entry, true, null, fakeLogger, err => {
                 assert.ifError(err);
                 assert(sendStub.calledOnce);
                 assert.deepStrictEqual(sendStub.firstCall.args[0].input.AccountId, '123456789012');
@@ -123,7 +336,7 @@ describe('ReplicateObject', () => {
             };
             task.targetRole = 'arn:aws:iam::123456789012:role/crr-role';
             sinon.stub(task.destConfig.auth, 'type').value('role');
-            task._putMetadataOnce(entry, true, fakeLogger, err => {
+            task._putMetadataOnce(entry, true, null, fakeLogger, err => {
                 assert.ifError(err);
                 assert(sendStub.calledOnce);
                 assert.strictEqual(sendStub.firstCall.args[0].input.AccountId, undefined);
@@ -198,6 +411,53 @@ describe('ReplicateObject', () => {
                     destination: 'arn:aws:s3:::bucket-a',
                 }),
                 'v-1');
+        });
+    });
+
+    describe('_processQueueEntry', () => {
+        it('should call _putMetadata with mdOnly=false for zero-byte objects', done => {
+            const destEntry = {
+                getBucket: () => 'dest-bucket',
+                getObjectKey: () => 'key',
+                getLogInfo: () => ({}),
+                setLocation: () => {},
+                getReplicationBackend: () => 'site',
+                getReplicationSiteStatus: () => 'PENDING',
+            };
+            const sourceEntry = {
+                getBucket: () => 'src-bucket',
+                getObjectKey: () => 'key',
+                getEncodedVersionId: () => encode(generateVersionId('test', 'RG001')),
+                getMicroVersionId: () => null,
+                getReplicationContent: () => ['DATA', 'METADATA'],
+                getContentLength: () => 0,
+                getLocation: () => [],
+                getReducedLocations: () => [],
+                getLastModified: () => new Date().toISOString(),
+                getIsDeleteMarker: () => false,
+                getReplicationBackend: () => 'site',
+                getLogInfo: () => ({}),
+                toReplicaEntry: () => destEntry,
+            };
+
+            task.metricsHandler = { rpo: () => {} };
+            task.mProducer = { publishMetrics: () => {} };
+
+            sinon.stub(task, '_setupRoles').callsFake((e, l, cb) => cb(null, 'srcRole', 'dstRole'));
+            sinon.stub(task, '_setTargetAccountMd').callsFake((e, r, l, cb) => cb(null));
+            sinon.stub(task, '_publishReplicationStatus');
+
+            const putMetadataStub = sinon.stub(task, '_putMetadata')
+                .callsFake((e, mdOnly, conflict, l, cb) => {
+                    assert.strictEqual(mdOnly, false,
+                        'zero-byte objects must use DATA,METADATA (create) mode, not METADATA-only (update) mode');
+                    cb(null);
+                });
+
+            task.processQueueEntry(sourceEntry, {}, () => {
+                sinon.assert.calledOnce(putMetadataStub);
+                done();
+            });
         });
     });
 
@@ -753,6 +1013,94 @@ describe('ReplicateObject', () => {
             assert.strictEqual(credentials.accessKeyId, 'accessKeyNoAssumeRole');
             assert.strictEqual(credentials.secretAccessKey, 'secretKeyNoAssumeRole');
             assert.strictEqual(endpoint, 'http://s3.zenko.local:80/');
+        });
+    });
+
+    describe('_shouldSkipMetadata', () => {
+        let sourceMvId, olderSourceMvId, encodedNewer, encodedOlder;
+
+        before(() => {
+            const ids = makeMicroVersionIds();
+            olderSourceMvId = ids.older;
+            sourceMvId = ids.newer;
+            encodedNewer = ids.newerEncoded;
+            encodedOlder = ids.olderEncoded;
+        });
+
+        it('returns false when conflict is null', () => {
+            assert.strictEqual(task._shouldSkipMetadata(sourceMvId, null, fakeLogger), false);
+        });
+
+        it('returns false when conflict is undefined', () => {
+            assert.strictEqual(task._shouldSkipMetadata(sourceMvId, undefined, fakeLogger), false);
+        });
+
+        it('returns false when conflict has no microVersionId field', () => {
+            assert.strictEqual(task._shouldSkipMetadata(sourceMvId, { isCollision: true }, fakeLogger), false);
+        });
+
+        it('returns false when conflict has null microVersionId', () => {
+            const conflict = { isCollision: true, microVersionId: null };
+            assert.strictEqual(task._shouldSkipMetadata(sourceMvId, conflict, fakeLogger), false);
+        });
+
+        it('returns false when conflict has undecodable microVersionId', () => {
+            const conflict = { isCollision: true, microVersionId: 'tooshort' };
+            assert.strictEqual(task._shouldSkipMetadata(sourceMvId, conflict, fakeLogger), false);
+        });
+
+        it('returns false when source microVersionId is null (cannot compare)', () => {
+            const conflict = { isCollision: true, microVersionId: encodedNewer };
+            assert.strictEqual(task._shouldSkipMetadata(null, conflict, fakeLogger), false);
+        });
+
+        it('returns true when source revision equals destination revision', () => {
+            const conflict = { isCollision: true, microVersionId: encodedNewer };
+            assert.strictEqual(task._shouldSkipMetadata(sourceMvId, conflict, fakeLogger), true);
+        });
+
+        it('returns true when source revision is older than destination revision', () => {
+            const conflict = { isCollision: true, microVersionId: encodedNewer };
+            assert.strictEqual(task._shouldSkipMetadata(olderSourceMvId, conflict, fakeLogger), true);
+        });
+
+        it('returns false when source revision is newer than destination revision', () => {
+            const conflict = { isCollision: true, microVersionId: encodedOlder };
+            assert.strictEqual(task._shouldSkipMetadata(sourceMvId, conflict, fakeLogger), false);
+        });
+    });
+
+    describe('_putMetadataOnce with conflict', () => {
+        it('skips the request when conflict revision is equal to source (already at destination)', done => {
+            sinon.stub(task, '_publishMetadataWriteMetrics').returns();
+            const { newer, newerEncoded } = makeMicroVersionIds();
+            const entry = QueueEntry.createFromKafkaEntry(replicationEntry);
+            sinon.stub(entry, 'getMicroVersionId').returns(newer);
+            const conflict = { isCollision: true, microVersionId: newerEncoded };
+            const sendStub = sinon.stub().resolves({});
+            task.backbeatDest = { send: sendStub };
+            task.targetRole = 'arn:aws:iam::123456789012:role/crr-role';
+            task._putMetadataOnce(entry, false, conflict, fakeLogger, err => {
+                assert.ifError(err);
+                sinon.assert.notCalled(sendStub);
+                done();
+            });
+        });
+
+        it('proceeds with the request when conflict revision is older than source', done => {
+            sinon.stub(task, '_publishMetadataWriteMetrics').returns();
+            const { olderEncoded, newer } = makeMicroVersionIds();
+            const entry = QueueEntry.createFromKafkaEntry(replicationEntry);
+            sinon.stub(entry, 'getMicroVersionId').returns(newer);
+            const conflict = { isCollision: true, microVersionId: olderEncoded };
+            const sendStub = sinon.stub().resolves({});
+            task.backbeatDest = { send: sendStub };
+            task.targetRole = 'arn:aws:iam::123456789012:role/crr-role';
+            task._putMetadataOnce(entry, false, conflict, fakeLogger, err => {
+                assert.ifError(err);
+                sinon.assert.calledOnce(sendStub);
+                done();
+            });
         });
     });
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "9.5.0-preview.4",
+  "version": "9.5.0-preview.5",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "@aws-sdk/client-s3": "^3.921.0",
     "@aws-sdk/client-sts": "^3.921.0",
     "@aws-sdk/credential-providers": "^3.921.0",
-    "@scality/cloudserverclient": "^1.0.8",
+    "@scality/cloudserverclient": "^1.0.9",
     "@smithy/node-http-handler": "^3.3.3",
     "JSONStream": "^1.3.5",
     "arsenal": "git+https://github.com/scality/arsenal#8.5.6",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1203,6 +1203,16 @@
     "@smithy/types" "^4.12.0"
     tslib "^2.6.2"
 
+"@aws-sdk/middleware-expect-continue@^3.972.8":
+  version "3.972.14"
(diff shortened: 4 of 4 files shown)


extensions/replication/tasks/ReplicateObject.js
SylvainSenechal, line 1: Can already check this pr, but should really be reviewed after all the other cascade prs, as changes in these pr would also mean changes here

tests/unit/replication/ReplicateObject.spec.js
SylvainSenechal, line 1: I think we can functional tests instead of just these,
But waiting for Arsenal/cloudserver to be merged, as it will be easier to make these tests (functional tests in backbeat rely on an image of cloudserver)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -654,6 +740,15 @@ class ReplicateObject extends BackbeatTask {
                 if (err.ObjNotFound || err.name === 'ObjNotFound') {
                     return cbOnce(err);
                 }
+                if (err.$metadata?.httpStatusCode === 409) {
```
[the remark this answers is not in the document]
SylvainSenechal: addressed, will use specific exception now

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -422,11 +425,16 @@ class ReplicateObject extends BackbeatTask {
         const mpuConcLimit = this.repConfig.queueProcessor.mpuPartsConcurrency;
         return mapLimitWaitPendingIfError(locations, mpuConcLimit, (part, done) => {
             this._getAndPutPart(sourceEntry, destEntry, part, log, done);
-        }, (err, destLocations) => {
+        }, (err, partResults) => {
+            const allPartsAlreadyAtDest = !err &&
+                partResults.length > 0 &&
+                partResults.every(result => result === partAlreadyAtDest);
+            const destLocations = allPartsAlreadyAtDest ? [] :
+                (partResults || []).filter(result => result && result !== partAlreadyAtDest);
```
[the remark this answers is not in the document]
SylvainSenechal: stale review

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -562,6 +578,64 @@ class ReplicateObject extends BackbeatTask {
...
+                                return doneOnce(cascadeDataComplete);
+                            }
+                            // proceed: source is newer, skip data write and update metadata
+                            log.info('cascade putData: data already at destination, ' +
+                                'proceeding with metadata update',
+                                {
+                                    method: 'ReplicateObject._getAndPutPartOnce',
+                                    entry: destEntry.getLogInfo(),
+                                });
+                            return doneOnce(null, partAlreadyAtDest);
```
francoisferrand: best to introduce a dedicated function here, which can go through the different checks and just return the error if/when needed: then we have a single log line in here. easier to test, easier to read, and less duplication...

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -641,10 +715,20 @@ class ReplicateObject extends BackbeatTask {
             // destination bucket has to be versioning enabled.
             VersioningRequired: true,
             RequestUids: log.getSerializedUids(),
+            MicroVersionId: entry.getMicroVersionId(),
         });
         const writeStartTime = Date.now();
         return this.backbeatDest.send(command)
             .then(data => {
+                if (data.ReplicationLoop) {
```
francoisferrand: did we not have a "conflict" exception in that case as well?
(i.e. does backbeat continue differently when cascadeLoopDetected vs cascadeDataComplete : does not seem to be the same in `_handleReplicationOutcome`, but maybe missing something)

tests/unit/replication/ReplicateObject.spec.js
francoisferrand: keeping unit test is good, functional test should just be an addition?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -654,6 +738,15 @@ class ReplicateObject extends BackbeatTask {
                 if (err.ObjNotFound || err.name === 'ObjNotFound') {
                     return cbOnce(err);
                 }
+                if (err instanceof StaleMicroVersionIdException) {
```
SylvainSenechal: @maeldonn Actually I'm rethinking this : I think from the design we said staleException on putMetada would be marked as failed. 
But 🤔 Really, do we want to mark an object replication status as failed in that case ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -422,11 +428,16 @@ class ReplicateObject extends BackbeatTask {
         const mpuConcLimit = this.repConfig.queueProcessor.mpuPartsConcurrency;
         return mapLimitWaitPendingIfError(locations, mpuConcLimit, (part, done) => {
             this._getAndPutPart(sourceEntry, destEntry, part, log, done);
-        }, (err, destLocations) => {
+        }, (err, partResults) => {
+            const allPartsAlreadyAtDest = !err &&
```
maeldonn: Move this constant declaration after the error check below
```suggestion
            const allPartsAlreadyAtDest =
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -531,23 +542,28 @@ class ReplicateObject extends BackbeatTask {
...
-                    Bucket: destEntry.getBucket(),
-                    Key: destEntry.getObjectKey(),
-                    CanonicalID: destEntry.getOwnerId(),
-                    ContentMD5: partObj.getPartETag(),
-                    Body: incomingMsg,
-                    // destination bucket has to be versioning enabled.
-                    VersioningRequired: true,
-                    RequestUids: log.getSerializedUids(),
-                });
+                const putCommand = attachExpectContinueMiddleware(
```
maeldonn: Should we set `expectContinueHeader` ?

https://github.com/scality/cloudserverclient/pull/25/changes#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R31

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -562,6 +578,64 @@ class ReplicateObject extends BackbeatTask {
                         return doneOnce(null, partObj.getValue());
                     })
                     .catch(err => {
+                        if (err instanceof VersionIdCollisionException) {
```
maeldonn: Could you extract this logic to a new function ? And add some tests.

extensions/replication/tasks/ReplicateObject.js
maeldonn: Isn't the field in camelCase now ?

maeldonn: I think a 200 status is appropriate in this case. It’s conceptually identical to a successful replication, except the data is already present on the destination. On the contrary, the stale case is a legitimate error, since the data was never actually replicated from the source to the destination

extensions/replication/tasks/ReplicateObject.js
maeldonn: To be discussed. We should introduce a new CONFLICT status to differentiate it from real replication errors, because in my opinion, it's not a SUCCESS.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -915,20 +1009,30 @@ class ReplicateObject extends BackbeatTask {
...
+                    { entry: sourceEntry.getLogInfo() });
+            } else if (err === cascadeDataComplete) {
+                log.info('replication completed: destination already holds ' +
+                    'this version with an equal or newer revision',
+                    { entry: sourceEntry.getLogInfo() });
+            } else {
+                log.debug('replication succeeded for object, publishing ' +
+                    'replication status as COMPLETED',
+                    { entry: sourceEntry.getLogInfo() });
+            }
```
maeldonn:
```suggestion
_handleReplicationOutcome(err, sourceEntry, destEntry, kafkaEntry, log, done) {
    const logInfo = { entry: sourceEntry.getLogInfo() };

    if (err === cascadeLoopDetected) {
        log.info('replication completed via cascade loop: object already at destination with the same revision', logInfo);
    } else if (err === cascadeDataComplete) {
        log.info('replication completed: destination already holds this version with an equal or newer revision', logInfo);
    } else if (!err) {
        log.debug('replication succeeded for object, publishing replication status as COMPLETED', logInfo);
    } else {
        // Handle actual replication errors here if needed, or just pass to done()
    }
}
```

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: i changed it, to be handled with an LoopException error, felt weird to have a successful api calls followed by an if to check if there is an error

SylvainSenechal: yeah but not pushed yet, anyways completely removed now

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: ah yeah, currently it's set to true so we will always use it, but maybe we shouldn't use it all the time, like only for object > 4mb or something like that ?

francoisferrand: > maybe we shouldn't use it all the time, like only for object > 4mb or something like that

Seems like a good idea, but for a followup: let's make it work first (which may actually be easier to reproduce with small objects), then we can improve.

The threshold will probably need to be fine-tuned: this is the point where request(s) change from being latency-bound to bandwidth-bound I guess, and depends on the network interface speed...

> it's set to true so we will always use it

is this kind of a breaking change: i.e. what happens if a (newer) backbeat is used to replicate to an older cloudserver which does not support `expectContinueHeader` ? Will the call be rejected, or will it just go through seamlessly?
Maybe handled through the automatic send if no continue, should we (can we) track the "continue" response, so we know if the remote supports it?

More generally, are we introducing some kind of breaking change with the "conflict handling" logic, or will it degrade nicely when the CRR target runs an older version?

(no need to fix here, but we should create a followup ticket to ensure we review/test this)

maeldonn: Can you create the follow-up ticket @SylvainSenechal ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -946,77 +1049,87 @@ class ReplicateObject extends BackbeatTask {
...
+            (location, allPartsAlreadyAtDest, next) => {
                 destEntry.setLocation(location);
-                this._putMetadata(destEntry, false, log, next);
+                this._putMetadata(destEntry, allPartsAlreadyAtDest, log, next);
             },
         ], err => this._handleReplicationOutcome(
             err, sourceEntry, destEntry, kafkaEntry, log, done));
     }
 
     _handleReplicationOutcome(err, sourceEntry, destEntry, kafkaEntry,
```
maeldonn, line 1027: This section has quite a bit of nested conditional logic and duplicate checks that make it hard to read and maintain.

Could we flatten this using guard clauses (early returns) and abstract the err.XYZ || err.name === 'XYZ' checks into a helper function? It would drastically reduce the cognitive load of this function. Let me know if you want to pair on it!

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +459,15 @@ class ReplicateObject extends BackbeatTask {
         const mpuConcLimit = this.repConfig.queueProcessor.mpuPartsConcurrency;
         return mapLimitWaitPendingIfError(locations, mpuConcLimit, (part, done) => {
             this._getAndPutPart(sourceEntry, destEntry, part, log, done);
-        }, (err, destLocations) => {
+        }, (err, partResults) => {
+            const destLocations = (partResults || [])
```
maeldonn: Could you extract this change into its own commit so you can explain the 'why' in the commit description ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -642,6 +668,63 @@ class ReplicateObject extends BackbeatTask {
             });
     }
 
+    _resolveVersionIdCollision(collisionErr, sourceEntry, destEntry, log) {
```
maeldonn: The logical flow here using early returns is great, but the function feels a bit hard to read due to the repetitive log metadata block being copied into every if statement.

We can dramatically clean this up by extracting method and destEntry.getLogInfo() into a single const logMeta object at the top of the function, then passing it directly to log.info/log.error. What do you think?

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I did, forgot to link it here.
https://scality.atlassian.net/browse/BB-793

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I agree its trash code and the diff is hard to read with the last else

I just changed it and tried something that i didn't want to do first but i think it's fine : for each condition, directly publish/return, without having to do it at the end of the function. I believe its quite readable this way

SylvainSenechal: Could still use some refactor maybe, although all the erro don't have the same form, I think the diff is reasonnable here

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: true the whole function is just logs

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: For now I added some comments to clarify this a bit

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -561,16 +577,20 @@ class ReplicateObject extends BackbeatTask {
...
-                    Bucket: destEntry.getBucket(),
-                    Key: destEntry.getObjectKey(),
-                    CanonicalID: destEntry.getOwnerId(),
-                    ContentMD5: partObj.getPartETag(),
-                    Body: incomingMsg,
-                    // destination bucket has to be versioning enabled.
-                    VersioningRequired: true,
-                    RequestUids: log.getSerializedUids(),
-                });
+                const putCommand = attachExpectContinueMiddleware(
```
francoisferrand: nit: would be more readable to attach the middleware _after_


```suggestion
                const putCommand =  new PutDataCommand({ ... });
                attachExpectContinueMiddleware(putCommand, ...);
                addContentLengthMiddleware(putCommand, ...);
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -592,6 +612,17 @@ class ReplicateObject extends BackbeatTask {
                         return doneOnce(null, partObj.getValue());
                     })
                     .catch(err => {
+                        if (err instanceof VersionIdCollisionException) {
+                            abortController.abort();
+                            sourceStreamAborted = true;
+                            if (incomingMsg.destroy) {
+                                incomingMsg.destroy();
+                            }
```
francoisferrand, line 630: this is already below


```suggestion
                        if (!destRequestAborted) {
                            // Abort the source stream
                            abortController.abort();
                            sourceStreamAborted = true;
                            if (incomingMsg.destroy) {
                                incomingMsg.destroy();
                            }
                        }
                        if (err instanceof VersionIdCollisionException) {
                            const resolved = this._resolveVersionIdCollision(
                                err, sourceEntry, destEntry, log);
                            return doneOnce(resolved.err, resolved.result);
                        }
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -592,6 +612,17 @@ class ReplicateObject extends BackbeatTask {
...
+                            const resolved = this._resolveVersionIdCollision(
```
francoisferrand: nit:
```suggestion
                            const { err, result } = this._resolveVersionIdCollision(
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +459,20 @@ class ReplicateObject extends BackbeatTask {
...
+            // exclude it so it is not stored as a new location or orphan-deleted.
+            const destLocations = (partResults || [])
+                .filter(result => result && result !== partAlreadyAtDest);
             if (err) {
                 return this._deleteOrphans(destEntry, destLocations, log, () => cb(err));
             }
-            return cb(null, destLocations);
+            // If all parts were already at destination, destLocations is [].
+            // Sending [] without metadata-only mode would wipe the object's
+            // location field at the destination, so force metadata-only.
```
francoisferrand: should we not pass `partResults` anyway? I mean for `deleteOrphans` we do indeed need to skip the parts already at destination ; but for continuation we need to pass _all_ parts from destination : so that we can construct the metadata of the MPU object ?

the issue I think of is racing to create the object: so 2 source replicate half the objects each, and eventually one of them must still write the metadata _after filling it with the actual part "ids"_.

→ if all parts at destination, maybe we can indeed skip metadata creation (or best to try it anyway? this cheap vs the upload of all parts anyway....)
→ if some pats are missing, we must still try to write the metadata I guess (contrary to simple objects, the case of a "partial" MPU is not the same!)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +459,20 @@ class ReplicateObject extends BackbeatTask {
...
+            const allPartsAlreadyAtDest = partResults.length > 0 &&
+                partResults.every(result => result === partAlreadyAtDest);
```
francoisferrand:
```suggestion
            const allPartsAlreadyAtDest = destLocations.length == 0
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -642,6 +673,44 @@ class ReplicateObject extends BackbeatTask {
...
+            return { err: cascadeLoopDetected };
+        }
+        if (cmp === Ordering.OLDER) {
+            log.info('cascade stale on putData: ' +
+                'destination already has a newer revision', logMeta);
+            return { err: cascadeDataStale };
+        }
+        log.info('cascade putData: data already at destination, ' +
+            'proceeding with metadata update', logMeta);
+        return { err: null, result: partAlreadyAtDest };
```
francoisferrand: nit: use a switch on `cmp` ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -642,6 +673,44 @@ class ReplicateObject extends BackbeatTask {
...
+            return { err: null, result: partAlreadyAtDest };
+        }
+        const destinationMicroVersionId = decode(collisionErr.microVersionId);
+        if (destinationMicroVersionId instanceof Error) {
+            log.error('failed to decode microVersionId from putData 409',
+                { ...logMeta, error: destinationMicroVersionId.message });
+            return { err: destinationMicroVersionId };
+        }
+        const sourceMicroVersionId = sourceEntry.getMicroVersionId();
+        if (!sourceMicroVersionId) {
```
francoisferrand: is this case not handled already by `compareMicroVersionId` ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -642,6 +673,44 @@ class ReplicateObject extends BackbeatTask {
...
 
+    _resolveVersionIdCollision(collisionErr, sourceEntry, destEntry, log) {
+        const logMeta = {
+            method: 'ReplicateObject._resolveVersionIdCollision',
+            entry: destEntry.getLogInfo(),
+        };
+        if (!collisionErr.microVersionId) {
+            log.info('cascade putData: data at destination, ' +
+                'no microVersionId, proceeding with putMetadata', logMeta);
+            return { err: null, result: partAlreadyAtDest };
```
francoisferrand: can this really happen?
→ if we get a collision, it means we have the "newer" cloudserver code, which returns the microVersionId : the field should always be set
→ so this case should either be an error (like decoding), or just the usual path of `compareMicroVersionId()` (since microVersionId is not set when "creating" object)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -684,6 +755,24 @@ class ReplicateObject extends BackbeatTask {
                 if (err.ObjNotFound || err.name === 'ObjNotFound') {
                     return cbOnce(err);
                 }
+                if (err instanceof MicroVersionIdAlreadyStoredException) {
```
francoisferrand: don't we have a single exception in the latest API, leaving the caller to check microVersionID to identify if this is a "loop" or "stale" ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -684,6 +755,24 @@ class ReplicateObject extends BackbeatTask {
...
+                }
+                if (err instanceof StaleMicroVersionIdException) {
+                    log.info('cascade stale on putMetadata: ' +
+                        'destination has a newer revision, marking COMPLETED',
+                        {
+                            method: 'ReplicateObject._putMetadataOnce',
+                            entry: entry.getLogInfo(),
+                        });
+                    return cbOnce(cascadeDataStale);
+                }
```
francoisferrand, line 737: this whole block seems duplicate: we just pass an error, which is processed by _handleReplicationOutcome
→ we may just continue passing the err unaffected here (drop the sentinels cascadeLoopDetected and cascadeDataStale), and use instanceOf in _handleReplicationOutcome
→ no need for extra logging here, there are logs as in _handleReplicationOutcome

francoisferrand (commented):
a few cleanup, but most importantly I am kind of worried on how to handle the conflict in case of MPU specifically...

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Thinking about this.
DeleteOrphans is fine as it is.

I think the big problem we have (although maybe it's not a expected product scenario) is
4 locations with a mix of cascade and multi destination, using MPU :

a -> b
a -> c

b -> d
c -> d

b and c will write to d at the same time, so let's say for 10 parts, half of them will be written from b and the other from c.
Then for the parts already written by the other location, the current location will get a list of partResults with some elements being errors partAlreadyAtDest, and other elements will be proper locations.
If we pass the whole unfiltered partResults down the call chain, we will get corrupted location data with beause of the errors in partResults.


I feel like a correct way to handle this would be : 
When doing an mpu, as soon as I see a partAlreadyAtDest errors, I stop because it means another location is already doing its replication, so I will let it do it fully. 

But it seems like this could get racy, when both sources detect a partAlreadyAtDest -_-

SylvainSenechal: And thinking some more I think what we would want is : 

When doing a putData, even when there is a 409, return the existing location for that part to backbeat
Then regardless of we had had collision on putData or not, at the end we have the data locations for all parts, backbeat can send putMEtaData with all parts locations, and so in our scenarios, both location b and c will call putMetaData, and one of the call will fail but its alright

cc @maeldonn

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I built it this way, prefer to have backbeat just checking error instead of reruning the whole microVersionId comparison that cloudserver already did

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: yes im gonna do this its a nice refactoring

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: yes this was redundant and too defensive

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: here microVersionId is the microVersionId at the destination, so it could be undefined.
I think it's unlikely but maybe a scenario like replication already setup before cascaded : we have a replica already but no microVersionId
This is quite defensive though but decode can crash if we don't check this 

Keeping this open because in cloudserver there is still discussion about microVersionId beging set when creating the object

extensions/replication/tasks/ReplicateObject.js
francoisferrand: > When doing a putData, even when there is a 409, return the existing location for that part to backbeat

409 is not "by part" : the 409 means "there is already an object" (i.e. metadata), it is not related to a single part.
each part upload is completely transparent.

> When doing an mpu, as soon as I see a partAlreadyAtDest errors, I stop because it means another location is already doing its replication, so I will let it do it fully.

agreed (though to be precise: once we detect it, it means the metadata was written: so it has already done the replication)

also need to delete **all** uploaded parts, since each replication source will upload their own parts independently until either success or conflict. On conflict all parts must be removed.

b and c may both write parts (putData) in parallel, but they will do so completely indepedently: the only moment when they will interact (i.e. conflict will be detectable) is when the metadata has been written by either → from that point the "remaining" putData will fail, and the loser must remove his "own" parts (**all** the successful putData he made) and skip the putMetadata...

---
this also means there is likely a followup here, to redesign MPU upload both for safety (avoid orphans) and optimization (detect conflict earlier and/or allow reusing parts already uploaded)

francoisferrand (commented):
multiple open topics open in https://github.com/scality/cloudserver/pull/6179, regarding "empty" versionId, shape of errors returned, ... → best to settle these before reviewing

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Oh ok I didn't understand that the 409 would only happen after we have done the putMetadata, but yeah it makes sense, it's when the data gets linked to the metadata.

In that case, our problem might become much easier, since what happens now is :
- one of the 2 location finishes its replication first, with multiple putData and 1 putMetedata
- The second one is slower, and as soon as the first one has done the putMetadata, this one starts receiving the "partAlreadyAtDestination". Then that second one eventually also tries to do a putMetadata, but because it's already there, it will get a MicroVersionIdAlreadyStoredException, and the error will trigger a deleteOrphans. 

So I think the only follow we may wanna do, which is not even fully related to this feature is : stopping the code execution earlier as soon as we receive one "partAlreadyAtDest"

SylvainSenechal: https://scality.atlassian.net/browse/BB-798

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +457,18 @@ class ReplicateObject extends BackbeatTask {
         const mpuConcLimit = this.repConfig.queueProcessor.mpuPartsConcurrency;
         return mapLimitWaitPendingIfError(locations, mpuConcLimit, (part, done) => {
             this._getAndPutPart(sourceEntry, destEntry, part, log, done);
-        }, (err, destLocations) => {
+        }, (err, partResults) => {
+            // partAlreadyAtDest signals data already at dest (cascade putData 409);
```
francoisferrand: `partAlreadyAtDest` name is not correct : we cannot detect that a "part" is already at destination, only that there is already an object (i.e. a document in mongo with the specified `key` and `versionId`).
→ should really be `objectAlreadyAtDest`

....and it changes the logic a bit : when it happen (even on a single part), all successful parts must be deleted.

```suggestion
if (err) {
    return this._deleteOrphans(destEntry, destLocations, log, () => cb(err));
}

const destLocations = (partResults || []).filter(result => result && result !== partAlreadyAtDest);
if (destLocations.length != partResults.length) {
    // object already exist, release all parts then check if metadata needs to be updated
    return this._deleteOrphans(destEntry, destLocations, log, () => 
cb(null, destLocations, true);
}

return this._deleteOrphans(destEntry, destLocations, log, () => 
cb(null, destLocations, false);
```

i.e. we cannot make a _partial_ write of object data. To create the metadata, all parts must be known: so  all of them must have been written successfully. If there is a "conflict" on any single part, it means another replicant has finished uploading parts and created the metadata document : so we must drop all the parts we wrote, and just update the object metadata if needed....

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -642,6 +662,43 @@ class ReplicateObject extends BackbeatTask {
             });
     }
 
+    _resolveVersionIdCollision(collisionErr, sourceEntry, destEntry, log) {
```
francoisferrand: this whole function is not correct in case of putData(Parts) : if we receive VersionIdCollision, it means there is already a object with this `versionId`.
- We **must not** _create_ or _replace_ the metadata with the new data (data is immutable)
- We **must** immediately delete whatever we already uploaded
- The `microVersionId` comparison can be used only after this, to decide if we need to proceed with the metadata update (e.g. not the location, but the other fields: tags, ...) or can skip it

→ here the function is used to compute `partAlreadyAtDest`, which would make `_getAndPutData` silently "hide" the error if there is already some data at the destination BUT the metadata we are trying to replicate is newer.
→ _getAndPutData() should probably return the max/last microVersionId it received (in case of conflict) or nothing it is wrote all data successfully (i.e. no conflict)

(the logic here is what should happen on microVersionId conflict on metadata, not here on VersionIdConflict)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -946,22 +1006,35 @@ class ReplicateObject extends BackbeatTask {
...
+            this._publishReplicationStatus(sourceEntry, 'COMPLETED', { kafkaEntry, log });
+            return done(null, { committable: false });
+        }
         if (!err) {
             log.debug('replication succeeded for object, publishing ' +
                 'replication status as COMPLETED',
                 { entry: sourceEntry.getLogInfo() });
-            this._publishReplicationStatus(
-                sourceEntry, 'COMPLETED', { kafkaEntry, log });
+            this._publishReplicationStatus(sourceEntry, 'COMPLETED', { kafkaEntry, log });
```
francoisferrand: we do exactly the same in all 3 branches : do we need 2 errors and different logs?

```
if (!err || err instanceof ...) {
    log.debug('replication succeeded for object, publishing ' +
                'replication status as COMPLETED',
                { entry: sourceEntry.getLogInfo(), err });
    this._publishReplicationStatus(sourceEntry, 'COMPLETED', { kafkaEntry, log });
}
```

extensions/replication/tasks/ReplicateObject.js
francoisferrand: also please create followups to
- ignore (other) error if a single part has a conflict
- abort other parts upload as soon as we identify a conflict (i.e. imagine a 1000 parts object: if we have a conflict on first part, no point trying the other parts)
- consider changing the putPartData protocol to create actual MPU parts (which can be garbage-collected by lifecycle after transfer is aborted), for extra safety (not strictly related to CRR, but best to track it)

francoisferrand (changes requested):
handling of MPU is not correct : each "replicant" must write their own object fully, there is no situation where 2 "sources" each replicate part of the data (nor a way for these to identify the data they have written). Each source create the metadata document with the parts they uploaded, so they MUST successfully upload all parts data and MUST not have a conflict on any part.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -946,22 +1006,35 @@ class ReplicateObject extends BackbeatTask {
...
+            return done(null, { committable: false });
+        }
+        if (err instanceof StaleMicroVersionIdException) {
+            log.info('replication completed: destination already holds ' +
+                'this version with a newer revision',
+                { entry: sourceEntry.getLogInfo() });
+            this._publishReplicationStatus(sourceEntry, 'COMPLETED', { kafkaEntry, log });
+            return done(null, { committable: false });
+        }
```
francoisferrand: there is a risk of creating orphans here, please **create a followup**:

1. new object was created → try to replicate DATA+META
2. putDataParts succeeds (no conflict)
3. before we could put metadata, another site was able to putData (same as us, not metadata → impossible to detect conflict) and putMetadata
4. putMetadata will thus fail with MicroVersionIdAlreadyStoredException/StaleMicroVersionIdException

→ in that case, data created in 2. must be deleted, i.e. call `_deleteOrphans`. `_deleteOrphans` MUST NOT be called if this was a META-only update / if we did not call putDataParts

extensions/replication/tasks/ReplicateObject.js
francoisferrand: also in that case the `x-scal-replication-content` sent to cloudserver MUST not be DATA+METADATA anymore, but be "downgraded" to METADATA (to keep the existing data)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +457,18 @@ class ReplicateObject extends BackbeatTask {
...
+            // exclude it so it is not stored as a new location or orphan-deleted.
+            const destLocations = (partResults || [])
+                .filter(result => result && result !== partAlreadyAtDest);
             if (err) {
                 return this._deleteOrphans(destEntry, destLocations, log, () => cb(err));
             }
-            return cb(null, destLocations);
+            // When no new data locations were written, force metadata-only mode (noNewDataLocations)
+            // so cloudserver preserves the destination's existing location field instead of overwriting it
+            const noNewDataLocations = destLocations.length === 0;
```
benzekrimaha: AFAIU this still does not implement what was agreed in the MPU thread: if some parts return partAlreadyAtDest and some were written by us, destLocations is a partial list, noNewDataLocations is false, and we proceed to putMetadata with only the parts we wrote → corrupted location metadata at destination. A conflict on any part means another replicant already wrote the metadata: we must delete all the parts we uploaded and downgrade to a metadata-only update. i.e. the check should be destLocations.length !== partResults.length, not === 0

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -920,17 +989,42 @@ class ReplicateObject extends BackbeatTask {
...
-            (destLocations, next) => {
+            (destLocations, noNewDataLocations, next) => {
+                // METADATA mode tells cloudserver to preserve the destination's existing
+                // data location instead of overwriting it. It applies when either:
+                //   - this is a metadata-only replication entry (mdOnly), or
+                //   - data was already at the destination from another path (noNewDataLocations,
+                //     i.e. cascade putData 409).
+                // Zero-byte objects have no location to preserve, so METADATA mode is
+                // always wrong for them regardless of the above flags.
+                const localMdOnly = (mdOnly || noNewDataLocations) && sourceEntry.getContentLength() > 0;
```
benzekrimaha: This silently changes the existing behavior for metadata-only entries of zero-byte objects: before, mdOnly=true was passed through; now we force a full metadata write with location: []. Since this is a behavior change on a path that predates cascade → why? and where is the test for both zero-byte branches?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -671,6 +728,8 @@ class ReplicateObject extends BackbeatTask {
             // destination bucket has to be versioning enabled.
             VersioningRequired: true,
             RequestUids: log.getSerializedUids(),
+            MicroVersionId: entry.getMicroVersionId()
+                ? encodeMicroVersionId(entry.getMicroVersionId()) : '',
```
benzekrimaha, line 722: the empty-string convention is one of the open topics on scality/cloudserver#6179. Is the shape settled now?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -1,21 +1,25 @@
...
 const { S3Client, GetBucketReplicationCommand, GetObjectCommand } = require('@aws-sdk/client-s3');
 
-const errors = require('arsenal').errors;
-const jsutil = require('arsenal').jsutil;
-const ObjectMDLocation = require('arsenal').models.ObjectMDLocation;
-const ReplicationConfiguration = require('arsenal').models.ReplicationConfiguration;
+const { errors, jsutil, versioning } = require('arsenal');
+const { ObjectMDLocation } = require('arsenal').models;
+const { ReplicationConfiguration } = require('arsenal').models;
+const { encode: encodeMicroVersionId, decode } = versioning.VersionID;
```
benzekrimaha: nit: rename both or neither; decode unqualified next to encodeMicroVersionId reads as if they operate on different things

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -875,6 +934,12 @@ class ReplicateObject extends BackbeatTask {
         }
 
         const mdOnly = !sourceEntry.getReplicationContent().includes('DATA');
+        log.warn('AAAAA 1: _processQueueEntry starting', {
```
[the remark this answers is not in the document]
benzekrimaha: still valid , think about the dopping the commit before merging

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Yes sorry I had not addressed comments yet when you did the review I was on cloudserver, it's updated now

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Yeah I rewrote this because I think from the start I overcomplicated it.
The flow is :
if any err or conflict : delete all written data, else we can all parts like usually.

For the follow up :
- I already created the follow up to do an early exit on multi part upload
- The error handling, I think no need for a follow up, I added it in the code by checking conflict before checking error (although maybe its a bit simplistic, it kinda raise some questions about whether we still want to act on the errors, at least logs, maybe more depending on the errors, not sure)
 - Created a ticket for putPartDataMPU https://scality.atlassian.net/browse/BB-801

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: yep this is very intentional, using undefined would cause the sdk to not even send the header and would break the cascade replication. A bit fragile if you ask me but this is the current design

extensions/replication/tasks/ReplicateObject.js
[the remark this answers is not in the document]
SylvainSenechal: yeah this was ongoing tests, removed

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Ok I see this is a left over from some missunderstanding that I had. 

This function can be completely removed, but before I do it, there is one thing i wanna confirm :
 
In the design, we said in case of putData collision, we would still do the putMetadata, and let cloudserver on putMetadata deal with stale microVersionId 

But as you said, since with the putData error we get a microVersionId returned, it is possible for backbeat to do the comparison directly here to avoid a useless putMetadata.

So I wanna decide which thing we should do : I would prefer skipping the putMetadata in backbeat if its useless, but also I wonder if there was a specific reason (that i can't find) we went with the design of sending putMetada regardless of the result of putData ?

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: You're right there is behavior change but it's needed, there is a comment explaining a bit what's happening. 

Side note, mdOnly is and was defined like this in backbeat 
`const mdOnly = !sourceEntry.getReplicationContent().includes('DATA');
`
In cloudserver, replicationContent is set like this : 

```
    const content = isMD || objSize === 0 ? ['METADATA'] : ['DATA', 'METADATA'];
    if (hasCloudBackend && operationType) {
        content.push(operationType);
    }
```

So the change makes sense. I added a small test though to cover the behavior

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I think thats what I had a few week ago + the logs but I got suggested a refactoring

Are you sure you want to drop log in case of conflicts on microVersionId ? Seems like something we might wanna know about in case of issue

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Yeah I see, created this : https://scality.atlassian.net/browse/BB-802
Looking from afar i think it shouldn't be too hard to handle, I will address it when we merge this pr otherwise this one will never be merged

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Ok I talked with Maël and review again, I'm removing this function and making the improvement to check the microversion id on conflict to avoid useless put metadata

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -684,6 +729,10 @@ class ReplicateObject extends BackbeatTask {
                 if (err.ObjNotFound || err.name === 'ObjNotFound') {
                     return cbOnce(err);
                 }
+                if (err instanceof MicroVersionIdAlreadyStoredException ||
+                    err instanceof StaleMicroVersionIdException) {
+                    return cbOnce(err);
+                }
```
francoisferrand: do we need the condition here? Can't we just log the error (like all other errors) and pass it up in these case as well?
(the main difference is that we would get a ERROR log instead of INFO... and log here instead of in the continuation callback)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -600,6 +624,25 @@ class ReplicateObject extends BackbeatTask {
...
+                            const destMvIdRaw = err.microVersionId
+                                ? decodeMicroVersionId(err.microVersionId) : null;
+                            if (destMvIdRaw instanceof Error) {
+                                log.error('failed to decode microVersionId from putData 409',
+                                    {
+                                        method: 'ReplicateObject._getAndPutPartOnce',
+                                        entry: destEntry.getLogInfo(),
+                                        error: destMvIdRaw.message,
+                                    });
+                                return doneOnce(destMvIdRaw);
```
francoisferrand: should still be a collision : even if we can't parse the microVersionID, we are still in the case "data already at destination", so we shoud still try to write the metadata anyway ?

(the only difference is that we can't really make the optimization of skipping the metadata updated if the microVersionId is already up to date on target)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -600,6 +624,25 @@ class ReplicateObject extends BackbeatTask {
                                 incomingMsg.destroy();
                             }
                         }
+
+                        if (err instanceof VersionIdCollisionException) {
+                            const destMvIdRaw = err.microVersionId
```
francoisferrand: nit: variable name is not clear, the abbreviation obfuscates it. Just `microVersoinId` or `decodedMicroVersoinId` is better?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +456,29 @@ class ReplicateObject extends BackbeatTask {
...
+                        if (hasPutDataConflict) {
+                            // Another replicant already stored the object data
+                            // Compare microVersionIds to decide what to do with metadata:
+                            // - source is newer : proceed in metadata-only mode
+                            // - same or older : skip putMetadata
+                            const destMvId = collisionResult.destMvIdRaw;
+                            const srcMvId = sourceEntry.getMicroVersionId() || null;
+                            const isLoop = srcMvId === destMvId;
+                            const isStale = destMvId !== null &&
+                                (srcMvId === null || srcMvId > destMvId);
```
francoisferrand: this is where `compareMicroVersionId` shines, please use the arsenal function:
- it makes the intent explicit (just comparing micro version id)
- the function should handle the corner cases itself ("null" micro version id...)
- we don't actually need to differentiate loop vs stale here

It would look something like this, i.e. we skip if we receive the destination micro version id (field present and no parse error) AND the micro version id is newer or equal at destination
```
const skipPutMetadata = destMvId !== undefined && compareMicroVersoinId(sourceEntry.getMicroVersionId(), destMvId) <= 0;
return cb(null, [], true, skipPutMetadata);
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -600,6 +624,25 @@ class ReplicateObject extends BackbeatTask {
                                 incomingMsg.destroy();
                             }
                         }
+
+                        if (err instanceof VersionIdCollisionException) {
+                            const destMvIdRaw = err.microVersionId
+                                ? decodeMicroVersionId(err.microVersionId) : null;
+                            if (destMvIdRaw instanceof Error) {
```
francoisferrand: nit: I wonder if it is best to do the decoding here, or just pass the field "verbatim" in the part and let `_getAndPutData()` decode

- it would avoid useless decoding of every parts' microVersionId
- it would increase locally (
- but the code here still needs to make sure if forwards both the microVersionId "value" AND the information that the field was indeed provided

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -600,6 +624,25 @@ class ReplicateObject extends BackbeatTask {
                                 incomingMsg.destroy();
                             }
                         }
+
+                        if (err instanceof VersionIdCollisionException) {
+                            const destMvIdRaw = err.microVersionId
+                                ? decodeMicroVersionId(err.microVersionId) : null;
```
francoisferrand: the `: null` here masks the absence of field. So if the request does not include the field, it will be the same value as a "null" microVersionId.
→ is this expected?
→ semantically, absence of field is similar to decode error: we don't know, so the safe (in-depth) approach is to perform the putMetadata anyway (we are not _sure_ it would conflict)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +456,29 @@ class ReplicateObject extends BackbeatTask {
...
-        }, (err, destLocations) => {
-            if (err) {
-                return this._deleteOrphans(destEntry, destLocations, log, () => cb(err));
+        }, (err, partResults) => {
+            const collisionResult = (partResults || []).find(r => r && r.isCollision);
+            const hasPutDataConflict = collisionResult !== undefined;
+            if (err || hasPutDataConflict) {
+                // On error or conflict, drop all parts written
+                return this._deleteOrphans(destEntry,
+                    (partResults || []).filter(r => r && !r.isCollision), log, () => {
```
francoisferrand: nit: I don' really like the 2 list traversals, I'd rather we do it only once. not sure how though...

(collisions are unlikely, so most of the time we would go through the whole list anyway: so `find()` is not faster than `filter()` in the nominal case... and in case of conflict, we do the filter anyway)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -920,13 +969,27 @@ class ReplicateObject extends BackbeatTask {
...
+                    });
+                    return next();
+                }
+                // METADATA mode tells cloudserver to preserve the destination's existing
+                // data location instead of overwriting it. It applies when either:
+                //   - this is a metadata-only replication entry (mdOnly), or
+                //   - data was already at the destination from another path (noNewDataLocations,
+                //     i.e. cascade putData 409).
+                // Zero-byte objects have no location to preserve, so METADATA mode is
+                // always wrong for them regardless of the above flags.
```
francoisferrand: `Zero-byte objects have no location to preserve` is correct, but the point of METADATA vs DATA+METADATA is not just preservation

- DATA+METADATA is expected to **_CREATE_** a new document (and hopefully it will fail if the document for that version already exists)
- METADATA is expected to **_UPDATE_** the existing document (and hopefully it will fail if there is no such document) - and will preserve the locations (which may be empty, but can be preserved as well)

Hence the the condition below is simpler:
```
const localMdOnly = (mdOnly || noNewDataLocations)
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -920,13 +969,27 @@ class ReplicateObject extends BackbeatTask {
                     return this._getAndPutData(sourceEntry, destEntry, log,
                                                next);
                 }
-                return next(null, []);
+                return next(null, [], false, false);
             },
             // update location, replication status and put metadata in
             // target bucket
-            (destLocations, next) => {
+            (destLocations, noNewDataLocations, skipMetadata, next) => {
```
francoisferrand: these 2 parameters are not independent, there are actually only 3 cases:
- no conflict (i.e. as before)
- conflict with metadata already up to date
- conflict but we have newer metadata which need to be replicated

not sure how best to handle this, maybe an object ("conflict") with the microVersionId : so we also improve locality by performing the microVersionId comparison (optimization) here?
→ `undefined` in the usual case (no conflict), don't even need the `false` like line 972
→ `{ microVersionId: ... }` (possibly with the "conflict" field you already have) otherwise ; skip if compareMicroVersionId() <= 0

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -939,29 +1002,54 @@ class ReplicateObject extends BackbeatTask {
...
+                if (skipMetadata) {
+                    log.info('skipping putMetadata: destination already has same or newer revision', {
+                        entry: destEntry.getLogInfo(),
+                    });
+                    return next();
+                }
+                // On retry, only keep METADATA mode when data was already at the destination
+                // from another path (cascade 409) and the object is non-zero-byte.
+                // For zero-byte there is no location to preserve; always write metadata directly.
```
francoisferrand: not correct, c.f. above

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -939,29 +1002,54 @@ class ReplicateObject extends BackbeatTask {
...
-
         return async.waterfall([
             next => this._getAndPutData(sourceEntry, destEntry, log, next),
             // update location, replication status and put metadata in
             // target bucket
-            (location, next) => {
-                destEntry.setLocation(location);
-                this._putMetadata(destEntry, false, log, next);
+            (destLocations, noNewDataLocations, skipMetadata, next) => {
+                if (skipMetadata) {
```
francoisferrand: this block is duplicated and not trivial (it handles putData conflict resolution) : should be written just once... does it makes sense do do this in `_putMetadata`, by adding it another parameter for putData conflict ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -939,29 +1002,54 @@ class ReplicateObject extends BackbeatTask {
...
+                { entry: sourceEntry.getLogInfo() });
+            this._publishReplicationStatus(sourceEntry, 'COMPLETED', { kafkaEntry, log });
+            return done(null, { committable: false });
+        }
+        if (err instanceof StaleMicroVersionIdException) {
+            log.info('replication completed: destination already holds ' +
+                'this version with a newer revision',
+                { entry: sourceEntry.getLogInfo() });
+            this._publishReplicationStatus(sourceEntry, 'COMPLETED', { kafkaEntry, log });
+            return done(null, { committable: false });
```
francoisferrand: this is the same code, dedup by using the err.message or logging the err object itself

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -657,9 +700,9 @@ class ReplicateObject extends BackbeatTask {
             accountId = _extractAccountIdFromRole(this.targetRole);
         }
 
-        // sends extra header x-scal-replication-content to the target
-        // if it's a metadata operation only
-        const replicationContent = (mdOnly ? 'METADATA' : undefined);
+        // Send x-scal-replication-content so cloudserver know the putMetadata api
+        // is used in the context of a replication
+        const replicationContent = (mdOnly ? 'METADATA' : 'DATA,METADATA');
```
SylvainSenechal: This is required as otherwise, in non mdOnly situation, there is no replicationContent header, and cloudserver has no way no know we are calling putMetadata in the context of a replication.

Anyways i think sending data, metadata for non mdOnly should've been added earlier.

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Yes, I updated the code, this will be a situation where we still do a putMetadata. 
I also moved the decoding to another place, more centralized

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I moved this but yes, the null will be treated as a situation where perform putMetadata

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I'm gonna refactor this but I prefer extracting a helper functoin that putting it in putMetadata. 
It looks very weird to call a function with an object that's used at the top of the function so that the function itself decides whether it should conitnue or not

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I updated it with a conflict object and move some logic around into a helper function as we still need it in 2 different places

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I think its nicer to keep, these arent unexpected errors, and they are logged already in the replication outcome.

I moved them with the if (err.ObjNotFound) above though, as its a similar situation : expected error

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I changed this to use a for loop. 
Side note, technically we could have multiple collision error on putData, and yet we only do conflict comparison on one microVersionId. But I think this is fine as we should expect all collision to happen on the same versionId

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -938,30 +996,63 @@ class ReplicateObject extends BackbeatTask {
...
     }
 
-    _processQueueEntryRetryFull(sourceEntry, destEntry, kafkaEntry, log, done) {
-        log.debug('reprocessing entry as full replication',
-            { entry: sourceEntry.getLogInfo() });
+    // Returns true if putMetadata can be skipped because the destination already
+    // holds this revision or a newer one. Returns false when there is no conflict,
+    // when the destination microVersionId is absent or can't be parsed (proceed
+    // conservatively), or when the source holds a newer revision.
+    _shouldSkipMetadata(sourceMicroVersionId, conflict, log) {
```
SylvainSenechal, line 995: Discuss : Could be moved into putMetada so that we don't have to call this from 2 places, and instead inline it.

But imo it's a weird pattern to add an extra param to a function, then call that function and use that extra param at the top to decide to potentially not run this function

SylvainSenechal:
The code is getting smaller after each review 😆 
Retested in codespace after last push

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -938,30 +996,63 @@ class ReplicateObject extends BackbeatTask {
...
+                log.warn('could not decode microVersionId from putData 409, ' +
+                    'proceeding to putMetadata without skip optimisation', {
+                    error: decoded.message,
+                });
+            } else {
+                destMvId = decoded;
+            }
+        }
+        const comparison = compareMicroVersionId(sourceMicroVersionId, destMvId);
+        return destMvId !== null &&
```
SylvainSenechal, line 1012: May completely remove the null check if I end up update the compareMicroVersionId function.
Still reviewing cloudserver, haven't done it/decided yet

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +461,32 @@ class ReplicateObject extends BackbeatTask {
...
-                return this._deleteOrphans(destEntry, destLocations, log, () => cb(err));
+        }, (err, partResults) => {
+            let collisionResult;
+            const nonCollisionParts = [];
+            for (const result of (partResults || [])) {
+                if (!result) {
+                    continue;
+                }
+                if (result.isCollision) {
+                    collisionResult = collisionResult || result;
```
francoisferrand, line 471: nit:
```suggestion
                    collisionResult ||= result;
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +461,32 @@ class ReplicateObject extends BackbeatTask {
         const mpuConcLimit = this.repConfig.queueProcessor.mpuPartsConcurrency;
         return mapLimitWaitPendingIfError(locations, mpuConcLimit, (part, done) => {
             this._getAndPutPart(sourceEntry, destEntry, part, log, done);
-        }, (err, destLocations) => {
-            if (err) {
-                return this._deleteOrphans(destEntry, destLocations, log, () => cb(err));
+        }, (err, partResults) => {
+            let collisionResult;
+            const nonCollisionParts = [];
```
francoisferrand: nit:
```suggestion
            const parts = [];
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -600,6 +632,17 @@ class ReplicateObject extends BackbeatTask {
...
                         }
+
+                        if (err instanceof VersionIdCollisionException) {
+                            log.info('cascade putData: data already at destination', {
+                                method: 'ReplicateObject._getAndPutPartOnce',
+                                entry: destEntry.getLogInfo(),
+                            });
+                            return doneOnce(null, {
+                                isCollision: true,
+                                destinationMicroVersionId: err.microVersionId,
```
francoisferrand: I would keep the name simple here : the `destination` prefix adds verbosity, but does not improve readability IMHO

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -452,11 +461,32 @@ class ReplicateObject extends BackbeatTask {
...
             }
-            return cb(null, destLocations);
+            const hasPutDataConflict = collisionResult !== undefined;
+            if (err || hasPutDataConflict) {
+                // On error or conflict, drop all parts written
+                return this._deleteOrphans(destEntry, nonCollisionParts, log, () => {
+                    if (hasPutDataConflict) {
+                        return cb(null, [], {
+                            microVersionId: collisionResult.destinationMicroVersionId,
+                        });
```
francoisferrand: nit: could even return the `collisionResult` directly


```suggestion
                        return cb(null, [], collisionResult);
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -920,16 +967,27 @@ class ReplicateObject extends BackbeatTask {
...
+                        entry: destEntry.getLogInfo(),
+                    });
+                    return next();
+                }
+                // METADATA tells cloudserver to UPDATE the existing metadata document
+                // (preserving the stored location). DATA,METADATA tells cloudserver to
+                // CREATE a new document. Use METADATA when:
+                //   - this is a metadata-only replication entry (mdOnly), or
+                //   - data was already written to the destination from another path (conflict).
+                const localMdOnly = mdOnly || conflict !== undefined;
```
francoisferrand: nit: would work the same if `conflict` is null ?
```suggestion
                const localMdOnly = mdOnly || !!conflict;
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -920,16 +967,27 @@ class ReplicateObject extends BackbeatTask {
...
+                    log.info('skipping putMetadata: destination already has same or newer revision', {
+                        entry: destEntry.getLogInfo(),
+                    });
+                    return next();
+                }
+                // METADATA tells cloudserver to UPDATE the existing metadata document
+                // (preserving the stored location). DATA,METADATA tells cloudserver to
+                // CREATE a new document. Use METADATA when:
+                //   - this is a metadata-only replication entry (mdOnly), or
+                //   - data was already written to the destination from another path (conflict).
```
francoisferrand: this comment seems too verbose and out-of-place here, does not really help the understand the code : kind of paraphrasing the code. Maybe more appropriate in `_putMetadata`, either in function doc or where we compute `const replicationContent = ` ?

```suggestion
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -920,16 +967,27 @@ class ReplicateObject extends BackbeatTask {
...
+                const localMdOnly = mdOnly || conflict !== undefined;
                 destEntry.setLocation(destLocations);
-                this._putMetadata(destEntry, mdOnly, log, err => {
+                return this._putMetadata(destEntry, localMdOnly, log, err => {
```
francoisferrand: `local` (in variable name) is a technicality, which makes the variable name weird.
maybe the cleaner way is just keep things simple:

```suggestion
                return this._putMetadata(destEntry, mdOnly || !!conflict, log, err => {
```

extensions/replication/tasks/ReplicateObject.js
francoisferrand: either are fine with me.

- while I agree on the param (especially the `skip(bool skip)` anti-pattern), in this case it seems fine to me : the parameter is not just an indication to skip, but really an input allowing to make the decision ("details of the conflict").
- in my mind, putting it is `putMetadata` makes it a further from the anti-pattern : to me it is not skipping the metadata write, but just raising the abstraction level. PutMetadata's goal is to ensure the metadata is in the system, and it would gain the ability to reach the goal without actually making a write (kind of magic!) - similar in a way to how a getter could either read from the actual DB or quickly return the data from a cache.

(but either way may be better to keep `_shouldSkipMetadata` separate -just decide if it is called from within `_putMetadata` or before it- so each path can be demonstrated and tested easily)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -938,30 +996,63 @@ class ReplicateObject extends BackbeatTask {
...
+            (destLocations, conflict, next) => {
+                if (this._shouldSkipMetadata(sourceEntry.getMicroVersionId(), conflict, log)) {
+                    log.info('skipping putMetadata: destination already has same or newer revision', {
+                        entry: destEntry.getLogInfo(),
+                    });
+                    return next();
+                }
+                const localMdOnly = conflict !== undefined;
+                destEntry.setLocation(destLocations);
+                return this._putMetadata(destEntry, localMdOnly, log, next);
```
francoisferrand: nit: same as above


```suggestion
                destEntry.setLocation(destLocations);
                return this._putMetadata(destEntry, !!conflict, log, next);
```

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -938,30 +996,63 @@ class ReplicateObject extends BackbeatTask {
...
             },
         ], err => this._handleReplicationOutcome(
             err, sourceEntry, destEntry, kafkaEntry, log, done));
     }
 
     _handleReplicationOutcome(err, sourceEntry, destEntry, kafkaEntry,
         log, done) {
+        if (err instanceof MicroVersionIdAlreadyStoredException ||
+            err instanceof StaleMicroVersionIdException) {
+            log.info('replication completed: object already at destination',
```
francoisferrand: nit: "object" may be ambiguous here, it is not about the object (~data) but really about the metadata revision ?
```suggestion
            log.info('replication completed: metadata already at destination',
```
or
```suggestion
            log.info('replication completed: metadata revision already at destination',
```
or skip the ambiguity (but maybe not clear)
```suggestion
            log.info('replication completed: already at destination',
```

tests/unit/replication/ReplicateObject.spec.js
```diff
@@ -72,6 +80,211 @@ describe('ReplicateObject', () => {
...
+            const realErr = Object.assign(new Error('network failure'), { origin: 'target' });
+            task._handleReplicationOutcome(
+                realErr, sourceEntry, destEntry, kafkaEntry, fakeLogger, () => {
+                    sinon.assert.calledWith(task._publishReplicationStatus,
+                        sourceEntry, 'FAILED', sinon.match.any);
+                    done();
+                });
+        });
+    });
+
```
francoisferrand, line 287: should add tests to verify `_shouldSkipMetadata` behavior directly in the different cases
- no conflict
- conflict with "invalid" microVersionId
- conflict with missing microVersionId
- conflict with newer/equal/older microVersionId
- conflict with null microVersionId (either in conflict or in destination)

francoisferrand (approved):

tests/unit/replication/ReplicateObject.spec.js
maeldonn, line 1: Please avoid callbacks on new isolated functions.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -1,21 +1,30 @@
 const async = require('async');
 const { S3Client, GetBucketReplicationCommand, GetObjectCommand } = require('@aws-sdk/client-s3');
 
-const errors = require('arsenal').errors;
-const jsutil = require('arsenal').jsutil;
-const ObjectMDLocation = require('arsenal').models.ObjectMDLocation;
-const ReplicationConfiguration = require('arsenal').models.ReplicationConfiguration;
+const { errors, jsutil, versioning } = require('arsenal');
+const { ObjectMDLocation } = require('arsenal').models;
+const { ReplicationConfiguration } = require('arsenal').models;
```
maeldonn: Collapse two imports

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -657,9 +700,9 @@ class ReplicateObject extends BackbeatTask {
             accountId = _extractAccountIdFromRole(this.targetRole);
         }
 
-        // sends extra header x-scal-replication-content to the target
-        // if it's a metadata operation only
-        const replicationContent = (mdOnly ? 'METADATA' : undefined);
+        // Send x-scal-replication-content so cloudserver know the putMetadata api
```
maeldonn:
```suggestion
        // Send x-scal-replication-content so cloudserver knows the putMetadata api
```

maeldonn (approved):
Some minor issues to fix

benzekrimaha (approved):
Nothing more to add after mael and françois's comments. Approving

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Yes, moved it when setting relpicationContent and made it shorter

tests/unit/replication/ReplicateObject.spec.js
SylvainSenechal: Ok I'm gonna do this in a separate pr where i will remove all callbacks from this file, ticket already opened and some work done on it

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Actually i dont think its supoprted currently : 

Run yarn run --silent lint

/home/runner/work/backbeat/backbeat/extensions/replication/tasks/ReplicateObject.js
  471:39  error  Parsing error: Unexpected token =

PR #2748, opened 2026-06-04
feat(lifecycle): skip index creation on large collections and mass-expiration rules

delthas:
## Summary

Add two pre-flight heuristics to the conductor's index-creation path so we keep lifecycle on **v1 (full scan)** instead of paying the v2 build cost when it wouldn't pay off:

1. **Collection too big** — if `count > 10M` OR `storageSize > 10 GiB`, skip auto-creating v2 lifecycle indexes. The build's transient disk usage on a large existing collection can exceed 5–10× the final index size during the WiredTiger external-merge-sort phase — risky on tight deployments. Both thresholds are env-overridable via `LIFECYCLE_MAX_AUTO_INDEX_DOC_COUNT` and `LIFECYCLE_MAX_AUTO_INDEX_STORAGE_BYTES` (read in `lib/Config.js` `_setLifecycleConductorOptions`, exposed as `config.lifecycleConductorOptions`).

2. **Mass-expiration rule (Days=0)** — skip auto-creating indexes if any enabled lifecycle rule is a whole-bucket `Expiration` with **`Days=0`** and no prefix/tag filter:

   ```
   Expiration.Days === 0  AND  no rulePrefix  AND  no tags
   ```

   `Days=0` is the explicit **"empty this bucket"** signal now supported end-to-end (arsenal **ARSN-597**, backbeat **BB-791**, management API **PSVAPI-128**). A bucket about to be wiped shouldn't pay for a v2 index that would be churned before it pays for itself. It's an unambiguous customer signal — unlike `Days=1`, which is also the syntax legitimate short-retention buckets (logs, etc.) use.

## Notes

- Rule helpers operate on the **MongoDB internal lifecycle format** (lowercase `ruleStatus`, `actions[]`, `filter.rulePrefix`) — distinct from the AWS S3 format that `util/rules.js` / `RulesReducer.js` consume. Precedent: `LifecycleOplogPopulatorUtils.isBucketExtensionEnabled`.
- One mass-expiration rule (no filter) is enough (`any`, not `every`); disabled rules don't contribute.
- The verdict is precomputed once per bucket in `listMongodbBuckets` and plumbed via `task.hasMassExpirationRule`.
- The collection-size check reuses the `collStats` fetch BB-753 already performs — no extra round trip.
- Both guards complement (do not replace) the BB-753 free-disk check and the BB-778 SosAPI check. Buckets that already have the v2 indexes continue to use v2 regardless.
- Two new metric tags on `LifecycleMetrics.onLegacyTask`: `'collectionTooBig'` and `'massExpirationRule'`.

Rebased onto `development/9.5`.

Issue: BB-779

### DIFF
--- a/extensions/lifecycle/conductor/LifecycleConductor.js
+++ b/extensions/lifecycle/conductor/LifecycleConductor.js
@@ -11,6 +11,7 @@ const BucketClient = require('bucketclient').RESTClient;
 const MongoClient = require('arsenal').storage
     .metadata.mongoclient.MongoClientInterface;
 
+const config = require('../../../lib/Config');
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 const ZookeeperManager = require('../../../lib/clients/ZookeeperManager');
@@ -25,6 +26,7 @@ const {
 } = require('../../../lib/constants');
 
 const { LifecycleMetrics } = require('../LifecycleMetrics');
+const { hasMassExpirationRule } = require('../util/mongoRules');
 const { BreakerState, CircuitBreaker } = require('breakbeat').CircuitBreaker;
 const {
     startCircuitBreakerMetricsExport,
@@ -277,6 +279,12 @@ class LifecycleConductor {
                 return cb(null, lifecycleTaskVersions.v1);
             }
 
+            if (task.hasMassExpirationRule) {
+                log.trace('skipping index creation: bucket has a mass-expiration rule');
+                LifecycleMetrics.onLegacyTask(log, 'massExpirationRule');
+                return cb(null, lifecycleTaskVersions.v1);
+            }
+
             if (!this.lcConfig.autoCreateIndexes) {
                 log.trace('skipping index creation: auto creation of indexes disabled');
                 LifecycleMetrics.onLegacyTask(log, 'noIndex');
@@ -328,6 +336,22 @@ class LifecycleConductor {
                         threshold: 3 * idIndexSize,
                     });
                     LifecycleMetrics.onLegacyTask(log, 'insufficientDiskSpace');
+                    return cb(null, lifecycleTaskVersions.v1);
+                }
+
+                const { maxAutoIndexDocCount, maxAutoIndexStorageBytes } =
+                    config.lifecycleConductorOptions;
+                const docCount = results.collStats.count || 0;
+                const storageSize = results.collStats.storageSize || 0;
+                if (docCount > maxAutoIndexDocCount || storageSize > maxAutoIndexStorageBytes) {
+                    log.warn('skipping index creation: collection too large', {
+                        bucket: task.bucketName,
+                        docCount,
+                        storageSize,
+                        maxDocCount: maxAutoIndexDocCount,
+                        maxStorageBytes: maxAutoIndexStorageBytes,
+                    });
+                    LifecycleMetrics.onLegacyTask(log, 'collectionTooBig');
                     return cb(null, lifecycleTaskVersions.v1);
                 }
 
@@ -893,6 +917,8 @@ class LifecycleConductor {
                                         canonicalId: doc.value.owner,
                                         isLifecycled: !!doc.value.lifecycleConfiguration,
                                         hasSosApi: !!doc.value.capabilities?.VeeamSOSApi,
+                                        hasMassExpirationRule: hasMassExpirationRule(
+                                            doc.value.lifecycleConfiguration?.rules),
                                     });
                                     nEnqueued += 1;
                                     lastSentId = doc._id;
--- a/extensions/lifecycle/util/mongoRules.js
+++ b/extensions/lifecycle/util/mongoRules.js
@@ -1,0 +1,28 @@
+'use strict';
+
+// Operates on the MongoDB internal lifecycle format (lowercase ruleStatus,
+// actions[], filter.rulePrefix) — NOT the AWS S3 format used by util/rules.js.
+
+// A whole-bucket Expiration with days=0 is the explicit "empty this bucket"
+// signal: treat it as mass-expiration so the conductor keeps lifecycle on v1.
+function isMassExpirationRule(rule) {
+    if (!rule || rule.ruleStatus !== 'Enabled') {
+        return false;
+    }
+    if (rule.filter && (rule.filter.rulePrefix
+            || (rule.filter.tags && rule.filter.tags.length > 0))) {
+        return false;
+    }
+    return (rule.actions || []).some(action => action
+        && action.actionName === 'Expiration'
+        && action.days === 0);
+}
+
+function hasMassExpirationRule(rules) {
+    return (rules || []).some(r => isMassExpirationRule(r));
+}
+
+module.exports = {
+    isMassExpirationRule,
+    hasMassExpirationRule,
+};
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -180,6 +180,19 @@ class Config extends EventEmitter {
         this.transientLocations = {};
 
         this._setTimeOptions();
+        this._setLifecycleConductorOptions();
+    }
+
+    _setLifecycleConductorOptions() {
+        const maxAutoIndexDocCount =
+            Number.parseInt(process.env.LIFECYCLE_MAX_AUTO_INDEX_DOC_COUNT, 10) || 10000000;
+        const maxAutoIndexStorageBytes =
+            Number.parseInt(process.env.LIFECYCLE_MAX_AUTO_INDEX_STORAGE_BYTES, 10)
+            || 10 * 1024 * 1024 * 1024;
+        this.lifecycleConductorOptions = {
+            maxAutoIndexDocCount,
+            maxAutoIndexStorageBytes,
+        };
     }
 
     _setTimeOptions() {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "9.5.0-preview.2",
+  "version": "9.5.0-preview.3",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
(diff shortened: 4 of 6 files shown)


extensions/lifecycle/util/mongoRules.js
```diff
@@ -0,0 +1,40 @@
...
+        if (action.days !== undefined && action.days <= 1) {
+            return true;
+        }
+        return false;
+    });
+}
+
+function allRulesAreMassExpiration(rules, currentDate = new Date()) {
+    return (rules || [])
+        .filter(r => r && r.ruleStatus === 'Enabled')
```
DarkIsDude: this is already checked inside the `isMassExpirationRule` ?

DarkIsDude (approved):

extensions/lifecycle/util/mongoRules.js
delthas: Good catch — addressed by dropping the filter entirely. `isMassExpirationRule` is now `ruleAllowsIndexSkip`, which returns `true` for disabled (and absent) rules, so `allRulesAreMassExpiration` is just `(rules || []).every(ruleAllowsIndexSkip)` — `.every()` ignores disabled rules without a separate `ruleStatus === 'Enabled'` pre-filter. No more duplicated check.

extensions/lifecycle/util/mongoRules.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+    // Allow a day of slack: a rule expiring within ~1 day still clears the
+    // whole collection, same as Days <= 1.
+    const dateCutoff = currentDate.getTime() + 24 * 60 * 60 * 1000;
+    return (rule.actions || []).some(action => {
+        if (!action || action.actionName !== 'Expiration') {
+            return false;
+        }
+        if (action.date && new Date(action.date).getTime() <= dateCutoff) {
+            return true;
```
francoisferrand: that is not correct:
- setting a date "10 years ago" will not remove a lot
- esp. once the "mass deletion" has been done, the rule deprecates...but will still prevent creating an index

→ this condition would be something like `new Date(action.date).getTime() >= currentDate.getTime() - 24 * 60 * 60 * 1000`

extensions/lifecycle/util/mongoRules.js
```diff
@@ -0,0 +1,38 @@
...
+        }
+        if (action.days !== undefined && action.days <= 1) {
```
francoisferrand: `some` is not appropriate for this case: e.g. we often want to expire non-current versions "quickly", but not current versions... If we want to define an heuristic on _days_, should be when _all_ actions match.

extensions/lifecycle/util/mongoRules.js
```diff
@@ -0,0 +1,38 @@
...
+    if (!rule || rule.ruleStatus !== 'Enabled') {
+        return true;
+    }
+    if (rule.filter && (rule.filter.rulePrefix
+            || (rule.filter.tags && rule.filter.tags.length > 0))) {
+        return false;
+    }
+    // Allow a day of slack: a rule expiring within ~1 day still clears the
+    // whole collection, same as Days <= 1.
+    const dateCutoff = currentDate.getTime() + 24 * 60 * 60 * 1000;
```
francoisferrand: this 24h should be another constant

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -35,6 +36,8 @@ const DEFAULT_CONCURRENCY = 10;
 const BUCKET_CHECKPOINT_PUSH_NUMBER = 50;
 const BUCKET_CHECKPOINT_PUSH_NUMBER_BUCKETD = 50;
 const ACCOUNT_SPLITTER = ':';
+const MAX_AUTO_INDEX_DOC_COUNT = 1000000;
+const MAX_AUTO_INDEX_STORAGE_BYTES = 1024 * 1024 * 1024;
```
francoisferrand: I wonder if we should make these "customizable" (i.e. try read from env) ?
i.e. could we have case where we want/need to override the heuristic threshold, at some customer?

(it seems we would only skip v1, max-expire objects, then create index : so not really needed, but I wonder if there could be corner cases...)

extensions/lifecycle/util/mongoRules.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+// Operates on the MongoDB internal lifecycle format (lowercase ruleStatus,
+// actions[], filter.rulePrefix) — NOT the AWS S3 format used by util/rules.js.
+
+function ruleAllowsIndexSkip(rule, currentDate = new Date()) {
```
francoisferrand: function name is not clear...
not sure I like `allRulesAreMassExpiration`, but at least it is clear → ruleIsMassExpiration ?

extensions/lifecycle/util/mongoRules.js
```diff
@@ -0,0 +1,38 @@
...
+        }
+        if (action.days !== undefined && action.days <= 1) {
+            return true;
+        }
+        return false;
+    });
+}
+
+function allRulesAreMassExpiration(rules, currentDate = new Date()) {
+    return (rules || []).every(r => ruleAllowsIndexSkip(r, currentDate));
```
francoisferrand: should it not be "any" ?
as long as we have 1 mass-expiration, everything will expire and index is not used... (for now, i.e. until the max expiration has been done)

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -35,6 +36,8 @@ const DEFAULT_CONCURRENCY = 10;
 const BUCKET_CHECKPOINT_PUSH_NUMBER = 50;
 const BUCKET_CHECKPOINT_PUSH_NUMBER_BUCKETD = 50;
 const ACCOUNT_SPLITTER = ':';
+const MAX_AUTO_INDEX_DOC_COUNT = 1000000;
```
francoisferrand: is 1'000'000 that many documents?
That would be about 1GB of total metadata usage, is it really a high-enough threshold?

(same for index: is 1GB of index really the turning point)

extensions/lifecycle/util/mongoRules.js
francoisferrand: is `days` a good criteria for mass-expiration?
- Some use-case may rely on (lots of) short-lived data, and thus use these even though they could benefit form index (lots of data created in the day)
- What we really need is one "pattern" we can document to expire the whole bucket: this is the one that really needs to be detected. Others are probably corner cases, and handled by index failure handling/...

francoisferrand (commented):
not sure if this should land in 9.4 or 9.5, let's confirm this with product team

extensions/lifecycle/util/mongoRules.js
delthas: Addressed via heuristic redesign in 09df4aa6. New rule fires only on `Expiration.Date > now()` with no filter — dropped both the past-date and the days legs. Old past dates no longer trigger the skip. Discussed offline.

extensions/lifecycle/util/mongoRules.js
delthas: Moot after 09df4aa6 — the `days`-based path is gone, so the some/every choice within a rule's actions no longer affects current-version Expiration.

extensions/lifecycle/util/mongoRules.js
delthas: Moot after 09df4aa6 — the 24h slack window is gone (heuristic is now strict `date > now()`, no slack).

extensions/lifecycle/conductor/LifecycleConductor.js
delthas: Done in b63010c8 — both thresholds now read from env overrides (`LIFECYCLE_MAX_AUTO_INDEX_DOC_COUNT`, `LIFECYCLE_MAX_AUTO_INDEX_STORAGE_BYTES`) with the constants as defaults.

extensions/lifecycle/util/mongoRules.js
delthas: Renamed in 09df4aa6: `ruleAllowsIndexSkip` → `isMassExpirationRule` and `allRulesAreMassExpiration` → `hasMassExpirationRule` (the wrapper now uses `any` semantics so `has...` fits better than `all...`).

extensions/lifecycle/util/mongoRules.js
delthas: Switched to `any` in 09df4aa6 (via the `hasMassExpirationRule` wrapper).

extensions/lifecycle/conductor/LifecycleConductor.js
delthas: Bumped in b63010c8 to 10M docs / 10 GiB. Also exposed both as env overrides (`LIFECYCLE_MAX_AUTO_INDEX_DOC_COUNT`, `LIFECYCLE_MAX_AUTO_INDEX_STORAGE_BYTES`) for further per-deployment tuning. The new values keep the worst-case index build comfortably inside a typical PVC while still falling back to v1 on truly large collections.

extensions/lifecycle/util/mongoRules.js
delthas: Dropped `days` entirely in 09df4aa6. The remaining detection requires the explicit "expire whole bucket" pattern: a future-dated `Expiration` action with no prefix and no tags. Logs-style buckets with `Days: 1` keep their indexes.

delthas:
Moved to 9.5, days=0

extensions/lifecycle/conductor/LifecycleConductor.js
```diff
@@ -331,6 +339,23 @@ class LifecycleConductor {
                     return cb(null, lifecycleTaskVersions.v1);
                 }
 
+                const { maxAutoIndexDocCount, maxAutoIndexStorageBytes } =
+                    config.lifecycleConductorOptions;
+                const docCount = results.collStats.count || 0;
+                const storageSize = results.collStats.storageSize || 0;
+                if (docCount > maxAutoIndexDocCount
+                        || storageSize > maxAutoIndexStorageBytes) {
```
francoisferrand: nit: fits on a single line

francoisferrand (approved):

extensions/lifecycle/conductor/LifecycleConductor.js
delthas: Done — joined onto one line.

PR #2749, opened 2026-06-05
BB-780: Configure CRR queue processor poll interval

anurag4DSB:
**Note to reviewers:** `lib tests` is red, but it's a **pre-existing flake on `development/9.0`, unrelated to this change.** The flaky test is `BackbeatConsumer main tests › should be able to read messages sent to the topic and publish topic metrics` (intermittent 30s timeout). Proof: full CI was run on an empty commits off *unmodified* `development/9.0` - https://github.com/scality/backbeat/actions/runs/27203064676

Lowering `mpu_parts_concurrency` (to stop large MPUs overloading Metadata) makes a single large-MPU replication run longer than Kafka's 5-min `max.poll.interval.ms`, so the queue processor's consumer gets evicted mid-transfer — the partition is redelivered and you get duplicate work, orphan sproxyd parts, and ghost processes. This exposes that timeout so an operator can raise it and let long transfers finish. It's the short-term mitigation from the "CRR on large MPUs" root-cause plan; the real fix (keep polling while a slow task runs) is separate, larger work.

Only the replication **queue processor's** Kafka consumer. The value rides through to `BackbeatConsumer`, which already accepted `maxPollIntervalMs` — so there are no `lib/` changes. Downstream: federation has to render the key for S3C to use it (S3C-11270); Zenko sets it through config directly.

Unset behaves exactly as today — `BackbeatConsumer`'s existing 300000 default applies, so deployments that set nothing are unchanged. Every other consumer (status processor, lifecycle, GC, notification, metrics, queue populator) is deliberately untouched: this is the shortest, lowest-risk fix for the incident, and the queue processor is the only consumer with the documented slow-task problem. Making the timeout global, or tuning the other consumers, is intentionally left for a separate consistent methodology rather than bolted on here.

`extensions.replication.queueProcessor.maxPollIntervalMs` is now an optional config key, bounded both ways: minimum 45000 (librdkafka's `session.timeout.ms` floor, below which the client refuses to start) and maximum 1800000 (30 min) so the timeout can't be set unbounded — past 30 min a transfer is wedged and we want kafka to evict and redeliver rather than mask it. Wired into the consumer, plus a `KAFKA_..._MAX_POLL_INTERVAL_MS` entrypoint mapping for container deployments.

New unit tests assert the value passes through from config, stays optional, and is rejected outside [45000, 1800000]. Built TDD-style: the first commit adds the failing tests, the second implements the schema + wiring (check out the test commit alone and the spec goes red, proving the tests bite); the third adds the 30-min ceiling with its own rejection test.


anurag4DSB:
> * docker-entrypoint.sh:245 — unquoted env var breaks the file's quoting convention; a non-numeric value would silently produce an empty config (see inline suggestion)Config wiring, Joi validation, BackbeatConsumer passthrough, and tests all look correct. The bootstrap consume-guard ([BB-782](https://scality.atlassian.net/browse/BB-782)) and warning log are solid improvements.Review by Claude Code

This is handled

[BB-782]: https://scality.atlassian.net/browse/BB-782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

BourgoisMickael (approved):

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -69,6 +72,7 @@ const joiSchema = joi.object({
         groupId: joi.string().required(),
         retry: qpRetryJoi,
         concurrency: joi.number().greater(0).default(10),
+        maxPollIntervalMs: joi.number().min(MAX_POLL_INTERVAL_MS_MIN),
```
BourgoisMickael: Should you also validate a maximum ?

What happens if we put 1 hour, or 1 day ? I suppose on a crash of the consumer, kafka will not notice it until the full interval is elasped, so we might wait a long time before another consumer can take its place

nicolas2bert (approved):
let's add a max limit to 30 minutes which let 1 hour for a task to finish (drain window) which should be more the enough for all our cases.

extensions/replication/ReplicationConfigValidator.js
anurag4DSB: We had originally decided not to have an upper bound on BB itself, but only on federation if we change our minds. I added it to BB, per Nico's comment below.

anurag4DSB:
ping

BourgoisMickael (approved):

anurag4DSB:
ping

nicolas2bert:
@bert-e bypass_build_status

PR #2752, opened 2026-06-05
BB-782: Fix bootstrap consumer stealing messages

anurag4DSB:

The "read messages … and publish topic metrics" functional test fails CI with a bare 30s
timeout on roughly half the runs on loaded runners — it blocked PR #2749 twice and fails on
pristine `development/9.0`. Root cause, confirmed by tracing: the test-only bootstrap mechanism
issues a kafka consume request every 200ms while each request waits up to 1s inside the client,
so a backlog of stale requests outlives bootstrap and steals the first real messages a test
produces — bootstrap's callback silently drops anything that isn't a bootstrap sentinel.

Only `_bootstrapConsumer` — a test-only path (`bootstrap: true` is set exclusively from
`tests/`; the option is documented "TEST ONLY"). Every bootstrap-using suite (lib, replication
queueProcessor) inherits the fix. No production data path changes.

The bootstrap protocol is unchanged: sentinel every 5s, commit past the sentinel, unsubscribe,
hand off to the real loop. The stacked requests were not useful parallelism — bootstrap fetches
one sentinel per 5s, and the regular consume loop already enforces the same one-request-in-flight
rule via its `_nConsumePendingRequests` gate. Production consumers never enter this code.

At most one in-flight bootstrap consume request, so the final sentinel necessarily arrives on
the only outstanding request and none survive bootstrap to steal messages — plus a warn log if
bootstrap ever drops a non-bootstrap message again. In the test itself: the ZK offsets check
polls with bounded attempts instead of a fixed 5s sleep + one-shot read, and the prometheus
check is chained into `done` (it was a floating promise whose assertion failures were lost as
unhandled rejections).

Reproduced locally at ~10% by CPU-throttling the CI kafka image; tracing captured the theft
directly (3 messages produced, the pipeline received 2, the commit skipped over the eaten one).
With the fix: 40 consecutive local passes under the same throttle and a green CI matrix. For
the before/after on CI itself: an empty commit on this branch failed `lib tests` with zero code
changes ([run 27012249826](https://github.com/scality/backbeat/actions/runs/27012249826)), the
fix went green ([run 27014743787](https://github.com/scality/backbeat/actions/runs/27014743787)).
If bootstrap ever eats a message again, the new warn makes it visible in the logs instead of
silent.

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -837,7 +837,24 @@ class BackbeatConsumer extends EventEmitter {
      * @return {boolean} true if paused
      */
     isPaused() {
-        return this._consumer.subscription().length === 0;
+        try {
+            return this._consumer.subscription().length === 0;
+        } catch (err) {
+            // subscription() throws "Erroneous state" if the consumer is
+            // disconnected/closing; such a consumer is not actively consuming,
+            // so report it as paused. Callers (the consume loop, offsetsStore)
+            // already rely on isPaused() to skip operations that would
+            // otherwise throw on an invalid consumer state. Logged at debug
+            // because this is an expected transient during teardown and would
+            // otherwise flood logs on every shutdown.
+            this._log.debug('could not read consumer subscription, treating ' +
+                'consumer as paused', {
+                error: err.message,
+                topic: this._topic,
+                groupId: this._groupId,
+            });
+            return true;
+        }
     }
 
     /**
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -130,7 +130,6 @@ class BackbeatConsumer extends EventEmitter {
         this._nConsumePendingRequests = 0;
         this._consumer = null;
         this._consumerReady = false;
-        this._readyEmitted = false;
         this._kafkaBacklogMetrics = null;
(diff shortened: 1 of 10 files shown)


lib/BackbeatConsumer.js
```diff
@@ -826,7 +826,16 @@ class BackbeatConsumer extends EventEmitter {
...
         let producerTimer;
         let consumerTimer;
+        // Run at most one consume request at a time, like the regular
+        // consume loop does with its _nConsumePendingRequests gate. A
+        // request can wait up to a second inside the kafka client before
+        // returning, so firing one every 200ms piles up requests that
+        // outlive bootstrap. A leftover request would then grab the first
+        // real messages produced after bootstrap, and this callback would
+        // drop them (BB-782).
+        let consumePending = false;
```
nicolas2bert: Not a big fan of the bootstrap "test only" logic implemented along production code.
Could it be an elegant way to get rid of it instead of complexifying it.

nicolas2bert: In the tests we could add a way to (after subscribing) wait until the consumer is ready to consume meaning broker has assigned it partitions. That way we are sure that all messages produced are seen/consumed.

anurag4DSB:
We are closing this for now. We'll come back to it later if needed.

PR #2751, opened 2026-06-05
feat(lifecycle): emit putBucketIndexesFailed metric on index creation errors

delthas:
Replaces [#2750](https://github.com/scality/backbeat/pull/2750) (auto-closed when the branch was renamed from `improvement/ZENKO-5286/...` to `improvement/BB-783/...`).

## Summary

Adds a dedicated metric tag (`putBucketIndexesFailed`) when the conductor's `putBucketIndexes` call returns an error. Provides the Prometheus signal that [BB-784](https://scality.atlassian.net/browse/BB-784)'s Backbeat-side `LifecycleIndexCreationFailed` alert and [ZENKO-5285](https://scality.atlassian.net/browse/ZENKO-5285)'s mongo-side alert will hook into.

## Why

Today the error callback at `LifecycleConductor.js:338-345` just logs a warning and falls back to v1 — no Prometheus signal for the failure. The existing `putBucketIndexes` tag was incremented before the call and therefore counted attempts indistinguishably from successes.

## Implementation

Routes the `LifecycleMetrics.onLegacyTask` call into the callback so the status is set from the outcome:
- success → `putBucketIndexes`
- failure → `putBucketIndexesFailed`

This makes the two statuses mutually exclusive (one increment per task), so `putBucketIndexes + putBucketIndexesFailed = attempts` and `putBucketIndexesFailed / (putBucketIndexes + putBucketIndexesFailed) = failure rate`.

The new tag is cause-agnostic — operators can dig into the warn log to find the underlying reason (MongoDB 7.1's `IndexBuildKilledByOutOfDiskSpace`, duplicate-name conflict, transient infra, etc.). A future ticket can map specific Arsenal errors and emit more granular tags if needed.

Two unit tests cover the routing: success path emits only `putBucketIndexes`, failure path emits only `putBucketIndexesFailed`.

## Related

- [BB-783](https://scality.atlassian.net/browse/BB-783) — this ticket
- [BB-784](https://scality.atlassian.net/browse/BB-784) — Backbeat-side Prometheus alert wired to this metric ([PR #2753](https://github.com/scality/backbeat/pull/2753))
- [ZENKO-5286](https://scality.atlassian.net/browse/ZENKO-5286) — parent ticket that originally tracked the work (\"Mongodb index creation error\")
- [ZENKO-5285](https://scality.atlassian.net/browse/ZENKO-5285) — mongo-side alert rule ([Zenko PR #2431](https://github.com/scality/Zenko/pull/2431))
[...]

PR #2753, opened 2026-06-05
feat(lifecycle): alert when index creation fails

delthas:
## Summary
- Add `LifecycleIndexCreationFailed` Prometheus alert (`monitoring/lifecycle/alerts.yaml`) wired to the `s3_lifecycle_legacy_tasks_total{status="putBucketIndexesFailed"}` counter introduced by BB-783 (#2751).
- Fires on any failure in a 5m window with severity `warning` — mirrors the `CircuitBreakerError` shape: the conductor falls back to v1 full-scan, so this is recoverable but always warrants investigation (typically a disk-fill / `indexBuildMinAvailableDiskSpaceMB` refusal).
- Companion fixture added to `alerts.test.yaml`.

Part of the [Empty bucket feature epic](https://scality.atlassian.net/browse/ARTESCA-16740). A symmetric MongoDB-side createIndexes-failed alert lives in the Zenko chart under [ZENKO-5285](https://scality.atlassian.net/browse/ZENKO-5285).

Issue: BB-784

[ZENKO-5285]: https://scality.atlassian.net/browse/ZENKO-5285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/monitoring/lifecycle/alerts.test.yaml
+++ b/monitoring/lifecycle/alerts.test.yaml
@@ -466,6 +466,49 @@ tests:
         eval_time: 14m
         exp_alerts: []
 
+  - name: LifecycleIndexCreationFailed
+    interval: 1h
+    input_series:
+      - series: s3_lifecycle_legacy_tasks_total{namespace="zenko",job="artesca-data-backbeat-lifecycle-producer-headless",origin="conductor",status="putBucketIndexesFailed"}
+        values: 0x5 1x12
+      - series: s3_lifecycle_legacy_tasks_total{namespace="zenko",job="artesca-data-backbeat-lifecycle-producer-headless",origin="conductor",status="noIndex"}
+        values: 0+1x17
+    alert_rule_test:
+      - alertname: LifecycleIndexCreationFailed
+        eval_time: 4h
+        exp_alerts: []
+      - alertname: LifecycleIndexCreationFailed
+        eval_time: 6h
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: zenko
+              job: artesca-data-backbeat-lifecycle-producer-headless
+              origin: conductor
+              status: putBucketIndexesFailed
+            exp_annotations:
+              zenko_service: backbeat-lifecycle-producer
+              description: |-
+                The lifecycle producer failed to create the v2 lifecycle indexes on at least one bucket in the last 12 hours. The producer will fall back to v1 (full-scan) processing for that bucket. Check producer logs for the bucket name and the underlying MongoDB error.
+              summary: "Lifecycle index creation failure"
+      - alertname: LifecycleIndexCreationFailed
+        eval_time: 15h
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: zenko
+              job: artesca-data-backbeat-lifecycle-producer-headless
+              origin: conductor
+              status: putBucketIndexesFailed
+            exp_annotations:
+              zenko_service: backbeat-lifecycle-producer
+              description: |-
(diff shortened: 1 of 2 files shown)


monitoring/lifecycle/alerts.test.yaml
```diff
@@ -466,6 +466,44 @@ tests:
...
+        exp_alerts: []
+      - alertname: LifecycleIndexCreationFailed
+        eval_time: 4m
+        exp_alerts: []
+      - alertname: LifecycleIndexCreationFailed
+        eval_time: 5m
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+            exp_annotations:
```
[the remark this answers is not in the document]
delthas: Good catch — fixed in 29521bd6. Added the four metric-carried labels (`namespace`, `job`, `origin`, `status`) to both `exp_alerts` blocks.

monitoring/lifecycle/alerts.test.yaml
```diff
@@ -466,6 +466,44 @@ tests:
...
+              zenko_service: backbeat-lifecycle-producer
+              description: |-
+                The lifecycle producer failed to create the v2 lifecycle indexes on at least one bucket in the last 5 minutes. The producer will fall back to v1 (full-scan) processing for that bucket. Check producer logs for the bucket name and the underlying MongoDB error.
+              summary: "Failed to create lifecycle indexes"
+      - alertname: LifecycleIndexCreationFailed
+        eval_time: 9m
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+            exp_annotations:
```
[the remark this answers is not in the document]
delthas: Fixed in 29521bd6 along with the 5m block.

SylvainSenechal (approved):

monitoring/lifecycle/alerts.yaml
```diff
@@ -82,6 +82,20 @@ groups:
         {{ ${lifecycle_latency_critical_threshold} | humanizeDuration }} ago.
       summary: "Lifecycle scan not executed in time"
 
+  - alert: LifecycleIndexCreationFailed
+    Expr: |
+      increase(s3_lifecycle_legacy_tasks_total{namespace="${namespace}", job="${job_lifecycle_producer}", status="putBucketIndexesFailed"}[5m]) > 0
```
francoisferrand: nit: should it be over more than 5min?
- If alert notifications are not configured, the alert would appear for 5 minutes (around midnight and noon), so may be easy to miss...
- Even if notification is configured, it may be difficult to troubleshoot: as it would be there anymore when you go check it out (this could be addressed through documentation, though)

→ may be better to use a longer period (12h or even 24h), matching lifecycle cron job? Or will the computation degrade in that case (or become costly)?
→ lifecycle cronjob is configurable, but not sure it's worth tweaking the alert based on that: just a period matching the timing of lifecycle semantics should be enough I guess

francoisferrand (commented):
- not urgent I think, should target 9.5
- esp. there is a risk this would create support calls, should not ship in product through a patch release

monitoring/lifecycle/alerts.yaml
delthas: Good call. Bumped to `[12h]` in e0dfc613 — matches the typical lifecycle cron cadence so a failure stays visible for the duration of the scan period. Description and test fixture updated accordingly (test now runs at interval 1h across an 18h window to exercise both the firing range and the post-12h clearance).

delthas: Checked, performance will not be an issue for 12h for this use case.

francoisferrand (approved):

PR #2760, opened 2026-06-10
BB-787: Restart CRR consumer stuck past drain timeout

anurag4DSB:

When a CRR queue processor task wedges past the rebalance drain timeout, the consumer
deliberately disconnects — the code's own comment says "so that the healthcheck will fail
and the process will get restarted". On Zenko, kubernetes does exactly that. On S3C,
supervisord only restarts processes that exit, so the worker stays down forever while its
stuck in-flight tasks can still complete late and race the partition's new owner (ghost
writes, orphaned parts). This makes the process exit on that state so supervisord's
existing autorestart brings it back into the consumer group, and kills the zombie work.

Replication queue processors only, by a two-layer gate: the lib emits a new
`rebalance.timeout` event that has no listener anywhere except QueueProcessor, and the
handler only acts when `REPLICATION_QUEUE_PROCESSOR_CRASH_ON_REBALANCE_TIMEOUT` is exactly
`"true"` — set per supervisord program by federation (scality/Federation#6929). Zenko never
sets the variable, so its probe-based restart path is untouched. Lifecycle, GC,
notification, populator, and the status processor see no behavior change.

With the variable unset, everything behaves as today — the only difference is richer
payloads on log lines that already exist (drain deadline/duration, and the identities of
the stuck tasks on the existing "consumer stuck, disconnecting" error). An exit budget
(3 per 6h per site, persisted across restarts in a per-site/per-topic tmp file, fail-open
on file errors) caps the new behavior: past the budget the processor stays disconnected and
today's exact escalation takes over — /_/live latches 500 and the platform healthcheck
alarms.

On the stuck path with the gate armed: a fatal log naming the stuck objects and the budget
state, then a hard `process.exit(1)` after a 1s log-flush grace. Hard exit rather than the
`CRASH_ON_BATCH_TIMEOUT`-style SIGTERM on purpose: the graceful-stop path waits on the very
stuck tasks being escaped and can hang forever. The log story is completed by a gate-off
error ("self-restart disabled… external restart required"), a budget-exhausted error, and a
boot-correlation warning that ties each respawn back to the exit that caused it — all
visible at the default `info` log level.

Unit: a 7-case spec for the handler (armed/unset/`"false"` gating, budget exhaustion,
stale-window pruning, corrupt-file fail-open, boot correlation) plus getInFlightTasks shape
tests. Functional: the existing stuck-rebalance test now also asserts the event fires with
the stuck-task identities; the full BackbeatConsumer functional file passes 15/15 against
the CI kafka image locally. If the mechanism ever misfires in the field, every decision
point logs which branch it took and why.

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -487,9 +487,15 @@ class QueueProcessor extends EventEmitter {
                 });
             return;
         }
+        const stuckAges = (queueState.stuckTasks || [])
+            .map(task => task.ageSeconds)
+            .filter(age => typeof age === 'number');
         this.logger.fatal(
             'consumer stuck after rebalance drain timeout, exiting for ' +
             'restart', {
+                stuckTaskCount: queueState.stuckTasks?.length,
+                oldestAgeSeconds:
+                    stuckAges.length ? Math.max(...stuckAges) : undefined,
                 queueLen: queueState.queueLen,
                 running: queueState.running,
                 exitsInWindow: budget.exitsInWindow,
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -1,6 +1,9 @@
 'use strict';
 
 const async = require('async');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
 const { EventEmitter } = require('events');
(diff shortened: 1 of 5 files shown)


anurag4DSB:
Superseded — splitting per review feedback into smaller PRs: the minimal exit mechanism is #2761; the observability/log story will follow as its own PR. The branch is kept for cherry-picking the log work.

PR #2761, opened 2026-06-11
BB-787: Exit consumer stuck past rebalance drain timeout

anurag4DSB:
Note: 1 test failing on this branch is a known and not related to this PR

On S3C nothing restarts a backbeat process whose consumer wedged past the rebalance
drain timeout: the lib disconnects it so the healthcheck fails, but supervisord only
restarts on exit — so the worker stays down, and its stuck tasks can still complete
late and write over the partition's new owner (ghost objects). This adds an env-gated
hard exit right after that disconnect, in the shared consumer so every backbeat
service is covered: when `CRASH_ON_REBALANCE_TIMEOUT === 'true'`, log a fatal and
`process.exit(1)` after a 1s log-flush grace. Same pattern as the existing
`CRASH_ON_BATCH_TIMEOUT` in LogReader.

The 1s grace leaves a window where a wedged task could wake and fire one last write.
That is benign: the next partition owner has not even finished the rebalance yet, so
its redo of the same uncommitted entry always lands later and supersedes the write.
The damaging ordering (a stale write landing after the redo) needs the stuck worker
alive minutes later, which is exactly what the exit removes.

Inert unless the env var is set: never set on Zenko/K8s (the liveness probe already
handles this state), set container-wide on S3C by scality/Federation#6929. Hard exit
rather than SIGTERM because the graceful stop path waits on the very tasks that are
stuck; strict `=== 'true'` so the var can be explicitly set to `false` to disarm.

Both endings of the drain window are pinned by functional tests against a real kafka
broker: gate unset → disconnect, process stays alive; gate armed → exit(1) fires after
the fatal. The tests stub `process.exit` before arming the gate, so the test process
can never actually die.

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -731,6 +731,21 @@ class BackbeatConsumer extends EventEmitter {
 
                 this._log.error('rdkafka.rebalance timeout: consumer stuck, disconnecting');
                 this._consumer.disconnect();
+                if (process.env.CRASH_ON_REBALANCE_TIMEOUT === 'true') {
+                    // On S3C, supervisord only restarts a program when
+                    // its process exits, and nothing reacts to the
+                    // healthcheck that fails after the disconnect above, so
+                    // we exit. On K8s the liveness probe handles that state,
+                    // and this env var is never set there. Hard exit rather
+                    // than a graceful stop: the stop path would wait on the
+                    // very tasks that are stuck.
+                    this._log.fatal('consumer disconnected from queue ' +
+                        'after rebalance drain timeout, exiting process',
+                        { topic: this._topic, groupId: this._groupId });
+                    // grace period so the fatal line flushes to stdout
+                    // before the exit
+                    setTimeout(() => process.exit(1), 1000);
+                }
             }, this._maxPollIntervalMs - 1000); // 1 second earlier, to be within the limit
         } else {
             this._log.error('rdkafka.rebalance', { err, assignment });
--- a/tests/functional/lib/BackbeatConsumer.js
+++ b/tests/functional/lib/BackbeatConsumer.js
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const async = require('async');
+const sinon = require('sinon');
 const werelogs = require('werelogs');
 
 const { metrics } = require('arsenal');
@@ -299,6 +300,8 @@ describe('BackbeatConsumer rebalance tests', () => {
 
     afterEach(function after(done) {
         this.timeout(10000);
+        delete process.env.CRASH_ON_REBALANCE_TIMEOUT;
+        sinon.restore();
         if (timer) {
             clearInterval(timer);
             timer = null;
@@ -356,6 +359,11 @@ describe('BackbeatConsumer rebalance tests', () => {
         assert(consumer.isReady());
         assert(consumer2.isReady());
 
+        // with the gate unset, the drain timeout must disconnect the
+        // consumer but leave the process alive; stub the exit so a
+        // gate regression cannot kill the test process
+        const exitStub = sinon.stub(process, 'exit');
+
         consumer.once('consumed.message', () => {
             // trigger rebalance during processing of first message
             consumer2.subscribe();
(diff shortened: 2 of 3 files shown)


anurag4DSB:
ping

lib/BackbeatConsumer.js
```diff
@@ -731,6 +731,12 @@ class BackbeatConsumer extends EventEmitter {
 
                 this._log.error('rdkafka.rebalance timeout: consumer stuck, disconnecting');
                 this._consumer.disconnect();
+                // Let owners decide what to do when stuck (e.g. exit on
+                // S3C, where nothing restarts a process on a failed
+                // healthcheck). disconnect() above is guarded and
+                // non-throwing, so this emit always runs.
+                // No listener => no behavior change.
+                this.emit('rebalance.timeout');
```
francoisferrand: why not make crash _here_, to fix the issue for all processes ?

- As far as I understand, *all* backbeat consumers would have the same issue, so we could just handle it once and be done with...
- Not much can be done in case of rebalance timeout to manage, so the event cannot really be useful for much else I think

```suggestion
		    if process.env.CRASH_ON_REBALANCE_TIMEOUT {
                this.logger.fatal('consumer stuck after rebalance drain timeout' [...]);
		        setTimeout(() => process.exit(1), 1000);
		    }
```

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -427,6 +428,34 @@ class QueueProcessor extends EventEmitter {
...
+     * tasks can still complete late and write over the partition's new
+     * owner (ghost objects, orphaned parts). Hard exit, not SIGTERM:
+     * the graceful stop path waits on the very tasks that are stuck.
+     * The env var is never set on K8s, where the liveness probe
+     * already handles this state.
+     *
+     * @return {undefined}
+     */
+    _onRebalanceTimeout() {
+        const armed = process.env
```
francoisferrand: nit: `armed` variable name feels kind of weird here

lib/BackbeatConsumer.js
```diff
@@ -731,6 +731,12 @@ class BackbeatConsumer extends EventEmitter {
 
                 this._log.error('rdkafka.rebalance timeout: consumer stuck, disconnecting');
                 this._consumer.disconnect();
+                // Let owners decide what to do when stuck (e.g. exit on
+                // S3C, where nothing restarts a process on a failed
+                // healthcheck). disconnect() above is guarded and
+                // non-throwing, so this emit always runs.
+                // No listener => no behavior change.
```
francoisferrand: > (e.g. exit on S3C where nothing restarts a process on a failed healthcheck)

useless : the owner decides this, so out of scope for backbeat consumer

> No listener => no behavior change.

useless (it's obvious!), and technically false because you actually always register a listener (in queue processor)

lib/BackbeatConsumer.js
anurag4DSB: Per discussion with the team, we want to scope the changes only for CRR and later on we will have a global strategy for all consumers.

lib/BackbeatConsumer.js
anurag4DSB: changed in this commit: https://github.com/scality/backbeat/pull/2761/changes/7f723393c37dec9e761a7f588c16617f6cac986b
Thanks

extensions/replication/queueProcessor/QueueProcessor.js
anurag4DSB: changed in this commit: https://github.com/scality/backbeat/pull/2761/changes/7f723393c37dec9e761a7f588c16617f6cac986b
Thanks

lib/BackbeatConsumer.js
anurag4DSB: we are re-thinking this.

anurag4DSB:
Do not review, making changes.

lib/BackbeatConsumer.js
anurag4DSB: Adopted in 8faf5cb4 - the crash now lives here in the lib, gated on `CRASH_ON_REBALANCE_TIMEOUT === 'true'` (same pattern and rationale as the existing `CRASH_ON_BATCH_TIMEOUT` in LogReader). The `rebalance.timeout` event is gone and QueueProcessor is untouched. Federation now sets the variable container-wide in /conf/env (scality/Federation#6929).

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -427,6 +428,34 @@ class QueueProcessor extends EventEmitter {
...
+        const isSelfRestartEnabled = process.env
+            .REPLICATION_QUEUE_PROCESSOR_CRASH_ON_REBALANCE_TIMEOUT === 'true';
+        if (!isSelfRestartEnabled) {
+            return;
+        }
+        this.logger.fatal('CRR consumer stuck after rebalance drain ' +
+            'timeout, exiting so the process supervisor restarts us',
+            { site: this.site });
+        // grace period so the fatal line flushes to stdout before the exit
+        setTimeout(() => process.exit(1), 1000);
```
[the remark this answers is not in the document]
anurag4DSB: Moot after 8faf5cb4: the handler this commented on is gone. The crash now sits in each `BackbeatConsumer` instance — if two consumers in one process are both stuck, each logs a fatal naming its own topic/groupId and the first exit wins, so the second line identifies the other stuck consumer rather than duplicating.

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -352,10 +355,16 @@ describe('BackbeatConsumer rebalance tests', () => {
         });
     }).timeout(40000);
 
-    it('should fail healthcheck on rebalance timeout', done => {
+    it('should exit on rebalance timeout when CRASH_ON_REBALANCE_TIMEOUT ' +
```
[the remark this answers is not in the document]
anurag4DSB: Both points addressed: the original disarmed test is restored alongside the armed one in 07e2dad0 (it now also stubs `process.exit` and asserts it was NOT called, so the gate staying inert is pinned explicitly), and the PR description was rewritten — the unit-spec claim is gone.

lib/BackbeatConsumer.js
```diff
@@ -729,8 +749,31 @@ class BackbeatConsumer extends EventEmitter {
...
+                    // the disconnect above trips. On K8s the liveness probe
+                    // does, and this env var is never set there. Hard exit
+                    // rather than a graceful stop: the stop path would wait
+                    // on the very tasks that are stuck.
+                    this._log.fatal('consumer disconnected from queue ' +
+                        'after rebalance drain timeout, exiting process',
+                        { topic: this._topic, groupId: this._groupId });
+                    // grace period so the fatal line flushes to stdout
+                    // before the exit
+                    setTimeout(() => process.exit(1), 1000);
```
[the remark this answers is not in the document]
anurag4DSB: so the handle is intentionally not stored, clearing it (e.g. from a close() racing the crash inside the 1s grace) would resurrect exactly the zombie state

extensions/replication/queueProcessor/QueueProcessor.js
anurag4DSB: we are not using this, so not applicable anymore

lib/BackbeatConsumer.js
anurag4DSB: we are not using this code , so not applicable anymore

francoisferrand (approved):

nicolas2bert (approved):

anurag4DSB:
added a new commit for just bumping package.json

anurag4DSB:
squashed and pushed, no changes

anurag4DSB:
ping

anurag4DSB:
Quick note, I am gonna add a bypass build status on this as discussed with @nicolas2bert for other 9.0 targetting PRs and same for https://github.com/scality/backbeat/pull/2763

PR #2762, opened 2026-06-11
BB-788: Log stuck task identities on rebalance timeout

anurag4DSB:

When BB-787's self-exit fires (scality/backbeat#2761), the logs say nothing about which
objects were stuck or whether the exit gate was even armed. Customer success debugging a
stuck CRR site from log files alone needs both: the identity of the wedged object, and the
difference between "this consumer will restart itself" and "this consumer is down until
someone acts".

**Replication (CRR) only — no other consumer changes behavior.** Three structural
guarantees, each grep-verifiable on this diff:

- The `rebalance.timeout` event has exactly **one listener in the repo**
  (`QueueProcessor.js`); for lifecycle, GC, notification, the populator, the status
  processor and all of Zenko, the emit is a no-op by Node EventEmitter semantics.
- `QueueProcessor.js` is loaded **only** by the replication queue/replay processor
  entrypoint (`extensions/replication/queueProcessor/task.js`) — the logger naming and the
  new CRR log lines cannot execute in any other process.
- **Zero per-message work is added** — the hot path (millions of objects/day) is untouched;
  every addition runs on the stuck path, at most once per drain timeout.

The one shared surface, stated openly: the pre-existing `'consumer stuck, disconnecting'`
error line in `lib/BackbeatConsumer.js` gains payload fields (topic, groupId, queue state,
stuck-task identities). Other consumers DO get those richer fields if **they** ever get
stuck — same line, same level, same moment, identical control flow; strictly additive JSON
on a rare incident-only error. That is an observability improvement for them, not a
behavior change.

All hot-path code is untouched — zero per-entry work added (deliberately: millions of
objects flow through these consumers). The revoke/un-assign rebalance lines are unchanged;
drain timing remains derivable from their existing timestamps. The rebalance.timeout event
gains a payload, which its only listener uses for logging.

The stuck error line now names the in-flight tasks — topic, partition, offset and the
kafka key, which for replication entries is the object name, the one identity no existing
log line carries (`slow task` logs the offset but not the key). Capped at 5 tasks, keys
truncated to 200 chars. The queue processor logs an error when the timeout fires with the
gate disarmed ("self-restart disabled; consumer stays disconnected until restarted externally"), and the fatal exit
line carries queue state and site.

Unit: getInFlightTasks shape/truncation/cap tests; the stuck-exit spec asserts the fatal
payload, the gate-off error, and exit timing across all gate states. Functional: the
existing stuck-rebalance test now asserts the event payload carries the stuck-task array
with real offsets from a genuinely wedged consumer against the CI kafka image.

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -390,6 +390,9 @@ class QueueProcessor extends EventEmitter {
         const groupId =
               `${this.repConfig.queueProcessor.groupId}-${this.site}`;
         const consumer = new BackbeatConsumer({
+            // names this consumer's log lines (the default is a generic
+            // 'BackbeatConsumer' shared by every backbeat service)
+            clientId: `Backbeat:Replication:QueueProcessor:${this.site}:Consumer`,
             zookeeper: {
                 connectionString: this.zkConfig.connectionString,
             },
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -497,6 +497,26 @@ class BackbeatConsumer extends EventEmitter {
         };
     }
 
+    /**
+     * Snapshot of the entries currently being processed, for
+     * observability (e.g. naming the tasks stuck past the rebalance
+     * drain timeout). Capped at 5 tasks; the key (the object name for
+     * replication entries) is truncated to 200 chars.
+     *
+     * @return {object[]} one item per in-flight task:
+     * { topic, partition, offset, key }
+     */
+    getInFlightTasks() {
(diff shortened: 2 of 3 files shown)


anurag4DSB:
Folded into #2761 per review simplification — the logs change now rides there as its own commit (ef318c70, BB-788: Log stuck task identities on rebalance timeout). Closing.

anurag4DSB:
Reopened — keeping the logs separate after all, so this PR has room to grow (the consumer logger naming should eventually cover the other backbeat services too, not just CRR).

anurag4DSB:
Correction to the comment above: GitHub would not reopen this after the head branch moved, so the logs change continues as #2763 (same branch, same commit).

PR #2763, opened 2026-06-11
BB-788: Log stuck task identities on rebalance timeout

anurag4DSB:
Note: 1 test failing on this branch is a known and not related to this PR

When the rebalance drain timeout fires, the stuck-consumer error line now says what the
consumer was stuck on: queue depth, running count, and the kafka coordinates + message
key of up to 10 in-flight tasks. Keys are logged whole (their size is already bounded
by arsenal's objectKeyByteLimit). The cap of 10 equals the default consumer
concurrency, so the line can list a full default in-flight set but never grows past
~11KB worst case; the list is oldest-first, so the wedged tasks always lead it, and
anything beyond ten would be the newest, least suspicious entries. Runs once, in the
drain-timeout path only — zero per-message work. Stacked on #2761.

A real line, captured from the functional suite (single line in the logs, wrapped here):

```json
{"name":"BackbeatConsumer","time":1781182870950,
 "topic":"backbeat-consumer-spec-shutdown","groupId":"bucket-processor-0.274398999",
 "queueLen":0,"running":1,
 "stuckTasks":[{"topic":"backbeat-consumer-spec-shutdown","partition":0,"offset":142,"key":"key"}],
 "level":"error","message":"rdkafka.rebalance timeout: consumer stuck, disconnecting",
 "hostname":"...","pid":76103}
```

On a CRR processor the same line carries `topic: backbeat-replication`,
`groupId: backbeat-replication-group-<site>` (configured groupId + site), and
`key: <bucket>/<object-key>` — the key the populator sets when publishing the entry
(`ReplicationQueuePopulator.js`). The key names the wedged object, partition/offset let
you fetch the exact message back from kafka, and grepping the processor log for the key
shows how far the task got.

Ten is also sufficient because nothing beyond it is lost: `running` minus the listed
tasks says how many more were in flight, and the remainder is recoverable two ways --
dump the partition from the lowest listed offset with kafka tools (every message key
names its object), or just wait: the next partition owner re-runs every uncommitted
entry and logs each one as it goes.

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -19,6 +19,9 @@ const { unassignStatus } = require('./constants');
 // controls the number of messages to process in parallel
 const CONCURRENCY_DEFAULT = 1;
 const CLIENT_ID = 'BackbeatConsumer';
+// max in-flight task identities included in a log line, so the line
+// stays small at any configured concurrency
+const IN_FLIGHT_TASKS_LOG_LIMIT = 10;
 const { withTopicPrefix } = require('./util/topic');
 
 /**
@@ -497,6 +500,27 @@ class BackbeatConsumer extends EventEmitter {
         };
     }
 
+    /**
+     * Snapshot of the entries currently being processed, for
+     * observability (e.g. naming the tasks stuck past the rebalance
+     * drain timeout). Capped at IN_FLIGHT_TASKS_LOG_LIMIT tasks; a
+     * wedged task never finishes, so it is always among them.
+     *
+     * @return {object[]} one item per in-flight task:
+     * { topic, partition, offset, key }; [] when there is no
+     * processing queue
+     */
+    getInFlightTasks() {
+        return (this._processingQueue?.workersList() || [])
+            .slice(0, IN_FLIGHT_TASKS_LOG_LIMIT)
+            .map(w => ({
+                topic: w.data.topic,
+                partition: w.data.partition,
+                offset: w.data.offset,
+                key: w.data.key && w.data.key.toString(),
+            }));
+    }
+
     _scheduleNextTryConsume() {
         this._tryConsumedTimeout = setTimeout(this._tryConsume.bind(this), 1000);
     }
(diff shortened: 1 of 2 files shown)


tcarmet (approved):

dvasilas (approved):

anurag4DSB:
rebased

anurag4DSB:
/bypass_build_status


*EDIT*: Update: This was issued with discussion with lead for only 1 test: like it was done [here](https://github.com/scality/backbeat/pull/2749#issuecomment-4659914715)
But on integration branches due to architecture change the tests started failing which was missed due to the bypass.

Learning: When using bypass, check all 5 integration branches' CI
Action: Update BB 9.3.x in Federation https://scality.atlassian.net/browse/S3C-11332 to avoid the overhead of reviewing 6 branches for 1 change

francoisferrand:
> /bypass_build_status
> 
> _EDIT_: Update: This was issued with discussion with lead for only 1 test: like it was done [here](https://github.com/scality/backbeat/pull/2749#issuecomment-4659914715) But on integration branches due to architecture change the tests started failing which was missed due to the bypass.
> 
> Learning: When using bypass, check all 5 integration branches' CI Action: Update BB 9.3.x in Federation https://scality.atlassian.net/browse/S3C-11332 to avoid the overhead of reviewing 6 branches for 1 change

- if one test is _flaky_, just re-running the build is enough to merge the PR ; and much safer the bypassing build status altogether
- we are all in the same boat here, when a flaky test is found we should try to fix it (even if not the same PR) instead of trying to work around it

anurag4DSB:
> * if one test is _flaky_, just re-running the build is enough to merge the PR ; and much safer the bypassing build status altogether'

We retried many times, it failed everytime, even on 9.0.


> * we are all in the same boat here, when a flaky test is found we should try to fix it (even if not the same PR) instead of trying to work around it

Indeed, we created this ticket for it and then it was decided to work on it later: https://scality.atlassian.net/browse/BB-782

PR #2764, opened 2026-06-15
Drop redundant stuck-task logging and stabilize topic-metrics test

maeldonn:
This PR removes the stuck-task snapshot added in BB-788. getInFlightTasks() was reading w.data.topic/partition/offset off async.queue's workers, but the processing queue is now a TaskScheduler that pushes contexts shaped as { entry }, so the field accesses threw and the rdkafka.rebalance timeout: consumer stuck, disconnecting log never carried the identities it was meant to surface. Investigation showed the snapshot was redundant anyway: the per-task slow task warning in _startProcessingTask fires at maxPollIntervalMs and the rebalance drain timeout fires at maxPollIntervalMs - 1000, with the consumer paused during draining, so every in-flight task at timeout has already been logged individually with topic, partition, consumerGroup and offset.

The only field the snapshot carried that slow task did not was the message key, which is now added to the slow-task log. The rebalance-timeout error keeps the cheap queueLen and running reads from the TaskScheduler so operators can still distinguish "wedged on running work" from "backed up in the queue" at the moment of disconnection. TaskScheduler stays agnostic of the kafka entry shape. No new public surface was added.

The PR also stabilizes the "publish topic metrics" functional test, which summed debounced consumed event counts to exactly messages.length and flaked under CI timing. It now checks the queueProcessor's own consumedMessages array, guarded by a flag so the assertion block runs once.

Issue: BB-789

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -478,13 +478,16 @@ class BackbeatConsumer extends EventEmitter {
      * @returns {function} - callback function to be called when task has been processed
      */
     _startProcessingTask(entry) {
-        const { topic, partition, offset } = entry;
+        const { topic, partition, offset, key } = entry;
         const consumerGroup = this._groupId;
         const taskStartTime = Date.now();
 
         let slow = false;
         const slowTimer = setTimeout(() => {
-            this._log.warn('slow task', { topic, partition, consumerGroup, offset });
+            this._log.warn('slow task', {
+                topic, partition, consumerGroup, offset,
+                key: key?.toString(),
+            });
             slow = true;
             KafkaBacklogMetrics.onSlowTask(topic, partition, consumerGroup);
         }, this._maxPollIntervalMs);
@@ -729,7 +732,12 @@ class BackbeatConsumer extends EventEmitter {
             this._drainProcessQueueTimeout = setTimeout(() => {
                 unassign(unassignStatus.TIMEOUT);
 
-                this._log.error('rdkafka.rebalance timeout: consumer stuck, disconnecting');
+                this._log.error('rdkafka.rebalance timeout: consumer stuck, disconnecting', {
+                    topic: this._topic,
+                    groupId: this._groupId,
+                    queueLen: this._processingQueue.length(),
+                    running: this._processingQueue.running(),
+                });
                 this._consumer.disconnect();
                 if (process.env.CRASH_ON_REBALANCE_TIMEOUT === 'true') {
                     // On S3C, supervisord only restarts a program when
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -19,9 +19,6 @@ const { unassignStatus } = require('./constants');
 // controls the number of messages to process in parallel
 const CONCURRENCY_DEFAULT = 1;
 const CLIENT_ID = 'BackbeatConsumer';
-// max in-flight task identities included in a log line, so the line
-// stays small at any configured concurrency
-const IN_FLIGHT_TASKS_LOG_LIMIT = 10;
 const { withTopicPrefix } = require('./util/topic');
 
 /**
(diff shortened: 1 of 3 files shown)


lib/BackbeatConsumer.js
```diff
@@ -156,6 +156,7 @@ class BackbeatConsumer extends EventEmitter {
         this._canary = canary;
         this._bootstrap = bootstrap;
         this._offsetLedger = new OffsetLedger();
+        this._inFlight = new Set();
```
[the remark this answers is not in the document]
maeldonn: No other consumer state (offset ledger, counters, processing queue) is cleared on revoke; entries drain as work completes, or the process exits via the rebalance-timeout disconnect. Leaving stuck entries in `_inFlight` is also useful so `getInFlightTasks()` can still name them after the timeout.

anurag4DSB (commented):
The first change related to log looks good to me, a 2nd change has been bundled in, lets add that context to the PR description /title for future history digging
I am guessing its related to this:
```
 1) BackbeatConsumer main tests
       should be able to read messages sent to the topic and publish topic metrics:
     Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/backbeat/backbeat/tests/functional/lib/BackbeatConsumer.js)
      at listOnTimeout (node:internal/timers:585:17)
      at process.processTimers (node:internal/timers:521:7)
```
For the second fix 9.0 is also affected, can we please backport it to that as well.

Note: The bypass was added for the above test on the original PR(after checking with the lead)
Learning: when issuing a bypass for known flakyness/depriortized failing tests, check all 5 integration branches for potential failures on integration PRs. 9.0 was not an issue here.

anurag4DSB (approved):

francoisferrand:
@anurag4DSB @maeldonn do we really need to add this (duplicated) tracking of tasks and logging?

- when a task takes too long, a log is printed c.f. https://github.com/scality/backbeat/blob/0a8e9dd96a8991ea76584eb0af65358f8bf6af00/lib/BackbeatConsumer.js . Is this not enough?
- the queued task are already tracked in the `TaskScheduler` (as you mentioned), so we can just add a simple function to list/iterate over the content if we really need this extra log : does not need to teach `TaskScheduler` what's in an entry, just reading what was pushed

anurag4DSB:
Update: we added an action in the sprint to update federation with BB 9.3.x https://scality.atlassian.net/browse/S3C-11332 to avoid the overhead of reviewing 6 branches for 1 change

SylvainSenechal (approved):

lib/tasks/TaskScheduler.js
```diff
@@ -39,6 +39,7 @@ class TaskScheduler {
         this._dedupeCache = {};
         this._inQueue = 0; // Number of tasks currently waiting in the queue
         this._running = 0; // Number of tasks currently running
+        this._inFlight = new Set(); // Contexts queued or running, in push order
```
francoisferrand: why? Aren't the tasks are already in `this._taskQueues` ?

francoisferrand: or are we interested in the "running" tasks only?

maeldonn: Good catch, dropped the Set and replaced it with the `inFlight()` generator over `workersList()` (running tasks only).

lib/BackbeatConsumer.js
```diff
@@ -799,15 +778,11 @@ class BackbeatConsumer extends EventEmitter {
...
-                const running = this._processingQueue?.running();
-                const stuckTasks = this.getInFlightTasks();
                 this._log.error('rdkafka.rebalance timeout: consumer stuck, disconnecting', {
                     topic: this._topic,
                     groupId: this._groupId,
-                    queueLen,
-                    running,
-                    stuckTasks,
+                    queueLen: this._processingQueue.length(),
+                    running: this._processingQueue.running(),
```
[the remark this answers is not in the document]
maeldonn: Updated the PR description.

lib/BackbeatConsumer.js
francoisferrand, line 1: nit: `git revert` is the more "visible" way to revert a commit... (and could be reverted directly in 9.0, not sure we need the log there either...)

francoisferrand (approved):

PR #2765, opened 2026-06-16
Stabilize "publish topic metrics" functional test

maeldonn:
The trigger relied on summing the debounced 'consumed' event counts to exactly messages.length, which is fragile under CI timing. Check the queueProcessor's own consumedMessages array instead, guarded by a flag so the block runs once even if 'consumed' fires again after we have enough.

Issue: BB-789


SylvainSenechal (approved):

anurag4DSB (approved):

benzekrimaha (approved):

maeldonn:
Replaced with #2764

PR #2766, opened 2026-06-18
feat(lifecycle): support Days=0 expiration (current, noncurrent, abort-MPU)

delthas:
## Summary

Support an explicit lifecycle expiration rule with `Days=0` (previously the minimum was 1) as the unambiguous "empty this bucket" intent — for current `Expiration.Days`, `NoncurrentVersionExpiration.NoncurrentDays`, and orphan delete-marker expiration.

The only backbeat code that dropped `Days=0` was the V2 listing optimizer (`RulesReducer`), which used truthy `if (r.Expiration.Days)` gates. These are switched to `!== undefined`. `days=0` then yields a listing with no `BeforeDate` (lists all objects) — the same internal state a past-dated `Expiration.Date` already produces. Per-object eligibility checks (`LifecycleTask` / `LifecycleTaskV2`) already handle `0`.

`AbortIncompleteMultipartUpload.DaysAfterInitiation=0` needs no backbeat change: MPUs are listed via `listMultipartUploads` (not the reducer), and the per-object check already supports `0`.

## Dependencies

This change is **self-contained** — the reducer reads the already-stored `Days` value and never calls arsenal's lifecycle parser, so backbeat needs no arsenal change. The earlier arsenal-bump commit has been **dropped**; this PR inherits `development/9.5`'s arsenal pin (`8.4.7`) with no version jump.

End-to-end, `Days=0` still requires the *storage* side to accept and persist a 0-day rule — arsenal **ARSN-597** (relaxed day-based validators, now on the 8.5 line) as consumed by cloudserver **CLDSRV-928**. That's a runtime/rollout ordering, not a build dependency of this PR.

## Notes

- `Days=0` deliberately diverges from AWS S3 (which requires ≥ 1); that divergence is what makes it a safe explicit intent signal.

Issue: BB-791

### DIFF
--- a/extensions/lifecycle/util/RulesReducer.js
+++ b/extensions/lifecycle/util/RulesReducer.js
@@ -142,8 +142,7 @@ class RulesReducer {
         let storageClass;
 
         if (r.Expiration) {
-            // NOTE: Expiration Days cannot be 0.
-            if (r.Expiration.Days) {
+            if (r.Expiration.Days !== undefined) {
                 days = this._decrementExpirationDay(r.Expiration.Days);
             } else if (r.Expiration.Date) {
                 if (r.Expiration.Date <= this._currentDate) {
@@ -200,8 +199,8 @@ class RulesReducer {
         let storageClass;
 
         if (r.NoncurrentVersionExpiration) {
-            // 'NoncurrentDays' for NoncurrentVersionExpiration action is a positive integer
-            if (r.NoncurrentVersionExpiration.NoncurrentDays) {
+            // 'NoncurrentDays' for NoncurrentVersionExpiration action is a nonnegative integer
+            if (r.NoncurrentVersionExpiration.NoncurrentDays !== undefined) {
                 days = this._decrementExpirationDay(r.NoncurrentVersionExpiration.NoncurrentDays);
             }
         }
@@ -246,7 +245,7 @@ class RulesReducer {
         // NOTE: When you specify the Days tag, Amazon S3 automatically performs ExpiredObjectDeleteMarker
         // cleanup when the delete markers are old enough to satisfy the age criteria.
         if (r.Expiration) {
-            if (r.Expiration.Days) {
+            if (r.Expiration.Days !== undefined) {
                 days = this._decrementExpirationDay(r.Expiration.Days);
             } else if (r.Expiration.Date) {
                 if (r.Expiration.Date <= this._currentDate) {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "@scality/cloudserverclient": "^1.0.8",
     "@smithy/node-http-handler": "^3.3.3",
     "JSONStream": "^1.3.5",
-    "arsenal": "git+https://github.com/scality/arsenal#8.3.9",
+    "arsenal": "git+https://github.com/scality/arsenal#f5d43abeb8139ec8835ef0354888ff13ddb14d8b",
```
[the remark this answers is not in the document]
delthas: I know

extensions/lifecycle/util/RulesReducer.js
```diff
@@ -200,8 +199,8 @@ class RulesReducer {
         let storageClass;
 
         if (r.NoncurrentVersionExpiration) {
-            // 'NoncurrentDays' for NoncurrentVersionExpiration action is a positive integer
-            if (r.NoncurrentVersionExpiration.NoncurrentDays) {
+            // 'NoncurrentDays' for NoncurrentVersionExpiration action is a nonnegative integer
+            if (r.NoncurrentVersionExpiration.NoncurrentDays !== undefined) {
```
SylvainSenechal, line 203: should we use >= 0 instead of !== undefined 🤔 

For this line but also the other in this file, I think result is the same but more readable and gatekeep negative values

SylvainSenechal (approved):

extensions/lifecycle/util/RulesReducer.js
delthas: Good point — for valid input the two are equivalent. I kept `!== undefined` for consistency: the transition checks in these same functions (`r.Transitions[0].Days !== undefined`, `r.NoncurrentVersionTransitions[0].NoncurrentDays !== undefined`) and all the per-object checks in `LifecycleTask`/`LifecycleTaskV2` already use this idiom, so switching only the three expiration gates would mix two styles within the same function.

Negative values can't reach the reducer anyway — arsenal validates day-based actions as nonnegative (ARSN-597 routes them through `_checkDays`), so this gate is really a presence check (Days vs Date vs ExpiredObjectDeleteMarker) rather than a re-validation.

benzekrimaha (approved):

PR #2768, opened 2026-06-24
Process master entry on ObjectRestore:Retry

maeldonn:
For null-version objects (master entry with versionId and isNull: true, typical of buckets that have had versioning Suspended at any point), a requeueRestore Kafka message without objectVersion updates the master metadata as expected but never produces a cold-restore-req message: the lifecycle queue populator's master-skip rule drops the resulting oplog event, and the retry never reaches DMF.

The skip rule was designed for the normal restore flow (s3:ObjectRestore:Post), where the populator receives events on both the master and the version entry and must dedup. The retrigger restore flow (s3:ObjectRestore:Retry) only updates the master entry, so the master event is the only event the populator will ever get for that retry.

This PR exempts s3:ObjectRestore:Retry from the master-skip rule. The normal s3:ObjectRestore:Post path is unchanged.

The current workaround is to pass the encoded internal VersionId returned by head-object in objectVersion, which routes the update through the version-entry path and avoids the skip rule. That requires the caller to know an internal versionId on objects that are otherwise non-versioned from the API's perspective. With this fix, omitting objectVersion works as expected on null-version objects, matching what the S3 API surface implies.

Issue: BB-794

### DIFF
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -283,8 +283,10 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         }
 
         // if entry is a versioned object and is the master entry, skip task as
-        // the non-master entry will be processed
-        if (this._isVersionedObject(value) && isMasterKey(entry.key)) {
+        // the non-master entry will be processed. The retrigger restore flow
+        // (originOp 's3:ObjectRestore:Retry') only updates the master entry, so
+        // the master event is the only one we will ever get — do not skip it.
+        if (this._isVersionedObject(value) && isMasterKey(entry.key) && operation !== 's3:ObjectRestore:Retry') {
             this.log.trace('skip processing of object master entry');
             return;
         }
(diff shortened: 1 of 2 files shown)


extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -283,8 +283,10 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         }
 
         // if entry is a versioned object and is the master entry, skip task as
-        // the non-master entry will be processed
-        if (this._isVersionedObject(value) && isMasterKey(entry.key)) {
+        // the non-master entry will be processed. The retrigger restore flow
+        // (originOp 's3:ObjectRestore:Retry') only updates the master entry, so
+        // the master event is the only one we will ever get — do not skip it.
+        if (this._isVersionedObject(value) && isMasterKey(entry.key) && operation !== 's3:ObjectRestore:Retry') {
```
francoisferrand, line 289: Introducing this asymmetry seems very weird : AFAIK we are able to restore both "master-only" null versions (i.e. object created before versioning was enabled) and "version-suspended" versions : so `_isVersionedObject()` seems to handle its job, and prevent duplicate handling of restore while still processing master document when needed.

Generally, lifecycle (cold transition & restore) work for these objects (tested in zenko I think, to be confirmed) :  so we need to understand exactly _what_ Sorbet passes in message, so we can properly "forge" the message (+maybe add missing checks to ensure a "bad" message does not make things worse)

> "only updates the master entry"

- if there is only a master, should be covered (already) by `_isVersionedObject` ?
- if there is both a master & a version documents, somehow the code ends up creating a desynchro between master & version : which is a separate but critical issue

francoisferrand (changes requested):

PR #2771, opened 2026-07-06
improvement: commit + log + metric on task failure in GC and lifecycle cold-archive

delthas:
**TL;DR** — Three task error paths returned `{ committable: false }` after their retry budget was exhausted without ever calling `onEntryCommittable`. The offset never advanced past those entries, they lingered in the `OffsetLedger`, and with the BB-777 rebalance-wait-for-drain fix now merged this turns every rebalance that catches one into a forced pod restart. This PR flips those sites to commit on final failure, emits a structured error log at the point of drop (including the individual data-store parts so ops can reclaim manually), and adds a per-extension counter that can be queried from Prometheus.

## What "final failure" means here

All three sites are wrapped by a `BackbeatTask.retry`-style helper with a ~5 minute budget and exponential backoff — they only reach the poison-pill path once retries are exhausted or the error is flagged non-retryable. Realistic triggers include:

- a warm cluster / metadata service / vault outage lasting longer than the retry budget
- non-retryable backend errors (`AccessDenied` after a role rotation, `Conflict` on `putMetadata`, `InvalidRequest`)
- the lifecycle orphan-cleanup path failing to publish because the cold producer can't reach its target topic
- `arsenal.errors.InternalError` when the vault-derived client can't be obtained

## Why not "kafka will redeliver"

That was the intent behind `committable: false`, but it doesn't work: Kafka only redelivers on the next assignment (rebalance / restart), and neither path benefits from an upstream cron. The cold-status message is a one-shot notification from the cold backend; the GC `deleteArchivedSourceData` entry is published exactly once by the archive task in fire-and-forget mode. On top of that, `LifecycleTask` explicitly skips objects with `transitionInProgress=true` on later bucket scans, so the regular pipeline won't pick a stuck object back up either. Effectively the current behavior is "hold the offset forever, ledger grows, and next rebalance forces a pod restart."

## What this PR does

For each of the three sites:

1. Stop returning `{ committable: false }` — just call `done(err)` so the consumer commits the offset like it would for any other completed task.
2. At the outer retry-wrapper's completion callback (i.e. the point where the retry gave up), emit an error-level log with the entry info, the error message, and the individual data-store parts / locations the task was trying to delete (so ops has enough to reclaim stuck data manually).
3. Increment a new counter recording the failure so operators can query it from Prometheus:
   - `s3_gc_failed_total{origin, location}` for GC
   - `s3_lifecycle_failed_total{origin, type, location}` for lifecycle

The log + metric live at the retry-wrapper's completion callback (not inside the task's own error handler) so they fire exactly once per final give-up, not once per attempt.

Sites covered:

- `GarbageCollectorTask._deleteArchivedSourceData` — the archive-side warm-copy cleanup. Parts come from `objMD.getLocation()` inside the async.waterfall; stashed on the entry via `entry.setAttribute('target.locations', ...)` so the outer failure log can surface them.
- `GarbageCollectorTask._executeDeleteData` — the hot transition / restore-expiration paths. Already committed silently on failure (no poison-pill), but had no log or metric. Parts are already on the entry.
- `LifecycleObjectTransitionProcessor.processColdStorageStatusEntry` — the cold-status archive handler. No parts concept (this task manipulates metadata + enqueues GC); log fires without a parts field. Uses `this.getProcessorType()` for the metric label rather than a hard-coded string.

For GC, the pre-existing `ObjNotFound` / `NoSuchBucket` carveout (return with commit, no retry) is simplified from `done(err, { committable: true })` to `done(err)` — same semantics, less noise.

## What this PR does not do

- No dashboards or alerts. The metric is a primitive for ad-hoc extraction on platforms where the ledger leak is observed; dashboards and alert rules are a follow-up.
- No retry topic, dead-letter topic, or sweeper. That's a separate design conversation. The consequence of this PR is that permanent failures are now visible as an error log + a metric datapoint rather than a silent ledger leak, but the underlying operation is not retried automatically.
- Does not change the behavior of Pattern A callers (`ReplicateObject`, `MultipleBackendTask`, `CopyLocationTask`, `LifecycleUpdateTransitionTask`) that use `committable: false` correctly with a producer-callback `onEntryCommittable`.

Issue: BB-772

### DIFF
--- a/extensions/gc/GarbageCollectorMetrics.js
+++ b/extensions/gc/GarbageCollectorMetrics.js
@@ -23,6 +23,14 @@ const gcDuration = ZenkoMetrics.createHistogram({
     buckets: [0.2, 0.1, 0.5, 2.5, 10, 50],
 });
 
+const gcFailed = ZenkoMetrics.createCounter({
+    name: 's3_gc_failed_total',
+    help: 'Total number of GC tasks that failed permanently after retries ' +
+        'were exhausted or on a non-retryable error. Offset is committed to ' +
+        'avoid ledger leaks; the underlying operation is not retried.',
+    labelNames: [GC_LABEL_ORIGIN, GC_LABEL_LOCATION],
+});
+
 class GarbageCollectorMetrics {
     static handleError(log, err, method) {
         if (log) {
@@ -51,6 +59,17 @@ class GarbageCollectorMetrics {
             }, duration / 1000);
         } catch (err) {
             GarbageCollectorMetrics.handleError(log, err, 'GarbageCollectorMetrics.onGcComplete');
+        }
+    }
+
+    static onGcFailed(log, process, location) {
+        try {
+            gcFailed.inc({
+                [GC_LABEL_ORIGIN]: process,
+                [GC_LABEL_LOCATION]: location,
+            });
+        } catch (err) {
+            GarbageCollectorMetrics.handleError(log, err, 'GarbageCollectorMetrics.onGcFailed');
         }
     }
 }
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -203,7 +203,20 @@ class GarbageCollectorTask extends BackbeatTask {
             actionFunc: cb => this._executeDeleteDataOnce(entry, log, cb),
             shouldRetryFunc: err => err.retryable,
             log,
-        }, done);
+        }, err => {
+            if (err) {
+                const ruleType = entry.getContextAttribute('ruleType');
+                log.error('task failed permanently after retries, committing offset', {
+                    method: 'GarbageCollectorTask._executeDeleteData',
+                    error: err.message,
+                    locations: entry.getAttribute('target.locations'),
+                    ...entry.getLogInfo(),
+                });
+                GarbageCollectorMetrics.onGcFailed(log, ruleType,
+                    entry.getAttribute('source.storageClass'));
+            }
+            return done(err);
+        });
     }
 
     _deleteArchivedSourceDataOnce(entry, log, done) {
@@ -250,6 +263,11 @@ class GarbageCollectorTask extends BackbeatTask {
                     }
 
                     if (err) {
+                        // Stash the parts on the entry so the outer
+                        // retry-wrapper's failure log (in
+                        // _deleteArchivedSourceData) can surface them for ops
+                        // to reclaim manually if the delete never succeeds.
+                        entry.setAttribute('target.locations', locations);
                         log.error('an error occurred on batchDelete backbeat route',
                             Object.assign({
                                 method: 'GarbageCollectorTask._deleteArchivedSourceData',
@@ -289,17 +307,7 @@ class GarbageCollectorTask extends BackbeatTask {
                     next(err);
                 });
             },
-        ], err => {
-            if (err) {
-                // if error occurs, do not commit offset unless the error is ObjNotFound
-                // because it means the object has been deleted by other means and we don't need to retry
-                    if (err.name === 'ObjNotFound' || err.name === 'NoSuchBucket') {
-                        return done(err, { committable: true });
-                }
-                return done(err, { committable: false });
-            }
-            return done();
-        });
+        ], err => done(err));
     }
 
     _deleteArchivedSourceData(entry, log, done) {
@@ -310,7 +318,19 @@ class GarbageCollectorTask extends BackbeatTask {
             actionFunc: cb => this._deleteArchivedSourceDataOnce(entry, log, cb),
             shouldRetryFunc: err => err.retryable,
             log,
-        }, done);
+        }, err => {
+            if (err && err.name !== 'ObjNotFound' && err.name !== 'NoSuchBucket') {
+                log.error('task failed permanently after retries, committing offset', {
+                    method: 'GarbageCollectorTask._deleteArchivedSourceData',
+                    error: err.message,
+                    locations: entry.getAttribute('target.locations'),
+                    ...entry.getLogInfo(),
+                });
+                GarbageCollectorMetrics.onGcFailed(log, 'archive',
+                    entry.getAttribute('target.oldLocation'));
+            }
+            return done(err);
+        });
     }
 
     /**
--- a/extensions/lifecycle/LifecycleMetrics.js
+++ b/extensions/lifecycle/LifecycleMetrics.js
@@ -194,6 +194,15 @@ const lifecycleKafkaPublish = {
     }),
 };
 
+const lifecycleFailed = ZenkoMetrics.createCounter({
+    name: 's3_lifecycle_failed_total',
+    help: 'Total number of lifecycle tasks that failed permanently after ' +
+        'retries were exhausted or on a non-retryable error. Offset is ' +
+        'committed to avoid ledger leaks; the underlying operation is not ' +
+        'retried.',
+    labelNames: [LIFECYCLE_LABEL_ORIGIN, LIFECYCLE_LABEL_TYPE, LIFECYCLE_LABEL_LOCATION],
+});
+
 class LifecycleMetrics {
     static handleError(log, err, method, params = {}) {
         if (log) {
@@ -419,6 +428,20 @@ class LifecycleMetrics {
         } catch (err) {
             LifecycleMetrics.handleError(log, err, 'LifecycleMetrics.onKafkaPublish', {
                 op, process, count, kafkaErr,
+            });
+        }
+    }
+
+    static onLifecycleFailed(log, process, type, location) {
+        try {
+            lifecycleFailed.inc({
+                [LIFECYCLE_LABEL_ORIGIN]: process,
+                [LIFECYCLE_LABEL_TYPE]: type,
+                [LIFECYCLE_LABEL_LOCATION]: location,
+            });
+        } catch (err) {
+            LifecycleMetrics.handleError(log, err, 'LifecycleMetrics.onLifecycleFailed', {
+                process, type, location,
             });
         }
     }
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
@@ -3,6 +3,7 @@ const assert = require('assert');
 const async = require('async');
 
 const ColdStorageStatusQueueEntry = require('../../../lib/models/ColdStorageStatusQueueEntry');
+const { LifecycleMetrics } = require('../LifecycleMetrics');
 const LifecycleObjectProcessor = require('./LifecycleObjectProcessor');
 const LifecycleUpdateExpirationTask = require('../tasks/LifecycleUpdateExpirationTask');
 const LifecycleUpdateTransitionTask = require('../tasks/LifecycleUpdateTransitionTask');
@@ -200,7 +201,20 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
             actionFunc: done => task.processEntry(coldLocation, entry, done),
             shouldRetryFunc: err => err.retryable,
             log: this._log,
-        }, done);
+        }, err => {
+            if (err) {
+                this._log.error('task failed permanently after retries, committing offset', {
+                    method: 'LifecycleObjectTransitionProcessor.processColdStorageStatusEntry',
+                    error: err.message,
+                    op: entry.op,
+                    coldLocation,
+                    ...entry.getLogInfo(),
+                });
+                LifecycleMetrics.onLifecycleFailed(this._log, this.getProcessorType(),
+                    entry.op, coldLocation);
+            }
+            return done(err);
+        });
     }
 
     getStateVars() {
--- a/tests/unit/gc/GarbageCollectorTask.spec.js
+++ b/tests/unit/gc/GarbageCollectorTask.spec.js
@@ -4,6 +4,7 @@ const sinon = require('sinon');
 const { ObjectMD } = require('arsenal').models;
 
 const GarbageCollectorTask = require('../../../extensions/gc/tasks/GarbageCollectorTask');
+const { GarbageCollectorMetrics } = require('../../../extensions/gc/GarbageCollectorMetrics');
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
 
 const {
@@ -47,6 +48,7 @@ describe('GarbageCollectorTask', () => {
             consumer: {
                 retry: {
                     maxRetries: 3,
+                    backoff: { min: 10, max: 20, factor: 1, jitter: 0 },
                 },
             },
         };
@@ -141,8 +143,9 @@ describe('GarbageCollectorTask', () => {
         });
     });
 
-    it('should not delete archived location info if gc failed', done => {
+    it('should commit and record a failure metric if gc failed', done => {
         backbeatClient.batchDeleteResponse = { error: { statusCode: 500 }, res: null };
+        const onGcFailedSpy = sinon.spy(GarbageCollectorMetrics, 'onGcFailed');
 
         const entry = ActionQueueEntry.create('deleteArchivedSourceData')
               .addContext({
@@ -169,7 +172,10 @@ describe('GarbageCollectorTask', () => {
 
         gcTask.processActionEntry(entry, (err, commitInfo) => {
             assert.strictEqual(err.statusCode, 500);
-            assert.deepStrictEqual(commitInfo, { committable: false });
+            assert.strictEqual(commitInfo, undefined);
+            assert.strictEqual(onGcFailedSpy.callCount, 1);
+            assert.deepStrictEqual(onGcFailedSpy.firstCall.args.slice(1),
+                ['archive', 'old-location']);
             assert.strictEqual(backbeatMetadataProxyClient.getReceivedMd(), null);
 
             const objMD = backbeatMetadataProxyClient.mdObj;
@@ -177,12 +183,14 @@ describe('GarbageCollectorTask', () => {
             assert.strictEqual(objMD.getDataStoreName(), 'old-location');
             assert.strictEqual(objMD.getAmzStorageClass(), 'old-location');
             assert.strictEqual(objMD.getTransitionInProgress(), true);
+            onGcFailedSpy.restore();
             done();
         });
     });
 
     it('should retry delete archived source data if gc failed with retryable error', done => {
         backbeatClient.batchDeleteResponse = { error: { statusCode: 500, retryable: true }, res: null };
+        const onGcFailedSpy = sinon.spy(GarbageCollectorMetrics, 'onGcFailed');
 
         const entry = ActionQueueEntry.create('deleteArchivedSourceData')
               .addContext({
@@ -212,13 +220,17 @@ describe('GarbageCollectorTask', () => {
         gcTask.processActionEntry(entry, (err, commitInfo) => {
             assert.strictEqual(batchDeleteDataSpy.callCount, 4);
             assert.strictEqual(err.statusCode, 500);
-            assert.deepStrictEqual(commitInfo, { committable: false });
+            assert.strictEqual(commitInfo, undefined);
+            assert.strictEqual(onGcFailedSpy.callCount, 1);
+            assert.deepStrictEqual(onGcFailedSpy.firstCall.args.slice(1),
+                ['archive', 'old-location']);
             batchDeleteDataSpy.restore();
+            onGcFailedSpy.restore();
             done();
         });
     });
 
-    it('should send committable error when object is not found', done => {
+    it('should commit without retry when object is not found', done => {
         backbeatMetadataProxyClient.error = { statusCode: 404, name: 'ObjNotFound' };
 
         const entry = ActionQueueEntry.create('deleteArchivedSourceData')
@@ -242,12 +254,12 @@ describe('GarbageCollectorTask', () => {
 
         gcTask.processActionEntry(entry, (err, commitInfo) => {
             assert.strictEqual(err.statusCode, 404);
-            assert.deepStrictEqual(commitInfo, { committable: true });
+            assert.strictEqual(commitInfo, undefined);
             done();
         });
     });
 
-    it('should send committable error when bucket is not found', done => {
+    it('should commit without retry when bucket is not found', done => {
         backbeatMetadataProxyClient.error = { statusCode: 404, name: 'NoSuchBucket' };
 
         const entry = ActionQueueEntry.create('deleteArchivedSourceData')
@@ -271,7 +283,7 @@ describe('GarbageCollectorTask', () => {
 
         gcTask.processActionEntry(entry, (err, commitInfo) => {
             assert.strictEqual(err.statusCode, 404);
-            assert.deepStrictEqual(commitInfo, { committable: true });
+            assert.strictEqual(commitInfo, undefined);
             done();
         });
     });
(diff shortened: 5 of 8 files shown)


extensions/gc/GarbageCollectorMetrics.js
```diff
@@ -53,6 +62,18 @@ class GarbageCollectorMetrics {
             GarbageCollectorMetrics.handleError(log, err, 'GarbageCollectorMetrics.onGcComplete');
         }
     }
+
+    static onGcFailed(log, process, location, retryable) {
+        try {
+            gcFailed.inc({
+                [GC_LABEL_ORIGIN]: process,
+                [GC_LABEL_LOCATION]: location,
+                [GC_LABEL_RETRYABLE]: retryable === true,
```
francoisferrand: is `retryable` really relevant here?
the metric will only be called if we exhausted the retries anyway, and details will only be in the logs...


```suggestion
```

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -200,7 +201,21 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
+            if (err) {
+                this._log.error('task failed permanently after retries, committing offset', {
+                    method: 'LifecycleObjectTransitionProcessor.processColdStorageStatusEntry',
+                    error: err.message,
+                    retryable: err.retryable === true,
+                    op: entry.op,
+                    coldLocation,
+                    ...entry.getLogInfo(),
+                });
+                LifecycleMetrics.onLifecycleFailed(this._log, 'transition-processor',
```
francoisferrand: should not use hard-coded `transition-processor`, there is a variable I think?
(a member of of LifecycleObjectProcessor, set in constructor by each derived subclass if I remember correctly)

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -194,6 +194,16 @@ const lifecycleKafkaPublish = {
...
 };
 
+const lifecycleFailed = ZenkoMetrics.createCounter({
+    name: 's3_lifecycle_failed_total',
+    help: 'Total number of lifecycle tasks that failed permanently after ' +
+        'retries were exhausted or on a non-retryable error. Offset is ' +
+        'committed to avoid ledger leaks; the underlying operation is not ' +
+        'retried.',
+    labelNames: [LIFECYCLE_LABEL_ORIGIN, LIFECYCLE_LABEL_TYPE,
+                 LIFECYCLE_LABEL_LOCATION, 'retryable'],
```
francoisferrand: same, I don't think the "retryable" label is useful

```suggestion
    labelNames: [
        LIFECYCLE_LABEL_ORIGIN,
        LIFECYCLE_LABEL_TYPE,
        LIFECYCLE_LABEL_LOCATION,
    ],
```

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -310,7 +300,20 @@ class GarbageCollectorTask extends BackbeatTask {
...
+                    method: 'GarbageCollectorTask._deleteArchivedSourceData',
+                    error: err.message,
+                    retryable: err.retryable === true,
+                    ...entry.getLogInfo(),
+                });
+                GarbageCollectorMetrics.onGcFailed(log, 'archive',
+                    entry.getAttribute('target.oldLocation'),
+                    err.retryable === true);
+            }
+            return done(err);
```
francoisferrand, line 332: don't we need to set `committable: true` explicitely in case of error ?

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -310,7 +300,20 @@ class GarbageCollectorTask extends BackbeatTask {
             actionFunc: cb => this._deleteArchivedSourceDataOnce(entry, log, cb),
             shouldRetryFunc: err => err.retryable,
             log,
-        }, done);
+        }, err => {
+            if (err && err.name !== 'ObjNotFound' && err.name !== 'NoSuchBucket') {
+                log.error('task failed permanently after retries, committing offset', {
```
francoisferrand, line 323: is error not logged already (in continuation callback) ?

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -310,7 +300,20 @@ class GarbageCollectorTask extends BackbeatTask {
             actionFunc: cb => this._deleteArchivedSourceDataOnce(entry, log, cb),
             shouldRetryFunc: err => err.retryable,
             log,
-        }, done);
+        }, err => {
```
francoisferrand, line 321: here you fix `_deleteArchivedSourceData` ; but don't we have the same issue in `_batchDeleteData()` as well (used for hot transitions) ?

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -310,7 +300,20 @@ class GarbageCollectorTask extends BackbeatTask {
             actionFunc: cb => this._deleteArchivedSourceDataOnce(entry, log, cb),
             shouldRetryFunc: err => err.retryable,
```
francoisferrand, line 319: what is the "retry budget" here?

we need to do something, but should we also increase this budget to ensure we don't just create orphans unexpectedly?
(note: we must increase the retry too much either, i.e. not go beyond Kafka's poll/rebalance deadlines : each task must still be processed within the expected time)

extensions/gc/tasks/GarbageCollectorTask.js
benzekrimaha: Not needed AFAIK the consumer commits by default  `BackbeatConsumer._onEntryProcessingDone` only skips onEntryCommittable() when the task explicitly returns `committable: false`.

tests/unit/gc/GarbageCollectorTask.spec.js
```diff
@@ -210,13 +217,17 @@ describe('GarbageCollectorTask', () => {
         const batchDeleteDataSpy = sinon.spy(gcTask, '_batchDeleteData');
 
         gcTask.processActionEntry(entry, (err, commitInfo) => {
-            assert.strictEqual(batchDeleteDataSpy.callCount, 4);
+            assert(batchDeleteDataSpy.callCount >= 1);
```
benzekrimaha: why weaken this? with `maxRetries: 3` the count is deterministic (4 attempts) and nothing in this PR changes the retry behavior. if the test is slow/flaky, we can set a small backoff  in the test's retry config rather than dropping the assertion and bumping the timeout to 15s.

tests/unit/gc/GarbageCollectorTask.spec.js
```diff
@@ -210,13 +217,17 @@ describe('GarbageCollectorTask', () => {
...
+            assert.deepStrictEqual(onGcFailedSpy.firstCall.args.slice(1),
+                ['archive', 'old-location', true]);
             batchDeleteDataSpy.restore();
+            onGcFailedSpy.restore();
             done();
         });
-    });
+    }).timeout(15000);
 
     it('should send committable error when object is not found', done => {
```
benzekrimaha: nit: test name is stale as there is no "committable error" anymore (same below)
```suggestion
    it('should commit without retry when object is not found', done => {
```

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -200,7 +201,21 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
             actionFunc: done => task.processEntry(coldLocation, entry, done),
             shouldRetryFunc: err => err.retryable,
             log: this._log,
-        }, done);
+        }, err => {
+            if (err) {
+                this._log.error('task failed permanently after retries, committing offset', {
```
benzekrimaha, line 206: could we add a unit test driving processColdStorageStatusEntry with a failing task and asserting the onLifecycleFailed call ? ( codecov is complaining as well )

extensions/gc/GarbageCollectorMetrics.js
```diff
@@ -23,6 +24,14 @@ const gcDuration = ZenkoMetrics.createHistogram({
     buckets: [0.2, 0.1, 0.5, 2.5, 10, 50],
 });
 
+const gcFailed = ZenkoMetrics.createCounter({
+    name: 's3_gc_failed_total',
```
benzekrimaha, line 27: the justification for this metric is alerting, but monitoring/ (alerts.yaml / dashboards) are untouched  do have a followup ticket for the alert definitions? otherwise the metric ships blind

extensions/gc/GarbageCollectorMetrics.js
delthas: The label distinguishes two operationally distinct failure modes that both land at this metric:
- `retryable=true` — retries genuinely exhausted after ~5 min (backend outage). `BackbeatTask.retry` only enters the retry loop when `shouldRetryFunc(err)` returns truthy, so this path corresponds to "we tried and gave up."
- `retryable=false` — the error was never retryable to begin with, so `BackbeatTask.retry` short-circuits on the first attempt (`BackbeatTask.js:79-81`) without any retry. Common triggers: `AccessDenied` after a role rotation, `InvalidRequest`/`Conflict` from cloudserver, `errors.InternalError` from arsenal when a client cannot be obtained, kafka producer send errors.

Same counter without the label conflates "backend was down for 5 minutes" (expected transient) with "we did not even try because the state was inconsistent" (config/auth problem). The two need different alert thresholds.

extensions/lifecycle/LifecycleMetrics.js
delthas: The label distinguishes two operationally distinct failure modes that both land at this metric:
- `retryable=true` — retries genuinely exhausted after ~5 min (backend outage). `BackbeatTask.retry` only enters the retry loop when `shouldRetryFunc(err)` returns truthy, so this path corresponds to "we tried and gave up."
- `retryable=false` — the error was never retryable to begin with, so `BackbeatTask.retry` short-circuits on the first attempt (`BackbeatTask.js:79-81`) without any retry. Common triggers: `AccessDenied` after a role rotation, `InvalidRequest`/`Conflict` from cloudserver, `errors.InternalError` from arsenal when a client cannot be obtained, kafka producer send errors.

Same counter without the label conflates "backend was down for 5 minutes" (expected transient) with "we did not even try because the state was inconsistent" (config/auth problem). The two need different alert thresholds.

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
delthas: Done in [`01006128`](https://github.com/scality/backbeat/pull/2771/commits/01006128) — using `this.getProcessorType()`.

extensions/gc/tasks/GarbageCollectorTask.js
delthas: +1 to Maha's note — `done(err)` and `done(err, { committable: true })` are equivalent since the consumer treats absence of `committable: false` as commit-able. The rest of the codebase uses the shorter form (`_executeDeleteData` in this same file does `done(err)` on failure).

extensions/gc/tasks/GarbageCollectorTask.js
delthas: There is overlap with `BackbeatTask.retry`'s `'giving up processing as retries ended'` (`BackbeatTask.js:90`), but only for the retryable-exhausted path. The non-retryable path skips the retry loop entirely (`shouldRetryFunc(err) === false` → `doneOnce(...args)` at `BackbeatTask.js:79-81`) and never hits that message — without this new log, non-retryable failures would only leave per-step logs (batchDelete error, putMetadata error) with no task-level "we gave up" signal.

extensions/gc/tasks/GarbageCollectorTask.js
delthas: Good catch. Done in [`01006128`](https://github.com/scality/backbeat/pull/2771/commits/01006128) — `_executeDeleteData` was already committing on failure (no poison-pill) but silently. Added the same log + metric pattern; reuses `s3_gc_failed_total` with `origin` = ruleType (`expiration`/`transition`/`restore`) and `location` from the target locations. Extended the existing retry unit test to assert the metric fires.

extensions/gc/tasks/GarbageCollectorTask.js
delthas: Current values: `BackbeatTask.retry` default `timeoutS: 300` (5 min), backoff min 1s, max 5 min, factor 1.5, jitter 0.1. `gcConfig.consumer.retry` can override. On the lifecycle side, `retryWrapper` is set up from the same shape via the object-processor config.

Ceiling: the sum of retry attempts must stay under `max.poll.interval.ms` (~5 min default in librdkafka), otherwise a rebalance that catches an in-flight retry hits the BB-777 outer timeout (`max.poll.interval - 1s`) and forces a pod disconnect. Bumping the retry budget in isolation moves us closer to that boundary; a wider budget would need `max.poll.interval` widened in step.

tests/unit/gc/GarbageCollectorTask.spec.js
delthas: Done in [`c7a81fb1`](https://github.com/scality/backbeat/pull/2771/commits/c7a81fb1). Restored the `callCount === 4` assertion, tightened the retry backoff in the test fixture (`{ min: 10, max: 20, factor: 1, jitter: 0 }`) so it runs in tens of ms, dropped the `.timeout(15000)`.

tests/unit/gc/GarbageCollectorTask.spec.js
delthas: Done in [`c7a81fb1`](https://github.com/scality/backbeat/pull/2771/commits/c7a81fb1) — renamed both to `should commit without retry when {object,bucket} is not found`.

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
delthas: Done in [`c7a81fb1`](https://github.com/scality/backbeat/pull/2771/commits/c7a81fb1) — added three tests in `tests/unit/lifecycle/LifecycleObjectTransitionProcessor.spec.js` under a `processColdStorageStatusEntry failure metric` describe: retryable=true after retries exhausted, retryable=false on non-retryable error, and no metric on success. Stubs `LifecycleColdStatusArchiveTask.prototype.processEntry` and swaps in a fast retry config so they run in ms.

extensions/gc/GarbageCollectorMetrics.js
delthas: Done in [`c7a81fb1`](https://github.com/scality/backbeat/pull/2771/commits/c7a81fb1) — added a "Task permanent failures" TimeSeries panel to the lifecycle dashboard's Lifecycle Tasks section, plotting both `s3_lifecycle_failed_total` (by type + retryable) and `s3_gc_failed_total` (by origin + retryable). `dashboard.json` regenerated with `grafanalib==0.6.3` + `scalgrafanalib==1.2.1` (the versions that produce a zero-diff regen of the currently committed JSON). No alerts.yaml changes for now — the metric is primarily for ad-hoc extraction on platforms where the ledger leak is observed; alert rules can be added on top of the counter later.

extensions/gc/tasks/GarbageCollectorTask.js
delthas: I think 5 min should be widely enough already. It would be pretty rare to have the system failing for 5 minutes but not 10. I'd say no change?

extensions/gc/GarbageCollectorMetrics.js
delthas: Correction to my previous message on this thread — I claimed cloudserver 5xx would hit the non-retryable path, but that is wrong. There is an adapter middleware `isRetryableMiddleware` at `lib/clients/utils.js:12-35`, applied to backbeat clients (`ClientManager.js:161`) and S3 clients (`utils.js:67-70`), that maps smithy classification to bare `err.retryable`:

```js
const retryable = TRANSIENT_ERROR_STATUS_CODES.includes(statusCode) ||  // 500/502/503/504
    NODEJS_TIMEOUT_ERROR_CODES.includes(code || '') ||  // ECONNRESET/ECONNREFUSED/EPIPE/ETIMEDOUT
    TRANSIENT_ERROR_CODES.includes(code || '') ||  // TimeoutError/RequestTimeout/RequestTimeoutException
    THROTTLING_ERROR_CODES.includes(code || '') ||
    statusCode === 429;

error.$retryable = retryable;
error.retryable = retryable;
```

So the corrected buckets:

**`retryable=true`** (~5 min of retries then give up):
- Cloudserver HTTP 5xx (500 / 502 / 503 / 504) that outlasted the retry budget
- HTTP 429 throttling that outlasted the retry budget
- Sustained Node network trouble (ECONNRESET / ECONNREFUSED / EPIPE / ETIMEDOUT)
- Smithy `TimeoutError` / `RequestTimeout` / `RequestTimeoutException`

**`retryable=false`** (immediate give-up, no retry):
- Cloudserver HTTP 4xx that isn't 404 or 412 — `AccessDenied` (after role rotation / bucket policy tightening), `Conflict` (version race on `putMetadata` between `_getMetadata` and `_putMetadata`), `InvalidRequest` / `InvalidArgument`. The middleware sets `err.retryable = false` on these, so BackbeatTask.retry short-circuits.
- Errors that never reach the SDK client, so the middleware never runs. These have `err.retryable === undefined` and BackbeatTask.retry's narrow auto-flag (only `ECONNRESET`/`EPIPE`/`ETIMEDOUT` in `err.code`) does not fire. Concrete instances:
  - `arsenal.errors.InternalError.customizeDescription('Unable to obtain client')` at `GarbageCollectorTask.js:47-51`, `82-93`, `120-128` when `getBackbeatMetadataProxy(accountId)` returns null (vault-derived client unavailable — vault down, bad account, IAM misconfig).
  - Malformed metadata blob at `_getMetadata:68-76` — `InternalError.customizeDescription('error parsing metadata blob')`.
  - Kafka producer send failure in `LifecycleColdStatusArchiveTask._deleteColdObject:66-77` (orphan-cleanup path) — node-rdkafka errors do not go through the SDK middleware.
  - Vault errors on `_getAccountId` unless they surface as bare Node network errors.
- 404 (`ObjNotFound`/`NoSuchBucket`) and 412 (precondition) are already carved out (commit + drop or success no-op) so they do not reach the metric.

So the label distinguishes "backend outage or throttling that persisted through the retry budget" from "we hit an inherent error and did not retry" — different operational signals, different alert thresholds.

**Note:** `err.retryable = true` is manually set in only two other places in the codebase (`ReplicateObject.js:544` for source-stream errors, `FailedCRRConsumer.js:147`). Nothing in the GC or lifecycle paths sets it explicitly — they rely entirely on the middleware + auto-flag.

delthas: Dropped the label in [`ed4e2781`](https://github.com/scality/backbeat/pull/2771/commits/ed4e2781). After discussing with the team: metric-only for now (no dashboard, no alerts), so distinguishing retryable vs non-retryable in a Prom label isn't load-bearing yet. Can add back if follow-up alerting needs it.

extensions/lifecycle/LifecycleMetrics.js
delthas: Dropped, same as the GC side — see [`ed4e2781`](https://github.com/scality/backbeat/pull/2771/commits/ed4e2781).

extensions/gc/GarbageCollectorMetrics.js
delthas: Dropped the dashboard changes in [`ed4e2781`](https://github.com/scality/backbeat/pull/2771/commits/ed4e2781). Metric-only in this PR; dashboards and alerts are a follow-up. The counters (`s3_gc_failed_total`, `s3_lifecycle_failed_total`) can be queried ad-hoc on platforms where the ledger leak is observed.

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -216,6 +230,10 @@ class GarbageCollectorTask extends BackbeatTask {
             }),
             (objMD, next) => {
                 const locations = objMD.getLocation();
+                // Stash the parts on the entry so the outer retry-wrapper's
+                // failure log (in _deleteArchivedSourceData) can surface them
+                // for ops to reclaim manually if the delete never succeeds.
+                entry.setAttribute('target.locations', locations);
```
francoisferrand: seems very dirty; could it be attached to the error instead, or do we really want this to behave the same as the other path?

if we keep it like this, should only be attached in case of error, though...

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -203,7 +203,21 @@ class GarbageCollectorTask extends BackbeatTask {
...
+                const ruleType = entry.getContextAttribute('ruleType');
+                const locations = entry.getAttribute('target')?.locations;
+                log.error('task failed permanently after retries, committing offset', {
+                    method: 'GarbageCollectorTask._executeDeleteData',
+                    error: err.message,
+                    locations,
+                    ...entry.getLogInfo(),
+                });
+                GarbageCollectorMetrics.onGcFailed(log, ruleType,
+                    locations?.[0]?.dataStoreName);
```
francoisferrand: `entry.getAttribute('target').oldLocation` is what is used elsewhere - should be the same normally (we don't usually have data stored in multiple datastores), but may be best to stay consistent?

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -203,7 +203,21 @@ class GarbageCollectorTask extends BackbeatTask {
             actionFunc: cb => this._executeDeleteDataOnce(entry, log, cb),
             shouldRetryFunc: err => err.retryable,
             log,
-        }, done);
+        }, err => {
+            if (err) {
+                const ruleType = entry.getContextAttribute('ruleType');
+                const locations = entry.getAttribute('target')?.locations;
```
francoisferrand:
```suggestion
                const locations = entry.getAttribute('target.locations');
```

extensions/gc/tasks/GarbageCollectorTask.js
delthas: Moved the `setAttribute('target.locations', ...)` into the error branch of the batchDelete callback in [`a0bddb69`](https://github.com/scality/backbeat/pull/2771/commits/a0bddb69), so the entry is only mutated when we're about to fail. Kept the pattern rather than attaching to the error object — mirrors what `LifecycleUpdateTransitionTask` / `LifecycleUpdateExpirationTask` already do when they publish deleteData GC entries (both call `.setAttribute('target.locations', locations)`).

extensions/gc/tasks/GarbageCollectorTask.js
delthas: Good point — `_executeDeleteData` entries do not have `target.oldLocation` (that field is only set by `LifecycleColdStatusArchiveTask._garbageCollectArchivedSource:34` for the archive path). Switched to `entry.getAttribute('source.storageClass')` in [`a0bddb69`](https://github.com/scality/backbeat/pull/2771/commits/a0bddb69), which is the equivalent for deleteData entries and is already used as the batch-delete `StorageClass` in `_executeDeleteDataOnce:155`.

extensions/gc/tasks/GarbageCollectorTask.js
delthas: Applied in [`a0bddb69`](https://github.com/scality/backbeat/pull/2771/commits/a0bddb69).

francoisferrand (approved):

delthas:
Requested @SylvainSenechal in place of Edouard Comtet, who is currently away.

benzekrimaha (approved):
LGTM

PR #2773, opened 2026-07-16
Downgrade node-rdkafka to 3.0.1

### LINKED ISSUE #5549: Consumer returns NULL key/value for legacy MessageSet format (message.format.version < 0.11) since v2.5.0
Description
===========
When consuming from a cluster or topic that stores messages in the legacy MessageSet format (`log.message.format.version` < 0.11, MsgVersion v0..v1), every message is delivered to the application with **NULL key and NULL value**: fetches succeed and offsets advance, but all payloads are NULL. Offsets and timestamps are correct. The consumer also raises errors such as:

```
Unsupported Message(Set) MagicByte 97 at offset 1525458452426752
```

Bisected to 6867b9f9 ("Fetch RPC prework: upgrade to v15", #4584, v2.5.0): since Fetch v12+ the response buffer is parsed as a flexible version, and the legacy MessageSet reader reads Message Key/Value with `rd_kafka_buf_read_kbytes()`, which switches to compact (varint) length parsing for flexible-version buffers. Legacy MessageSets always use fixed-width int32 length prefixes, so the first `0x00` length byte parses as a compact NULL and the reader drifts into the message bytes (the `MagicByte 97` above is a payload byte). On v2.5.x fetches fail/stall instead; from v2.6.0 onwards messages are delivered with NULL key/value as described. Still present on v2.15.0/master.

Found in production through node-rdkafka 3.6.1 (librdkafka 2.12.0), consuming from Apache Kafka 3.9.1 brokers that pin `log.message.format.version=0.10.2.2` for rolling-upgrade compatibility.

How to reproduce
================
Run any broker with `log.message.format.version=0.10.2.2` (e.g. Apache Kafka 3.9 in ZooKeeper mode with `inter.broker.protocol.version=0.10.2.2`), produce keyed messages, consume them: all messages arrive with NULL key/value. The fix PR adds a mock-broker regression test (0187) that reproduces this by capping the ProduceRequest version to <= v2 so the producer writes MsgVersion v1/v0 MessageSets.

Checklist
=========

 - [x] librdkafka version: v2.15.0 (also 2.6.0, 2.8.0, 2.11.1, 2.12.0, 2.14.2; regression since v2.5.0, bisected to 6867b9f9)
 - [x] Apache Kafka version: 3.9.1 with `inter.broker.protocol.version=0.10.2.2`, `log.message.format.version=0.10.2.2`
 - [x] librdkafka client configuration: default, `auto.offset.reset=earliest`
 - [x] Operating system: Linux (x64)
 - [x] Provide logs: `Unsupported Message(Set) MagicByte 97 at offset 1525458452426752` (`_NOT_IMPLEMENTED`) on every fetch
 - [ ] Provide broker log excerpts: no broker-side errors (messages stored intact, verified with `DumpLogSegments`)
 - [ ] Critical issue

delthas:
Downgrades `node-rdkafka` from `^3.6.0` to exactly `3.0.1`.

## Why

librdkafka >= 2.5.0 cannot consume messages stored in the legacy v0/v1 MessageSet format: when the FetchResponse uses a flexible version (v12+), the legacy reader parses the Key/Value length prefixes as compact varints instead of fixed int32, so every message is delivered with **NULL key and value** (plus `Unsupported Message(Set) MagicByte` errors). Reported upstream with a fix: confluentinc/librdkafka#5549 / confluentinc/librdkafka#5550.

S3C federation pins `log.message.format.version=0.10.2.2` on the backbeat-queue brokers, so **all** messages are stored/served in the legacy v1 format: since #2728 (node-rdkafka `^3.6.0` / librdkafka 2.12.0), backbeat cannot process any kafka entry on S3C — every entry fails with `malformed JSON in kafka entry — Cannot read properties of null (reading 'bootstrapId')` and replication stays `PENDING`. This currently blocks the S3C 10 component bump (scality/Federation#6993).

## Why 3.0.1

- Newest node-rdkafka release bundling librdkafka **2.3.0** — the exact librdkafka backbeat ran until 9.3.x (no release ever bundled the last-good 2.4.x; 3.1.0 jumps to the broken 2.5.0).
- 3.0.0 is 2.18.0 (the pre-#2728 lockfile resolution) with only EOL Node versions dropped; 3.0.1 adds one additive OAuthBearer API. Binding-side changes up to 3.6.1 are not relevant to backbeat (TS typings, cooperative rebalance — unused, we pin eager `range,roundrobin` — Node 23/24 support).
- All app-side fixes from #2728 (bootstrap chained-setTimeout, `partition.assignment.strategy` pin) are kept.
- Verified against a Kafka 3.9.1 broker configured with `log.message.format.version=0.10.2.2`: 3.0.1 consumes legacy v1 topics with intact key/value; 3.1.0+ does not.

**The pin must not be raised past 3.0.1** until node-rdkafka bundles a librdkafka release containing the fix from confluentinc/librdkafka#5550. Note: Node >= 24 support will require node-rdkafka >= 3.6.1 (nan bump), so the re-bump must precede or accompany any Node 24 migration.

Issue: BB-806

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "minimatch": "^10.0.1",
     "mongodb": "^6.11.0",
     "node-forge": "^1.3.1",
-    "node-rdkafka": "^3.6.0",
+    "node-rdkafka": "3.0.1",
     "node-rdkafka-prometheus": "^1.0.0",
     "node-schedule": "^2.1.1",
     "node-zookeeper-client": "^1.1.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "9.4.0",
+  "version": "9.4.1",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -9147,10 +9147,10 @@ nan@^2.14.0, nan@^2.18.0, nan@^2.3.2:
   resolved "https://registry.yarnpkg.com/nan/-/nan-2.23.1.tgz#6f86a31dd87e3d1eb77512bf4b9e14c8aded3975"
   integrity sha512-r7bBUGKzlqk8oPBDYxt6Z0aEdF1G1rwlMcLk8LCOMbOzf0mG+JUfUzG4fIMWwHWP0iyaLWEQZJmtB7nOHEm/qw==
 
-nan@^2.24.0:
-  version "2.26.2"
-  resolved "https://registry.yarnpkg.com/nan/-/nan-2.26.2.tgz#2e5e25764224c737b9897790b57c3294d4dcee9c"
-  integrity sha512-0tTvBTYkt3tdGw22nrAy50x7gpbGCCFH3AFcyS5WiUu7Eu4vWlri1woE6qHBSfy11vksDqkiwjOnlR7WV8G1Hw==
+nan@^2.19.0:
+  version "2.28.0"
+  resolved "https://registry.yarnpkg.com/nan/-/nan-2.28.0.tgz#126717fd359d5a03d3edf7c44e6ce9b707fb57f5"
+  integrity sha512-fTsDz99OTq2sVePhGdp4qQhggZFtKr64ZNVyVajRKtMOkJxYekplBh577PiJB12v/D3s2E5cGtOI45LWp6rnLQ==
 
 napi-macros@~2.0.0:
   version "2.0.0"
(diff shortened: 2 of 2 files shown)


delthas:
Requested @francoisferrand in place of Maha Benzekri, who is currently on PTO.

francoisferrand (approved):
(note: description is a bit hard to follow, kind of mixing librdkafka version with node-rdkafka versions)

leif-scality (approved):

leif-scality:
Can you bump the package.json so we can create a new preview directly

PR #2774, opened 2026-07-20
Gatekeep expect continue 100 header on 128kb object size

SylvainSenechal:
Issue: BB-793

This is an optimization

### DIFF
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -36,7 +36,7 @@ const RoleCredentials = require('../../../lib/credentials/RoleCredentials');
 const { metricsExtension, metricsTypeQueued, metricsTypeCompleted, replicationStages } = require('../constants');
 
 const ObjectQueueEntry = require('../../../lib/models/ObjectQueueEntry');
-const { authTypeAssumeRole } = require('../../../lib/constants');
+const { authTypeAssumeRole, replicationExpectContinueThreshold } = require('../../../lib/constants');
 
 const errorAlreadyCompleted = {};
 
@@ -599,10 +599,14 @@ class ReplicateObject extends BackbeatTask {
                     RequestUids: log.getSerializedUids(),
                     VersionId: sourceEntry.getEncodedVersionId(),
                 });
-                attachExpectContinueMiddleware(putCommand, this.backbeatDest.config?.requestHandler);
                 addContentLengthMiddleware(
                     putCommand,
                     response.ContentLength,
+                );
+                attachExpectContinueMiddleware(
+                    putCommand,
+                    this.backbeatDest.config?.requestHandler,
+                    replicationExpectContinueThreshold,
                 );
                 const writeStartTime = Date.now();
                 return this.backbeatDest.send(putCommand, { abortSignal: abortController.signal })
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -88,6 +88,9 @@ const constants = {
         // controls the max number of messages to queue for processing.
         MAX_QUEUED_DEFAULT: 1000,
     },
+    // Objects below this size skip the Expect: 100-continue handshake as the
+    // extra round-trip is a bigger overhead than the bandwidth saved
+    replicationExpectContinueThreshold: 1024 * 1024, // 1 mb
 };
 
 module.exports = constants;
--- a/tests/unit/replication/ReplicateObject.spec.js
+++ b/tests/unit/replication/ReplicateObject.spec.js
@@ -14,6 +14,8 @@ const {
     MicroVersionIdAlreadyStoredException,
 } = require('@scality/cloudserverclient');
 
+const { HttpRequest } = require('@smithy/protocol-http');
+const { replicationExpectContinueThreshold } = require('../../../lib/constants');
 const { replicationEntry } = require('../../utils/kafkaEntries');
 const fakeLogger = require('../../utils/fakeLogger');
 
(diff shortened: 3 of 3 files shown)


lib/constants.js
```diff
@@ -88,6 +88,9 @@ const constants = {
         // controls the max number of messages to queue for processing.
         MAX_QUEUED_DEFAULT: 1000,
     },
+    // Objects below this size skip the Expect: 100-continue handshake as the
+    // extra round-trip is a bigger overhead than the bandwith saved
+    replicationExpectContinueThreshold: 128 * 1024, // 128 kb
```
SylvainSenechal: I put 128kb, imo we don't really need to fight over this value, the most important is to avoid a round trip request on empty objects and small object (which im pretty sure with some clients represent a very big part of their objects already), which 128kb will be perfect for.

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal, line 1: Didnt add any tests 👀 
I asked claude to suggests tests, it was a big mess of mocks and 90% of setup, all that to assert a boolean, at this point its just not really good tests and not worth it

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -599,11 +599,15 @@ class ReplicateObject extends BackbeatTask {
...
                 });
-                attachExpectContinueMiddleware(putCommand, this.backbeatDest.config?.requestHandler);
                 addContentLengthMiddleware(
                     putCommand,
                     response.ContentLength,
                 );
+                attachExpectContinueMiddleware(
+                    putCommand,
+                    this.backbeatDest.config?.requestHandler,
+                    response.ContentLength >= replicationExpectContinueThreshold,
```
maeldonn: Could be simplified: https://github.com/scality/cloudserverclient/pull/25/changes#diff-8312f488094debc6dd0bda2b6fa72c33835e56b27b5d8b733adc1790e2ec2344R162

lib/constants.js
maeldonn: Avoid hardcoding or using a random value here. This should be computed dynamically based on actual production metrics (bandwidth, latency, system load) so it reflects real runtime conditions. @francoisferrand did something similar for another dev.

extensions/replication/tasks/ReplicateObject.js
maeldonn: This is already tested in `cloudserverclient`, so I don't think we need an additional test here. The existing coverage handles this case well.

lib/constants.js
SylvainSenechal: We can wait for more input here, we have a few possibilities : 
- Try to come up with a number based on more evidence
- Completely close this pr, marked this whole thing as premature/useless optimization
- Make this constant configurable through CR (don't like this one, in practice it will never be configured)

extensions/replication/tasks/ReplicateObject.js
benzekrimaha: AFAIK cloudserverclient tests the middleware itself, not that Backbeat passes the expected threshold. We can use an additional test below the threshold and one at the threshold. The current suite would remain green if this third argument were accidentally removed.

lib/constants.js
benzekrimaha: Every successful replication above this value now pays an additional round trip.... Do we have any measurements to select this threshold: latency, bandwidth, object-size distribution and rejection rate. Otherwise this could be a premature optimization

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -599,11 +599,15 @@ class ReplicateObject extends BackbeatTask {
...
                 });
-                attachExpectContinueMiddleware(putCommand, this.backbeatDest.config?.requestHandler);
                 addContentLengthMiddleware(
                     putCommand,
                     response.ContentLength,
                 );
+                attachExpectContinueMiddleware(
+                    putCommand,
+                    this.backbeatDest.config?.requestHandler,
+                    replicationExpectContinueThreshold,
```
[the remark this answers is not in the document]
SylvainSenechal: wrong, its available in 1.0.9 🤷‍♂️

lib/constants.js
SylvainSenechal: No before this pr, we were doing that additional round trip on all requests
Now we do it only for objects with more than 128kb of data, so its already an improvement.

 I'm updating it to 1mb because i think its more reasonnable but its still based on feelings.

As i wrote somewhere else, I doubt we should spend that much time on this, the whole point is to avoid the round trip on metadata-only objects or super small objects which should already be a very nice improvement

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: I added tests for the threshold

lib/constants.js
maeldonn: We can compute a real cutoff instead of guessing:

```
threshold = (RTT × bandwidth) / P(reject)
```

- **RTT** — round-trip to the destination; the cost of one `Expect: 100-continue` handshake.
- **bandwidth** — upload throughput to the destination. `RTT × bandwidth` = bytes we could send in that one round-trip (the break-even if every PUT were rejected).
- **P(reject)** — fraction of PUTs the destination declines. The handshake only pays off on a reject, so a lower rate pushes the threshold up.

P(reject) depends heavily on topology: near zero for one-way replication, but high for bidirectional/cascade CRR, where the destination declines PUTs for data it already has (`VersionIdCollisionException`) precisely to avoid re-uploading. That's the case the handshake is built for, and it pushes the threshold down.

We don't need exact numbers: RTT/bandwidth vary per customer, but a rough estimate lands in the right range. Being a bit low is safe (just a few extra handshakes on mid-size objects); being too high risks wasting a full upload on a decline. So 1MB is a sound conservative default.

To do better, this should be configurable for per-customer tuning, and a very large value effectively disables the handshake for one-way setups that don't need it (new backbeat + old cloudserver).

SylvainSenechal: Ok, still think its not a good idea to make it configurable

Overall agree with the formula, and so yeah 1mb is probably decent.

Anyways keep in mind, any value greater than will already be better than our current solution which always sends this header

So yeah keeping 1mb

tests/unit/replication/ReplicateObject.spec.js
```diff
@@ -14,6 +14,8 @@ const {
     MicroVersionIdAlreadyStoredException,
 } = require('@scality/cloudserverclient');
 
+const { HttpRequest } = require('@smithy/protocol-http');
```
benzekrimaha, line 17: Better added to `devDependencies`

benzekrimaha (approved):
nothing blocking on my end, approving

SylvainSenechal:
Created this follow up https://scality.atlassian.net/jira/software/c/projects/OS/boards/268?selectedIssue=BB-827

lib/constants.js
maeldonn: @SylvainSenechal Could you create a follow-up for that ?

maeldonn (approved):

PR #2775, opened 2026-07-20
handle orphans on processQueueEntryRetryFull

SylvainSenechal:
Issue: BB-802

This fixes a small risk that was raised during crr cascade code reviews : When calling putMetadata after having written data, if putMetadata errors out, there is a situation where we don't delete the orphan data.
This risk is not even fully related to crr cascade, as the bug could also happen before.

### DIFF
--- a/extensions/replication/tasks/ReplicateObject.js
+++ b/extensions/replication/tasks/ReplicateObject.js
@@ -1018,7 +1018,17 @@ class ReplicateObject extends BackbeatTask {
             next => this._getAndPutData(sourceEntry, destEntry, log, next),
             (destLocations, conflict, next) => {
                 destEntry.setLocation(destLocations);
-                return this._putMetadata(destEntry, false, conflict, log, next);
+                return this._putMetadata(destEntry, false, conflict, log, err => {
+                    if (err) {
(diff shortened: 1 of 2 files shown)


extensions/replication/tasks/ReplicateObject.js
```diff
@@ -1022,7 +1022,12 @@ class ReplicateObject extends BackbeatTask {
             next => this._getAndPutData(sourceEntry, destEntry, log, next),
             (destLocations, conflict, next) => {
                 destEntry.setLocation(destLocations);
-                return this._putMetadata(destEntry, false, conflict, log, next);
+                return this._putMetadata(destEntry, false, conflict, log, err => {
+                    if (err) {
+                        return this._deleteOrphans(destEntry, destLocations, log, () => next(err));
```
maeldonn, line 1028: Should we add a log ?

maeldonn (approved):

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: yay

delthas (approved):

PR #2776, opened 2026-07-20
Add async/await migration guidance for Claude and code reviews

delthas:
Encode the "migrate when you touch" rule from the [Scality async/await migration guide](https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration) into the repo's Claude guidance:

- `CLAUDE.md`: new **Async code style** section — new code uses async/await, significantly-changed functions get migrated (in a dedicated commit when large), `util.promisify`/`util.callbackify` at callback-API boundaries, and the main pitfalls (`return await`, no async `forEach`, never invoke a callback inside `try/catch`). Also fixes the repo summary line that described callback style as the convention.
- `.github/copilot-instructions.md`: new **Async/await migration suggestions** section telling the CI review when to suggest a migration (PR substantially reworks a callback-based function, new callback-style functions) and when to stay silent (minor edits, callback shapes imposed by external APIs such as node-rdkafka event handlers, test-only changes), capped at one suggestion per function. The existing verification row gains the same pitfall checks.

Issue: BB-805

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -9,7 +9,7 @@ When reviewing a PR, analyze the changes against these criteria:
 | Area | What to check |
 |------|---------------|
 | Async error handling | Uncaught promise rejections, missing error callbacks, swallowed errors in streams. Double callbacks in try/catch blocks (callback called in try then again in catch) |
-| Async/await usage | New or modified code should use async/await instead of callbacks when possible. When code is migrated from callbacks to async/await, verify: no leftover callback or next params, no mixed callback + promise patterns, proper try/catch around awaited calls, errors are re-thrown or handled (not silently swallowed). Watch for the anti-pattern: `try { cb(); } catch(err) { cb(err); }` where an exception after the first `cb()` triggers a second call |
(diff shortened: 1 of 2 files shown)


maeldonn (approved):

.github/copilot-instructions.md
```diff
@@ -20,3 +20,13 @@ When reviewing a PR, analyze the changes against these criteria:
 | Extension architecture | Changes respect the pluggable extension pattern, no cross-extension coupling |
 | Security | Command injection, prototype pollution, unsafe deserialization, credential exposure in config/env vars, OWASP-relevant issues for Node.js |
 | Breaking changes | Anything that changes public APIs, Kafka message formats, inter-service contracts, or oplog/change stream projections |
+
+## Async/await migration suggestions
```
SylvainSenechal, line 24: I have an ongoing migration for one backbeat services, and there is one thing we need to be extra careful about, its the class inheritence.

For example, we have 
class MultipleBackendTask extends ReplicateObject

Updating the function signature of ReplicateObject could have impact on MultiplebackendTask

SylvainSenechal (approved):

.github/copilot-instructions.md
delthas: Good point — made this explicit (amended): the suggestions section now says a migration suggestion on a class method must cover the whole hierarchy (overrides + `super` calls, using your `MultipleBackendTask extends ReplicateObject` example), and to flag partial migrations where a base method became async but overrides still use callbacks. Added the matching rule to CLAUDE.md's async style section.

PR #2777, opened 2026-07-20
Guard LogConsumer against offsetsStore failures during rebalance

delthas:
The queue populator's Kafka `LogConsumer.storeOffsets()` called `offsetsStore()` unguarded. When a rebalance or shutdown revokes partitions between batch processing and offset storage, the call throws `LibrdKafkaError: Local: Erroneous state`, which escapes as an unhandled exception and kills the process. The pod exits without a clean consumer-group leave, so its session lingers and the delayed rebalance hits surviving pods at arbitrary points — seeding the drain stalls that freeze oplog consumption (replication entries, bucket notifications). Observed twice in one Zenko end2end run ([run 29589236290 attempt 1](https://github.com/scality/Zenko/actions/runs/29589236290/attempts/1)).

This ports the protection BackbeatConsumer has had since BB-758 (#2731) to the LogConsumer: skip the store when the consumer is not connected, and log instead of crashing if the call still throws. Offsets not stored are simply re-consumed by whichever consumer owns the partitions next — the populator pipeline is at-least-once and tolerates duplicates. In both cases the processed batch is released (`_topicPartition` cleared) so consumption never wedges on a batch whose offsets can no longer be stored, and `_pendingCommit` is only raised when a store actually happened.

Note that #2731 was the defense-in-depth layer of BB-758, behind the structural rebalance fixes #2739 (BB-776) and #2740 (BB-777); the LogConsumer's structural analog (`_drainAndUnassign`/`_onOffsetCommit` drain deadlock) and the replication status consumer's missing `fromOffset` are intentionally out of scope here, tracked separately.

Issue: BB-823

### DIFF
--- a/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
+++ b/lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
@@ -314,14 +314,36 @@ class LogConsumer extends EventEmitter {
             });
             return;
         }
-        this._consumer.offsetsStore(this._topicPartition);
+        const topicPartition = this._topicPartition;
+        this._topicPartition = null;
+        // only store offsets while connected; skipped offsets are
+        // re-consumed by the next partition owner.
+        if (!this._consumer.isConnected()) {
+            this._log.info('Skipping offsets store', {
+                method: 'LogConsumer.storeOffsets',
+                reason: 'consumer not connected',
+                topicPartition,
+                consumerGroupId: this._consumerGroupId,
+            });
+            return;
+        }
+        try {
+            this._consumer.offsetsStore(topicPartition);
+        } catch (e) {
+            this._log.error('offsetsStore failed', {
+                method: 'LogConsumer.storeOffsets',
+                error: e.toString(),
+                topicPartition,
+                consumerGroupId: this._consumerGroupId,
+            });
+            return;
+        }
         this._log.info('Offsets stored', {
             method: 'LogConsumer.storeOffsets',
             consumerGroupId: this._consumerGroupId,
-            topicPartition: this._topicPartition,
+            topicPartition,
         });
         this._pendingCommit = true;
-        this._topicPartition = null;
     }
 
     _hasUnprocessedMessages() {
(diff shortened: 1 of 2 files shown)


lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
```diff
@@ -314,14 +314,39 @@ class LogConsumer extends EventEmitter {
...
             return;
         }
-        this._consumer.offsetsStore(this._topicPartition);
+        const topicPartition = this._topicPartition;
+        this._topicPartition = null;
+        // ensure consumer is active when calling offsetsStore() on
+        // it, to avoid raising an exception (invalid state): the
+        // partitions may have been revoked while the batch was being
+        // processed. Offsets not stored here are simply re-consumed
+        // by whichever consumer owns the partitions next (BB-758)
```
maeldonn: nit:
```suggestion
        // only store offsets while connected; skipped offsets are
        // re-consumed by the next partition owner.
```

maeldonn (approved):

SylvainSenechal (approved):
@delthas Seems like a recurrent problem for multiple backbeat services. 

Have you run claude to check if this also happens in the other backbeat consumers ? Might be worth it to create a ticket to verify it

lib/queuePopulator/KafkaLogConsumer/LogConsumer.js
delthas: Applied in 07249c22, thanks.

delthas:
@SylvainSenechal audited the whole repo: `BackbeatConsumer` and this `LogConsumer` are the only two classes owning a raw node-rdkafka consumer — every backbeat service (replication, lifecycle, notification, GC, ingestion) consumes through `BackbeatConsumer`, which has had the equivalent guard since #2731. So no other consumer needs this fix.

The two related gaps found by the same investigation are already ticketed: the LogConsumer drain deadlock (the ~5-min freeze this crash seeds) is BB-825, and the `fromOffset`/`auto.offset.reset` audit across extension consumers is BB-826.

One residual bare call site exists in the bootstrap/canary path (`BackbeatConsumer.js:1026`: `offsetsStore()` + `commit()` during bootstrap) — only reachable if a rebalance lands inside the short bootstrap window, so noting it here as known/low-risk rather than filing a ticket.

PR #2779, opened 2026-07-23
BB-828 Restore Kerberos GSSAPI support in the image

anurag4DSB:
The image convergence dropped libsasl2-dev from the build stage and the Kerberos runtime packages from the final image, so node-rdkafka compiles librdkafka without sasl_gssapi and Kerberos notification destinations fail at connect. Restore the packages and add CI checks on both images so GSSAPI support cannot silently regress again.

[We added some tests in integration for this](https://github.com/scality/Integration/pull/1459)
Passing integration build with this new backbeat: https://github.com/scality/Integration/actions/runs/30070161868/job/89409177612

Regression introduced in: https://github.com/scality/backbeat/pull/2600

### DIFF
--- a/.github/scripts/check_image_kerberos_support.sh
+++ b/.github/scripts/check_image_kerberos_support.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+# Image test: verify a backbeat image is Kerberos-capable.
+#
+# Checks inside the given image:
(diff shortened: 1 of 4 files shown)


anurag4DSB:
ping

BourgoisMickael (approved):

jonathan-gramain (approved):

PR #2780, opened 2026-07-23
Refactor replicate object with async await

SylvainSenechal:
Issue: BB-803

**Review hints** : 
- The diff is hard to read on github because indentations changes are marked as a diff, in my code editor, when looking at the diff, the display is different and it is **significantly** easier to review  
- Reviewable commit by commit

Imo this refactor is super overdue and not so complicated to do.

Refactor replicate object with async await

- Also updated the async parallel util function into async await
- Also made minimal mandatory updates to MultipleBackend class, as this class inherits ReplicateObject


lib/util/runTasksWithConcurrency.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * Runs an async task for each item in a collection, up to `limit` at a time.
+ * On error, no new tasks are started, but already-running tasks are awaited
+ * before the function resolves. Returns [firstError, results] so that callers
```
SylvainSenechal, line 4: Return the first error only : Keeping same behavior as before with mapLimitWaitPendingIfError.
 
Could've returned all errors as a list but not needed now and prefer to keep existing behavior.
Cannot aggregate errors neither, as we need to keep the original error to check its properties to determine if its retryable or not

lib/util/mapLimitWaitPendingIfError.js
SylvainSenechal, line 1: refactored with async await, in the file runTaskWithConcurrency

extensions/replication/tasks/MultipleBackendTask.js
SylvainSenechal, line 1: Migrations here aren't the main point of this pr, but were kinda forced as MultipleBackend extends ReplicateObject.

I did the minimum required migration on this file

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -911,85 +896,57 @@ class ReplicateObject extends BackbeatTask {
...
+                : [[], undefined];
+
+            destEntry.setLocation(destLocations);
+            try {
+                await this._putMetadata(destEntry, mdOnly, conflict, log);
+            } catch (err) {
+                await this._deleteOrphans(destEntry, destLocations, log);
+                throw err;
+            }
+            return this._handleReplicationOutcome(null, sourceEntry, destEntry, kafkaEntry, log, done);
```
benzekrimaha: `_handleReplicationOutcome(null, …, done)` is called from inside the `try`: if it throws after calling `done`, the `catch` calls it again and `done` fires twice. We can compute the error in the try/catch and call the handler exactly once after it. (same at line 909 and in `_processQueueEntryRetryFull`)

benzekrimaha (changes requested):
Good refactor overall, but one real issue with the outcome handler being called from inside the `try`.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -552,152 +545,149 @@ class ReplicateObject extends BackbeatTask {
...
+                if (!sourceStreamAborted && !destRequestAborted) {
                     abortController.abort();
                     destRequestAborted = true;
                 }
-                // eslint-disable-next-line no-param-reassign
-                err.origin = 'source';
                 if (err.$metadata?.httpStatusCode === 404) {
-                    return doneOnce(err);
+                    // eslint-disable-next-line no-param-reassign
+                    err.origin = 'source';
```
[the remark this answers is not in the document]
SylvainSenechal: this is intentional

SylvainSenechal:
Sorry I took some time to respond, you can review again next week @benzekrimaha @maeldonn

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -1017,42 +975,41 @@ class ReplicateObject extends BackbeatTask {
...
-                    }
-                    return next();
+    async _processQueueEntryRetryFull(sourceEntry, destEntry, kafkaEntry, log) {
+        let destLocations = null;
+        try {
+            let conflict;
+            [destLocations, conflict] = await this._getAndPutData(sourceEntry, destEntry, log);
+            destEntry.setLocation(destLocations);
+            await this._putMetadata(destEntry, false, conflict, log);
+            return this._handleReplicationOutcome(null, sourceEntry, destEntry, kafkaEntry, log);
```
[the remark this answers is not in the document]
SylvainSenechal: using option 2

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -1186,7 +1187,7 @@ class MultipleBackendTask extends ReplicateObject {
     _setupClients(entry, log, cb) {
         // Sets up source clients using the role from the replication
         // configuration if the authentication type is as such.
-        return this._setupRoles(entry, log, cb);
+        return this._setupRoles(entry, log).then(() => cb()).catch(cb);
```
DarkIsDude: This is always weird to read.

```suggestion
        return this._setupRoles(entry, log).then((_result) => cb()).catch(cb);
```

`_` is a convention in JS for unused parameter. The linter should not complain, what do you think ? I guess this can be an error if the callback trigger an error ?

lib/tasks/BackbeatTask.js
```diff
@@ -121,6 +121,16 @@ class BackbeatTask {
         };
         actionFunc(_handleRes, nbRetries);
     }
+
+    retryAsync(args) {
```
DarkIsDude: We usually prefer to keep the function `retry` and the last argument (`done`) is optional ?

lib/util/runTasksWithConcurrency.js
DarkIsDude, line 1: Did you get a try to some library to do that ?

lib/util/runTasksWithConcurrency.js
DarkIsDude: `but already-running tasks are awaited` what happens if an error is triggered here ? Can make sense to stop new flows on error, but for the current one we can miss an error ?

lib/util/runTasksWithConcurrency.js
```diff
@@ -0,0 +1,49 @@
...
+async function runTasksWithConcurrency(task, limit, coll) {
+    if (coll.length === 0) {
+        return [null, []];
+    }
+    const results = new Array(coll.length);
+    let nextIdx = 0;
+    const errors = [];
+
+    const worker = async () => {
+        while (nextIdx < coll.length) {
```
DarkIsDude, line 31: Let say 
- `A` `while (nextIdx < coll.length) {`
- `B` `const idx = nextIdx++;`

With 2 workers you can have A1 B1 A1 A2 B1 B2 and then trigger an error as coll is only size 2 ?

lib/util/runTasksWithConcurrency.js
DarkIsDude: Ok I saw your comment. I'm not sure it's the right approach and we should return all errors ?

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -911,85 +896,62 @@ class ReplicateObject extends BackbeatTask {
...
-                    return next();
-                });
-            },
-        ], err => this._handleReplicationOutcome(
-            err, sourceEntry, destEntry, kafkaEntry, log, done));
+            } catch (err) {
+                outcome = err;
+            }
+            const result = await this._handleReplicationOutcome(outcome, sourceEntry, destEntry, kafkaEntry, log);
+            return result === null ? done() : done(null, result);
```
DarkIsDude: This is what we don't want to do in async migration. done should not be called inside the async function 🙏. We did a skill to ease migration (and a confluence page) where we explain the why

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -1196,18 +1197,17 @@ class MultipleBackendTask extends ReplicateObject {
 
         return async.waterfall([
             next => this._setupClients(sourceEntry, log, next),
-            next => this._refreshSourceEntry(sourceEntry, log, (err, res) => {
-                if (err && err.name === 'ObjNotFound' &&
-                    sourceEntry.getReplicationIsNFS() && !sourceEntry.getIsDeleteMarker()) {
+            next => this._refreshSourceEntry(sourceEntry, log)
```
DarkIsDude, line 1200: This will runs the next waterfall stage synchronously, and that stage is not async — metricsHandler.rpo() can all throw synchronously for example. A throw there lands in .catch, which calls next again; async.waterfall's onlyOnce then throws "Callback was already called" as an uncaught exception in the queue processor, masking the real error. Use the two-argument then so the success path can't feed the rejection handler

```
next => this._refreshSourceEntry(sourceEntry, log)
                .then(res => next(null, res), err => {
```

extensions/replication/tasks/MultipleBackendTask.js
DarkIsDude: And I think we have the same issue here with `.catch`

DarkIsDude (changes requested):
Please try to create dedicated commit for prettier format as mention in the best practice of the migration 🙏. You can also use "Hide whitespace" on Github to ease diff if you want.

Image

CI is red, maybe you should have a look ?

lib/util/runTasksWithConcurrency.js
SylvainSenechal: I asked ai to create this function, although I also looked online at how people do it and the solutions were very similar.
Definitly there are libraries for this, but taking a look rapidly I only small non popular one, can't find a proper "do parallel stuff with js" thats very popular and battle tested. 
Here we also have some specific behavior, about continuing ongoing task when one error is found, so not sure the available libraries would offer that.

I think its fine to not use a lib

lib/util/runTasksWithConcurrency.js
SylvainSenechal: Mhh its replicating initial pre migration behavior, I think its fine anyways as replicateObject will try to delete all orphans as soon as a single error is found

Although maybe there is a question raised about reusability of this function, maybe its built to be too custom for replicateObject and should be more generic (return array of errors) ?

lib/util/runTasksWithConcurrency.js
SylvainSenechal: no I think this is fine, its js single threaded code even if it kinda looks weird, this wont run at the same time 🤔

lib/tasks/BackbeatTask.js
SylvainSenechal: Yeah im moving this at the top of retry

extensions/replication/tasks/MultipleBackendTask.js
SylvainSenechal: aiya its more complicated than i thought to make these refactoring

extensions/replication/tasks/MultipleBackendTask.js
SylvainSenechal: yeah its nicer to have _result, I also see i messed up some stuff around the catches, its tricky

edit : no : Error:   1190:51  error    '_result' is defined but never used                     no-unused-vars

SylvainSenechal:
> Please try to create dedicated commit for prettier format as mention in the best practice of the migration 🙏. You can also use "Hide whitespace" on Github to ease diff if you want.
> 
> Image
> CI is red, maybe you should have a look ?

I didnt run the linter though, the diff is just re indentation mostly 🤔 
The failure were unrelated, its rerunning but it was fine the other time so should be good here

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -81,89 +82,89 @@ class MultipleBackendTask extends ReplicateObject {
...
+            log.error('role in replication entry for source does not ' +
+            'match role in bucket replication configuration', {
+                method: 'MultipleBackendTask._setupRolesOnce',
+                entry: entry.getLogInfo(),
+                entryRole: entryRoles[0],
+                bucketRole: roles[0],
+            });
+            throw errors.BadRole;
+        }
+        return [this.sourceRole, entryRoles[1]];
```
benzekrimaha: `entryRoles[1]` is undefined for the single-role external-backend case , the value is discarded by _setupClients, but returning a knowingly-undefined slot is confusing; a comment or returning only what's meaningful would help.

SylvainSenechal: Thanks for the reviews, i didnt expect to mess it up so much -_-

SylvainSenechal: I'm gonna remove it, just like it was before as it's not used

lib/tasks/BackbeatTask.js
```diff
@@ -119,8 +128,9 @@ class BackbeatTask {
             }
             return doneOnce(...args);
         };
-        actionFunc(_handleRes, nbRetries);
+        return actionFunc(_handleRes, nbRetries);
```
benzekrimaha: nothing consumes that return value, why change it?

lib/tasks/BackbeatTask.js
```diff
@@ -119,8 +128,9 @@ class BackbeatTask {
...
     }
+
```
benzekrimaha:
```suggestion
```

lib/tasks/BackbeatTask.js
```diff
@@ -30,6 +30,15 @@ class BackbeatTask {
     }
 
     retry(args, done) {
+        if (!done) {
+            return new Promise((resolve, reject) =>
+                this.retry(args, (err, ...rest) => {
+                    if (err) {
+                        return reject(err);
+                    }
+                    return resolve(rest.length <= 1 ? rest[0] : rest);
```
benzekrimaha: the promise overload resolves a *variadic* shape: a bare value if the callback got one result arg, an array if it got several. `_setupRoles` only works because the bridge passes a single array arg ,a future action calling `done(null, a, b)` resolves the exact same shape. We should either resolve a single well-defined value, or fully convert `retry` to async; at minimum we can document the shape.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -103,39 +103,40 @@ class ReplicateObject extends BackbeatTask {
...
 
-    _setupRoles(entry, log, cb) {
-        this.retry({
+    async _setupRoles(entry, log) {
+        return await this.retry({
             actionDesc: 'get bucket replication configuration',
             logFields: { entry: entry.getLogInfo() },
-            actionFunc: done => this._setupRolesOnce(entry, log, done),
+            actionFunc: done => this._setupRolesOnce(entry, log)
+                .then(roles => done(null, roles), done),
```
benzekrimaha, line 111: all these `done => this._xxxOnce(...).then(r => done(null, r), done)` bridges keep `retry` callback-based inside an otherwise async class , fine as a transition, but can we please create a followup ticket to convert `retry` internals so we don't keep both paradigms. (applies to all similar cases)

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -552,152 +545,149 @@ class ReplicateObject extends BackbeatTask {
...
+                if (!sourceStreamAborted && !destRequestAborted) {
                     abortController.abort();
                     destRequestAborted = true;
                 }
-                // eslint-disable-next-line no-param-reassign
-                err.origin = 'source';
                 if (err.$metadata?.httpStatusCode === 404) {
-                    return doneOnce(err);
+                    // eslint-disable-next-line no-param-reassign
+                    err.origin = 'source';
```
benzekrimaha: behavior change: the old code rejected `errors.ObjNotFound` *without* `origin`, so `_handleReplicationOutcome` took the “target object not found, retrying with full data write” path; setting `origin = 'source'` here makes it take the skip path instead. Skipping looks correct for a source-side 404 mid-stream, but is the change intended? IMO it deserves a test.

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -552,152 +545,149 @@ class ReplicateObject extends BackbeatTask {
...
+                    // eslint-disable-next-line no-param-reassign
+                    err.ObjNotFound = true;
+                    // eslint-disable-next-line no-param-reassign
+                    err.name = 'ObjNotFound';
```
benzekrimaha: why mutate the SDK error instead of rejecting `errors.ObjNotFound` like before? this satisfies the `err.name === 'ObjNotFound'` checks but loses `err.is.*`, so any future `err.is.ObjNotFound` check will silently miss it. arsenal errors is the canonical module...

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -911,85 +896,63 @@ class ReplicateObject extends BackbeatTask {
...
+            }
+        } catch (err) {
+            return this._handleReplicationOutcome(err, sourceEntry, destEntry, kafkaEntry, log);
+        }
+        return this._handleReplicationOutcome(null, sourceEntry, destEntry, kafkaEntry, log);
+    }
+
+    processQueueEntry(_sourceEntry, kafkaEntry, done) {
+        this._processQueueEntry(_sourceEntry, kafkaEntry).then(
+            result => result === null ? done() : done(null, result),
```
benzekrimaha, line 949: does the queue processor actually distinguish `done()` from `done(null, undefined)`? if not, the `null` sentinel and this ternary can go away I think

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -351,140 +276,209 @@ class ReplicateObject extends BackbeatTask {
...
-                        peer = this.destBackbeatHost;
-                        if (this.destConfig.auth.vault) {
-                            const { host, port } = this.destConfig.auth.vault;
-                            if (host) {
-                                // no proxy is used, log the vault host/port
-                                peer = { host, port };
-                            }
-                        }
+        const lookupAccountAttributes = promisify(
+            this.s3destCredentials.lookupAccountAttributes.bind(this.s3destCredentials));
```
benzekrimaha, line 363: nit: the promisified helper is rebuilt on every call we can hoist it (module level or constructor).

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -81,89 +82,88 @@ class MultipleBackendTask extends ReplicateObject {
...
-                return cb(errors.InternalError.
-                    customizeDescription('error parsing metadata blob'));
-            }
-            const refreshedEntry = new ObjectQueueEntry(sourceEntry.getBucket(),
-                sourceEntry.getObjectVersionedKey(), parsedEntry.result)
-                .setReplicationBackend(sourceEntry.getReplicationBackend());
-            return cb(null, refreshedEntry);
-        });
+        const getMetadata = promisify(
+            this.backbeatSourceProxy.getMetadata.bind(this.backbeatSourceProxy));
```
benzekrimaha: same (hoist the promisified helper)

extensions/replication/tasks/MultipleBackendTask.js
```diff
@@ -1196,18 +1196,16 @@ class MultipleBackendTask extends ReplicateObject {
 
         return async.waterfall([
```
benzekrimaha, line 1198: still `async.waterfall` with promise shims (`_setupClients`, `_refreshSourceEntry` bridges) while the parent class is fully async I think it is ok as “minimal mandatory updates”, but we should probably create a followup ticket to finish the migration

lib/util/runTasksWithConcurrency.js
```diff
@@ -0,0 +1,49 @@
...
+ * requests, so that we have the full list of orphans to delete
+ * afterwards without missing the in-progress requests at the time the
+ * error occurs.
+ *
+ * @param {Function} task       - async function (item) => result
+ * @param {number}   limit     - maximum number of concurrent tasks
+ * @param {Array}    coll      - collection to iterate over
+ * @return {Promise<[Error|null, Array]>} - always resolves, never rejects
+ */
+async function runTasksWithConcurrency(task, limit, coll) {
```
benzekrimaha: two things:
- why swap the parameter order? both `async.mapLimit` and the old `mapLimitWaitPendingIfError` use `(coll, limit, iteratee)` ,keeping the conventional order would avoid silent argument swaps.
- a never-rejecting `[err, results]` tuple is easy to misuse: a caller who forgets to check `[0]` silently ignores errors... The partial-results need is real, so OK if kept, but the doc comment should be kept prominent

lib/util/runTasksWithConcurrency.js
```diff
@@ -0,0 +1,49 @@
...
+    let nextIdx = 0;
+    const errors = [];
+
+    const worker = async () => {
+        while (nextIdx < coll.length) {
+            const idx = nextIdx++;
+            try {
+                results[idx] = await task(coll[idx]);
+            } catch (err) {
+                errors.push(err);
```
benzekrimaha: nit: `errors` is only ever read at index 0 , a single `firstError` variable says what it means.

tests/functional/replication/streamedCopy.spec.js
```diff
@@ -222,7 +222,9 @@ describe('streamed copy functional tests', () => {
            repTask._setupDestClients('dummyrole', log);
            repTask._getAndPutPartOnce(
                mockSourceEntry, mockSourceEntry, mockPartInfo,
-               log.newRequestLogger(), cb);
+               log.newRequestLogger())
+               .then(r => cb(null, r))
+               .catch(cb);
```
benzekrimaha: if `cb` throws, `.catch(cb)` calls it a second time we should use the two-argument form like `_setupClients` does:
```suggestion
               .then(r => cb(null, r), cb);
```

tests/unit/replication/ReplicateObject.spec.js
```diff
@@ -246,39 +238,36 @@ describe('ReplicateObject', () => {
...
             });
             task.backbeatDest = { send: sinon.stub().rejects(loopErr) };
-            task._putMetadataOnce(entry, false, null, fakeLogger, err => {
-                assert.ok(err instanceof MicroVersionIdAlreadyStoredException);
-                sinon.assert.notCalled(metricsStub);
-                done();
-            });
+            return task._putMetadataOnce(entry, false, null, fakeLogger)
+                .then(() => assert.fail('expected error'))
+                .catch(err => {
```
benzekrimaha: use `await assert.rejects(...)`: as written, when the task unexpectedly succeeds, the `assert.fail` AssertionError is swallowed by the following `.catch` and re-asserted, so the test fails with a confusing message instead of “expected error”. (applies to all similar cases)

tests/unit/replication/MultipleBackendTask.js
```diff
@@ -143,14 +140,15 @@ describe('MultipleBackendTask', function test() {
                 }),
             };
 
-            task._setupRolesOnce(makeEntry(), fakeLogger, err => {
-                assert(err);
-                assert.strictEqual(err.is.PreconditionFailed, true);
-                done();
-            });
+            return task._setupRolesOnce(makeEntry(), fakeLogger)
+                .then(() => assert.fail('expected error'))
```
benzekrimaha: same (`assert.rejects`)

extensions/replication/tasks/MultipleBackendTask.js
DarkIsDude: not really just a bit of habits to have. After 3/4 migrations you'll be an expoert

DarkIsDude (approved):
LGTM. Just comments from Maha 🙏

lib/tasks/BackbeatTask.js
SylvainSenechal: I added it because of the "consistent-return" linter but yeah better to return undefined

lib/tasks/BackbeatTask.js
SylvainSenechal: I'm just gonna handle single "result" objects, like its currently used across the codebase + document the function parameters

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: yes this was discussed and the change is small and intended, i added 2 tests

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Yeah I think i did this because right after the if statement use similar error rewrite pattern but we can use the arsenal error

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: yeah and i think i would even create a follow to fully move retry to be async, and to take an async action function because its really annoying to deal with this transition state

lib/util/runTasksWithConcurrency.js
SylvainSenechal: yeah i did that

lib/util/runTasksWithConcurrency.js
SylvainSenechal: Yes both legit and addressed

tests/unit/replication/ReplicateObject.spec.js
SylvainSenechal: yeah and many similar patterns actually

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Ah we can't do it for this one, this.s3destCredentials returns different things depending on config

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: yeah its fine

extensions/replication/tasks/ReplicateObject.js
SylvainSenechal: Addressed in another similar comment, I messed up, I think wanted to follow similar pattern to the else branch but yeah we can use arsenal error

extensions/replication/tasks/MultipleBackendTask.js
SylvainSenechal: yeah it would be the next migration I would do

extensions/replication/tasks/ReplicateObject.js
```diff
@@ -552,152 +545,145 @@ class ReplicateObject extends BackbeatTask {
...
+                if (!sourceStreamAborted && !destRequestAborted) {
                     abortController.abort();
                     destRequestAborted = true;
                 }
-                // eslint-disable-next-line no-param-reassign
-                err.origin = 'source';
                 if (err.$metadata?.httpStatusCode === 404) {
-                    return doneOnce(err);
+                    const objNotFound = errors.ObjNotFound;
+                    objNotFound.origin = 'source';
```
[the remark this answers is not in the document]
SylvainSenechal: false, errors.ObjNotFound invokes this getter : 
static errors() → { get: () => new ArsenalError(...) })

benzekrimaha (approved):

PR #2786, opened 2026-07-23
BB-830: fix notification config write race that silently drops events

tcarmet:
Bucket notification events could be silently and permanently lost for a bucket whenever the one-time write of its notification configuration to Zookeeper lost a race and failed, because the delivery path treats a missing configuration as "nothing to deliver" and drops every event for that bucket without a trace. This change makes that configuration write resilient so a transient failure no longer strands a bucket's notifications, and surfaces the previously silent drop in the logs. The race was found through intermittent failures of the bucket-notification end-to-end suite in CI (BB-830).

### DIFF
--- a/extensions/notification/configManager/ZookeeperConfigManager.js
+++ b/extensions/notification/configManager/ZookeeperConfigManager.js
@@ -9,6 +9,11 @@ const BaseConfigManager = require('./BaseConfigManager');
 const safeJsonParse = require('../../../lib/util/safeJsonParse');
 const constants = require('../constants');
 
+// bounded retry with backoff for config writes to zookeeper, so a transient
+// failure does not drop the configuration
+const CONFIG_WRITE_MAX_ATTEMPTS = 5;
+const CONFIG_WRITE_RETRY_BASE_MS = 200;
+
 const paramsJoi = joi.object({
     zkClient: joi.object().optional(),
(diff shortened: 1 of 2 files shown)


extensions/notification/configManager/ZookeeperConfigManager.js
```diff
@@ -209,23 +222,55 @@ class ZookeeperConfigManager extends BaseConfigManager  {
...
+                    }
+                    return this._zkClient.setData(zkPath, Buffer.from(data),
+                        -1, retryErr => cb(retryErr));
+                });
+            }
+            return cb(err);
+        });
+    }
+
+    _verifyBucketNotifConfigNode(zkPath, data, cb) {
```
jonathan-gramain: Can you clarify the need for a verify function, in my sense, if the backend says it's written, then it's 100% sure it's written so we don't need to check?

tcarmet: I might have trust issues :sweat_smile:, if it's 100% sure then I'll take it. I lacked knowledge in this area so that's my bad. The actual important part here in the PR is to catch a `NO_NODE` when trying to `setData`, create the node  retrying if we hit it, and retrying as a whole if anything fail and not fail silently. 

Removed it.

jonathan-gramain (approved):

anurag4DSB (approved):

PR #2788, opened 2026-07-28
Consume the replication status topic from the earliest offset

delthas:
### The problem

The replication status processor consumes the `backbeat-replication-status` topic and writes each replication's outcome back to the source object's metadata (COMPLETED/FAILED per site); FAILED entries also feed the failed-CRR listing so the object can be retried.

Its Kafka consumer never specified where to start reading when a partition has **no saved position** (no committed offset). In that case the client default applies: start at the **end** of the log, skipping whatever is already there. That default is never noticed in normal operation, because the group commits as it consumes and always finds its saved position. But it turns one specific race into permanent data loss:

1. the status processor restarts (rolling update, crash) — for a few seconds the consumer group has no member;
2. during that gap, a status message is produced onto a partition that never had a saved position (typically its first message ever — common on fresh deployments where every partition starts empty);
3. the new pod comes up, finds no saved position for that partition, starts at the end of the log — **after** the message.

Nobody ever reads that status, and nothing re-sends it: the data processor produced it, got its delivery acknowledgment, and moved on. The object's metadata says `PENDING` forever, and since the failure never reached the failed-CRR listing, it cannot even be retried.

### Impact on Zenko CI

This is the second half of the top CTST Replication flake, alongside the populator freeze fixed by #2787. In [run 29589236290 attempt 1](https://github.com/scality/Zenko/actions/runs/29589236290/attempts/1): the status-processor pod was rolled at 15:37:36; the FAILED status for the test's object landed on partition 4, offset 0 — first message ever on that partition — at 15:38:36, during the member gap; the replacement pod joined at 15:38:57 and started at offset 1. The "Re-replicate objects that failed to replicate" scenario (23 of 257 runs failed over two months) then waited its full 500s for a FAILED status that no longer existed anywhere — no test timeout can save this case, which is what makes it the *unrecoverable* variant of the flake.

The two fixes are complementary: #2787 removes the freezes and zombie pods that cause most of the red runs — and most of the restarts that open these gaps in the first place; this PR ensures that when a restart does happen (deploys never stop), a status produced during the gap is processed late instead of never. Together they should close out the Replication cluster (49 of 257 runs). Outside CI, the same race exists on any production rolling update, with the same silent consequence.

### The fix

Set `fromOffset: 'earliest'` on the status consumer: with no saved position, read the partition from the beginning. Steady state is unchanged (a saved position always wins over this setting), and the oplog LogConsumer already pins `earliest` for the same reason. In the worst case — a group that lost all its saved positions, e.g. after sitting empty past the offset retention — the consumer replays what the topic still retains (bounded by topic retention, broker default 7 days), and `UpdateReplicationStatus` re-reads the object's current metadata and skips anything already in a final state, so a replay mostly heals stuck-PENDING objects at the cost of temporary consumer lag.

The consumer options move to a `_getConsumerOptions()` method so a unit test can pin the option on this component, and two functional tests pin the semantics against a real kafka: an `'earliest'` consumer picks up a message produced before its group existed, and a default-configured consumer provably skips it — the exact loss mode this PR removes.

Issue: BB-826

### DIFF
--- a/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
+++ b/extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
@@ -347,6 +347,29 @@ class ReplicationStatusProcessor {
     }
 
     /**
+     * Build the options of the status topic consumer
+     * @param {Object} [options] - options object as given to start()
+     * @return {Object} consumer options
+     */
+    _getConsumerOptions(options) {
+        return {
+            kafka: {
+                hosts: this.kafkaConfig.hosts,
+                site: this.kafkaConfig.site,
+                compressionType: this.kafkaConfig.compressionType,
+                requiredAcks: this.kafkaConfig.requiredAcks,
+            },
+            topic: this.repConfig.replicationStatusTopic,
+            groupId: this.repConfig.replicationStatusProcessor.groupId,
+            concurrency: this.repConfig.replicationStatusProcessor.concurrency,
+            maxQueued: this.repConfig.replicationStatusProcessor.maxQueued,
+            queueProcessor: this.processKafkaEntry.bind(this),
+            bootstrap: (options && options.bootstrap) || false,
+            fromOffset: 'earliest',
+        };
+    }
+
+    /**
      * Start kafka consumer
      *
      * @param {object} [options] - options object (only used for tests
@@ -386,20 +409,8 @@ class ReplicationStatusProcessor {
             },
             done => {
                 let consumerReady = false;
-                this._consumer = new BackbeatConsumer({
-                    kafka: {
-                        hosts: this.kafkaConfig.hosts,
-                        site: this.kafkaConfig.site,
-                        compressionType: this.kafkaConfig.compressionType,
-                        requiredAcks: this.kafkaConfig.requiredAcks,
-                    },
-                    topic: this.repConfig.replicationStatusTopic,
-                    groupId: this.repConfig.replicationStatusProcessor.groupId,
-                    concurrency: this.repConfig.replicationStatusProcessor.concurrency,
-                    maxQueued: this.repConfig.replicationStatusProcessor.maxQueued,
-                    queueProcessor: this.processKafkaEntry.bind(this),
-                    bootstrap: (options && options.bootstrap) || false,
-                });
+                this._consumer = new BackbeatConsumer(
+                    this._getConsumerOptions(options));
                 this._consumer.on('error', () => {
                     if (!consumerReady) {
                         this.logger.fatal('error starting a backbeat consumer');
--- a/tests/functional/lib/BackbeatConsumer.js
+++ b/tests/functional/lib/BackbeatConsumer.js
@@ -257,7 +257,7 @@ describe('BackbeatConsumer rebalance tests', () => {
         // Bootstrap just once at the beginning of the test suite
         const bootstrapConsumer = new BackbeatConsumer({
             zookeeper: zookeeperConf,
-            kafka: { hosts: consumerKafkaConf.hosts }, groupId, topic,
+            kafka: consumerKafkaConf, groupId, topic,
             queueProcessor,
             bootstrap: true,
         });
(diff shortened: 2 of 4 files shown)


extensions/replication/replicationStatusProcessor/ReplicationStatusProcessor.js
```diff
@@ -355,6 +355,32 @@ class ReplicationStatusProcessor {
      * is complete
      * @return {undefined}
      */
+    /**
+     * Build the options of the status topic consumer
+     * @param {Object} [options] - options object as given to start()
+     * @return {Object} consumer options
+     */
+    _getConsumerOptions(options) {
```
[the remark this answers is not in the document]
delthas: Fixed: start()'s JSDoc moved back to directly precede start().

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -1274,3 +1274,133 @@ describe('BackbeatConsumer shutdown tests', () => {
...
+            next => startConsumer('earliest', next),
+            // wide budget: a consumer group's first join can take a
+            // session timeout (45s) to settle when a rebalance hits
+            // its joining window
+            next => waitFor(() => consumedMessages.includes(marker), 75000,
+                'the pre-existing message to be consumed', next),
+        ], done);
+    });
+
+    it('should skip messages produced before the group existed by default',
```
benzekrimaha: This validates librdkafka's default, not our code, and costs up to ~2.5 min in an already long suite. Is it really needed? The earliest test alone pins the behavior we ship; if the intent is to document the contrast, a comment on the first test is cheaper than a flaky-prone timing test IMO

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -1274,3 +1274,133 @@ describe('BackbeatConsumer shutdown tests', () => {
...
+
+    afterEach(function afterEach(done) {
+        this.timeout(90000);
+        async.parallel([
+            innerDone => producer.close(innerDone),
+            innerDone => (consumer ? consumer.close(innerDone) : innerDone()),
+        ], () => {
+            consumer = null;
+            done();
+        });
```
benzekrimaha: Let's avoid swallowing closing errors here
```suggestion
        ], err => {
            consumer = null;
            done(err);
        });
```

benzekrimaha:
Does the same gap exist in the other backbeat consumers (queue processor, lifecycle, GC, notification)? Whichever ones don't pin fromOffset have the same latent behavior on a fresh group. If out of scope here, can we please have a followup ticket for the audit?

tests/functional/lib/BackbeatConsumer.js
delthas: Fair point — dropped in 7ffd0cc8. The earliest test keeps a short comment stating what the default would do, and the unit test still pins that this processor passes the option.

tests/functional/lib/BackbeatConsumer.js
delthas: Applied in 7ffd0cc8.

delthas:
Good question — the audit is done, results are in a table on the BB-826 ticket. Summary: no extension consumer pins `fromOffset` today; by loss class, the **notification queue processor** (lost message = notification never delivered) and the **GC service consumer** (lost entry = orphaned storage) share this bug's failure mode, while the lifecycle processors are self-healing (the conductor re-enumerates every cycle) and the replication data processor needs its own assessment. Follow-up filed as BB-831 for the notification + GC consumers, reusing this PR's pattern.

SylvainSenechal:
> Good question — the audit is done, results are in a table on the [BB-826](https://scality.atlassian.net/browse/BB-826) ticket. Summary: no extension consumer pins `fromOffset` today; by loss class, the **notification queue processor** (lost message = notification never delivered) and the **GC service consumer** (lost entry = orphaned storage) share this bug's failure mode, while the lifecycle processors are self-healing (the conductor re-enumerates every cycle) and the replication data processor needs its own assessment. Follow-up filed as [BB-831](https://scality.atlassian.net/browse/BB-831) for the notification + GC consumers, reusing this PR's pattern.

interesting that you mention notification queue processor has the same issue, as this is one of the test that's really often flaky

[BB-826]: https://scality.atlassian.net/browse/BB-826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BB-831]: https://scality.atlassian.net/browse/BB-831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

SylvainSenechal (approved):
when all related pr are merged, try to release a backbeat version so we can check if this was really the reason for the flaky tests, it should be visible quickly by doing 2/3 zenko ci run.

Imo this is probably it, as I previously saw the backbeat pods were restarting very often, so i figured some messages were dropped/mishandled during these events

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -1274,3 +1274,102 @@ describe('BackbeatConsumer shutdown tests', () => {
...
+            }
+            if (Date.now() > deadline) {
+                return cb(new Error(`timed out waiting for ${description}`));
+            }
+            return setTimeout(check, 200);
+        };
+        check();
+    }
+
+    function startConsumer(fromOffset, cb) {
```
benzekrimaha: nit: `fromOffset` now has a single caller passing `'earliest'` (leftover from the dropped `latest` test) ,could be inlined into the consumer options.

tests/functional/lib/BackbeatConsumer.js
```diff
@@ -1274,3 +1274,102 @@ describe('BackbeatConsumer shutdown tests', () => {
...
+
+    before(function before(done) {
+        this.timeout(60000);
+        // settle the topic once (bootstrap canary produced and
+        // consumed) so per-test consumers join an established topic;
+        // per-test groups stay fresh so their partitions have no
+        // committed offset
+        const bootstrapConsumer = new BackbeatConsumer({
+            zookeeper: zookeeperConf,
+            kafka: { hosts: consumerKafkaConf.hosts },
```
benzekrimaha: nit: inconsistent with `startConsumer`, which passes the full `consumerKafkaConf`:
```suggestion
            kafka: consumerKafkaConf,
```

benzekrimaha (commented):
A few nits only.

benzekrimaha (approved):

tests/functional/lib/BackbeatConsumer.js
delthas: Inlined.

tests/functional/lib/BackbeatConsumer.js
delthas: Applied.

delthas:
> interesting that you mention notification queue processor has the same issue, as this is one of the tests that's really often flaky

Good catch — plausible second contributor there indeed (CI clusters have all-fresh partitions + constant pod rolls, so the loss window is much easier to hit than in production). Noted, BB-831 covers it.

delthas:
Release summary:
* `development/9.4`: releasing [`9.4.2`](https://github.com/scality/backbeat/actions/runs/30632170333)
* `development/9.5`: releasing [`9.5.0-preview.8`](https://github.com/scality/backbeat/actions/runs/30632176852)

PR #2789, opened 2026-08-03
Configurable kafka producer parameters

SylvainSenechal:
Issue: BB-685

Related Zenko Operator : https://github.com/scality/zenko-operator/pull/625

### DIFF
--- a/extensions/ingestion/IngestionConfigValidator.js
+++ b/extensions/ingestion/IngestionConfigValidator.js
@@ -1,5 +1,6 @@
 const joi = require('joi');
 const { probeServerJoi } = require('../../lib/config/configItems.joi');
+const { KAFKA_PRODUCER_PARAMS_SCHEMA } = require('../../lib/config.joi');
 
 const joiSchema = joi.object({
     auth: joi.object({
@@ -17,6 +18,7 @@ const joiSchema = joi.object({
     processor: joi.object({
         circuitBreaker: joi.object().optional(),
     }).optional(),
+    producerParams: KAFKA_PRODUCER_PARAMS_SCHEMA,
 });
 
 function configValidator(backbeatConfig, extConfig) {
--- a/lib/BackbeatProducer.js
+++ b/lib/BackbeatProducer.js
@@ -14,6 +14,7 @@ const {
     KAFKA_PRODUCER_MESSAGE_MAX_BYTES,
     KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE,
     KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS,
+    KAFKA_PRODUCER_PARAMS_SCHEMA,
 } = require('./config.joi');
 const FLUSH_TIMEOUT = 5000;
 
@@ -58,6 +59,7 @@ class BackbeatProducer extends EventEmitter {
                 maxRequestSize: joi.number().default(KAFKA_PRODUCER_MESSAGE_MAX_BYTES),
                 compressionType: joi.string().default(KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE),
                 requiredAcks: joi.number().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
+                producerParams: KAFKA_PRODUCER_PARAMS_SCHEMA,
             }
         );
     }
@@ -75,7 +77,8 @@ class BackbeatProducer extends EventEmitter {
     }
 
     get producerConfig() {
-        const producerParams = {
+        const config = {
+            ...this._producerParams,
             'metadata.broker.list': this._kafkaHosts,
             'message.max.bytes': this._maxRequestSize,
             'dr_cb': true,
@@ -84,10 +87,10 @@ class BackbeatProducer extends EventEmitter {
         };
 
         if (process.env.RDKAFKA_DEBUG_LOGS) {
-            producerParams.debug = process.env.RDKAFKA_DEBUG_LOGS;
+            config.debug = process.env.RDKAFKA_DEBUG_LOGS;
         }
 
-        return producerParams;
+        return config;
     }
 
     get topicConfig() {
@@ -125,6 +128,7 @@ class BackbeatProducer extends EventEmitter {
             maxRequestSize,
             compressionType,
             requiredAcks,
+            producerParams,
         } = joiResult;
         this._kafkaHosts = kafka.hosts;
         this._topic = topic && withTopicPrefix(topic);
@@ -132,6 +136,7 @@ class BackbeatProducer extends EventEmitter {
         this._maxRequestSize = maxRequestSize;
         this._compressionType = compressionType;
         this._requiredAcks = requiredAcks;
+        this._producerParams = producerParams;
     }
 
     connect() {
--- a/lib/config.joi.js
+++ b/lib/config.joi.js
@@ -15,6 +15,13 @@ const {
 const KAFKA_PRODUCER_MESSAGE_MAX_BYTES = 5000020;
 const KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE = 'Zstd';
 const KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS = -1; // all brokers
+const KAFKA_PRODUCER_PARAMS_SCHEMA = joi.object({
+    'metadata.broker.list': joi.forbidden(),
+    'dr_cb': joi.forbidden(),
+    'message.max.bytes': joi.forbidden(),
+    'compression.type': joi.forbidden(),
+    'statistics.interval.ms': joi.forbidden(),
+}).unknown(true).default({});
 const logSourcesJoi = joi.string().valid('bucketd', 'ingestion', 'dmd', 'kafka');
 
 const joiSchema = joi.object({
@@ -34,6 +41,7 @@ const joiSchema = joi.object({
         site: joi.string(),
         compressionType: joi.string().default(KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE),
         requiredAcks: joi.number().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
+        producerParams: KAFKA_PRODUCER_PARAMS_SCHEMA,
     },
     transport: transportJoi,
     s3: hostPortJoi.optional(),
@@ -107,4 +115,5 @@ module.exports = {
     KAFKA_PRODUCER_MESSAGE_MAX_BYTES,
     KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE,
     KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS,
+    KAFKA_PRODUCER_PARAMS_SCHEMA,
 };
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -163,6 +163,10 @@ class IngestionPopulator {
             maxRequestSize: this.kafkaConfig.maxRequestSize,
             compressionType: this.kafkaConfig.compressionType,
             requiredAcks: this.kafkaConfig.requiredAcks,
+            producerParams: {
+                ...this.kafkaConfig.producerParams,
+                ...this.ingestionConfig.producerParams, // Extension params override global params
+            },
             topic,
             pollIntervalMs: POLL_INTERVAL_MS,
         });
--- a/tests/unit/backbeatProducer.js
+++ b/tests/unit/backbeatProducer.js
@@ -111,6 +111,73 @@ describe('backbeatProducer', () => {
             'custom-topic', [{ key: 'foo', message: 'bar' }], () => {});
     });
 
+    describe('producerParams', () => {
+        it('should include extra producerParams in producerConfig', () => {
+            const producer = new BackbeatProducer({
+                kafka,
+                producerParams: {
+                    'queue.buffering.max.kbytes': 1048576,
+                    'queue.buffering.max.messages': 200000,
+                },
+            });
+            const config = producer.producerConfig;
+            assert.strictEqual(config['queue.buffering.max.kbytes'], 1048576);
+            assert.strictEqual(config['queue.buffering.max.messages'], 200000);
+        });
+
+        it('should reject critical built-in params via Joi', () => {
+            assert.throws(
+                () => new BackbeatProducer({
+                    kafka,
+                    producerParams: { 'metadata.broker.list': 'attacker:9092' },
+                }),
+                /metadata\.broker\.list/,
+            );
+            assert.throws(
+                () => new BackbeatProducer({
+                    kafka,
+                    producerParams: { 'compression.type': 'gzip' },
+                }),
+                /compression\.type/,
+            );
+        });
+
+        it('should default to empty producerParams when not provided', () => {
+            const producer = new BackbeatProducer({ kafka });
+            assert.deepStrictEqual(producer._producerParams, {});
+        });
+
+        it('should throw on invalid librdkafka producerParams keys', () => {
+            let err;
+            try {
+                new BackbeatProducer({
+                    kafka,
+                    producerParams: { 'not.a.valid.librdkafka.option': 'value' },
+                });
+            } catch (e) {
+                err = e;
+            }
+            assert(err, 'expected BackbeatProducer to throw on invalid producerParams key');
+            assert.match(err.message,
+                /No such configuration property: "not\.a\.valid\.librdkafka\.option"/);
+        });
+
+        it('should throw on out-of-range librdkafka producerParams values', () => {
+            let err;
+            try {
+                new BackbeatProducer({
+                    kafka,
+                    producerParams: { 'queue.buffering.max.kbytes': -1 },
+                });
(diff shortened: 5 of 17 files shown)


lib/config.joi.js
```diff
@@ -35,6 +35,7 @@ const joiSchema = joi.object({
         site: joi.string(),
         compressionType: joi.string().default(KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE),
         requiredAcks: joi.number().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
+        producerParams: joi.object().unknown(true).default({}),
```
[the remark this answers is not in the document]
SylvainSenechal: This is intentional, the objective is to have config that will later be extensible

extensions/ingestion/IngestionConfigValidator.js
```diff
@@ -16,6 +16,7 @@ const joiSchema = joi.object({
     processor: joi.object({
         circuitBreaker: joi.object().optional(),
     }).optional(),
+    producerParams: joi.object().unknown(true).default({}),
```
DarkIsDude: Why should we accept unknown ? Here we don't really define them neither validate them, why ? We should define what is behind object ?

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -173,6 +173,10 @@ class IngestionPopulator {
             maxRequestSize: this.kafkaConfig.maxRequestSize,
             compressionType: this.kafkaConfig.compressionType,
             requiredAcks: this.kafkaConfig.requiredAcks,
+            producerParams: {
+                ...this.kafkaConfig.producerParams,
+                ...this.ingestionConfig.producerParams, // Extension params override global params
```
DarkIsDude, line 168:
```suggestion
                ...this.ingestionConfig.producerParams,
```

lib/BackbeatProducer.js
```diff
@@ -58,6 +58,7 @@ class BackbeatProducer extends EventEmitter {
                 maxRequestSize: joi.number().default(KAFKA_PRODUCER_MESSAGE_MAX_BYTES),
                 compressionType: joi.string().default(KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE),
                 requiredAcks: joi.number().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
+                producerParams: joi.object().unknown(true).default({}),
```
DarkIsDude: Same here, we should define them ? Maybe also factorise them between both (three with kafka params?)?

lib/BackbeatProducer.js
```diff
@@ -75,7 +76,8 @@ class BackbeatProducer extends EventEmitter {
     }
 
     get producerConfig() {
-        const producerParams = {
+        const config = {
+            ...this._producerParams,
```
DarkIsDude, line 81: Should `_producerParams` overwrite the default one ?

lib/config.joi.js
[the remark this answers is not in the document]
DarkIsDude: That's a good call from claude, should we define config per producer type or having a global one. What do you mean by `extensible` ?

lib/BackbeatProducer.js
DarkIsDude: `should not let producerParams override critical built-in params` what do you mean by `critical` ?

tests/unit/backbeatProducer.js
```diff
@@ -111,6 +111,39 @@ describe('backbeatProducer', () => {
...
+                    'queue.buffering.max.kbytes': 1048576,
+                    'queue.buffering.max.messages': 200000,
+                },
+            });
+            const config = producer.producerConfig;
+            assert.strictEqual(config['queue.buffering.max.kbytes'], 1048576);
+            assert.strictEqual(config['queue.buffering.max.messages'], 200000);
+        });
+
+        it('should not let producerParams override critical built-in params', () => {
```
DarkIsDude: Should it be rejected by joi directly by disallow some keys ?

DarkIsDude (changes requested):

extensions/ingestion/IngestionConfigValidator.js
SylvainSenechal: there are dozens of potentials parameters here : https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html

So we went with this to not have to enumerate all, we are only using one or two for now, but wanna be able to use more without having to do another backbeat pr

Honestly debatable though, an ai could write down instantly the ~50 different params available 🤔

lib/queuePopulator/IngestionPopulator.js
SylvainSenechal: I wanna keep this one, i know its more obvious for you but not for me 🧐

extensions/ingestion/IngestionConfigValidator.js
SylvainSenechal: @francoisferrand Should we enumerate all params, or enumerate only the ones we use now, or keep it that way. 


```suggestion
    producerParams: joi.object({
    'queue.buffering.max.kbytes': joi.number().optional(),
    'queue.buffering.max.messages': joi.number().optional(),
}).optional().default({}),
```

lib/BackbeatProducer.js
SylvainSenechal: discussion ongoing in the other topic

lib/config.joi.js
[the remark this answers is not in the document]
SylvainSenechal: The idea is we have 
kafka.producerParams
extensions.ingestion.producerParams

extensions override the first one when provided. But for now we only have extension for ingestion, and aren't gonna bother doing it for all backbeat extensions.

I think its fine as is, but yeah eventually we should update all other backbeat extension to use this global kafka params, else its a bit weird @francoisferrand

tests/unit/backbeatProducer.js
SylvainSenechal: Let's rediscuss after we've finalized joi config, some ongoing discussion on it

lib/BackbeatProducer.js
SylvainSenechal: critical is not so important, tbh its ai generated, but the intent is "don't touch hardcoded existing parameters because they are here for a good reason (broker.list, max.bytes, dr_cb)".

Considering we probably don't wanna touch the existing 3 configured values configurations, i would say no we don't want producerParams to override them. Especially kafkaHosts.

lib/config.joi.js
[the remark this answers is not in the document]
DarkIsDude: Agree, can you do a follow up ?

extensions/ingestion/IngestionConfigValidator.js
DarkIsDude: No enumerate all of them. We use only a few, and we can use only a few.  So just them should be enough. The goal of Joi is to validate, here we just skip it. Also we can make sure for example that we don't insert critical one (I remember the host later that is skipped for example).

lib/queuePopulator/IngestionPopulator.js
DarkIsDude: 🤝 let's go then. You just describe the code but why not

lib/BackbeatProducer.js
DarkIsDude: 🤝 it's a design choice. For me it's a good idea to overwrite them because you will be more flexible (general for everyone and then you can override with specific). But we can keep it. Resolved

DarkIsDude (changes requested):
Keeping request changes. Waiting for Joi decision to finish the review 🙏

SylvainSenechal:
@francoisferrand requested as I want your opinion on some comments : 
Keep the kafka config very generic, or enumerate the specific kafka options we want to configure

extensions/ingestion/IngestionConfigValidator.js
SylvainSenechal: Rechecked with François, we said the purpose of the ticket is to not bother enumerating all kafka possible options. There is also an associated ticket that we talked about for this sprint which is about using env variables in some kind of "override debugging mode" that will make use of this.

What François was also saying is normally, here, on kafka client creation, if a provided option is wrong, it will be rejected, so we are using librdkafka as some kind of validator (gonna add a test for this though)

francoisferrand: indeed, the whole point is to avoid the "double validation" : librdkafka already validates its parameter, so there is no point validating them ourself first.

lib/config.joi.js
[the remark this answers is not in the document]
francoisferrand: as for catch from Claude : yes, we should pass `kafka.producerParams` directly to all extensions, no reason to "filter" the params and pass specific ones only.

in this PR or the next does not really matter, but must be done "soon" for consistency.

(though I find it hard to say "_Add producerParams support to BackbeatProducer_" without binding it in each place BakcbeatProducer is used)

---
as for having override for every extension, not needed now indeed, but should we have a followup ?
* the producer params are configured globally - which as far as I know was enough for all uses cases....
* ...except for ingestion, which has specific requirement/constraints (e.g. to accommodate reading large "batch" from metadata)

If the need for override in other extension arises, we can certainly add it there; but not sure we need to be proactive on this matter : so I fear if we create a followup ticket we would just forget it, and create a duplicate if/when we eventually need to do the same for another extension...

Do you feel there is a need to implement override in every extensions (i.e. we should implement that followup "soonish" as well) ?

lib/BackbeatProducer.js
francoisferrand: the point is to allow passing "any" kafka param, and let librdkafka or noderdkafka do the validation - not us → having an "unknown" field is the goal.

we could pass a callback to joi so we explicitely let the underlying lib validate this param - but I am not sure such a function exists ; and anyway the validation will be done a bit later, when creating the kafka consumer (which is done on process startup anyway)

As for the other (older) parameters, they are both kind of required (we need the defaults, not sure they can be changed in isolation) and removing them would break compatibility: so I'd rather not.

francoisferrand (commented):
- the "unknown" parameter is actually the goal of the ticket: as it is validated by underlying lib already ; a test verifying that invalid params are caught may be good however
- as we are adding a generic `producerParams` to the config and BackbeatProducer, we should pass it to every BackbeatProducer to consistency, to avoid any confusion
- it should not replace the pre-existing params (config/env) IMHO: both for retro-compatibility and because the default values are specific to our use case, and should be kept to avoid any risk of regression (`producerParams` is empty however, so by default we keep the exact same behavior as before ; and would be used for tuning parameters, not to make the architecture work)
- no need for "override" in other extension at this point I think : can be added later if/when needed

extensions/ingestion/IngestionConfigValidator.js
SylvainSenechal: I did some testing, invalid parameters will be rejected with a clear message on producer creation

lib/config.joi.js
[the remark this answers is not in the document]
SylvainSenechal: I'm adding the base kafka producerParams to all extensions.
Imo no need to add dedicated kafkaParams to the extensions if not used now, we will add it if we need them 🤔

tests/unit/backbeatProducer.js
SylvainSenechal: So with the current design it will be rejected on extension creation by the kafka lib

extensions/ingestion/IngestionConfigValidator.js
DarkIsDude: @SylvainSenechal can you add a test to make sure the lib will reject an invalid configuration and update the Joi config to exclude parameters that we should not have 🙏

DarkIsDude: @SylvainSenechal also what do you think to define a `const KAFKA_EXTENSIONS_PARAMETERS_SCHEMA = joi.object().unknown(true).default({});` to avoid duplicate with `lib/BackbeatProducer.js` ?

DarkIsDude: `https://github.com/scality/backbeat/pull/2789/changes#diff-06f2fd2bcbeb368f3a5e7912e093e08bfec3f778c423168ebf48249ef01361a2R146` rejection is tested here, perfect 🙏

tests/unit/backbeatProducer.js
francoisferrand: > So with the current design it will be rejected on extension creation by the kafka lib

kafka lib will reject _invalid_ parameters.
But some fields will be overridden by backbeat itself (like `metadata.broker.list` or `message.max.bytes`...):
```
            ...this._producerParams,
            'metadata.broker.list': this._kafkaHosts,
            'message.max.bytes': this._maxRequestSize,
            'dr_cb': true,
            'compression.type': this._compressionType,
            'statistics.interval.ms': 1000,
```

→ as discussed in Slack, the joi config should verify that these keys ARE NOT set in the "unknown" field
maybe using [.without()](https://joi.dev/api/18.x.x#objectwithoutkey-peers-options), [.invalid()](https://joi.dev/api/18.x.x#anyinvalidvalues---aliases-disallow-not) or by adding a callback...

extensions/ingestion/IngestionConfigValidator.js
SylvainSenechal: Yeah we have a test although if I'm being honest, the lib rejected non existent params and negative params, but when i tried to use 9999999999 for a max of 2147483647, it was silently truncated 🧐 

I added joi.forbidden for the 5 parameters that are hardcoded I think its fine since they have their own ways of being configured if needed

tests/unit/backbeatProducer.js
francoisferrand: seems to work with `.forbidden()` :

```
Joi.object({
  a: Joi.any().forbidden(),
}).unknown(true).default({})
```

c.f. https://joi.dev/tester/18.x.x#share=eyJkYXRhIjoiLy9JbnNlcnQgZGF0YSB0byB2YWxpZGF0ZSBoZXJlXG57XG4gIHVzZXJuYW1lOiBcImFiY1wiLFxuICBhOiBcImZvb1wiXG59Iiwic2NoZW1hIjoiLy9JbnNlcnQgeW91ciBqb2kgc2NoZW1hIGhlcmVcbkpvaS5vYmplY3Qoe1xuICBhOiBKb2kuYW55KCkuZm9yYmlkZGVuKCksXG59KS51bmtub3duKHRydWUpLmRlZmF1bHQoe30pIn0=

francoisferrand (commented):
following the discussion, missing the "rejection" of fields which we override or set via other means) - and which would never reach librdkafka ; c.f. https://github.com/scality/backbeat/pull/2789#discussion_r3767057219

SylvainSenechal:
> following the discussion, missing the "rejection" of fields which we override or set via other means) - and which would never reach librdkafka ; c.f. [#2789 (comment)](https://github.com/scality/backbeat/pull/2789#discussion_r3767057219)

Yes, this is added already

lib/BackbeatProducer.js
```diff
@@ -58,6 +59,13 @@ class BackbeatProducer extends EventEmitter {
                 maxRequestSize: joi.number().default(KAFKA_PRODUCER_MESSAGE_MAX_BYTES),
                 compressionType: joi.string().default(KAFKA_PRODUCER_DEFAULT_COMPRESSION_TYPE),
                 requiredAcks: joi.number().default(KAFKA_PRODUCER_DEFAULT_REQUIRED_ACKS),
+                producerParams: KAFKA_PRODUCER_PARAMS_SCHEMA.keys({
+                    'metadata.broker.list': joi.forbidden(),
+                    'dr_cb': joi.forbidden(),
+                    'message.max.bytes': joi.forbidden(),
+                    'compression.type': joi.forbidden(),
+                    'statistics.interval.ms': joi.forbidden(),
+                }),
```
francoisferrand: should these "forbidden" keys not be moved _inside_ the `KAFKA_PRODUCER_PARAMS_SCHEMA` variable, so the config is validated directly in both `extensions/ingestion/IngestionConfigValidator.js` and `lib/config.joi.js` ?

SylvainSenechal: yeah we can do that since all classes end up instantiating backbeatProducer anyways

DarkIsDude (approved):
LGTM

tests/unit/backbeatProducer.js
```diff
@@ -111,6 +111,73 @@ describe('backbeatProducer', () => {
...
+                });
+            } catch (e) {
+                err = e;
+            }
+            assert(err, 'expected BackbeatProducer to throw on invalid producerParams key');
+            assert.match(err.message,
+                /No such configuration property: "not\.a\.valid\.librdkafka\.option"/);
+        });
+
+        it('should throw on out-of-range librdkafka producerParams values', () => {
```
francoisferrand, line 165: not sure this is something we really want to assert : overall what we are doing with this PR is just delegate all validation to librdkafka, so we don't really care how (thoroughly) it does it and we will just follow its behavior...

francoisferrand (approved):

tests/unit/backbeatProducer.js
SylvainSenechal: it shows our intent to rely librdkafka instead of joi for validation 🤔

PR #2793, opened 2026-08-05
Retry stale service-credential 403s on the internal backbeat routes

delthas:
## Problem

Artesca regenerates builtin service-account keys on every overlay write (PSVAPI-134). Backbeat and cloudserver each snapshot authdata at process start and neither reloads it, so during the rolling restarts that follow a rotation cloudserver returns `403 InvalidAccessKeyId` on internal calls.

Smithy classifies a 403 as non-retryable (`lib/clients/utils.js`), every retry wrapper gates on `err.retryable`, and `BackbeatConsumer` commits the offset regardless — so a replication-status entry is destroyed outright. The failure lands in `_refreshSourceEntry`, upstream of every FAILED bookkeeping step, so the object keeps its `PENDING` status, never enters the failed-CRR sorted set, and is invisible to crrRetry.

A census on the ticket (10 CTST runs, 100 rejected requests, 8 skew episodes) found 95 % of these are cloudserver-stale — so a retry in backbeat can succeed — and that staleness is a per-request property: internal-cloudserver runs 4 replicas to backbeat's 1 and rolls a pod at a time, so a later attempt can be served by an already-updated pod. One backbeat pod was observed both failing and succeeding within the same second.

## Changes

**`lib/util/staleCredentialError.js`** (new) — `isStaleCredentialError()` matching `InvalidAccessKeyId` and `SignatureDoesNotMatch`, and `retryOnStaleCredentials()`, a stateful predicate that preserves the existing `err.retryable` behaviour and adds a bounded number of stale-credential retries. `AccessDenied` is deliberately excluded: on an internal route it is the existing signal for a misconfigured source role (`lib/util/replicationPermissionError.js`) and must keep failing fast.

**`lib/BackbeatMetadataProxy.js`** — the three source-metadata calls (`getMetadata`, `putMetadata`, `headLocation`) use that predicate. A fresh one per `retry()` call, so the budget applies to a single logical operation.

**`extensions/replication/tasks/MultipleBackendTask.js`** — `_handleReplicationOutcome` listed `InvalidAccessKeyId` as a permanent failure and returned a bare `done()`, committing the offset with no status published. `ReplicateObject` already omits it and publishes FAILED with `committable: false`; the subclass now matches, so an entry whose retries are exhausted stays visible to crrRetry.

## Notes for review

- **Timing.** The proxy calls `super()` with no retryParams, so it uses the defaults (`min: 1000, factor: 1.5`): attempts at 1 s, 2.5 s, 4.75 s, 8.1 s, 13.2 s — ~13 s total, ~20x under `max.poll.interval.ms` (300 s). This matters because the failure is correlated across all `concurrency: 10` slots during a skew window. A unit test guards the property.
- **The backend-specific budgets are not reachable.** The `60000` ms backoff / `900 s` timeout picked in `ReplicateObject.js:67-71` belongs to the task instances and only governs retries those tasks start themselves. `MultipleBackendTask._refreshSourceEntry` calls the proxy directly, so a stale-credential retry runs on the proxy's own params regardless of destination type.
- **`committable: false` defers the commit, it does not withhold it** — `_publishReplicationStatus` calls `onEntryCommittable` in the producer callback whether the produce succeeded or not, and this is already the path taken by every COMPLETED and FAILED outcome. No new stall mode; the change removes a special case that bypassed the normal path.
- No change to `isRetryableMiddleware`, so a customer-destination 403 stays permanent by construction. The four call sites that opt in only ever talk to the internal endpoint.

## Deliberately out of scope

- The 24 % of census rejections that arrived on the ordinary S3 route (`GetBucketReplication` in `_setupRolesOnce`) are not retried. With the second commit they become visible FAILED entries rather than silent losses.
- The ~5 % backbeat-stale direction still needs a pod restart; an `operatorBackend` credential reload would close it.
- What the *status* processor should do when retries are exhausted — collides with BB-772's commit-on-poison-pill decision.
- Backports: 9.2 can take the same patch; 9.1 is on aws-sdk v2 and needs a variant.

Issue: BB-832

### DIFF
--- a/extensions/replication/tasks/MultipleBackendTask.js
+++ b/extensions/replication/tasks/MultipleBackendTask.js
@@ -1283,8 +1283,7 @@ class MultipleBackendTask extends ReplicateObject {
         if (err.BadRole || err.name === 'BadRole' ||
             (err.origin === 'source' &&
              (err.NoSuchEntity || err.name === 'NoSuchEntity' ||
-              err.AccessDenied || err.name === 'AccessDenied' ||
-              err.InvalidAccessKeyId || err.name === 'InvalidAccessKeyId'))) {
+              err.AccessDenied || err.name === 'AccessDenied'))) {
             log.error('replication failed permanently for object, ' +
                       'processing skipped',
                 { failMethod: err.method,
--- a/lib/BackbeatMetadataProxy.js
+++ b/lib/BackbeatMetadataProxy.js
@@ -7,6 +7,7 @@ const RoleCredentials = require('./credentials/RoleCredentials');
 const { getAccountCredentials } = require('./credentials/AccountCredentials');
 const { http: HttpAgent } = require('httpagent');
 const { lifecycleListing: { NON_CURRENT_TYPE, CURRENT_TYPE, ORPHAN_DM_TYPE } } = require('./constants');
+const { retryOnStaleCredentials } = require('./util/staleCredentialError');
 const { 
     BackbeatRoutesClient,
     GetBucketIndexesCommand,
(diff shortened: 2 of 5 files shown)


delthas:
Deferring this as too hard and not needed as is right now.

PR #2794, opened 2026-08-06
Consume from the earliest offset where a lost message is unrecoverable

delthas:
Pins `fromOffset: 'earliest'` on the three consumers where a message produced during a consumer group gap is permanently lost **and** the group is service-wide rather than created per site or per destination:

| Topic | Consumer |
| --- | --- |
| `backbeat-lifecycle-transition-tasks` | transition processor |
| `cold-status-<location>` | transition processor, per location |
| gc topic | `GarbageCollector` |

Four consumers deliberately stay on the default: the lifecycle object (expiration) tasks topic (self-healing via the conductor's re-enumeration), the notification internal topic and the data-mover topic (both use a per-instance group id on a shared topic, so a new destination or location would replay the full retention window), and the replication topic itself (replay cost needs its own decision). A unit test pins the expiration exclusion so it is not "fixed" by accident.

The per-consumer rationale — why each loss is or is not recoverable, and the replay-safety analysis behind each decision — is in BB-831.

### Validation

On a 30-run census with an earlier build of this fix, the Azure Archive location CRUD scenario went from 8/20 runs failing (40 %) to 1/30 (3.3 %), and no skip of this kind appears in any of the dumps inspected. Pre-fix, cold-status messages produced but never consumed ran at 15–34 % across two censuses, concentrated in the rollout window and absent outside it, and the mechanism was traced end to end in every case examined — all of them cold-status. Caveats: an unrelated fix touching cold transitions landed between censuses, and the validated build's consumer set is wider than what this PR now pins, so the attribution rests on the traced mechanism rather than on the rate alone.

### Implementation note

The gc consumer options move into `_getConsumerOptions()` for unit coverage, following #2788. Generic `fromOffset` semantics are already covered by the functional tests #2788 added to the `BackbeatConsumer` suite.

Issue: BB-831

### DIFF
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
@@ -99,6 +99,8 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
     getConsumerParams() {
         const consumerParams = super.getConsumerParams(this._lcConfig.transitionTasksTopic);
 
+        consumerParams[this._lcConfig.transitionTasksTopic].fromOffset = 'earliest';
+
         const locations = require('../../../conf/locationConfig.json') || {};
 
         this._lcConfig.coldStorageTopics.forEach(topic => {
@@ -133,6 +135,7 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
                 maxQueued: this._processConfig.maxQueued,
                 queueProcessor: this.processColdStorageStatusEntry.bind(this),
                 circuitBreaker,
+                fromOffset: 'earliest',
             };
         });
 
--- a/extensions/gc/GarbageCollector.js
+++ b/extensions/gc/GarbageCollector.js
@@ -121,14 +121,11 @@ class GarbageCollector extends EventEmitter {
     }
 
     /**
-     * Start kafka consumer. Emits a 'ready' event when
-     * consumer is ready.
-     *
-     * @return {undefined}
+     * Build the options of the garbage collector topic consumer
+     * @return {Object} consumer options
      */
-    start() {
-        let consumerReady = false;
-        this._consumer = new BackbeatConsumer({
+    _getConsumerOptions() {
+        return {
             kafka: {
                 hosts: this._kafkaConfig.hosts,
                 site: this._kafkaConfig.site,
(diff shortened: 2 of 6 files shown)


SylvainSenechal (approved):
Approved but this looks again like quite a big oversight (maybe its mostly fixing issues with ci as if I understand correctly, 'earliest' will only apply to never commited partition which we have plenty in tests ci but probably not many in prod 🤔 )
You may wanna wait a few days for François to take a look though as he is back soon

Also, while its probably a bit annoying to do, I think this would really deserve a functional ci test in backbeat where we setup kafka, start a backbeat consumer then kill it and restart it, while manually producing a message in kafka in between, to make sure the behavior is correct 🤔

delthas:
Agree this is mainly for CI, though this can happen when creating locations in production.

For functional tests we already have functional tests for `earliest` consumers added in #2488. Kill & restart wouldn't work as is because the partition offset would be kept and wouldn't restart from earliest.

delthas:
Waiting for another reviewer :)

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -756,7 +757,8 @@ class QueueProcessor extends EventEmitter {
                 this._dataMoverConsumer = this._createConsumer(
                     this.repConfig.dataMoverTopic,
                     this.processDataMoverEntry.bind(this),
-                    Object.assign({ enableBacklogMetrics: true }, options));
+                    // per-consumer: the replication topic shares _createConsumer and stays on the default
```
DarkIsDude: This comment make sense when reading the PR but without PR context, I guess it's WTF comment ?

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -99,6 +99,9 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
     getConsumerParams() {
         const consumerParams = super.getConsumerParams(this._lcConfig.transitionTasksTopic);
 
+        // not on the base params: the expiration tasks topic is self-healing and stays on the default
```
DarkIsDude:
```suggestion
```

DarkIsDude (approved):
Nothing for me. It's clear enough.

But I have a weird feeling that I can't explain: like we did a fix but it's the right one, something else must be done. I don't know why, I don't know what. Maybe something when we create a location for example 🤷. Anyway, code looks right. For example:

- backbeat-data-mover is one shared topic, but the group is ${groupId}-${this.site}, and every site filters by toLocation === this.site. Adding a location at runtime spawns a fresh QueueProcessor, hence a fresh group with no committed offsets — which under earliest now starts at offset 0 of the entire retained log.
- extensions/notification/queueProcessor/QueueProcessor.js : Same. Group is ${groupId}-${destinationId}; when internalTopic is unset, all destinations share one topic. Adding a destination creates a new group on that old topic, and processKafkaEntry filters by destination ID, prefix, suffix, and event type — never by time. So the full retention window of a bucket's events gets delivered to the new destination as if fresh, breaking the S3 guarantee that only post-configuration events are notified.

Also this can have an impact on Ring, maybe better to check with them and asking for a review ?

delthas:
Thanks.

I'm dropping notification and data-mover from this PR to keep it low-risk. They're not affecting CI anyway (no notification issues in the Zenko run census, and no data-mover tests).

delthas:
So only backbeat-lifecycle-transition-tasks, cold-status-<location>, gc remain. Those should be safe.

delthas:
!release

delthas:
Release summary:
* `development/9.5`: releasing [`9.5.0-preview.9`](https://github.com/scality/backbeat/actions/runs/31401498851)

PR #2795, opened 2026-08-07
Drop the unused "mongo" log source from the queue populator

delthas:
The `"mongo"` log source made the queue populator tail MongoDB's internal replica-set oplog (`local.oplog.rs`) through arsenal's `mongoclient.LogConsumer`. It is broken on MongoDB >= 7.1 (the oplog `h` field used as a cursor was removed — ARSN-545, closed Won't Fix) and unused by every deployment: Federation sets `bucketd`, and zenko-operator overrides its `mongo` fallback to `kafka` whenever the oplog populator is enabled (the default on all mongo-backed deployments).

Changes:
- Delete `MongoLogReader` and its `case 'mongo'` in `QueuePopulator._setupLogSources()`, plus related jsdoc and the log-source error message.
- Remove `'mongo'` from the `logSource` joi enum.
- Remove the mongo-specific opaque-string offset handling in `LogReader._readLogOffset()` — stored offsets are numeric for all remaining log sources.
- Switch the sample config and unit-test fixtures from `logSource: "mongo"` to `"kafka"`.

The `queuePopulator.mongo` config section is intentionally kept: it is the shared MongoDB connection configuration used by the lifecycle bucket processor, notification queue processor, mongo processor and oplog populator.

Follow-ups tracked separately: removing `mongoclient/LogConsumer` + `ListRecordStream` from arsenal, and (optionally) the `logSource: "mongo"` fallback in zenko-operator.

Issue: BB-836


DarkIsDude (approved):

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -47,8 +47,6 @@ class IngestionPopulator {
      *   configuration (mandatory if logSource is "bucket")
      * @param {Object} [qpConfig.dmd] - dmd source
      *   configuration (mandatory if logSource is "dmd")
-     * @param {Object} [qpConfig.mongo] - mongo source
-     *   configuration (mandatory if logSource is "mongo")
```
francoisferrand, line 51: nit: this looks like it got copied from `QueuePopulator`. IngestionPopulator only supports reading from bucketd AFAIK (but probably not the place to reverse engineer the code to fix the doc)

francoisferrand (approved):

lib/queuePopulator/IngestionPopulator.js
delthas: Indeed — checked, and it goes further than the doc: in the ingestion path `qpConfig` is only passed through (`IngestionPopulator` → `IngestionReader` → `IngestionProducer`) and none of its fields are ever read, so the whole per-field block was stale copy-paste. Trimmed in 21dca584.

delthas:
!done 3h

delthas:
[`BB-836`](https://scality.atlassian.net/browse/BB-836): logged `3h`; transitioned to Done (`Fixed`)

PR #2798, opened 2026-08-14
TaskScheduler: include the bucket and the eTag in the action dedupe key

delthas:
The `TaskScheduler` dedupe key of an `ActionQueueEntry` omitted the bucket, so two actions targeting the same object key in different buckets were considered duplicates and the second one was silently dropped — it is acknowledged and its offset committed, so nothing replays it.

Two changes, in this order:

**1. A distinct task scheduler for the data mover.** A queue processor pushed the entries of both its consumers into one scheduler — replication entries from the replication topic (`QueueProcessor.js:936`) and `copyLocation` actions from the data-mover topic (`:993`) — so entries of the two kinds serialized on the same queue key and, worse, deduplicated against each other, although they are distinct work. Each consumer now gets its own. Note this also means a replication entry and a `copyLocation` on the same `bucket/key` are no longer serialized against each other; that guard was weak, as `_applyTransitionRule` already skips objects whose replication status is `PENDING`/`PROCESSING`/`FAILED`.

**2. The dedupe key itself.** What it is made of, before and after:

| Component | Before | After | Why |
| --- | --- | --- | --- |
| bucket | no | yes | The object key alone is not unique: `index.html` in two buckets gave the same key, and the second action was dropped. |
| object key | yes | yes | |
| version | yes | yes | Identifies the object version — but is `undefined` for objects in non-versioned buckets, hence the eTag below. |
| eTag | read `contentMd5`, always `undefined` | yes | No producer has ever set `target.contentMd5`, in any version: the content of a `copyLocation` action is carried by `target.eTag`, so that component was dead and the read is now gone. Without it `bucket/key` is a mutable identity — a non-versioned object overwritten between two scans collides with its own pending action, and the pending one then fails `CopyLocationTask._checkObjectState` (`:845-853`) with &#34;object contents have changed&#34;, so neither action transitions the object. Quotes are stripped so the same content always yields the same key. |
| `toLocation` | no | no | Not needed: `processDataMoverEntry` only builds a task when `toLocation === this.site` (`:980`), so it is constant within a given scheduler. |

The `ObjectQueueEntry` branch is unchanged: it already carried the bucket (via `getCanonicalKey()`) and a real `contentMd5`.

Nothing to migrate: a dedupe cache is per scheduler and in-memory, cleared when the task ends, and the key is derived from attributes already published.

Tasks skipped by deduplication are now logged, which was so far silent. The logger takes the constructor slot of a concurrency parameter that `TaskScheduler` never had — that limit is enforced by the consumer, in `BackbeatConsumer._getAvailableSlotsInPipeline()`.

Follow-up in BB-855: rather than letting a stale action run and fail its state check, deduplication could replace the queued task with the newer one and keep only the most up-to-date message. That needs `TaskScheduler` to support replacing a queued task, which it cannot do today.

Issue: BB-810

### DIFF
--- a/extensions/replication/queueProcessor/taskSchedulerHelpers.js
+++ b/extensions/replication/queueProcessor/taskSchedulerHelpers.js
@@ -14,15 +14,15 @@ function getTaskSchedulerQueueKey(entry) {
 
 function getTaskSchedulerDedupeKey(entry) {
     if (entry instanceof ObjectQueueEntry) {
-        const key = entry.getCanonicalKey();
+        const canonicalKey = entry.getCanonicalKey();
         const version = entry.getVersionId();
         const contentMd5 = entry.getContentMd5();
-        return `${key}:${version || ''}:${contentMd5}`;
+        return `${canonicalKey}:${version || ''}:${contentMd5}`;
     }
     if (entry instanceof ActionQueueEntry) {
-        const { key, version, contentMd5 } =
-              entry.getAttribute('target');
-        return `${key}:${version || ''}:${contentMd5}`;
+        const { bucket, key, version, eTag } = entry.getAttribute('target');
+        const objectETag = (eTag || '').replace(/"/g, '');
+        return `${bucket}/${key}:${version || ''}:${objectETag}`;
     }
     return undefined;
 }
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -161,7 +161,7 @@ class BackbeatConsumer extends EventEmitter {
             (ctx, done) => this._processTask(ctx.entry, done),
             orderByFunc,
             null,
-            this._concurrency,
+            this._log,
         );
 
         // Single drain callback fed by both signals (processing queue
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -264,7 +264,13 @@ class QueueProcessor extends EventEmitter {
             }
         }
 
-        this.taskScheduler = new TaskScheduler(
+        // The two consumers' entries must not deduplicate against each other
+        this.taskScheduler = this._createTaskScheduler();
+        this.dataMoverTaskScheduler = this._createTaskScheduler();
+    }
+
+    _createTaskScheduler() {
+        return new TaskScheduler(
             (ctx, done) => ctx.task.processQueueEntry(
                 ctx.entry, ctx.kafkaEntry, done),
             ctx => getTaskSchedulerQueueKey(ctx.entry),
@@ -989,9 +995,10 @@ class QueueProcessor extends EventEmitter {
             this.logger.debug('data mover entry is being pushed', {
                 entry: actionEntry.getLogInfo(),
             });
-            return this.taskScheduler.push({ task, entry: actionEntry,
-                                             kafkaEntry },
-                                           done);
+            return this.dataMoverTaskScheduler.push(
+                { task, entry: actionEntry, kafkaEntry },
+                done,
+            );
         }
         this.logger.debug('skip data mover entry', {
             entry: actionEntry.getLogInfo(),
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -274,7 +274,8 @@ class QueueProcessor extends EventEmitter {
             (ctx, done) => ctx.task.processQueueEntry(
                 ctx.entry, ctx.kafkaEntry, done),
             ctx => getTaskSchedulerQueueKey(ctx.entry),
-            ctx => getTaskSchedulerDedupeKey(ctx.entry));
+            ctx => getTaskSchedulerDedupeKey(ctx.entry),
+            this.logger);
     }
 
     _setupVaultclientCache() {
--- a/lib/tasks/TaskScheduler.js
+++ b/lib/tasks/TaskScheduler.js
@@ -29,11 +29,14 @@ class TaskScheduler {
      * @param {function} [getDedupeKeyFunc] - function to get the dedupe
      *   key of a task: getDedupeKeyFunc(ctx) -> {string} key
      *   (see: {@link TaskScheduler.push()})
+     * @param {Logger} [logger] - logger object, used to trace the tasks
+     *   skipped by deduplication
      */
-    constructor(processingFunc, getQueueKeyFunc, getDedupeKeyFunc) {
+    constructor(processingFunc, getQueueKeyFunc, getDedupeKeyFunc, logger) {
         this._processingFunc = processingFunc;
         this._getQueueKeyFunc = getQueueKeyFunc;
         this._getDedupeKeyFunc = getDedupeKeyFunc;
+        this._logger = logger;
         this._drainFunc = null;
         this._taskQueues = {};
         this._dedupeCache = {};
@@ -87,6 +90,9 @@ class TaskScheduler {
             }
             if (dedupeKey !== undefined) {
                 if (this._dedupeCache[dedupeKey]) {
+                    this._logger?.debug('task skipped: a task with the ' +
+                                        'same dedupe key is in progress ' +
+                                        'or queued', { dedupeKey });
                     return process.nextTick(done);
                 }
                 this._dedupeCache[dedupeKey] = true;
--- a/tests/unit/lib/tasks/TaskScheduler.spec.js
+++ b/tests/unit/lib/tasks/TaskScheduler.spec.js
@@ -360,6 +360,30 @@ describe('TaskScheduler', () => {
                 }, 10);
             });
         });
+        it('should log the tasks skipped by deduplication', done => {
+            let blockedTaskResolve;
+            const blockedTaskPromise = new Promise(resolve => { blockedTaskResolve = resolve; });
+            const processingFunc = sinon.stub().callsFake((ctx, cb) => {
+                if (ctx.id === 'block') {
+                    return blockedTaskPromise.then(() => cb());
+                }
+                return cb();
+            });
+            const logger = { debug: sinon.spy() };
+            const taskScheduler = new TaskScheduler(
+                processingFunc, null, ctx => ctx.dedupeKey, logger);
+
+            taskScheduler.push({ id: 'block', dedupeKey: 'key' }, () => {});
+            taskScheduler.push({ dedupeKey: 'key' }, () => {});
+
+            process.nextTick(() => {
+                assert(logger.debug.calledOnce);
+                assert.deepStrictEqual(logger.debug.firstCall.args[1],
+                                       { dedupeKey: 'key' });
+                blockedTaskResolve();
+                done();
+            });
+        });
         it('should remove dedup key after task completion', done => {
             const processingFunc = sinon.stub().yields();
             const taskScheduler = new TaskScheduler(processingFunc, ctx => ctx.queueKey, null);
--- a/tests/unit/replication/QueueProcessor.spec.js
+++ b/tests/unit/replication/QueueProcessor.spec.js
@@ -305,6 +305,36 @@ describe('Queue Processor', () => {
         });
     });
 
+    describe('processDataMoverEntry', () => {
+        beforeEach(() => {
+            qp.taskScheduler = { push: sinon.stub().callsArgWith(1, null) };
+            qp.dataMoverTaskScheduler = {
+                push: sinon.stub().callsArgWith(1, null),
+            };
+            qp._mProducer = { getProducer: () => null };
+        });
+
+        it('dispatches copyLocation actions to the data mover scheduler',
+        done => {
+            const kafkaEntry = {
+                value: JSON.stringify({
+                    action: 'copyLocation',
+                    toLocation: 'site-crr',
+                    target: {
+                        bucket: 'src-bucket',
+                        key: 'obj',
+                        eTag: '"d41d8cd98f00b204e9800998ecf8427e"',
+                    },
+                }),
+            };
+            qp.processDataMoverEntry(kafkaEntry, () => {
+                sinon.assert.calledOnce(qp.dataMoverTaskScheduler.push);
+                sinon.assert.notCalled(qp.taskScheduler.push);
+                done();
+            });
+        });
+    });
+
     describe('constructor', () => {
         it('should use s3c site\'s host as a destination host', () => {
             const config = getQueueProcessorConfig();
--- a/tests/unit/replication/taskSchedulerHelpers.spec.js
+++ b/tests/unit/replication/taskSchedulerHelpers.spec.js
@@ -7,54 +7,63 @@ const { ObjectMD } = require('arsenal').models;
 const { getTaskSchedulerQueueKey, getTaskSchedulerDedupeKey } = require(
     '../../../extensions/replication/queueProcessor/taskSchedulerHelpers');
 
-function makeObjectQueueEntry({ key, versionId, contentMd5 }) {
+function makeObjectQueueEntry({ bucket, key, versionId, contentMd5 }) {
     const objMd = new ObjectMD().setKey(key);
-    return new ObjectQueueEntry('test-bucket-name', key, objMd)
+    return new ObjectQueueEntry(bucket, key, objMd)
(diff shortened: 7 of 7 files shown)


extensions/replication/queueProcessor/taskSchedulerHelpers.js
```diff
@@ -13,16 +13,21 @@ function getTaskSchedulerQueueKey(entry) {
 }
 
 function getTaskSchedulerDedupeKey(entry) {
+    // Both entry types share a single dedupe cache: keep the
+    // namespaces of their keys disjoint
```
DarkIsDude, line 17:
```suggestion
```

extensions/replication/queueProcessor/taskSchedulerHelpers.js
```diff
@@ -13,16 +13,21 @@ function getTaskSchedulerQueueKey(entry) {
...
-        return `${key}:${version || ''}:${contentMd5}`;
+        return `object:${key}:${version || ''}:${contentMd5}`;
     }
     if (entry instanceof ActionQueueEntry) {
-        const { key, version, contentMd5 } =
+        const { bucket, key, version, contentMd5, eTag } =
               entry.getAttribute('target');
-        return `${key}:${version || ''}:${contentMd5}`;
+        // contentMd5 and eTag hold the same value, the latter quoted
+        const content = (contentMd5 || eTag || '').replace(/"/g, '');
```
DarkIsDude, line 28:
```suggestion
        const content = contentMd5 || eTag.replace(/"/g, '') || '';
```

Better then ?

DarkIsDude (approved):

extensions/replication/queueProcessor/taskSchedulerHelpers.js
francoisferrand, line 1: (not specific to this file - using inline comment to allow threaded replies...)

> No producer sets target.contentMd5; createCopyLocationAction sets target.eTag, which holds the same value, quoted. Taken from either attribute, unquoted. Without it bucket/key is a mutable identity: a non-versioned object overwritten between two scans collides with its own pending action, and the pending one then fails CopyLocationTask._checkObjectState with "object contents have changed", so neither action transitions the object.

I don't understand this, can you clarify?

Do you mean there is a second problem, i.e. `createCopyLoation` may be used in non-versioned buckets, where `bucket/key` is not "unique" (it may relate to different revision/content of the object, if it was replaced) - yet dedup would engage anyway?
→ Is this really an issue? Dedup happens only if both entries are in the queue I guess (so nothing processed yet); and there is only a single "object" to process anyway (if it was replaced, we don't have access to previous content anymore): so whatever message is processed with replicate whatever is there?
→ Or do we have an extra guard when processing the message, to ensure we only replicate exactly what we tried to replicate? (i.e. we actually expect to fail the first replication with incorrect eTag; and succeed the second one _if it happens_ - to avoid some potential race conditions maybe?). In that case dedup should actually still happen - but make sure we keep the latest message only (that is the one with the most "up to date" eTag, so the only one which may succeed) ?

francoisferrand, line 1: > Both entry types share a single dedupe cache. Now that both keys carry an unquoted content, this prefix is what keeps a replication entry and a copyLocation on the same object version apart — they are distinct work.

What do you mean "distinct work" ?

Either they are in the same kafka topic / BackbeatConsumer → they are the same work (a single entry may be replicated to multiple destinations) ; or they are indeed distinct work (separate topic / consumer group), should use a distinct dedup cache instead?

extensions/replication/queueProcessor/taskSchedulerHelpers.js
```diff
@@ -13,16 +13,21 @@ function getTaskSchedulerQueueKey(entry) {
 }
 
 function getTaskSchedulerDedupeKey(entry) {
```
francoisferrand, line 15: can we consistently & systematically use "eTag" instead of "content" ?

That is _the_ actual business vocabulary here, unambiguously:
- content could be the data or just about anything;
- contentMD5 is not necessarily the eTag, esp. in case of MPU;
- `eTag` is the S3 standard field designed explicitly to identify "changes" to the content (however it may be computed)

extensions/replication/queueProcessor/taskSchedulerHelpers.js
```diff
@@ -13,16 +13,21 @@ function getTaskSchedulerQueueKey(entry) {
...
         const contentMd5 = entry.getContentMd5();
-        return `${key}:${version || ''}:${contentMd5}`;
+        return `object:${key}:${version || ''}:${contentMd5}`;
     }
     if (entry instanceof ActionQueueEntry) {
-        const { key, version, contentMd5 } =
+        const { bucket, key, version, contentMd5, eTag } =
               entry.getAttribute('target');
-        return `${key}:${version || ''}:${contentMd5}`;
+        // contentMd5 and eTag hold the same value, the latter quoted
```
francoisferrand: so the eTag/contentMd5 was already (theorically) handled - just not working consistently ? Not a real _change_ (like the bucket part), but more of a fix to standardize the format and fallback to eTag?

In which case would we have eTag but not contentMd5 ?

extensions/replication/queueProcessor/taskSchedulerHelpers.js
```diff
@@ -13,16 +13,21 @@ function getTaskSchedulerQueueKey(entry) {
...
 
 function getTaskSchedulerDedupeKey(entry) {
+    // Both entry types share a single dedupe cache: keep the
+    // namespaces of their keys disjoint
     if (entry instanceof ObjectQueueEntry) {
         const key = entry.getCanonicalKey();
         const version = entry.getVersionId();
         const contentMd5 = entry.getContentMd5();
-        return `${key}:${version || ''}:${contentMd5}`;
+        return `object:${key}:${version || ''}:${contentMd5}`;
```
francoisferrand: shouldn't we have ${bucket} here as well?

extensions/replication/queueProcessor/taskSchedulerHelpers.js
francoisferrand: Also wondering about `Both entry types share a single dedupe cache` : the code indeed supports both kinds of message, but I don't think (at least I hope...) that a single kafka topic can actually contains both.

So each _instance_ if the cache would only get a single type, and the "content" prefix is really just defensive, with no real-life impact?

lib/BackbeatConsumer.js
```diff
@@ -161,7 +161,7 @@ class BackbeatConsumer extends EventEmitter {
             (ctx, done) => this._processTask(ctx.entry, done),
             orderByFunc,
             null,
-            this._concurrency,
```
francoisferrand, line 164: not your change, but it is weird that `_concurrency` was not used ; can you double check what happened (in git history), to make sure we did not unexpectedly break something there? (i.e. it is really just an incomplete change, not a broken feature...)

extensions/replication/queueProcessor/taskSchedulerHelpers.js
```diff
@@ -13,16 +13,21 @@ function getTaskSchedulerQueueKey(entry) {
...
     }
     if (entry instanceof ActionQueueEntry) {
-        const { key, version, contentMd5 } =
+        const { bucket, key, version, contentMd5, eTag } =
```
francoisferrand: is there always a `bucket` ? (i.e. accross all uses of "ActionQueueEntry")
does it degrade cleanly if that is not the case?

DarkIsDude:
Reading back Francois review, whats happens with old data and new data ? I mean, can we have conflicts and/or dedup not working because the key changes at deployment time ?

extensions/replication/queueProcessor/taskSchedulerHelpers.js
delthas: Done, removed.

extensions/replication/queueProcessor/taskSchedulerHelpers.js
delthas: That form throws a `TypeError` when `target.eTag` is unset: `.replace` is evaluated before the `|| ''` can catch it, and unset eTags do occur (`tests/functional/replication/queueProcessor.js:730` builds a `copyLocation` action without one).

It is moot now anyway: `contentMd5` is never set by any producer, so the read is gone and it reads `(eTag || '').replace(/"/g, '')`.

extensions/replication/queueProcessor/taskSchedulerHelpers.js
delthas: Adopted, the local is now `objectETag`.

extensions/replication/queueProcessor/taskSchedulerHelpers.js
delthas: It is already there: `getCanonicalKey()` is `${bucket}/${key}` (`lib/models/ObjectQueueEntry.js:152-154`). I renamed the local to `canonicalKey` so that reads as such.

extensions/replication/queueProcessor/taskSchedulerHelpers.js
delthas: Only `createCopyLocationAction` produces this target shape, and it always sets `bucket` from `params.bucketName`. The other `ActionQueueEntry` producers (`deleteData`, `deleteObject`, `deleteMPU`, `deleteArchivedSourceData`) go to other topics and processors, and `processDataMoverEntry` only builds a task for `copyLocation` with `toLocation === this.site`.

The assumption is pre-existing too: `getTaskSchedulerQueueKey` already builds `${bucket}/${key}` for actions. And it degrades cleanly — a missing bucket gives `undefined/<key>`, i.e. the grouping we had before, rather than throwing.

extensions/replication/queueProcessor/taskSchedulerHelpers.js
delthas: Always — no producer has ever set `target.contentMd5`, in any version (zero occurrences in `ReplicationAPI.js` on 7.70, 8.6, 9.0 and 9.5, and `git log --all -S"contentMd5"` on that file returns no commit).

So it was not inconsistent, it was dead: the third component of the key was the literal string `undefined` for every action ever published. I removed the read entirely, and the key now takes the eTag.

extensions/replication/queueProcessor/taskSchedulerHelpers.js
delthas: Yes, this is a second problem, and your second reading is the right one: there *is* an extra guard, and that is exactly what makes the drop harmful.

`CopyLocationTask._checkObjectState` compares `target.eTag` with the object's current `content-md5` and rejects a mismatch with `InvalidObjectState` ("object contents have changed"). So an action published *before* the overwrite can never succeed — only the most recent one can. Without the eTag in the dedupe key, the two actions collide, dedupe keeps the **older** one and drops the newer: the old one then fails its own state check and is skipped as committable, and nothing transitions the object until `LifecycleResetTransitionInProgressTask` clears the stale flag. With the eTag, both run — the stale one costs one metadata read and is skipped, the newest one succeeds.

The preconditions are narrow, to be clear: a non-versioned bucket, the object re-qualifying after the overwrite (immediate with a `Date`-based rule, since `getTransitionTimestamp` ignores `LastModified`), and a data-mover backlog spanning a scan interval.

Your "keep the latest message only" is strictly better than letting both run, and I filed BB-855 for it. It needs `TaskScheduler` to replace an already-queued task, which it cannot do today, and the class is shared by every consumer — so it did not feel right to fold into this fix.

delthas: You are right that a single topic holds a single type — but the cache is per *process*, not per topic, and one process consumes both topics.

The `backbeat-replication-data-processor` deployment runs `npm run queue_processor` with no topic argument (zenko-operator `pkg/controller/zenko/reconcile_backbeat_replication_data_processor_deployment.go:87-96`). `QueueProcessor` then creates `_consumer` on the replication topic (`:744`) **and** `_dataMoverConsumer` on the data-mover topic (`:757`), and both push into the single `this.taskScheduler` (`:934` for replication entries, `:991` for actions). The replay processor is the exception: it gets a topic argument, so `:752` skips the data-mover consumer and it only ever sees `ObjectQueueEntry`.

That said, I agree on the practical impact: a live collision would also need the same object/version/eTag in flight on both paths at once, and `_applyTransitionRule` skips objects whose replication status is `PENDING`/`PROCESSING`/`FAILED`. So this is hygiene rather than a bug I can demonstrate.

lib/BackbeatConsumer.js
delthas: Checked: the argument is simply unused — `TaskScheduler` never had a fourth parameter. The limit itself was moved, not lost.

Before BB-645 the processing queue was `async.queue(this._queueProcessor, this._concurrency)`, so the concurrency lived in the queue. #2626 replaced it with `TaskScheduler` and moved the bound into `BackbeatConsumer._getAvailableSlotsInPipeline()`, which now gates how many messages are fetched: `this._concurrency - this._processingQueue.running() - this._nConsumePendingRequests`. Your own thread on that PR (r1995146317) is where it was settled.

So nothing is broken, only the argument was left behind at the call site — and the logger now takes its slot.

delthas:
> whats happens with old data and new data ? I mean, can we have conflicts and/or dedup not working because the key changes at deployment time ?

No conflict is possible: the dedupe cache is per-process and in-memory, and it only holds tasks that are currently queued or in flight. Keys are computed at push time, from the message, by whichever process consumes it — they are never persisted, and never compared between processes. A process that restarts or is upgraded starts with an empty cache.

Mixed producer versions are not a problem either. Nothing has ever written `target.contentMd5` — zero occurrences in `ReplicationAPI.js` on 7.70, 8.6, 9.0 and 9.5, and `git log --all -S"contentMd5"` on that file returns no commit — so there is no legacy message shape to be compatible with, and the fields the key now uses (`bucket`, `key`, `version`, `eTag`) have always been published by `createCopyLocationAction`.

The worst case during a rolling upgrade is that an old and a new process each deduplicate within their own window, so a duplicate pair could be processed twice instead of once. That is the at-least-once behaviour the pipeline already handles: `CopyLocationTask._checkObjectState` skips an object that has already been transitioned.

delthas:
Moved to 9.6.

extensions/replication/queueProcessor/taskSchedulerHelpers.js
francoisferrand: > the cache is per process, not per topic, and one process consumes both topics.

question still stands: should we keep the cache per-process, and use some "trick" to somehow partition it (as you did), or just make sure we create multiple caches, e.g. one for each actual consumer?

- if we have multiple consumers, the cache should not invalidate "cross-consumer" anyway, AFAIK
- we have multiple consumers in the process indeed; but do we have multiple consumers with different _types_ ?

also looking at the code I see we create an instance of `TaskScheduler` in each `BackbeatConsumer` and in `QueueProcessor` : so it does look like it is not per-process, but per consumer indeed ?

francoisferrand (changes requested):

extensions/replication/queueProcessor/taskSchedulerHelpers.js
delthas: You're right. Moved each consumer to its own scheduler and dropped the prefix. (First commit).

First, my "per process" was sloppy, and you are right that there are two layers:

- each `BackbeatConsumer` builds its own `TaskScheduler` (`this._processingQueue`), but with `null` as the dedupe function (`BackbeatConsumer.js:163`) — those only order by key, they never deduplicate;
- the dedupe cache in question was the separate one in `QueueProcessor` (`:267`), and `task.js:179-236` builds one `QueueProcessor` per destination site, so it was per site, not per process.

> we have multiple consumers in the process indeed; but do we have multiple consumers with different _types_ ?

That was the case, within one `QueueProcessor`: `_consumer` on the replication topic pushed `ObjectQueueEntry` (`:936`) and `_dataMoverConsumer` on the data-mover topic pushed `ActionQueueEntry` (`:993`), into that same scheduler

delthas:
Moved each consumer to its own scheduler and dropped the prefix. (First commit).

francoisferrand (approved):

delthas:
!done 4h

delthas:
[`BB-810`](https://scality.atlassian.net/browse/BB-810): logged `4h`; transitioned to Done (`Fixed`)

PR #2805, opened 2026-08-18
Copy location task reads source location with s3 sdk with STS

SylvainSenechal:
This pr handles the necessary modification to copyLocation task, so that the clean room can use it to pull object data from the production site.
The main change is basically modifying the way we set up the s3 clients in copyLocationTask, depending on the location of the object. This requires storing and exposing the location configuration which we didn't do before.

Issue: BB-812

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -15,6 +15,7 @@ const RoundRobin = require('arsenal').network.RoundRobin;
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
 const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
 const VaultClientCache = require('../../../lib/clients/VaultClientCache');
+const CredentialsManager = require('../../../lib/credentials/CredentialsManager');
 const QueueEntry = require('../../../lib/models/QueueEntry');
 const TaskScheduler = require('../../../lib/tasks/TaskScheduler');
 const { getTaskSchedulerQueueKey,
@@ -228,6 +229,12 @@ class QueueProcessor extends EventEmitter {
 
         this.logger = new Logger(
             `Backbeat:Replication:QueueProcessor:${this.site}`);
+
+        this.assumedRoleCredentialsManager = new CredentialsManager(
+            'replication-copy-location', this.logger);
+        this.assumedRoleS3Clients = {};
+        this.assumedRoleHTTPAgent = new HttpAgent.Agent({ keepAlive: true });
+        this.assumedRoleHTTPSAgent = new HttpsAgent.Agent({ keepAlive: true });
 
         // global variables
         if (sourceConfig.transport === 'https') {
@@ -690,6 +697,10 @@ class QueueProcessor extends EventEmitter {
             destHTTPAgent: this.destHTTPAgent,
             vaultclientCache: this.vaultclientCache,
             accountCredsCache: this.accountCredsCache,
+            assumedRoleCredentialsManager: this.assumedRoleCredentialsManager,
+            assumedRoleS3Clients: this.assumedRoleS3Clients,
+            assumedRoleHTTPAgent: this.assumedRoleHTTPAgent,
+            assumedRoleHTTPSAgent: this.assumedRoleHTTPSAgent,
             replicationStatusProducer: this.replicationStatusProducer,
             mProducer: this._mProducer,
             logger: this.logger,
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -15,7 +15,6 @@ const RoundRobin = require('arsenal').network.RoundRobin;
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
 const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
 const VaultClientCache = require('../../../lib/clients/VaultClientCache');
-const CredentialsManager = require('../../../lib/credentials/CredentialsManager');
 const QueueEntry = require('../../../lib/models/QueueEntry');
 const TaskScheduler = require('../../../lib/tasks/TaskScheduler');
 const { getTaskSchedulerQueueKey,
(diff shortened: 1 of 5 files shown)


extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -3,12 +3,14 @@ const { v4: uuid } = require('uuid');
 
 const { errors, jsutil, models } = require('arsenal');
 const { ObjectMD } = models;
+const { S3Client: AwsS3Client, GetObjectCommand: AwsGetObjectCommand } =
```
SylvainSenechal: Can discuss these renaming
It's just that we are using getObject from both the official aws SDK, and from our cloudserverClient

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -229,6 +230,12 @@ class QueueProcessor extends EventEmitter {
         this.logger = new Logger(
             `Backbeat:Replication:QueueProcessor:${this.site}`);
 
+        this.assumedRoleCredentialsManager = new CredentialsManager(
```
SylvainSenechal: Not sure if this is the best place : queueProcessor does create a new copyLocation Task for each kafka entry so we can't put these in CopyLocationTask class otherwise its useless, but we may also have multiple instances of queue processor depending on the config which means we would have multiple credentials managers/clients for the same authentification

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +105,114 @@ class CopyLocationTask extends BackbeatTask {
...
+     * or directly from a CRR source location's own S3
+     * endpoint via an assumed role.
+     * @param {ActionQueueEntry} actionEntry - the action entry
+     * @param {ObjectMD} objMD - metadata object
+     * @param {Object} [range] - byte range to request, or undefined for the whole object
+     * @param {Werelogs} log - the logger instance
+     * @param {AbortController} abortController - abort controller for the GET request
+     * @return {Promise} resolves to the GetObject response
+     */
+    async _sendGetObject(actionEntry, objMD, range, log, abortController) {
```
SylvainSenechal, line 235: For those who wanna understand the pr's main idea :

Before : We only used our backbeatClient (cloudserver client) to get object, that backbeat client asked cloudserver to deal with reading the data based on the location.

But Cloudserver is not capable of reading data from external CRR locations.

Now : When the object's location is "CRR", we use a classic S3 client with sts to directly get the data without going through cloudserver

lib/management/operatorBackend.js
```diff
@@ -78,6 +78,7 @@ function initManagement(params, done) {
         }));
     const locations = require('../../conf/locationConfig.json') || {};
 
+    config.setLocationConstraints(locations);
```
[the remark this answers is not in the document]
SylvainSenechal: I don't think this is relevant, we don't really care bout backbeat running in orbit mode 🤔

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +105,117 @@ class CopyLocationTask extends BackbeatTask {
...
```
francoisferrand, line 235: this function does 2 things, which are orthogonal concerns:
- it retrieves/builds a client, depending on location : either the (exisitng/global) backbeatClient, or the new STS client if required.
- it send the actual command

it would seem more appropriate to split responsability, and introduce a "getClient" function - then fallthrough to the existing code.......but I see you don't pass the same parameter, to the command: see https://github.com/scality/backbeat/pull/2805/changes#r3819951224

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +105,117 @@ class CopyLocationTask extends BackbeatTask {
...
+                err.retryable = true;
+                throw err;
+            }
+            const s3Client = this._getAssumedRoleS3Client(locationConfig, part.role, log);
+            const command = new AwsGetObjectCommand({
+                Bucket: part.bucket,
+                Key: objMD.getKey(),
+                VersionId: part.dataStoreVersionId,
+                Range: range && `bytes=${range.start}-${range.end}`,
+            });
```
francoisferrand: the parameters should be mostly the same as the regular ones:
- `bucket`, `key` and `version` shoud come from the message's `actionEntry.getAttribute('target')`, same as the other case
- need to pass the `requestUids` as well

the only difference is the locationConstraint indeed.....and the object type AwsGetObjectCommand vs BackbeatRoutesGetObjectCommand
- I wonder if/why `locationConstraint` is used here, should not be needed most of the time - but maybe in case of transient location or similarly advanced/corner case. So indeed should not be changed I guess (for the regular transition path), let's not take risk ; and must not be added to the STS path indeed.
- the "target" of CopyLocationTask is not a random S3 server, this is really a cloudserver. So we can and should use CloudServerClient and our own 'extensions' (RequestUids)

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +105,117 @@ class CopyLocationTask extends BackbeatTask {
...
+        const locationConfig = config.getLocationConstraint(objMD.getDataStoreName());
+        if (locationConfig?.isCRR === true) {
+            const locations = objMD.getLocation();
+            const part = locations && locations[0];
+            if (!part || !part.role) {
+                const err = errors.AccessDenied.customizeDescription(
+                    'missing role on location part for isCRR source location');
+                err.retryable = true;
+                throw err;
+            }
```
francoisferrand: I don't have the full design in mind anymore, but I think params should be in the kafka message instea?

when we create the message, we will anyway parse the `ObjMD` : so this location validation/... would better be done there only (in particular adding the role to `actionEntry.getAttribute('target')`), so in CopyLocationTask we only need to do it.

this also matches the current 'design' of CopyLocationTask : it copies data from `actionEntry.getAttribute('target')` to the object's location, and does try to check what is currently in ObjMD's location - which may (or may not) help to avoid some race conditions, not sure...

(STS could even be 'triggered' not just by `locationConfig?.isCRR` but by `actionEntry.getAttribute('target').role` ? Either way, will need to access the location for creds/hosts...)

What do you think? Can you evaluate impact and confirm?

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +105,117 @@ class CopyLocationTask extends BackbeatTask {
...
+     * Get a cached S3 client, authenticated with the assumed-role
+     * credentials for the role carried on a location part.
+     * @param {Object} locationConfig - the location's config
+     * @param {String} roleArn - the role ARN carried by the location part
+     * @param {Werelogs} log - the logger instance
+     * @return {AwsS3Client} the client
+     * @throws {ArsenalError} AccessDenied (retryable) if credentials
+     * could not be obtained for the role
+     */
+    _getAssumedRoleS3Client(locationConfig, roleArn, log) {
```
francoisferrand: don't we have this (or similar) function in CRR ?
can't we dedup and use the same function?

extensions/replication/tasks/CopyLocationTask.js
francoisferrand: best to read from ObjMD, so we always act on fresh data (e.g. in case of message replay...)

extensions/replication/tasks/CopyLocationTask.js
francoisferrand: my bad, shoudl indeed come from the ObjMD.

PR #2806, opened 2026-08-18
Retry the setup of an ingestion reader when it fails

benzekrimaha:
When a reader's setup fails (typically the getRaftId call to the source), it was dropped from the pending list and never set up again: the reader stays registered in _ingestionSources, so applyUpdates considers the bucket already handled and only refreshes it. The bucket was therefore left out of ingestion until the pod was restarted.

Queue the reader again instead, so it is retried on the next cycle. This follows what _processLogReaderEntries already does with read failures, and keeps the other locations ingesting, which restarting the pod on a failed healthcheck would not.

The reader is only queued again if its source is still configured, as it may have been removed while its setup was in progress.

Issue: BB-846

### DIFF
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -621,6 +621,11 @@ class IngestionPopulator {
             // its source may have been removed or replaced in the meantime. It
             // must then neither be activated nor retried.
             if (this._ingestionSources[zenkoBucket] !== logReader) {
+                this.log.info('skipping log reader after setup: source was removed or replaced', {
+                    method: 'IngestionPopulator._setupUpdatedReaders',
+                    zenkoBucket,
+                    location: logReader.getLocationConstraint(),
+                });
                 return cb();
             }
             if (err) {
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -619,10 +619,21 @@ class IngestionPopulator {
             if (err) {
                 // if setup fails for a log reader, don't add it to `logReaders`
                 // log the error and continue setting up others
-                this.log.fatal('error setting up log reader', {
+                const zenkoBucket = logReader.getTargetZenkoBucketName();
+                this.log.error('error setting up log reader, retrying later', {
                     method: 'IngestionPopulator._setupUpdatedReaders',
+                    zenkoBucket,
+                    location: logReader.getLocationConstraint(),
                     error: err,
                 });
+                // Setup failures are usually transient (source unreachable,
+                // invalid credentials...): queue the reader again so that it
+                // is retried on the next cycle, unless it is no longer the
+                // reader registered for its bucket, meaning its source was
+                // removed or replaced in the meantime.
+                if (this._ingestionSources[zenkoBucket] === logReader) {
+                    this.logReadersUpdate.push(logReader);
+                }
             } else {
                 this.logReaders.push(logReader);
             }
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -616,10 +616,16 @@ class IngestionPopulator {
         const newReaders = this.logReadersUpdate;
         this.logReadersUpdate = [];
         async.each(newReaders, (logReader, cb) => logReader.setup(err => {
+            const zenkoBucket = logReader.getTargetZenkoBucketName();
+            // The reader is in neither list while its setup is in flight, so
+            // its source may have been removed or replaced in the meantime. It
+            // must then neither be activated nor retried.
+            if (this._ingestionSources[zenkoBucket] !== logReader) {
+                return cb();
+            }
             if (err) {
                 // if setup fails for a log reader, don't add it to `logReaders`
                 // log the error and continue setting up others
-                const zenkoBucket = logReader.getTargetZenkoBucketName();
                 this.log.error('error setting up log reader, retrying later', {
                     method: 'IngestionPopulator._setupUpdatedReaders',
                     zenkoBucket,
@@ -628,12 +634,8 @@ class IngestionPopulator {
                 });
                 // Setup failures are usually transient (source unreachable,
                 // invalid credentials...): queue the reader again so that it
-                // is retried on the next cycle, unless it is no longer the
-                // reader registered for its bucket, meaning its source was
-                // removed or replaced in the meantime.
-                if (this._ingestionSources[zenkoBucket] === logReader) {
-                    this.logReadersUpdate.push(logReader);
-                }
+                // is retried on the next cycle.
+                this.logReadersUpdate.push(logReader);
             } else {
                 this.logReaders.push(logReader);
             }
@@ -735,6 +737,12 @@ class IngestionPopulator {
      * @return {undefined}
      */
     _closeLogState(key) {
+        // Unregister the source first: a reader whose setup is in flight is in
+        // neither list, and leaving its entry here would let it register itself
+        // once the setup completes.
+        const reader = this._ingestionSources[key];
+        delete this._ingestionSources[key];
+
         if (this._checkAndRemoveLogReader(key, this.logReadersUpdate)) {
             // if removed from `this.logReadersUpdate`, zookeeper setup has not
             // been performed yet
@@ -742,7 +750,6 @@ class IngestionPopulator {
                 method: 'IngestionPopulator._closeLogState',
                 bucket: key,
             });
-            delete this._ingestionSources[key];
         }
         if (this._checkAndRemoveLogReader(key, this.logReaders)) {
             // if removed from `this.logReaders`, we must cleanup zookeeper
@@ -753,8 +760,6 @@ class IngestionPopulator {
             });
             // we should first validate this reader is not currently processing
             // a batch of entries before removing zookeeper state
-            const reader = this._ingestionSources[key];
-            delete this._ingestionSources[key];
             const path = `${this.ingestionConfig.zookeeperPath}/${key}`;
             this._removeReaderState(reader, path);
         }
--- a/tests/unit/ingestion/IngestionPopulator.js
+++ b/tests/unit/ingestion/IngestionPopulator.js
@@ -278,6 +278,102 @@ describe('Ingestion Populator', () => {
         });
     });
 
+    describe('_setupUpdatedReaders', () => {
+        const FAILING_BUCKET = 'failing-zenko-bucket';
+        const WORKING_BUCKET = 'working-zenko-bucket';
+
+        /**
+         * @param {string} zenkoBucket - target zenko bucket of the reader
+         * @param {Error|null} setupError - error to fail `setup` with
+         * @return {object} the stubbed reader
+         */
+        function createLogReaderMock(zenkoBucket, setupError) {
+            const logReader = sinon.createStubInstance(IngestionReader);
+            logReader.getTargetZenkoBucketName.returns(zenkoBucket);
+            logReader.setup.yieldsAsync(setupError);
+            return logReader;
+        }
+
+        beforeEach(() => {
+            ip.logReaders = [];
+            ip.logReadersUpdate = [];
+        });
+
+        it('should activate a log reader once its setup succeeds', done => {
+            const logReaderMock = createLogReaderMock(WORKING_BUCKET, null);
+            ip._ingestionSources[WORKING_BUCKET] = logReaderMock;
+            ip.logReadersUpdate = [logReaderMock];
+
+            ip._setupUpdatedReaders(err => {
+                assert.ifError(err);
+                assert.deepStrictEqual(ip.logReaders, [logReaderMock]);
+                assert.deepStrictEqual(ip.logReadersUpdate, []);
+                done();
+            });
+        });
+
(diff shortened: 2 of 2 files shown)


lib/queuePopulator/IngestionPopulator.js
```diff
@@ -618,12 +618,22 @@ class IngestionPopulator {
...
+                // sources that are already registered, so a reader dropped
+                // here would never be set up again.
+                const zenkoBucket = logReader.getTargetZenkoBucketName();
+                this.log.error('error setting up log reader, retrying later', {
                     method: 'IngestionPopulator._setupUpdatedReaders',
+                    zenkoBucket,
                     error: err,
                 });
+                // the source may have been removed from the configuration
+                // while its setup was in progress
```
francoisferrand: this comment is misleading: it is not that it "may have been removed", it is certain that is was removed c.f. the beginning of the function:

```
        const newReaders = this.logReadersUpdate;
        this.logReadersUpdate = [];
```

→ here you want to explain why you re-add it
```suggestion
                // Setup failures are usually transient (source unreachable,
                // invalid credentials...): queue the reader again so that it
                // is retried on the next cycle
```

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -618,12 +618,22 @@ class IngestionPopulator {
...
         async.each(newReaders, (logReader, cb) => logReader.setup(err => {
             if (err) {
-                // if setup fails for a log reader, don't add it to `logReaders`
-                // log the error and continue setting up others
-                this.log.fatal('error setting up log reader', {
+                // Setup failures are usually transient (source unreachable,
+                // invalid credentials...). Queue the reader again so that it
+                // is retried on the next cycle: `applyUpdates` only refreshes
+                // sources that are already registered, so a reader dropped
+                // here would never be set up again.
```
francoisferrand: not relevant to the code: the old comment was -and still is- correct, we don't add this one to logReaders and continue setting up others...

```suggestion
                // if setup fails for a log reader, don't add it to `logReaders`
                // log the error and continue setting up others
```

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -618,12 +618,22 @@ class IngestionPopulator {
...
+                const zenkoBucket = logReader.getTargetZenkoBucketName();
+                this.log.error('error setting up log reader, retrying later', {
                     method: 'IngestionPopulator._setupUpdatedReaders',
+                    zenkoBucket,
```
francoisferrand: I think we should log the _location_ instead of the bucket, seems more appropriate - or in addition to the bucket, for consistency with other logs in this file:

```suggestion
                this.log.error('error setting up log reader, retrying later', {
                    method: 'IngestionPopulator._setupUpdatedReaders',)
                    zenkoBucket: logReader.getTargetZenkoBucketName(),
                    location: logReader.getLocationConstraint(),
```

francoisferrand:
not the right target branch: no point fixing so far back in time: from 9.5 is enough
(also this is not an improvement, but a bug. so branch name is not right)

lib/queuePopulator/IngestionPopulator.js
benzekrimaha: Restored.

lib/queuePopulator/IngestionPopulator.js
benzekrimaha: Done, `location` sits next to `zenkoBucket` now. I kept the `zenkoBucket` const since the guard right below reads it.

lib/queuePopulator/IngestionPopulator.js
benzekrimaha: Reworded to say why the reader is queued again.

One precision : that comment was about `_ingestionSources`, the registry of configured sources, not about `logReadersUpdate` which the function drains at the top. Those are two different structures, and the guard reads the former.
It was imprecise for another reason though. `_closeLogState` only deletes the `_ingestionSources` entry when it finds the reader in `logReadersUpdate` or in `logReaders`, and while its setup is in progress the reader is in neither, so a source removed right then still leaves its entry behind. What the guard actually catches is a reader that is no longer the one registered for its bucket, which is what the new comment says.

benzekrimaha:
Retargeted to `development/9.5` and rebased on it, so the diverged-branch warning is gone.

On the branch prefix: BB-846 is already typed Bug in Jira, only the branch name says otherwise. Renaming it means closing this PR and opening a new one, which also moves the base of #2821 stacked on top, so I left the branch name as is.

francoisferrand:
What happens when the location is deleted?

Please confirm there no risk of path or race condition where we would keep retrying to setup a location which was removed...

benzekrimaha:
Fixed here as suggested, in a second commit.
`_closeLogState` now unregisters the source unconditionally, which makes `_ingestionSources` authoritative, and the guard moved above the error branch so it covers the success path too.
The point about the extra ingestion round is addressed rather than just bounded: the source is unregistered as soon as it is removed, so a reader whose setup was in flight is neither queued again nor activated. The extra round does not happen at all.
Additionnal validation made by reverting the fix while keeping the tests gives 6 failures and 24 passes, and the 6 failures are exactly the new cases, the pre-existing behaviours pass either way.
One detail worth flagging for future readers: the active-reader test pins that `_removeReaderState` receives the reader object. The reader has to be read before its entry is deleted, since `_removeReaderState` polls it on a timer until its batch completes, so a later simplification that drops that ordering would throw every 2s instead of failing loudly.

delthas (approved):
LGTM high level, but I'm not very familiar with that part of the code, so needs a proper second review 😛

tests/unit/ingestion/IngestionPopulator.js
```diff
@@ -278,6 +278,322 @@ describe('Ingestion Populator', () => {
...
+            return logReader;
+        }
+
+        beforeEach(() => {
+            ip.logReaders = [];
+            ip.logReadersUpdate = [];
+        });
+
+        it('should activate a log reader once its setup succeeds', done => {
+            const logReaderMock = createLogReaderMock(WORKING_BUCKET, null);
```
francoisferrand: nit: calling `createLogReaderMock()` is not very readable, the usual pattern is to setup a logReader in the fixture (`beforeEach`),and let each test setup expectations (`logReader.setup.yieldsAsync(setupError)`)

Of you stick with this form to reduce dedup, than should the next 2 lines as well into that helper

```js
function createLogReaderMock(zenkoBucket, setupError) {
	const logReader = sinon.createStubInstance(IngestionReader);
	logReader.getTargetZenkoBucketName.returns(zenkoBucket);
	logReader.setup.yieldsAsync(setupError);

	ip._ingestionSources[zenkoBucket] = logReader;
	ip.logReadersUpdate = [logReader];

	return logReader;
}
```

(mixing both approaches makes it a bit weird)

tests/unit/ingestion/IngestionPopulator.js
```diff
@@ -278,6 +278,322 @@ describe('Ingestion Populator', () => {
...
+            ip._setupUpdatedReaders(err => {
+                assert.ifError(err);
+                assert.deepStrictEqual(ip.logReaders, [logReaderMock]);
+                assert.deepStrictEqual(ip.logReadersUpdate, []);
+                done();
+            });
+        });
+
+        it('should queue a log reader again when its setup fails', done => {
+            const logReaderMock =
```
francoisferrand: line break not needed I guess, fits within max column in our prettier config?
(same in following tests)

tests/unit/ingestion/IngestionPopulator.js
```diff
@@ -278,6 +278,322 @@ describe('Ingestion Populator', () => {
...
+            ip._setupUpdatedReaders(err => {
+                assert.ifError(err);
+                assert.deepStrictEqual(ip.logReaders, [workingReader]);
+                assert.deepStrictEqual(ip.logReadersUpdate, [failingReader]);
+                done();
+            });
+        });
+    });
+
+    describe('removing a source while its setup is in flight', () => {
```
francoisferrand: this describe section is weird: this is still testing `_setupUpdatedReaders()`

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -613,13 +613,26 @@ class IngestionPopulator {
         const newReaders = this.logReadersUpdate;
         this.logReadersUpdate = [];
         async.each(newReaders, (logReader, cb) => logReader.setup(err => {
+            const zenkoBucket = logReader.getTargetZenkoBucketName();
+            // The reader is in neither list while its setup is in flight, so
+            // its source may have been removed or replaced in the meantime. It
+            // must then neither be activated nor retried.
+            if (this._ingestionSources[zenkoBucket] !== logReader) {
+                return cb();
```
francoisferrand, line 629: should add a log message (just info is enough), would help to troubleshoot if something ever happens in this corner case...

tests/unit/ingestion/IngestionPopulator.js
```diff
@@ -278,6 +278,322 @@ describe('Ingestion Populator', () => {
         });
     });
 
+    describe('_setupUpdatedReaders', () => {
+        const FAILING_BUCKET = 'failing-zenko-bucket';
+        const WORKING_BUCKET = 'working-zenko-bucket';
```
francoisferrand: nit: nothing working or failing for these buckets. Only ingestion is failing, not the buckets

putting the name here (which may be needed because the mocks are hidden deep down c.f. https://github.com/scality/backbeat/pull/2806/changes#r3880675313) gives a false impression there is something, where in fact it is just the ingestion setup mock which fail...

tests/unit/ingestion/IngestionPopulator.js
francoisferrand: or even just just use the helper to create/add the reader ; but keep the useful mock call (`setup`) in the test

```js
function addLogReader(zenkoBucket) {
	const logReader = sinon.createStubInstance(IngestionReader);
	logReader.getTargetZenkoBucketName.returns(zenkoBucket);

	ip._ingestionSources[zenkoBucket] = logReader;
	ip.logReadersUpdate = [logReader];

	return logReader;
}

it(....) {
   addLogReader("bucket1").setup.yieldsAsync(errors.InternalError);
}
```

In addition to making the mocks more explicit, it would also make the tests more readable, by removing extraneous list modifications - like the following, where the test needs to manually edit only to simulate a race condition (nothing needed for simple case):
```js
// should not queue a log reader again when its setup fails and its source is no longer configured
createLogReaderMock("bucket1")
delete ip._ingestionSources["bucket1"]; // remove from the config
```

tests/unit/ingestion/IngestionPopulator.js
francoisferrand: it tests the interaction between _setupUpdatedReaders and _closeLogState, but is it not redundant?

e.g. if you validate on one side the way `_setupUpdatedReaders` behaves (vs the various contents of `_ingestionSources` /  `logReadersUpdate`) and the way _closeLogState updates these : then there is nothing left.... if however you want a more "integrated" test, should use the other methods to setup the readers (like `applyUpdates` and/or `addNewLogSource`)

francoisferrand (approved):

francoisferrand:
should this land on `9.5` or just delay until `9.6` ?

tests/unit/ingestion/IngestionPopulator.js
benzekrimaha: Dropped that describe. `_setupUpdatedReaders` already covers missing/replaced `_ingestionSources`, and `_closeLogState` already covers unregistration including in-flight setup.

tests/unit/ingestion/IngestionPopulator.js
benzekrimaha: Renamed to `bucket1` / `bucket2`.

tests/unit/ingestion/IngestionPopulator.js
benzekrimaha: Switched to `addLogReader` and left `setup.yieldsAsync(...)` in each test.

lib/queuePopulator/IngestionPopulator.js
benzekrimaha: Added an info log with `zenkoBucket` and `location`.

tests/unit/ingestion/IngestionPopulator.js
benzekrimaha: Gone with the helper.

lib/queuePopulator/IngestionPopulator.js
```diff
@@ -613,13 +613,32 @@ class IngestionPopulator {
         const newReaders = this.logReadersUpdate;
         this.logReadersUpdate = [];
         async.each(newReaders, (logReader, cb) => logReader.setup(err => {
+            const zenkoBucket = logReader.getTargetZenkoBucketName();
+            // The reader is in neither list while its setup is in flight, so
+            // its source may have been removed or replaced in the meantime. It
+            // must then neither be activated nor retried.
+            if (this._ingestionSources[zenkoBucket] !== logReader) {
+                this.log.info('skipping log reader after setup: source was ' +
+                    'removed or replaced', {
```
francoisferrand: line break not needed (esp in the middle of the string)

francoisferrand (approved):

PR #2807, opened 2026-08-18
Accept object-valued value and overheadFields in kafka entries

delthas:
`createFromKafkaEntry` parses the kafka message body and then parses the nested `value` — and, for `del` entries, `overheadFields` — a second time, so producers have to send them as JSON text.

The D/R metadata stream is built by a MongoDB change-stream pipeline ([ZKOP-562]), which cannot serialise a document to a string: there is no `$toJson`, and `$convert` rejects documents. Manufacturing one for the consumer to parse straight back would mean either `$function` (per-document server-side JS on a production secondary, inside a change-stream pipeline) or a custom Kafka Connect SMT shipped in the kafka-connect image.

The string requirement only exists because every producer so far builds entries in JavaScript — `ListRecordStream` does `value: JSON.stringify(objectMd)` on the oplog path, and S3C ingestion stringifies upstream. Nothing consumes the string form; it is parsed on receipt.

So decode these fields only when they are strings, and pass objects through. That is a superset of the existing format: producers that stringify take the identical path, so ingestion is unaffected.

The first commit moves the `QueueEntry` spec to `tests/unit/lib/models/`, next to the model it tests.

Issue: BB-847

### DIFF
--- a/lib/models/QueueEntry.js
+++ b/lib/models/QueueEntry.js
@@ -5,6 +5,26 @@ const BucketQueueEntry = require('./BucketQueueEntry');
 const BucketMdQueueEntry = require('./BucketMdQueueEntry');
 const DeleteOpQueueEntry = require('./DeleteOpQueueEntry');
 
+/**
+ * Decode a nested field of a kafka entry, which producers may send either
+ * as a JSON string or as an already-decoded object.
+ *
+ * @param {string|object} field - JSON string or decoded object
+ * @return {object} the decoded field
+ * @throws {Error} if the field is neither a JSON string nor an object; the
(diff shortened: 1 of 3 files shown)


francoisferrand (approved):
note: should land on 9.6 branch (waiting for "flaky" fix and 9.5.0 release) to create it

delthas:
> note: should land on 9.6 branch (waiting for "flaky" fix and 9.5.0 release) to create it

Done

benzekrimaha (approved):

delthas:
!done 4h

delthas:
[`BB-847`](https://scality.atlassian.net/browse/BB-847): logged `4h`; transitioned to Done (`Fixed`)

PR #2808, opened 2026-08-18
Consume the ingestion topic from the earliest offset

delthas:
`MongoQueueProcessor` did not set `fromOffset`, so it inherited librdkafka's default `latest`: a message produced onto a partition with no committed offset while the group had no member — a rolling update, a crash-restart — is skipped forever once the replacement consumer resolves its start position.

BB-831 swept this defect class and left this consumer unpinned, as having "no known permanent-loss path". That does not hold. `LogReader` persists its own log position in Zookeeper and advances it after publishing a batch, and the ingestion snapshot phase runs once at source creation — so nothing re-reads that stretch of the raft log, and a skipped message means the object metadata is never ingested.

Pinned unconditionally rather than behind a D/R mode, since the D/R metadata sink reuses this processor and loses entries the same way. The permanent-loss argument, the group geometry and the replay-safety analysis are on BB-848.

The consumer options move into `_getConsumerOptions()` for unit coverage, following #2788.

Issue: BB-848

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -147,6 +147,7 @@ class MongoQueueProcessor {
                 },
                 concurrency: this.mongoProcessorConfig.concurrency,
                 maxQueued: this.mongoProcessorConfig.maxQueued,
+                fromOffset: 'earliest',
             });
             this._consumer.on('error', () => {
                 MongoProcessorMetrics.onIngestionKafkaConsume('error');
--- a/tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
+++ b/tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
@@ -1,9 +1,12 @@
 const assert = require('assert');
+const sinon = require('sinon');
 
 const MongoQueueProcessor =
     require('../../../extensions/mongoProcessor/MongoQueueProcessor');
 const ObjectQueueEntry =
     require('../../../lib/models/ObjectQueueEntry');
+const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
+const Config = require('../../../lib/Config');
 
 function _makeProcessor(bootstrapList) {
     const proc = Object.create(MongoQueueProcessor.prototype);
@@ -232,5 +235,41 @@ describe('MongoQueueProcessor._updateReplicationInfo', () => {
         assert.strictEqual(bySite['cloud-b'].destination, undefined);
         assert.strictEqual(bySite['cloud-b'].role, undefined);
         assert.strictEqual(bySite['cloud-b'].status, 'PENDING');
+    });
+});
+
+describe('MongoQueueProcessor.start', () => {
+    afterEach(() => {
+        sinon.restore();
+    });
+
+    function startProcessor() {
+        sinon.stub(BackbeatConsumer.prototype, '_init');
+        sinon.stub(Config, 'getBootstrapList').returns([]);
+        sinon.stub(Config, 'on');
+
+        const proc = _makeProcessor([]);
+        proc.logger = { info: () => {}, error: () => {}, fatal: () => {} };
+        proc.kafkaConfig = { hosts: 'localhost:9092' };
+        proc.mongoProcessorConfig = {
+            topic: 'backbeat-ingestion',
+            groupId: 'backbeat-ingestion-group',
+            concurrency: 5,
+        };
+        proc._setupMetricsClients = cb => cb();
+        proc._mongoClient = { setup: cb => cb() };
+
+        proc.start();
+
+        return proc;
+    }
+
+    it('starts the consumer at the earliest offset', done => {
+        const proc = startProcessor();
+
+        setImmediate(() => {
+            assert.strictEqual(proc._consumer._fromOffset, 'earliest');
+            done();
+        });
     });
 });


tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
```diff
@@ -234,3 +234,16 @@ describe('MongoQueueProcessor._updateReplicationInfo', () => {
         assert.strictEqual(bySite['cloud-b'].status, 'PENDING');
     });
 });
+
+describe('MongoQueueProcessor._getConsumerOptions', () => {
```
DarkIsDude: Test seems very light on this topic ? You just test that a function return an object, with a hardcoded value. Not sure that's very useful. We should have a more high level test ? Also, if you create a function, you should create all tests for it (so each config). In that case I'm not sure it's very useful but that should be the clean way.

DarkIsDude (changes requested):

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
delthas: Strengthened it to cover the config→option mapping rather than the single field — `fromOffset`, `topic`, `groupId` and `concurrency` — following `ReplicationStatusProcessor`'s `::_getConsumerOptions` test (`tests/unit/replication/ReplicationStatusProcessor.js:134`, from #2788), which asserts four named fields the same way.

I stopped short of asserting the whole options object: that becomes a change-detector test — copy the struct, assert it equals itself — which fails on every future field addition without catching anything. No precedent for it here either; the two existing `_getConsumerOptions` tests assert one field (`tests/unit/gc/GarbageCollector.spec.js:67`, from #2794) and four (above).

On a higher-level test: that already exists. #2788 added `describe('BackbeatConsumer fromOffset tests')` in `tests/functional/lib/BackbeatConsumer.js:1278`, which produces onto a partition predating the consumer group and asserts the message is still read — the actual behaviour, tested once at the level where it's meaningful. BB-831 documented deliberately not repeating that per consumer.

So the unit assertion here is narrow on purpose: it pins the *decision* that this consumer is one of the ones on `earliest`, rather than one of the four BB-831 deliberately left on `latest`. That's what a future "cleanup for consistency" could silently undo, and it's the part the functional test can't express.

delthas: Reworked, and you were right that something was missing.

The gap was not the assertion count — it was that a test on `_getConsumerOptions()`'s return value never proves those options are the ones `start()` actually hands to the consumer. That link was the line codecov flags as uncovered. So there are now three tests, and they sit with the commits they belong to:

**Commit 1** (the extraction) — `builds the consumer options from the configuration` covers the config→option mapping (`topic`, `groupId`, `concurrency`, `kafka.hosts`), and `is what start() builds the consumer from` stubs `BackbeatConsumer.prototype._init` and asserts the constructed consumer carries those options. I checked that second one fails if `start()` is decoupled from the getter, so it is a real wiring test rather than a restatement.

**Commit 2** (the behaviour change) — a single `fromOffset` assertion, which is all this commit adds.

I still stopped short of asserting the whole options object: that becomes a change-detector test — copy the struct, assert it equals itself — failing on every future field addition without catching anything. No precedent for it either; the two existing `_getConsumerOptions` tests assert one field (`tests/unit/gc/GarbageCollector.spec.js:67`, #2794) and four (`tests/unit/replication/ReplicationStatusProcessor.js:134`, #2788).

On a higher-level test: one exists. #2788 added `describe('BackbeatConsumer fromOffset tests')` in `tests/functional/lib/BackbeatConsumer.js:1278`, which produces onto a partition predating the consumer group and asserts the message is still read. It needs a randomised fresh group and a produce before the consumer exists, because the reset policy is only consulted for a partition with no committed offset — which is why BB-831 notes a kill-and-restart test would *not* reproduce the bug: by then the partition has a committed offset, and that wins over `auto.offset.reset`. Repeating that per consumer would re-verify librdkafka behaviour that does not vary between them, at the cost of a broker plus each consumer's dependency stack — so the per-consumer test pins the decision, and the functional test pins the semantics.

delthas: Fair point, though the real gap wasn't the assertion count: a test on what `_getConsumerOptions()` returns never proved those options are what `start()` actually hands to the consumer.

Now three tests, each with the commit it belongs to. Commit 1: one on the config→option mapping, one asserting `start()` builds the consumer from the getter (checked it fails if they're decoupled). Commit 2: just the `fromOffset` assertion it adds.

I left out asserting the whole options object — copying the struct to assert it equals itself fails on every future field addition without catching anything. The end-to-end behaviour is already covered once in the functional suite (`BackbeatConsumer fromOffset tests`, #2788).

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
```diff
@@ -234,3 +237,63 @@ describe('MongoQueueProcessor._updateReplicationInfo', () => {
...
+
+        const options = proc._getConsumerOptions();
+
+        assert.strictEqual(options.topic, 'backbeat-ingestion');
+        assert.strictEqual(options.groupId, 'backbeat-ingestion-group');
+        assert.strictEqual(options.concurrency, 5);
+        assert.strictEqual(options.kafka.hosts, 'localhost:9092');
+    });
+
+    it('is what start() builds the consumer from', done => {
```
francoisferrand: "title" of test is hard to read and not very clear...

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
```diff
@@ -234,3 +237,63 @@ describe('MongoQueueProcessor._updateReplicationInfo', () => {
         assert.strictEqual(bySite['cloud-b'].status, 'PENDING');
     });
 });
+
+describe('MongoQueueProcessor._getConsumerOptions', () => {
```
francoisferrand: the tests calling `_getConsumerOptions()` do not really help, they just duplicate the content of the function : it gives code coverage, but nothing else...

I'd rather you remove `_getConsumerOptions()` (as the code was previously), and do a real test: publish to kafka then start consumer and see a message ; or at start consumer and verify that kafka client itself has the right config... This would not only test _getConsumerOptions, but verify it is properly called, params not overridden, ...

francoisferrand (commented):
This may create another issue ? E.g. if an OOB location is created, then deleted, then another (OOB) location with the same name is created : I am not sure everything will be clean'ed up (topic, ...), could there be a risk that we re-ingest data from the previous location ?
→ Please double check impact of this architecture change.
→ Can you please check if we have some OOB design docs (here or in citadel), maybe this is discussed already and/or should be added there

note: this is mostly for D/R and not urgent, should land on dev/9.6 ?

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
DarkIsDude: I agree with Francois on the test, not the removal of the function https://github.com/scality/backbeat/pull/2808#discussion_r3811518722 as here `In that case I'm not sure it's very useful`.

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
```diff
@@ -234,3 +237,63 @@ describe('MongoQueueProcessor._updateReplicationInfo', () => {
...
+            groupId: 'backbeat-ingestion-group',
+        };
+        proc._setupMetricsClients = cb => cb();
+        proc._mongoClient = { setup: cb => cb() };
+
+        proc.start();
+
+        setImmediate(() => {
+            assert(proc._consumer instanceof BackbeatConsumer);
+            assert.strictEqual(proc._consumer._topic, 'backbeat-ingestion');
```
DarkIsDude: should we also assert the offset ? Not sure what this test want to test ?

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
DarkIsDude: Checked the functional coverage: BackbeatConsumer fromOffset tests passes fromOffset: 'earliest' as a literal, so it pins librdkafka's semantics — not this consumer's config. Remove the option from _getConsumerOptions() and it still passes. (Also only one of the two tests #2788's description promised landed; there's no negative control.)

Did I miss something ?

DarkIsDude (commented):
I also have concern about the use describe by Francois 🙏

delthas:
Agree on development/9.6 (it just wasn't created when I opened the PR)

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
delthas: Kept `_getConsumerOptions()` but moved the assertions off it: both tests now start the processor and assert on the `BackbeatConsumer` instance `start()` ends up with, rather than on the returned object. So they cover the config→options mapping, that it is actually called, and that BackbeatConsumer keeps the values through its own joi.

@DarkIsDude you were right about the functional coverage — `BackbeatConsumer fromOffset tests` passes `fromOffset: 'earliest'` as a literal, so it pins librdkafka's semantics, not this consumer's config. The new test is the negative control: I checked that deleting `fromOffset` from `_getConsumerOptions()` fails it. The offset you asked about is now asserted, on `_consumer._fromOffset`.

Titles reworded, and down to two tests, one per commit.

delthas:
> if an OOB location is created, then deleted, then another (OOB) location with the same name is created

Checked this. The ingestion topic and consumer group are fixed per Zenko instance rather than per location (`backbeat-ingestion` / `backbeat-ingestion-group` in the operator's config), so deleting and recreating a location never mints a fresh consumer group: the committed offsets persist and `earliest` is not consulted.

The populator does clean its own state per source — `_closeLogState` → `_removeReaderState` → `zkClient.removeRecur('<zookeeperPath>/<zenkoBucket>')`, keyed per Zenko bucket, matching the `/ingestion/<zenko_bucket>/logState/<RaftID>/logOffset` path in the OOB design doc. So a recreated location re-ingests from scratch, but that is the populator *producing* new entries, which happens identically with `latest`. Likewise entries already sitting in the topic for a removed location are applied either way, since the group's offset is behind them. Neither is changed by this PR.

`earliest` only takes effect when the group has no committed offset: the first ever start (topic empty, so a no-op), or after `offsets.retention.minutes` — 7 days by default of the group being *empty*, i.e. the processor down that long while the populator keeps producing. There, `latest` skips everything produced during the outage and that metadata is lost permanently, since the populator's own logOffset has advanced past it; `earliest` replays the retained window instead and the processor's writes are idempotent. That is the case this change is for.

On docs: the OOB design doc exists in both places (`docs/oob-s3-ring.md` here, `docs/design/out-of-band/oob-s3-ring.md` in citadel — same document). It documents the populator's zookeeper log offset but says nothing about the mongo-processor's Kafka consumer offsets, so this is not covered today.

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
delthas: You didn't miss anything — that's right, and it's why the assertion moved. The offset is now checked on the consumer instance (`_consumer._fromOffset`) rather than on the returned options, so deleting `fromOffset` from `_getConsumerOptions()` fails the test. I verified that. More detail in the thread below.

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
delthas: Reworded, and there are two tests now rather than three: `configures the consumer from the mongo-processor configuration` and `starts the consumer at the earliest offset`.

tests/unit/mongoProcessor/MongoQueueProcessor.spec.js
delthas: Both addressed. The offset is asserted now, on `_consumer._fromOffset` in the commit that introduces it. The test you were reading is gone — it and the options-object one are merged into a single test asserting the consumer instance's topic, group id, concurrency and hosts.

delthas:
Moved to 9.6.

DarkIsDude (approved):

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -93,6 +93,27 @@ class MongoQueueProcessor {
         this._mProducer.setupProducer(cb);
     }
 
+    _getConsumerOptions() {
```
francoisferrand: not needed with the latest change, let's keep the change lean (introducing the function also de-couples the options from consumer, which creates a risk if reused and inheriting incorrect params)

DarkIsDude: The code is cleaner and easier to read, IMO better to keep it but open to Thomas's opinion

delthas: Census of the nine `new BackbeatConsumer(...)` sites on `development/9.6`:

| shape | sites |
|---|---|
| helper | `GarbageCollector`, `ReplicationStatusProcessor`, `LifecycleObjectProcessor` (via `BackbeatConsumerManager`) |
| inline | `LifecycleBucketProcessor`, `MongoQueueProcessor`, `notification/QueueProcessor`, `FailedCRRConsumer`, `replication/QueueProcessor`, `MetricsConsumer` |

Inline is the majority, so dropped it for conciseness. (I personally don't mind much either way.)

francoisferrand (approved):

delthas:
!done 5h

delthas:
[`BB-848`](https://scality.atlassian.net/browse/BB-848): logged `5h`; transitioned to Done (`Fixed`)

PR #2809, opened 2026-08-18
Configure backbeat from the environment, without rewriting the config file

francoisferrand:
The docker entrypoint applied env-var configuration by rewriting conf/config.json in place with jq. That breaks when the config is mounted read-only, which is what zenko-operator does: setting any mapped variable made the container exit at startup. The mapping was also a hand-maintained shell table duplicated from the joi schema, and had drifted from it — LOG_LEVEL rejected the warn and error levels the schema accepts, and a few variables whose config section had since been removed were written back in, so setting one stopped the process from starting.

Overrides are now applied in memory before validation, with their names derived from the schemas themselves: a new configuration field is settable from the environment as soon as it is declared. Names that don't follow from the config path are annotated next to the field they name, which is how the historic ones keep working, and the handful of variables setting several fields at once stay in an explicit table. Values go through the schema like any other configuration, so an invalid one fails the startup instead of being silently ignored. docs/configuration.md documents the whole thing.

Every variable the entrypoint applied is pinned by a test, since zenko-operator and Federation set them and Federation forwards arbitrary ones from the field. Two deliberate exceptions, both already dead: MONGODB_HOSTS no longer selects the mongo log source dropped in BB-836, and the lifecycle rules and local cache variables are ignored rather than fatal.

### Removed variables

The following variables are not supported anymore, and will now be ignored. They used to be **worse than useless**: the entrypoint wrote them into configuration sections the schema no longer knows, and validation then rejected the whole file, so setting one of them stopped the process from starting.

- `EXTENSIONS_LIFECYCLE_RULES_EXPIRATION_ENABLED`,
  `EXTENSIONS_LIFECYCLE_RULES_NC_VERSION_EXPIRATION_ENABLED`,
  `EXTENSIONS_LIFECYCLE_RULES_TRANSITIONS_ENABLED`,
  `EXTENSIONS_LIFECYCLE_RULES_ABORT_INCOMPLETE_MPU_ENABLED`: the rules supported by the lifecycle extension are configured with `EXTENSIONS_LIFECYCLE_SUPPORTED_LIFECYCLE_RULES`.
- `REDIS_LOCALCACHE_HOST`, `REDIS_LOCALCACHE_PORT`: the local cache is not part of the configuration.

`MONGODB_HOSTS` keeps setting `queuePopulator.mongo.replicaSetHosts`, but no longer selects the `mongo` log source, which the queue populator dropped: a deployment relying on it has to set `queuePopulator.logSource` itself.

Issue: BB-808


Dockerfile
```diff
@@ -62,6 +60,6 @@ COPY --from=builder /usr/local/bin/dockerize /usr/local/bin/
 
 ENV AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE=1
 
-ENTRYPOINT ["tini", "-g", "--", "/usr/src/app/docker-entrypoint.sh"]
+ENTRYPOINT ["tini", "-g", "--"]
```
DarkIsDude, line 63: `tini` is still useful ?

extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -7,14 +7,15 @@ const {
...
+const { extensionConfigValidator } = require('../../lib/config/extensionConfigValidator');
 
 const { backbeatConsumer: { MAX_QUEUED_DEFAULT } } = require('../../lib/constants');
 const { ValidLifecycleRules: supportedLifecycleRules } = require('arsenal').models;
 
 const joiSchema = joi.object({
     zookeeperPath: joi.string().required(),
-    bucketTasksTopic: joi.string().required(),
-    objectTasksTopic: joi.string().required(),
+    bucketTasksTopic: joi.string().required().meta({ envAlias: 'BUCKET_TASK_TOPIC' }),
```
DarkIsDude: I would use envVar alias or just envVar. Nice idea to use meta for that 👏

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -13,10 +14,13 @@ const {
 
 const { MAX_QUEUED_DEFAULT }  = require('../../lib/constants').backbeatConsumer;
 
+// the historic env var names put the backend before `RETRY`, e.g.
```
DarkIsDude, line 38: What the goal of the comment ?

lib/config/configItems.joi.js
```diff
@@ -40,21 +40,22 @@ const bootstrapListJoi = joi.array()
...
-
-const logJoiOptional =
-          joi.object({
-              logLevel: joi.alternatives().try(...LOG_LEVELS).required(),
-              dumpLevel: joi.alternatives().try(...LOG_LEVELS).required(),
-          }).optional();
+const logLevelJoi = joi.alternatives()
+          .try('error', 'warn', 'info', 'debug', 'trace');
+
+// the levels default individually, so that setting one of them, from the
```
DarkIsDude: It was a decision made by Sylvain to required both or none. Why did you change that ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
+'use strict';
+
+/* eslint-disable no-param-reassign */
```
DarkIsDude: That's a bit aggresive ? Why do you want to disable it globally ?

docs/configuration.md
```diff
@@ -0,0 +1,120 @@
...
+  `EXTENSIONS_REPLICATION_DEST_AUTH_TYPE`, and `s3.host` by `CLOUDSERVER_HOST`.
+- `.meta({ envAlias: 'MONGODB' })` adds a name replacing the path of the field
+  within its schema: `queuePopulator.mongo.database` is set by
+  `MONGODB_DATABASE`, as well as by `QUEUE_POPULATOR_MONGO_DATABASE`. This is
+  how the historic names are kept working.
+
+The fields of an object with unconstrained keys, such as `kafka.producerParams`,
+have no derived name, and neither have the fields the schema forbids.
+
+## Values
```
DarkIsDude: I'm not sure we need that, it's pretty standard ?

docs/configuration.md
```diff
@@ -0,0 +1,120 @@
...
+
+- `kafka.hosts` -> `KAFKA_HOSTS`
+- `queuePopulator.batchMaxRead` -> `QUEUE_POPULATOR_BATCH_MAX_READ`
+- `extensions.gc.topic` -> `EXTENSIONS_GC_TOPIC`
+- `extensions.lifecycle.conductor.concurrency` ->
+  `EXTENSIONS_LIFECYCLE_CONDUCTOR_CONCURRENCY`
+
+Two annotations in the schemas tune these names, next to the field they name:
+
+- `.meta({ envName: 'DEST' })` renames the segment a field and its children
```
DarkIsDude: This is not very clear for the example. Can you provide a full schema with such `foo` and `bar` to make it clear and decorelate it from the code 🙏. Same for the next one ?

`segment a field and its children` but I understand it should be `EXTENSIONS_REPLICATION_DEST` ? Same, I don't understand from where `s3.host` example come from ?

The next one I don't understand how `MONGODB_DATABASE` can be used with `name replacing the path of the field`. 

IMO this is the most important logic of your PR but it's not very clear at all from the doc.

docs/configuration.md
```diff
@@ -0,0 +1,120 @@
...
+
+- `TIME_PROGRESSION_FACTOR`: decreases the weight of a day, to expedite the
+  lifecycle of objects.
+- `EXPIRE_ONE_DAY_EARLIER`, `TRANSITION_ONE_DAY_EARLIER`: deprecated in favor of
+  `TIME_PROGRESSION_FACTOR`.
+- `BACKBEAT_ECHO_TEST_MODE`, `BACKBEAT_INJECT_REPLICATION_ERROR_RATE`,
+  `BACKBEAT_INJECT_REPLICATION_ERRORS`, `CI`: fault  injection and test
+  fixtures.
+
+## Removed variables
```
DarkIsDude: Do you think this should be documented ? Maybe tracked in a dedicated JIRA to do the cleanup but not very useful to document them ?

docs/configuration.md
```diff
@@ -0,0 +1,120 @@
...
+- `CRASH_ON_BATCH_TIMEOUT`, `CRASH_ON_REBALANCE_TIMEOUT`: exit when a batch or a
+  consumer rebalance times out, rather than waiting for the liveness probe to
+  report it. Set on S3C, where supervisord only restarts a program on exit.
+- `RDKAFKA_DEBUG_LOGS`: librdkafka debug contexts to enable, comma separated.
+
+The following are meant for testing only:
+
+- `TIME_PROGRESSION_FACTOR`: decreases the weight of a day, to expedite the
+  lifecycle of objects.
+- `EXPIRE_ONE_DAY_EARLIER`, `TRANSITION_ONE_DAY_EARLIER`: deprecated in favor of
```
DarkIsDude, line 120: Should we just remove them ?

docs/configuration.md
```diff
@@ -0,0 +1,120 @@
...
+  `LIFECYCLE_MAX_AUTO_INDEX_STORAGE_BYTES`: limits above which lifecycle
+  indexes are not built automatically.
+- `BATCH_TIMEOUT_SECONDS`: how long a queue populator batch may run before it
+  is reported as stuck, 300 by default.
+- `CRASH_ON_BATCH_TIMEOUT`, `CRASH_ON_REBALANCE_TIMEOUT`: exit when a batch or a
+  consumer rebalance times out, rather than waiting for the liveness probe to
+  report it. Set on S3C, where supervisord only restarts a program on exit.
+- `RDKAFKA_DEBUG_LOGS`: librdkafka debug contexts to enable, comma separated.
+
+The following are meant for testing only:
```
DarkIsDude, line 116: We don't use `TIME_PROGRESSION_FACTOR` for some clients to increase the lifecycle ?

docs/configuration.md
```diff
@@ -0,0 +1,120 @@
...
+
+- `TIME_PROGRESSION_FACTOR`: decreases the weight of a day, to expedite the
+  lifecycle of objects.
+- `EXPIRE_ONE_DAY_EARLIER`, `TRANSITION_ONE_DAY_EARLIER`: deprecated in favor of
+  `TIME_PROGRESSION_FACTOR`.
+- `BACKBEAT_ECHO_TEST_MODE`, `BACKBEAT_INJECT_REPLICATION_ERROR_RATE`,
+  `BACKBEAT_INJECT_REPLICATION_ERRORS`, `CI`: fault  injection and test
```
DarkIsDude:
```suggestion
  `BACKBEAT_INJECT_REPLICATION_ERRORS`, `CI`: fault injection and test
```

Should we explain them instead of a general comment ? Also do you think it's useful to explain them ? I mean, we can check the code to understand. Document here is a risk of drift (when you add one, you forget to add it here, same for deletion and worst, if you change the behaviour ?)

docs/configuration.md
DarkIsDude: Not very clear the naming : `envVarFullPath` for example

`envAlias` can be `envVarSegmentPath` ?

docs/configuration.md
```diff
@@ -0,0 +1,120 @@
...
+- `REDIS_HOST`, `REDIS_PORT`: standalone `redis.host` and `redis.port` (6379 by
+  default). Both are ignored when sentinels are configured.
+- `EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST`: the servers of the `zenko` site
+  of the replication bootstrap list, comma separated.
+  `EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST_MORE` holds the additional sites,
+  as raw JSON objects, e.g. `{ "site": "aws", "type": "aws_s3" }`.
+
+## Other variables
+
+These are not part of the configuration schema:
```
DarkIsDude: Maybe it's important to explain why it's not part of the schema ? Pretty obvious for `BACKBEAT_CONFIG_FILE` but for example why not `KAFKA_TOPIC_PREFIX` ? Deserve an explanation 🙏

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * Configuration overrides from environment variables.
```
DarkIsDude: We should have a pointer to your markdown file ? Having logic explained at 2 places is not very efficient

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+ *   In an extension schema, the alias keeps the EXTENSIONS_<name> prefix.
+ *
+ * Annotating a shared schema at its use site, e.g. `hostPortJoi.meta(...)`,
+ * only affects that use: joi schemas are immutable.
+ *
+ * Vars setting several fields at once are listed in `compositeEnvVars`, and are
+ * excluded from the derived mapping. LIVENESS_PROBE_PORT is a special case: it
+ * fans out over the probe servers of every schema.
+ *
+ * Objects with unconstrained keys (e.g. `kafka.producerParams`) have no
```
DarkIsDude: What is `unconstrained` ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+        .replace(/([a-z0-9])([A-Z])/g, '$1_$2')
+        .replace(/([A-Z]+)([A-Z][a-z])/g, '$1_$2')
+        .toUpperCase();
+}
+
+/**
+ * @param {string[]} path - config path
+ * @returns {string} env var name derived from the path
+ */
+function envVarName(path) {
```
DarkIsDude, line 32:
```suggestion
function pathToEnvVar(path) {
```

envVar is a name, no need to mention it ? Same, we convert a path to an envVarName.

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+        throw new Error(`invalid JSON value for ${name}: ${err.message}`);
+    }
+}
+
+/**
+ * @param {Object} config - configuration to read from
+ * @param {string[]} path - config path
+ * @returns {*} value at the given path, undefined if any node is missing
+ */
+function getPath(config, path) {
```
DarkIsDude:
```suggestion
function getContentAtPath(config, path) {
```

we don't get a path ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+    return path.reduce((node, key) => node?.[key], config);
+}
+
+/**
+ * Sets a config value, creating the missing intermediate nodes. A node holding
+ * something else is reported rather than replaced: an array of per site probe
```
DarkIsDude: `an array of per site probe servers is not a section a value belongs in.` what do you mean ? I'm not sure it's useful to use site probe servers is this context ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+ * servers is not a section a value belongs in.
+ *
+ * @param {Object} config - configuration to update
+ * @param {string[]} path - config path
+ * @param {*} value - value to set
+ * @returns {undefined}
+ */
+function setPath(config, path, value) {
```
DarkIsDude: setValueAtPath

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+const compositeEnvVars = {
+    /** Standalone redis: the port has an implicit default, and sentinels win. */
+    REDIS_HOST: (config, host, env) => {
+        if (env.REDIS_SENTINELS) {
+            return;
+        }
+        setPath(config, ['redis', 'host'], host);
+        setPath(config, ['redis', 'port'], env.REDIS_PORT || '6379');
+    },
+    REDIS_PORT: (config, port, env) => {
```
DarkIsDude: env is not used ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+}
+
+/**
+ * Env vars setting several fields at once, or fields the schema cannot name.
+ * They are applied before the derived ones, which take precedence for the
+ * fields they set.
+ */
+const compositeEnvVars = {
+    /** Standalone redis: the port has an implicit default, and sentinels win. */
+    REDIS_HOST: (config, host, env) => {
```
DarkIsDude: why env and not port directly ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+     * standalone host and port, which the schema then forbids.
+     */
+    REDIS_SENTINELS: (config, sentinels, env) => {
+        setPath(config, ['redis', 'sentinels'], sentinels);
+        setPath(config, ['redis', 'name'], env.REDIS_HA_NAME || 'mymaster');
+        delete config.redis.host;
+        delete config.redis.port;
+    },
+    /**
+     * The historic name of `queuePopulator.mongo.replicaSetHosts`. The
```
DarkIsDude: Why documenting `logSource` ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+    }, config);
+    parent[path[path.length - 1]] = value;
+}
+
+/**
+ * Env vars setting several fields at once, or fields the schema cannot name.
+ * They are applied before the derived ones, which take precedence for the
+ * fields they set.
+ */
+const compositeEnvVars = {
```
DarkIsDude: all of them contains function. In your code before you describe arguments, it's not the case here. Why ?

lib/Config.js
```diff
@@ -65,19 +66,31 @@ class Config extends EventEmitter {
      * @returns {undefined}
      */
     _parseConfig(config) {
+        // env var overrides are applied before validation, so that the config
+        // file is left untouched: extension fields are handled by their own
+        // validator, from their own schema
```
DarkIsDude:
```suggestion
```

lib/Config.js
```diff
@@ -65,19 +66,31 @@ class Config extends EventEmitter {
...
+        applyCompositeEnvOverrides(config);
+        applyEnvOverrides(config, backbeatConfigJoi);
+
         const parsedConfig = joi.attempt(config, backbeatConfigJoi);
 
+        // the configuration an extension is validated against: the extension
+        // configurations are left out, as they are validated one after the
+        // other, and an extension must not depend on another one
```
DarkIsDude:
```suggestion
        // Extensions configurations are validated on their own
```

If you really wants to keep it

lib/config/envOverrides.js
DarkIsDude: I would extract that to a dedicated file to ease readability.

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+}
+
+/**
+ * Applies the env vars setting several config fields at once, in place.
+ *
+ * @param {Object} config - backbeat configuration to update
+ * @param {Object} [env] - environment to read the overrides from
+ * @returns {Object} updated configuration
+ */
+function applyCompositeEnvOverrides(config, env = process.env) {
```
DarkIsDude:
```suggestion
function applyCompositeEnvVars(config, env = process.env) {
```

is enough ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+ *
+ * @param {Object} config - configuration to update, matching the schema
+ * @param {joi.Schema} schema - configuration schema
+ * @param {string[]} [prefix] - config path of the schema root
+ * @param {Object} [env] - environment to read the overrides from
+ * @returns {Object} updated configuration
+ */
+function applyEnvOverrides(config, schema, prefix = [], env = process.env) {
+    if (config === null || typeof config !== 'object') {
+        // leave the invalid configuration for joi to report
```
DarkIsDude:
```suggestion
```

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+ */
+function envVarMappings(schema, prefix = []) {
+    const mappings = new Map();
+    const root = envVarName(prefix);
+    collectMappings(schema.describe(), [root], [], root, mappings);
+    return mappings;
+}
+
+/**
+ * A container exposes a single probe endpoint, shared by all the probe servers
```
DarkIsDude: This does not really describe whats happens in the function ? Why this comment ? Should be on the const `LIVENESS_PROBE_PORT` ? A dev will grep for the env var and expect to find the why directly there, not on the function ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+        // leave the invalid configuration for joi to report
+        return config;
+    }
+    const mappings = envVarMappings(schema, prefix);
+
+    if (env[LIVENESS_PROBE_PORT]) {
+        applyLivenessProbePort(config, mappings, env[LIVENESS_PROBE_PORT]);
+    }
+
+    mappings.forEach(({ path, type }, name) => {
```
DarkIsDude: I'm thinking about that. Here you first parse the schema to find all env var. That can be costly. Is it not better to check each env var and check if this env var can override a value ? Is it not more efficient ? Why did you choose one and not the other one ?

```
env | grep -v ARTESCA | grep -v DATA | grep -v KUBERNETES | grep -v ZENKO
```

We have a very few env var eligible now.

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+        case 'alternatives':
+            (description.matches || []).forEach(match =>
+                [match.schema, match.then, match.otherwise]
+                    .filter(alternative => alternative)
+                    .forEach(alternative =>
+                        collectMappings(alternative, names, path, root, mappings)));
+            break;
+        default:
+            // arrays are leaves: their items have no name to derive from
+            registerMapping(names, path, description.type, mappings);
```
DarkIsDude: Maybe is it useful to derive array by index (REDIS_HOST_0_PORT) ? Is it useful ? Should it be a follow up ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * Env vars setting several fields at once, or fields the schema cannot name.
+ * They are applied before the derived ones, which take precedence for the
+ * fields they set.
+ */
+const compositeEnvVars = {
+    /** Standalone redis: the port has an implicit default, and sentinels win. */
+    REDIS_HOST: (config, host, env) => {
+        if (env.REDIS_SENTINELS) {
```
DarkIsDude: https://github.com/scality/backbeat/pull/2809/changes#diff-5e31d6dd99646b03fb47254b4248523a947243b87ef511b12c1bb6354ed5c2a9R104 redis sentinels can be in the config ? Maybe it's not useful as we'll first ignore this one, and overide in the sentinels ?

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+            return;
+        }
+        setPath(config, ['redis', 'port'], port);
+    },
+    /**
+     * Sentinels come with their own group name (REDIS_HA_NAME), and replace the
+     * standalone host and port, which the schema then forbids.
+     */
+    REDIS_SENTINELS: (config, sentinels, env) => {
```
DarkIsDude: Can we have `REDIS_HA_NAME` and no `REDIS_SENTINELS` ?

lib/Config.js
DarkIsDude, line 1: We should have tests here ?

lib/config.joi.js
DarkIsDude, line 1: We should have tests 🙏

extensions/gc/GarbageCollectorConfigValidator.js
```diff
@@ -18,9 +19,4 @@ const joiSchema = joi.object({
     vaultAdmin: hostPortJoi,
 });
 
-function configValidator(backbeatConfig, extConfig) {
-    const validatedConfig = joi.attempt(extConfig, joiSchema);
-    return validatedConfig;
-}
-
-module.exports = configValidator;
+module.exports = extensionConfigValidator('gc', joiSchema);
```
DarkIsDude, line 22: This is never tested and can be a huge source of regression (same for other extentions)

lib/config/extensionConfigValidator.js
DarkIsDude, line 1: No test at all ?

tests/unit/lib/config/envOverrides.spec.js
```diff
@@ -0,0 +1,624 @@
...
+    it('should not derive a name for a field the schema forbids', () => {
+        const schema = joi.object({
+            kafka: joi.object({ hosts: joi.forbidden(), site: joi.string() }),
+        });
+
+        assert.deepStrictEqual([...envVarMappings(schema).keys()], ['KAFKA_SITE']);
+    });
+
+    it('should not derive a name for the params BackbeatProducer sets itself', () => {
+        const names = [...envVarMappings(backbeatConfigJoi).keys()];
```
DarkIsDude: Not really easy to read this test. Better to use your own config and don't reuse production one

lib/config/envOverrides.js
```diff
@@ -0,0 +1,368 @@
...
+/**
+ * @param {string} value - JSON document
+ * @param {string} name - env var it comes from, for error reporting
+ * @returns {*} parsed value
+ */
+function parseJSON(value, name) {
+    try {
+        return JSON.parse(value);
+    } catch (err) {
+        throw new Error(`invalid JSON value for ${name}: ${err.message}`);
```
DarkIsDude, line 45: this is never tested ?

lib/config/envOverrides.js
DarkIsDude, line 1: I'll let you decide but, ready the test file, we could split this one with:
- One for generic function (getPath, setPath, ...)
- One for managing composite env var
- One for managing replacement
- One for the extention logic
- Of course an index to expose what is needed to validation

I'm not sure it's useful, so I'll let you decide

DarkIsDude (changes requested):
Sorry for the review. I thought it'll be a fast one but finally not really.

With what we have, if we have for example SERVER_PORT as an env var and a extension.Replication.Server.Port, the env var can have an impact. Those very short env var name and generic, can be very dangerous. Setting a short env var car have a wide impact ? What do you think ? Maybe I'm not very clear, let me know by slack if it's the case.

DarkIsDude:
Should land to the next release ?

Dockerfile
francoisferrand: It is, to propagate signals to the `node` process, as the image is typically run with a `yarn` or `npm` command. Nothing really changed here: the entrypoint.sh used to `exec` the command, this PR just removed the middleman, but we still have the same command running (`npm rum ...`) and it still needs `tiny` to propagate signals/...

extensions/replication/ReplicationConfigValidator.js
francoisferrand: to explain why we need the alias, which is really "just" changing the order from `QUEUE_PROCESSOR_RETRY_<locationType>` (auto generated from the structure of the config) to `QUEUE_PROCESSOR_<locationType>_RETRY ` (the historical variable name)

if it was not there before, we would most likely have just kept the new name only: so best to let future readers be aware.

lib/config/configItems.joi.js
francoisferrand: discussed with him,
- his reflexion was mostly at ZKOP level, which is not changed here ;
- no specific reason to have chosen the existing behavior in backbeat ;
- at the component level, the behavior was counter-intuitive - and especially error prone when using variables, where we typically want to set a single variable and would otherwise need to set both

docs/configuration.md
francoisferrand: no, this is purely for testing or CI. Some customer may tweak the `cronRule`, so that we _check_ more often if object are elligible, but the computation is still performed in _days_, matching AWS semantics

This option changes the meaning of "days", by applying a multiplier to the duration : i.e. stretching the time... Definitely not for production, only for testing.

lib/config/envOverrides.js
francoisferrand: parts of the schema where we don't actually put any constraints, like `joi.object().unknown(true)`

lib/config/envOverrides.js
francoisferrand: that comment explains a non-obvious early return: i.e. why we don't throw. Not just the code, it provides the intent.

lib/config/envOverrides.js
francoisferrand: - it is not possible to "invert" the env name unambiguously: MIN_MPU_SIZE_MB could be min.mpu.size.mb, min.mpu_size.mb, min.mpu.size_mb....
- the traversing also yields the joi type for coercion and detects colliding names at startup.
- performance is not really an issue here: it happens only once on startup, and the env mapping adds around 30ms (on my machine)...

extension validators re-derive the schema it on every call (i.e. once per loaded extension), so we could shave a few tens of millisecond by memoizing the `envVarMappings` per schema indeed.

(also not sure about the "very few" : in Zenko we could never use it, so indeed; but I think it was used in S3C...)

lib/config/envOverrides.js
francoisferrand: This was not supported by the previous startup script, and I don't know of any need for this.

So I think this is already complex enough (but hopefully better than manual mapping, in the long run), next PR also already adds a generic way "patch" the JSON... and if really needed it can be added later

docs/configuration.md
francoisferrand: still supported (even if deprecated) in ZKOP / Zenko CR, so can't remove them yet.
not in this PR anyway, this PR is not about changing the behavior or surface, but just about changing the way the env variables are picked'up in the config.

docs/configuration.md
francoisferrand: pretty standard (as in "expected") mapping, but really just our own implementation (parsing the env variable value -a string- into the type which hopefully will path joi validation) : so worth documenting, but I will reword to make it clearer.

docs/configuration.md
francoisferrand: no cleanup needed here, these are actually useless: as the entry point would write a (now) invalid config if they were used. Removed, kept in the PR and GIt commit messages.

docs/configuration.md
francoisferrand: no perfect solution here indeed:
- `envName` is how this _field_ is used when computing the env variable name, for itself and its nested fields. This is similar to tags in golang structures.
- `envAlias` is used as an additional (hence alias) env variable name, but will still be appended with the nested field names

So we really need both, and both are used in path. I will change to `env` and `envVarAlias`, which has multiple benefits:
- `env` is short on purpose: it reads well inline in a joi chain, matches how the "feel" of Go struct tags, and being deliberately under-specified it sends a dev to the doc instead of assuming it is the whole variable name. The explicit alternatives (`envSegment`, `envFragment`) are jargon from the implementation, and no clearer to a newcomer.
- `envVarAlias` conveys both that it adds a name, and that the name is a whole environment variable rather than an env fragment. `envVarFullPath` doesn't hold here: on an object, an alias like MONGODB is a prefix that nested fields extend, not a full variable name.

Still not perfect, but it does not sound interchangeable anymore.

lib/config/envOverrides.js
francoisferrand: An environment variable has a name _and_ a value: this function returns only the name, so `pathToEnvVar` would promise an entity we don't build here.

It is also not really a conversion, more a constructor IMO: we build a new thing from another one, so naming it after what it returns says it — and it avoids a preposition (in my experience, a preposition in a name is often a symptom of not having found the precise word yet). On the other hand, `toSnakeCase` above is indeed a sort of conversion (same value, different spelling) and follows the usual `toString`/`toUpperCase` idiom.

It also matches the convention in this file: the pure derivations are noun-named (`envVarName`, `nodeNames`, `envVarMappings`), while the verb-named ones do more than return a value — they mutate or they walk (`registerMapping`, `collectMappings`, `applyEnvOverrides`).

So I'd rather keep `envVarName`.

lib/config/envOverrides.js
francoisferrand: it is used not line: `if (env.REDIS_SENTINELS) {`

Both standalone handlers ignore their variable when sentinels are configured in the environment, which is what the entrypoint did — if [[ "$REDIS_PORT" ]] && [[ -z "$REDIS_SENTINELS" ]]. No functional change here, same behavior implemented in pure javascript.

lib/config/envOverrides.js
francoisferrand: this table contains handlers for custom/advanced variable handling, which are called anyway with the value of the associated env variable: `REDIS_HOST` gets the host, `REDIS_PORT` gets the port.

But this highlights something that did not feel right, but failed to pursue: this table of callback is way too complex for what it is, this is really about something like:

```js
if (env.REDIS_SENTINELS) {
    … sentinels, name, drop host and port
} else if (env.REDIS_HOST || env.REDIS_PORT) {
    … host if set, port defaulting to 6379
}
```

and the others can be handled separately as well → refactoring/simplifying this.

lib/config/envOverrides.js
francoisferrand: removed altogether

lib/config/envOverrides.js
francoisferrand: removed altogether

lib/config/envOverrides.js
francoisferrand: no, and fixed (name is actually used for sentinels)

lib/config/envOverrides.js
francoisferrand: moved to joi config itself

lib/Config.js
francoisferrand: was already present, moved to its own test file for visibility

lib/config.joi.js
francoisferrand: added test for anything Joi logic was changed/added

extensions/gc/GarbageCollectorConfigValidator.js
francoisferrand: was tested implicitly (Config loads all extensions), but added explicit test in addition.

lib/config/extensionConfigValidator.js
francoisferrand: was tsted, but not in its own file. Moved to dedicated file for clarity

lib/config/envOverrides.js
francoisferrand: split to match existing convention: one spec file per source file

tests/unit/lib/config/envOverrides.spec.js
francoisferrand: modified so they don't use the backbeatConfigJoi anymore

docs/configuration.md
DarkIsDude: Clear for me, I thought it was use for CNES 🙏

docs/configuration.md
DarkIsDude: @francoisferrand do we have a follow-up for that ?

lib/config.joi.js
```diff
@@ -80,32 +82,40 @@ const joiSchema = joi.object({
...
         host: joi.string().required(),
         port: joi.number().default(8900),
     },
-    redis: {
-        host: joi.string().when('sentinels', {
-            is: joi.exist(),
-            then: joi.forbidden(),
-            otherwise: joi.required(),
+    // redis is reached either through its sentinels or directly: the two modes
+    // are exclusive, and each has the settings of its own
```
DarkIsDude:
```suggestion
```

It's just what happens one line after 🙏

lib/config.joi.js
```diff
@@ -80,32 +82,40 @@ const joiSchema = joi.object({
...
+                joi.string().custom(value => value.split(',').map(sentinel => {
+                    const [host, port] = sentinel.split(':');
+                    return { host, port: Number.parseInt(port, 10) };
+                })),
+                joi.array().items(joi.object({
+                    host: joi.string().required(),
+                    port: joi.number().required(),
+                })),
+            ]).required(),
+            // group name of the master the sentinels watch
```
DarkIsDude:
```suggestion
```

lib/config/configItems.joi.js
```diff
@@ -40,21 +40,22 @@ const bootstrapListJoi = joi.array()
...
+// configuration or from the environment, leaves the other one alone
```
DarkIsDude:
```suggestion
```

lib/config/configItems.joi.js
```diff
@@ -40,21 +40,22 @@ const bootstrapListJoi = joi.array()
...
+const logJoi = joi.object({
+    logLevel: logLevelJoi.default('info').meta({ env: 'LEVEL' }),
+    dumpLevel: logLevelJoi.default('error'),
+}).default();
+
+// logJoi for an extension: the levels it does not configure are inherited from
+// the global log config, passed as the validation context
```
DarkIsDude:
```suggestion
```

extensions/replication/ReplicationConfigValidator.js
```diff
@@ -13,10 +15,34 @@ const {
 
 const { MAX_QUEUED_DEFAULT }  = require('../../lib/constants').backbeatConsumer;
 
+const BOOTSTRAPLIST_MORE = 'EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST_MORE';
+
+/**
+ * Under its historic name, the bootstrap list holds the servers of the single
```
DarkIsDude: Even if I get the point, I have the feeling that this comment is a more a context that you want to output from your brain than a real useful comment. The comment don't describe what happens for the value provided ?

docs/configuration.md
DarkIsDude: Agree, that's better, thanks

lib/config/envOverrides.js
```diff
@@ -0,0 +1,259 @@
+'use strict';
+/**
+ * Configuration overrides from environment variables, with the names of the
+ * variables derived from the joi schemas.
+ */
+
+const { getField, setField } = require('./fields');
+
+// A container exposes a single probe endpoint, shared by all the probe servers
```
DarkIsDude, line 9: https://github.com/scality/backbeat/pull/2809#discussion_r3811978909 ping

lib/config/envOverrides.js
DarkIsDude: If you really wants to keep it, move it under the function description then or the return value ?

lib/config/envOverrides.js
DarkIsDude: Do we need a follow-up ?

DarkIsDude (commented):
Check also the CI even if I guess it's related to flakyness

extensions/lifecycle/LifecycleConfigValidator.js
```diff
@@ -7,14 +7,15 @@ const {
...
+const { extensionConfigValidator } = require('../../lib/config/extensionConfigValidator');
 
 const { backbeatConsumer: { MAX_QUEUED_DEFAULT } } = require('../../lib/constants');
 const { ValidLifecycleRules: supportedLifecycleRules } = require('arsenal').models;
 
 const joiSchema = joi.object({
     zookeeperPath: joi.string().required(),
-    bucketTasksTopic: joi.string().required(),
-    objectTasksTopic: joi.string().required(),
+    bucketTasksTopic: joi.string().required().meta({ env: 'BUCKET_TASK_TOPIC' }),
```
DarkIsDude, line 17: One last thing that came out the review. Should we add tests for that ? I'm mitigated as this is pure config `joi` but if not caught can be a pain (not sure we have tests for all them in Zenko...).

francoisferrand: by "this" do you mean the meta/env annotation? in that case, we actually have tests: these variables are the "legacy" variables, and there are test validating that they indeed are still working as expected.

docs/configuration.md
francoisferrand: As explained, we cannot remove them from backbeat as long as they are in ZKOP.
And we cannot remove them from ZKOP without breaking CR compatibility - i.e. it requires introducing a new Zenko CR version and implementing the CR a migration hook.

I don't think this is small effort, nor that we will work on it soon; do you think it's still worth creating the tickets now?

francoisferrand:
> Check also the CI even if I guess it's related to flakyness

this is the (known) issue about librdkafka revert breaking the newest tests, shall be fixed by rebasing on the fix once merged, just waiting confirmation for which way to go.

docs/configuration.md
```diff
@@ -0,0 +1,124 @@
...
+  sentinels are a comma separated list of `host:port`, e.g.
+  `REDIS_SENTINELS=sentinel1:26379,sentinel2:26379`.
+- `REDIS_HOST`, `REDIS_PORT`: standalone `redis.host` and `redis.port` (6379 by
+  default). Both are ignored when sentinels are configured.
+- `EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST`: the servers of the `zenko` site
+  of the replication bootstrap list, comma separated.
+  `EXTENSIONS_REPLICATION_DEST_BOOTSTRAPLIST_MORE` holds the additional sites,
+  as raw JSON objects, e.g. `{ "site": "aws", "type": "aws_s3" }`.
+
+## Other variables
```
delthas, line 84: I fear this list will get out of date? Should we drop it, or perhaps note it in CLAUDE.md st it is kept up to date?

delthas (approved):
Nice!

Should we squash a bit? The commits feel iterative / I see a refactor of a previous commit in a later commit

lib/config/configItems.joi.js
francoisferrand: is it really so obvious what `joi.ref('$log.dumpLevel')` means, especially when this field is just a "fragment", and not defined in the same joi struct (tree) ?

lib/config.joi.js
francoisferrand: this documents what the `name` field actually means : is it really clear here ?

extensions/lifecycle/LifecycleConfigValidator.js
DarkIsDude: My bad then, I didn't see tests for all of them. If it's the case, then we are good and can resolve

lib/config.joi.js
DarkIsDude: IMO yes, we are in the joi then of redis sentinels, so pretty clear ?

lib/config/configItems.joi.js
DarkIsDude: IMO yes, it's just before and pretty clear on how it's working ?

lib/config.joi.js
francoisferrand: don't know : I am not familiar with redis sentinel, so I can't tell if `name` is obviously the `group name of the master`, or if it could mean anything else... e.g. could a reader think this is "_a name for this redis connection_" or "_the sentinel's name_" ?

(i.e. we know it is a name related to redis sentinel, but not _what_ is named)

maybe shortening and moving the comment, so there is less risk of getting outdated?
```suggestion
name: joi.string().default('mymaster').meta({ env: 'HA_NAME' }), // sentinel master group
```

lib/config/configItems.joi.js
francoisferrand: the "defaulting" is indeed pretty clear, what is more subtle is the `$` (i.e. reference to the "validation context" - not sure everybody is familiar) and what we expect the context to hold (i.e. it is set only for extensions, and contain the global `config`).

in the end it indeed just refers to the joi block 2 lines up, but the mechanism to get there is actually subtly not so trivial... So maybe a more focused comment, what do you think?
```suggestion
// `$log` is the global log config, passed as the validation context
const logJoiOptional = joi.object({
    logLevel: logLevelJoi.default(joi.ref('$log.logLevel')).meta({ env: 'LEVEL' }),
    dumpLevel: logLevelJoi.default(joi.ref('$log.dumpLevel')),
}).optional();
```

There is a also a comment on `extensionConfigValidator` which explains this more clearly:
```
 * The global backbeat configuration is passed as the validation context, so
 * that a field can default to a global one, e.g. `joi.ref('$log.logLevel')`. It
 * is already validated when an extension is: its own defaults are set.
```
but I fear it may be difficult to find it from reading the joi structs themselves... Do you think that one is enough?

lib/config/envOverrides.js
francoisferrand: AFAIK we have no _need_ at this point, so I don't think we need a followup : no point creating tickets that will just wait forever in backlog, and we will create one if/when we see a need for this.

docs/configuration.md
francoisferrand: I think it is good to have it (and hopefully we should use the joi config more consistently in the feature, so we don't need such tracker), adding a mention in Claude.md to avoid drift is a good idea.

DarkIsDude (approved):
LGTM, thanks for all the works done 🙏

PR #2813, opened 2026-08-19
Support mTLS on the Kafka consumer

delthas:
> Stacked on #2808 (BB-848). Base needs retargeting to `development/9.6` once that merges.

`BackbeatConsumer` builds its rdkafka config from a fixed set of fields, so there is no way to pass `security.protocol` or `ssl.*`, and no consumer can reach a broker that requires client certificates. Producers already have this: `kafka.producerParams` is an existing config field that `BackbeatProducer` spreads.

Add `kafka.consumerParams` with the same shape — `.unknown(true)`, with exactly the keys a consumer sets itself forbidden, so a caller gets a validation error instead of a setting that is silently overwritten — and forward it from every Kafka client in the tree, including the oplog consumer, which builds its own `kafka.KafkaConsumer`.

The D/R metadata sink ([ZKOP-562]) is what needs it: it consumes from brokers on the production site, whose exposed listener is configured with `SSLClientAuth: SSLClientAuthRequired`. The operator half, which generates the `ssl.*` values from the KafkaUser and CA secrets, is [ZKOP-577].

Issue: BB-850

[ZKOP-577]: https://scality.atlassian.net/browse/ZKOP-577

### DIFF
--- a/extensions/gc/GarbageCollector.js
+++ b/extensions/gc/GarbageCollector.js
@@ -131,6 +131,7 @@ class GarbageCollector extends EventEmitter {
                 site: this._kafkaConfig.site,
                 compressionType: this._kafkaConfig.compressionType,
                 requiredAcks: this._kafkaConfig.requiredAcks,
+                consumerParams: this._kafkaConfig.consumerParams,
             },
             topic: this._gcConfig.topic,
             groupId: this._gcConfig.consumer.groupId,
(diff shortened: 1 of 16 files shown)


francoisferrand (approved):

benzekrimaha (approved):

delthas:
!done 5h

delthas:
[`BB-850`](https://scality.atlassian.net/browse/BB-850): logged `5h`; transitioned to Done (`Fixed`)

PR #2821, opened 2026-08-20
Improvement/bb 605 alert for ingestion processing

benzekrimaha:
Issue: BB-605

### DIFF
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -616,10 +616,21 @@ class IngestionPopulator {
             if (err) {
                 // if setup fails for a log reader, don't add it to `logReaders`
                 // log the error and continue setting up others
-                this.log.fatal('error setting up log reader', {
+                const zenkoBucket = logReader.getTargetZenkoBucketName();
+                this.log.error('error setting up log reader, retrying later', {
                     method: 'IngestionPopulator._setupUpdatedReaders',
+                    zenkoBucket,
+                    location: logReader.getLocationConstraint(),
                     error: err,
                 });
+                // Setup failures are usually transient (source unreachable,
+                // invalid credentials...): queue the reader again so that it
+                // is retried on the next cycle, unless it is no longer the
+                // reader registered for its bucket, meaning its source was
+                // removed or replaced in the meantime.
+                if (this._ingestionSources[zenkoBucket] === logReader) {
+                    this.logReadersUpdate.push(logReader);
+                }
             } else {
                 this.logReaders.push(logReader);
             }
--- a/monitoring/ingestion/ingestion-producer-alert.yaml
+++ b/monitoring/ingestion/ingestion-producer-alert.yaml
@@ -84,7 +84,7 @@ groups:
       description: "Ingestion producer source error rate has increased to 3% in the last 5 minutes"
       summary: "Ingestion producer source error rate has increased"
 
-  - alert: IngestionProducerSourceErrorRate3Percent
+  - alert: IngestionProducerSourceErrorRate5Percent
     Expr: |
       sum(rate(s3_ingestion_source_operations_total{status="error", namespace="${namespace}", job="${job_ingestion_producer}"}[5m]))
         /
--- a/monitoring/ingestion/ingestion-producer-alert.yaml
+++ b/monitoring/ingestion/ingestion-producer-alert.yaml
@@ -98,6 +98,22 @@ groups:
       description: "Ingestion producer source error rate has increased to 5% in the last 5 minutes"
       summary: "Ingestion producer source error rate has increased"
 
+  # The setup of a reader is retried on every cron tick, so a source that cannot
+  # be resolved fails continuously. It is a handful of errors against the reads
+  # of every healthy source, so it only shows up on the absolute rate: the ratio
+  # alerts above stay far below their threshold when a single source is stuck.
+  - alert: IngestionProducerSourceSetupFailing
+    Expr: |
+      sum(rate(s3_ingestion_source_operations_total{op="getRaftId", status="error", namespace="${namespace}", job="${job_ingestion_producer}"}[5m]))
+      > 0
+    For:   "10m"
+    Labels:
+      severity: warning
+    Annotations:
+      zenko_service: backbeat-ingestion-producer
+      description: "Ingestion producer has been unable to start reading from a source for the last 10 minutes, so the objects of that location are not ingested. Check the producer logs for the bucket it fails to resolve."
+      summary: "Ingestion producer cannot start reading from a source"
+
   - alert: IngestionProducerZookeeperErrorRate3Percent
     Expr: |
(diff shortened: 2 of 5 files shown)


lib/queuePopulator/IngestionPopulator.js
francoisferrand, line 1: this branch is on top of an outdated (merged) branch
- appears to include 4 commits, which do not belong in this branch
- the changes are unrelated
- the branch conflicts with upstream

→ rebase before asking for review

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -98,6 +98,22 @@ groups:
...
+    Expr: |
+      sum(rate(s3_ingestion_source_operations_total{op="getRaftId", status="error", namespace="${namespace}", job="${job_ingestion_producer}"}[5m]))
+      > 0
+    For:   "10m"
+    Labels:
+      severity: warning
+    Annotations:
+      zenko_service: backbeat-ingestion-producer
+      description: "Ingestion producer has been unable to start reading from a source for the last 10 minutes, so the objects of that location are not ingested. Check the producer logs for the bucket it fails to resolve."
+      summary: "Ingestion producer cannot start reading from a source"
```
francoisferrand, line 115:
```suggestion
      description: "Ingestion producer has been unable to read from a source for the last 10 minutes, so the objects of that location are not ingested. Check the producer logs for the bucket it fails to resolve."
      summary: "Ingestion producer cannot read from a source"
```

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -98,6 +98,22 @@ groups:
       description: "Ingestion producer source error rate has increased to 5% in the last 5 minutes"
       summary: "Ingestion producer source error rate has increased"
 
+  # The setup of a reader is retried on every cron tick, so a source that cannot
+  # be resolved fails continuously. It is a handful of errors against the reads
+  # of every healthy source, so it only shows up on the absolute rate: the ratio
+  # alerts above stay far below their threshold when a single source is stuck.
+  - alert: IngestionProducerSourceSetupFailing
+    Expr: |
+      sum(rate(s3_ingestion_source_operations_total{op="getRaftId", status="error", namespace="${namespace}", job="${job_ingestion_producer}"}[5m]))
```
francoisferrand, line 107: should we sum _by_ bucket or location? So the alert points to the specific bucket/location which does not work?

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -98,6 +98,22 @@ groups:
       description: "Ingestion producer source error rate has increased to 5% in the last 5 minutes"
       summary: "Ingestion producer source error rate has increased"
 
+  # The setup of a reader is retried on every cron tick, so a source that cannot
+  # be resolved fails continuously. It is a handful of errors against the reads
+  # of every healthy source, so it only shows up on the absolute rate: the ratio
+  # alerts above stay far below their threshold when a single source is stuck.
+  - alert: IngestionProducerSourceSetupFailing
```
francoisferrand, line 105: this alert is a sub-class of the previous one (`IngestionProducerSourceErrorRate5Percent`) :
- Why does it not report issues? Is this because of the threshold, or because one location's error may be lost in the other location's success, or because other calls are incorrectly reducing the ratio?
- Should we adjust the threshold instead? Or skip some ops which are not relevant to the alert? Or sum by bucket/location?

francoisferrand (changes requested):

monitoring/ingestion/ingestion-producer-alert.yaml
benzekrimaha: I did not take “read”: this alert is getRaftId, i.e. reader setup, not a journal read. I Worded it as “cannot set up a source” so it does not collide with getRaftLog

monitoring/ingestion/ingestion-producer-alert.yaml
benzekrimaha: Not possible today. s3_ingestion_source_operations_total only has origin, op, status. No bucket and no location sum by would be empty and locationNameRaftID exists on the journal gauges, not on this counter. We can add a locationName label, but it would apply to every source op (getRaftLog, metadata, …), not only getRaftId. I would keep this alert cluster-wide for now and add the label in a follow-up if we want the firing alert to name the location.

monitoring/ingestion/ingestion-producer-alert.yaml
benzekrimaha: For me it is the aggregation. A stuck source retries getRaftId every few seconds and those errors are drowned by getRaftLog successes from healthy sources. Lowering the ratio would noisy-alert on read errors and filtering the existing ratio to op="getRaftId" would fire, but then “error rate 5%” would mean setup failure...
The dedicated absolute rate on getRaftId, with for: 10m, is the smallest change that names this failure mode in my opinion

benzekrimaha:
Assigning @DarkIsDude for review since Mael is on PTO

DarkIsDude (approved):

monitoring/ingestion/ingestion-producer-alert.yaml
francoisferrand: you're right actually, but not this reason: we could add the label here, no reason to delay ; but this would actually create a (possibly) high-cardinality metric. best to avoid.

monitoring/ingestion/ingestion-producer-alert.yaml
francoisferrand: > A stuck source retries getRaftId every few seconds and those errors are drowned by getRaftLog successes from healthy sources.

- most often there are not many sources : like a few, and often just one. in any case, each source is one location, and we documented limit is 10 locations
- also in most cases, the locations are actually on the same server : so unlikely that one source fails while the others are fine

→ the aggregation could explain some corner case, but does not seem likely to explain the problem in every case?
→ the question still stands: we don't really care specifically that getRaftId failed, just that ingestion is not working. If the 5% alert is not working, it should be fixed, and we should avoid overlapping alerts : so we need to better understand what happens here before introducing a new alert which handles one case but leaves us "un-alerted" if it fails in the next line...

benzekrimaha: I dropped IngestionProducerSourceSetupFailing and changed the 3% / 5% rules to sum by (op). A stuck setup is 100% errors on getRaftId; a later getRaftLog failure is the same alert on that op.

monitoring/ingestion/ingestion-producer-alert.yaml
```diff
@@ -84,11 +86,11 @@ groups:
       description: "Ingestion producer source error rate has increased to 3% in the last 5 minutes"
       summary: "Ingestion producer source error rate has increased"
 
-  - alert: IngestionProducerSourceErrorRate3Percent
+  - alert: IngestionProducerSourceErrorRate5Percent
```
francoisferrand: having the same name is actually expected: this is how alerts are deduplicated, and the alert severity raises (from warning to critical) instead of having 2 alerts

(note: the dedup is not visible in the tests, this is setup by metalk8s in alert manager)

PR #2828, opened 2026-08-24
Never garbage-collect data on an isCRR location

francoisferrand:
Data living on an `isCRR` location is production data owned by a remote site. We may read it (that is the whole point of localizing a clean room object), but we must never delete it. That is a general rule keyed on the location type, not a clean room conditional and not behind a config flag.

It needs to hold at two independent points, one commit each.

**BB-813 - do not trigger the GC.** The copy engine reuses the lifecycle transition pipeline: `CopyLocationTask` copies the data locally, then `LifecycleUpdateTransitionTask` merges the new location into the metadata and publishes a `deleteData` action for the old one. Against an `isCRR` source that would wipe the remote production copy, so unlike a transition, the localization merge leaves the from-location alone and publishes nothing. This also gives us replay safety for duplicate copy actions: the second merge supersedes the first and collects its copy, which is only safe because the first never touched the remote source.

**BB-818 - guard the GC service.** The same rule applied where data actually gets deleted, whatever published the action. The service only has data to delete, and it has no way to delete data living on a remote location. Getting such an action is a bug upstream, hence the warning rather than silence, but there is nothing to do about it here beyond skipping it. No completion metric is emitted for a delete that did not happen.

Both check the location of the data being collected rather than the object storage class, because the two can disagree. If the object changes while the copy is in flight - the queue populator captures its `eTag` and `lastModified`, and the merge re-reads metadata later - the copy is discarded and we collect the local data we just wrote. Metadata was never updated at that point, so the storage class still names the `isCRR` location while the data to collect is local. Keying on the storage class would skip that collection and leak the local copy, which nothing else would ever pick up since metadata still points at the remote data.

Issue: BB-813
Issue: BB-818

### DIFF
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -6,6 +6,7 @@ const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 const { BatchDeleteCommand } = require('@scality/cloudserverclient');
 const { GarbageCollectorMetrics } = require('../GarbageCollectorMetrics');
 const { TRANSITION_ATTEMPT_MD } = require('../../../lib/util/transitionAttempt');
+const locationsConfig = require('../../../conf/locationConfig.json') || {};
 /** @typedef { import('../GarbageCollector.js') } GarbageCollector */
 
 class GarbageCollectorTask extends BackbeatTask {
@@ -143,6 +144,18 @@ class GarbageCollectorTask extends BackbeatTask {
     _executeDeleteDataOnce(entry, log, done) {
         const { locations } = entry.getAttribute('target');
         const ruleType = entry.getContextAttribute('ruleType');
+        // The service can only delete local data: data on a CRR location lives on
+        // the remote site, out of reach. Getting one here means a bug upstream.
+        const { dataStoreName } = locations[0] || {};
+        if (locationsConfig[dataStoreName]?.isCRR) {
+            log.warn('refusing to delete data on a CRR location', {
+                method: 'GarbageCollectorTask._executeDeleteDataOnce',
+                dataStoreName,
+                ruleType,
+                ...entry.getLogInfo(),
+            });
+            return process.nextTick(done);
+        }
         const params = {
             Locations: locations.map(location => ({
                 key: location.key,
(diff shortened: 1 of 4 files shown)


delthas:
> `lib tests`: all failures are in `tests/functional/lib/BackbeatConsumer.js` ("ledger drain on rebalance", then a cascade of `Local: Erroneous state` poisoning the following suites). PR [Skip non-localized entries in replication and lifecycle #2827](https://github.com/scality/backbeat/pull/2827) and PR [Trigger localization of clean room objects #2826](https://github.com/scality/backbeat/pull/2826) fail with the exact same assertion ("expected committed offset to advance from N to N+1 after deferred onEntryCommittable") and neither touches BackbeatConsumer either, so it looks like a flaky/broken test on development/9.6.

Now fixed, rebase and enjoy :)

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -143,6 +144,18 @@ class GarbageCollectorTask extends BackbeatTask {
     _executeDeleteDataOnce(entry, log, done) {
         const { locations } = entry.getAttribute('target');
         const ruleType = entry.getContextAttribute('ruleType');
+        // Last line of defense: whoever published this entry, data on a CRR
```
DarkIsDude: `Last line of defense:` this is out of scope. Last line of defense when ? who ? context ? Someone reading this comment will not understand.

// Data on a CRR location must never be deleted

But then you just describe the code...

extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
```diff
@@ -114,6 +115,19 @@ class LifecycleUpdateTransitionTask extends BackbeatTask {
 
     _garbageCollectLocation(entry, locations, log, done) {
         const { bucket, key, version, eTag, accountId, owner } = this.getTargetAttribute(entry);
+        // Data stored on a CRR location belongs to the remote site: the copy we
+        // just made is an extra local copy, the source must be left untouched.
```
DarkIsDude:
```suggestion
```

DarkIsDude (approved):

extensions/gc/tasks/GarbageCollectorTask.js
francoisferrand: the point is that this is defensive (hence "last line of defense" :-) ): we should normally not reach here, but if it happens drop the message instead of even trying to do the GC...

I will try to see how to reformulate, but I think it is important to express the intent here: so people are not tempted to remove it ("it cannot happen, see the design") nor extrapole from this that it can happen ("I remember GC skips these objects, so..."), without having to dig through history to get to the commit or PR message...

extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
francoisferrand: clarified the important part, not visibile in the code : `Data stored on a CRR location` means this is pull replication, not transition; therefore we must skip deleting

maeldonn (approved):

PR #2829, opened 2026-08-24
Pull replication of objects still on the source location

francoisferrand:
Objects are created locally but their metadata still points at the source cluster: the data itself has not been copied over yet. Something has to notice them and ask for it to be pulled in.

The queue populator is the natural place for that, since bootstrap, re-bootstrap and streamed updates all go through the same oplog. When an object lands on a location flagged `isCRR`, publish a `copyLocation` action and let the existing data mover/transition workflow do the actual copy. No activation switch: the flag on the location is the trigger. It is skipped without lifecycle -S3C runs replication without it- as there would be nothing to merge the copy into the object metadata.

This is unrelated to `replicationInfo`, which tracks replication of a *local* object to remote sites, hence the check before any of its conditions.

The destination comes from the object metadata, where the source-side rewrite stamps it next to the bucket and role the copy already needs. If it names a location we do not know, we fall back to the first local one and print a : copying the data elsewhere beats leaving it on the source forever.

The populator only knows the object owner's canonical id, and a Vault lookup per entry would throttle it, so the account id is resolved by the transition processor instead, once per action and only when missing.

Pull replication reuses the transition pipeline, but remains a separate workflow to report on and to troubleshoot, so it gets its own metrics type. It is also counted as queued on the replication metrics: without it a growing backlog would be indistinguishable from an idle cluster.

### Along the way

- the transition attempt metadata key was open-coded in five places across lifecycle and gc, one of which would throw on user metadata that does not parse
- `BACKBEAT_QUEUEPOPULATOR_EXTENSIONS` was applied by dropping the other extensions from the configuration, so anything reading their settings got nothing in the populator processes

Issue: BB-814

### DIFF
--- a/extensions/lifecycle/LifecycleMetrics.js
+++ b/extensions/lifecycle/LifecycleMetrics.js
@@ -12,6 +12,9 @@ const LIFECYCLE_LABEL_CONDUCTOR_SCAN_ID = 'conductor_scan_id';
 
 const LIFECYCLE_MARKER_METRICS_LOCATION = '-delete-marker-';
 
+const TRANSITION_TYPE = 'transition';
+const PULL_REPLICATION = 'pullReplication';
+
 // Keep per-scan series long enough for scraping and debugging recent overlap,
 // but remove them from prom-client after a configurable retention interval.
 // We intentionally do not cap the number of tracked scan IDs: if overlapping
@@ -447,9 +450,21 @@ class LifecycleMetrics {
     }
 }
 
+/**
+ * Metrics type of a copyLocation action.
+ * @param {ActionQueueEntry} actionEntry - copyLocation action
+ * @return {string} metrics type
+ */
+function getCopyLocationMetricsType(actionEntry) {
+    return actionEntry.getAttribute('metrics.origin') === PULL_REPLICATION ?
+        PULL_REPLICATION : TRANSITION_TYPE;
+}
+
 module.exports = {
     DEFAULT_SCAN_METRIC_RETENTION_S,
     LifecycleMetrics,
     LIFECYCLE_MARKER_METRICS_LOCATION,
+    PULL_REPLICATION,
+    getCopyLocationMetricsType,
     resetLifecycleScanMetricCleanupTimers,
 };
--- a/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
+++ b/extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
@@ -1,6 +1,7 @@
 'use strict';
 const assert = require('assert');
 const async = require('async');
+const { errors } = require('arsenal');
 
 const ColdStorageStatusQueueEntry = require('../../../lib/models/ColdStorageStatusQueueEntry');
 const { LifecycleMetrics } = require('../LifecycleMetrics');
@@ -14,6 +15,9 @@ const { updateCircuitBreakerConfigForImplicitOutputQueue } = require('../../../l
 const { LifecycleRetriggerRestoreTask } = require('../tasks/LifecycleRetriggerRestoreTask');
 const BackbeatProducer = require('../../../lib/BackbeatProducer');
 const GarbageCollectorProducer = require('../../gc/GarbageCollectorProducer');
+const VaultClientWrapper = require('../../utils/VaultClientWrapper');
+const { AccountIdCache } = require('../../utils/AccountIdCache');
+const { authTypeAssumeRole } = require('../../../lib/constants');
 
 class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
 
@@ -44,9 +48,68 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
      * @param {Number} s3Config.port - s3 endpoint port
      * @param {String} [transport="http"] - transport method ("http"
      *  or "https")
+     * @param {Object} [vaultAdminConfig] - vault admin endpoint, used to
+     *  resolve canonical ids into account ids
      */
-    constructor(zkConfig, kafkaConfig, lcConfig, s3Config, transport = 'http') {
+    constructor(zkConfig, kafkaConfig, lcConfig, s3Config, transport = 'http',
+        vaultAdminConfig = undefined) {
         super(zkConfig, kafkaConfig, lcConfig, s3Config, transport);
+
+        const authConfig = this.getAuthConfig(this._lcConfig);
+        if (authConfig.type === authTypeAssumeRole) {
+            this.vaultClientWrapper = new VaultClientWrapper(
+                `lifecycle:${this.getProcessorType()}`,
+                vaultAdminConfig,
+                authConfig,
+                this._log,
+            );
+            this._accountIdCache = new AccountIdCache(
+                this._processConfig.concurrency);
+        }
+    }
+
+    /**
+     * Resolve the account id of a canonical id. Actions published by the
+     * lifecycle conductor already carry the account id; those published by the
+     * queue populator (pull replication) only know the canonical id.
+     * @param {String} ownerId - canonical id of the object owner
+     * @param {Logger} log - logger instance
+     * @param {Function} cb - callback: cb(err, accountId)
+     * @return {undefined}
+     */
+    getAccountId(ownerId, log, cb) {
+        if (!this.vaultClientWrapper) {
+            log.debug('skipping: not assume role auth type');
+            return process.nextTick(cb);
+        }
+
+        // A cached miss must fail like a fresh lookup would: `isKnown()` is also
+        // true for misses, and `get()` would then hand back `undefined`.
+        if (this._accountIdCache.isMiss(ownerId)) {
+            log.error('canonical id does not exist (cached)', { ownerId });
+            return process.nextTick(cb, errors.NoSuchEntity);
+        }
+
+        if (this._accountIdCache.has(ownerId)) {
+            return process.nextTick(cb, null, this._accountIdCache.get(ownerId));
+        }
+
+        return this.vaultClientWrapper.getAccountId(ownerId, (err, accountId) => {
+            if (err) {
+                if (err.NoSuchEntity) {
+                    log.error('canonical id does not exist', { error: err, ownerId });
+                    this._accountIdCache.miss(ownerId);
+                } else {
+                    log.error('could not get account id', { error: err, ownerId });
+                }
+                return cb(err);
+            }
+
+            this._accountIdCache.set(ownerId, accountId);
+            this._accountIdCache.expireOldest();
+
+            return cb(null, accountId);
+        });
     }
 
     /**
@@ -56,6 +119,7 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
      * @return {undefined}
      */
     start(done) {
+        this.vaultClientWrapper?.init();
         async.waterfall([
             next => super.start(next),
             next => {
@@ -226,7 +290,12 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
             ...super.getStateVars(),
             coldProducer: this._coldProducer,
             gcProducer: this._gcProducer,
+            getAccountId: this.getAccountId.bind(this),
         };
+    }
+
+    isReady() {
+        return super.isReady() && (!this.vaultClientWrapper || this.vaultClientWrapper.tempCredentialsReady());
     }
 }
 
--- a/extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
+++ b/extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
@@ -6,6 +6,10 @@ const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
 const ObjectMD = require('arsenal').models.ObjectMD;
 const { LifecycleMetrics } = require('../LifecycleMetrics');
+const {
+    TRANSITION_ATTEMPT_MD,
+    getTransitionAttempt,
+} = require('../../../lib/util/transitionAttempt');
 /** @typedef { import('../objectProcessor/LifecycleObjectProcessor.js') } LifecycleObjectProcessor */
 
 class LifecycleUpdateTransitionTask extends BackbeatTask {
@@ -71,9 +75,7 @@ class LifecycleUpdateTransitionTask extends BackbeatTask {
             .setDataStoreName(newLocationName)
             .setAmzStorageClass(newLocationName)
             .setOriginOp('s3:LifecycleTransition')
-            .setUserMetadata({
-                'x-amz-meta-scal-s3-transition-attempt': undefined,
-            })
+            .setUserMetadata({ [TRANSITION_ATTEMPT_MD]: undefined })
             .setTransitionInProgress(false);
     }
 
@@ -232,20 +234,10 @@ class LifecycleUpdateTransitionTask extends BackbeatTask {
                 next(err, objMD);
             }),
             (objMD, next) => {
-                const userMDStr = objMD.getUserMetadata() || '{}';
-                const userMD = JSON.parse(userMDStr);
-
-                let tryCount = userMD['x-amz-meta-scal-s3-transition-attempt'];
-                if (tryCount === undefined) {
-                    tryCount = 1;
-                } else {
-                    tryCount = parseInt(tryCount, 10) + 1;
-                }
+                const tryCount = (getTransitionAttempt(objMD) || 0) + 1;
 
                 objMD.setTransitionInProgress(false)
-                    .setUserMetadata({
-                        'x-amz-meta-scal-s3-transition-attempt': tryCount,
-                    });
+                    .setUserMetadata({ [TRANSITION_ATTEMPT_MD]: tryCount });
 
                 return this._putMetadata(entry, objMD, log, next);
             },
--- a/extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
+++ b/extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
@@ -5,7 +5,7 @@ const errors = require('arsenal').errors;
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
 const ActionQueueEntry = require('../../../lib/models/ActionQueueEntry');
 const ObjectMD = require('arsenal').models.ObjectMD;
-const { LifecycleMetrics } = require('../LifecycleMetrics');
+const { LifecycleMetrics, getCopyLocationMetricsType } = require('../LifecycleMetrics');
 const {
     TRANSITION_ATTEMPT_MD,
     getTransitionAttempt,
@@ -202,7 +202,7 @@ class LifecycleUpdateTransitionTask extends BackbeatTask {
                     const transitionTime = entry.getAttribute('metrics.transitionTime') ||
                         objMD.getTransitionTime();
                     const locationName = entry.getAttribute('toLocation');
-                    LifecycleMetrics.onLifecycleCompleted(log, 'transition',
+                    LifecycleMetrics.onLifecycleCompleted(log, getCopyLocationMetricsType(entry),
                         locationName, Date.now() - Date.parse(transitionTime));
                     next(err);
                 });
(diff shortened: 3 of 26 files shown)


extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -61,9 +65,62 @@ class LifecycleObjectProcessor extends EventEmitter {
...
+    /**
+     * Resolve the account id of a canonical id. Actions published by the
+     * lifecycle conductor already carry the account id; those published by the
+     * queue populator (clean room localization) only know the canonical id.
+     * @param {String} ownerId - canonical id of the object owner
+     * @param {Logger} log - logger instance
+     * @param {Function} cb - callback: cb(err, accountId)
+     * @return {undefined}
+     */
+    getAccountId(ownerId, log, cb) {
```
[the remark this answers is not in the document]
francoisferrand: code is heavily callback based in this - not the place or time change it.

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -61,9 +65,81 @@ class LifecycleObjectProcessor extends EventEmitter {
...
```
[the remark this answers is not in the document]
francoisferrand: code is heavily callback based in this - not the place or time change it.

extensions/lifecycle/objectProcessor/LifecycleObjectProcessor.js
```diff
@@ -61,9 +65,81 @@ class LifecycleObjectProcessor extends EventEmitter {
...
+     * Whether this processor can resolve canonical ids through Vault. Only the
+     * transition processor receives actions published without an account id
+     * (clean room localization), and only it is configured with a Vault admin
+     * endpoint - so leave the Vault client alone everywhere else.
+     * @return {Boolean} true if account id lookups are available
+     */
+    _accountIdLookupEnabled() {
+        const authConfig = this.getAuthConfig(this._lcConfig);
+        return authConfig.type === authTypeAssumeRole &&
+            !!(this._processConfig.vaultAdmin || authConfig.vault);
```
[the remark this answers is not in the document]
francoisferrand: this is asserted with joi in Config; following the same pattern as all other modules.

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -73,6 +86,14 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
         if (sanityCheckRes) {
             return;
         }
+        const locationConfig = locationsConfig[queueEntry.getDataStoreName()] || {};
+        // Data still on the source location has to be fetched first. This is
+        // unrelated to replicationInfo, which tracks replication of a *local*
+        // object to remote sites, hence the check before any of its conditions.
+        if (locationConfig.isCRR && this.transitionTasksTopic) {
```
benzekrimaha, line 99: this always returns, including when _publishPullReplicationAction published nothing.
For empty objects CopyLocationTask already puts 0-byte without a GET, skipping them leaves `dataStoreName` on the source forever, and push CRR never runs either... Same for delete markers as those usually must be replicated, and they have no data to pull. We could either publish a copyLocation for the 0-byte case (so metadata moves to the local location), or only return after an actual publish I guess ?. Delete markers should fall through to the existing CRR path...

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -124,6 +140,136 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+        // Those buckets are versioned, and the metadata layer will repair the
+        // master key once the version has been copied over.
+        if (isMasterKey(queueEntry.getObjectVersionedKey())) {
+            return;
+        }
+        if (queueEntry.getIsDeleteMarker()) {
+            return;
+        }
+        const locations = queueEntry.getLocation();
+        if (!locations || locations.length === 0) {
```
benzekrimaha: “nothing to fetch” is true for the bytes, not for the metadata. the object still advertises the source location...

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -1,18 +1,31 @@
...
 class ReplicationQueuePopulator extends QueuePopulatorExtension {
     constructor(params) {
         super(params);
         this.repConfig = params.config;
         this.metricsHandler = params.metricsHandler;
+        this.transitionTasksTopic = config.extensions?.lifecycle?.transitionTasksTopic;
+
+        // Where the data is fetched to when the object metadata does not name a usable
+        // target: the first location that can actually hold a local copy.
+        this.defaultLocalLocation = Object.keys(locationsConfig).find(
```
benzekrimaha: is the ordering guaranteed? Object.keys on locationConfig.json is whatever was written first, not the location this object belongs to...

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -124,6 +140,136 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+        if (!this.defaultLocalLocation) {
+            this.log.error('invalid target location and no local location ' +
+                'to fall back to, skipping pull replication', {
+                method: 'ReplicationQueuePopulator._getPullReplicationTarget',
+                ...queueEntry.getLogInfo(),
+                targetLocation,
+            });
+            return undefined;
+        }
+        this.log.error('invalid target location in object metadata', {
```
benzekrimaha: This path then succeeds with a fallback , better go`warn`?

lib/util/transitionAttempt.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Read the transition attempt count of an object.
+ * @param {ObjectMD} objMD - object metadata
+ * @return {Number|undefined} attempt count, or undefined if the object was
+ * never transitioned
+ */
+function getTransitionAttempt(objMD) {
+    const attempt = Number.parseInt(objMD.getValue()[TRANSITION_ATTEMPT_MD], 10);
+    return Number.isInteger(attempt) ? attempt : undefined;
```
DarkIsDude, line 14: Should we return 0 or 1 ? Not sure we should as the caller code will be much complex

extensions/replication/ReplicationMetric.js
```diff
@@ -65,8 +60,9 @@ class ReplicationMetric {
     }
 
     publish() {
-        // Lifecycle metrics not yet implemented.
-        if (this._isLifecycleAction()) {
```
DarkIsDude, line 69: maybe a good idea to keep the function and rename it with "canPublish"

extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
```diff
@@ -232,26 +234,53 @@ class LifecycleUpdateTransitionTask extends BackbeatTask {
...
+    _resolveAccountId(entry, log, cb) {
+        const { accountId, owner } = this.getTargetAttribute(entry);
+        if (accountId) {
+            return process.nextTick(cb);
+        }
+
+        if (!owner) {
+            // Every publisher sets one or the other, so this is a malformed
+            // entry: log it, and let the task fail on its own further down
+            // rather than retrying something that cannot be fixed.
```
DarkIsDude:
```suggestion
            // Every publisher sets one or the other, so this is a malformed
            // entry
```

DarkIsDude: To avoid the comment I would do:

```
if (!accountId && !owner) {
  log.error('cannot resolve account id: entry has no account id nor owner');
            return process.nextTick(cb);
}

if (accountId)
  ...

this.getAccountId
```

The needed value of accountId and owner is explicit instead of implicit and the comment not needed

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -44,9 +48,68 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
+        const authConfig = this.getAuthConfig(this._lcConfig);
+        if (authConfig.type === authTypeAssumeRole) {
+            this.vaultClientWrapper = new VaultClientWrapper(
+                `lifecycle:${this.getProcessorType()}`,
+                vaultAdminConfig,
+                authConfig,
+                this._log,
+            );
+            this._accountIdCache = new AccountIdCache(
+                this._processConfig.concurrency);
```
DarkIsDude, line 67: Where does this value come from ? Default is 10, is it fine / enough ?

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
```diff
@@ -44,9 +48,68 @@ class LifecycleObjectTransitionProcessor extends LifecycleObjectProcessor {
...
+     */
+    getAccountId(ownerId, log, cb) {
+        if (!this.vaultClientWrapper) {
+            log.debug('skipping: not assume role auth type');
+            return process.nextTick(cb);
+        }
+
+        // A cached miss must fail like a fresh lookup would: `isKnown()` is also
+        // true for misses, and `get()` would then hand back `undefined`.
+        if (this._accountIdCache.isMiss(ownerId)) {
```
DarkIsDude, line 88: I checked a little bit this one, miss entry never expire. Are we fine with that ? It's an old code, so I guess but should we expire them to allow a retry without restart ?

extensions/replication/ReplicationMetric.js
```diff
@@ -65,8 +60,9 @@ class ReplicationMetric {
...
+        // Metrics API/routes only support CRR
+        const { origin } = this._entry.getContext();
+        if (['lifecycle', 'pullReplication'].includes(origin)) {
```
DarkIsDude: Should we use the new const ?

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -124,6 +140,136 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+    _publishPullReplicationAction(entry, queueEntry, value) {
+        // Those buckets are versioned, and the metadata layer will repair the
+        // master key once the version has been copied over.
+        if (isMasterKey(queueEntry.getObjectVersionedKey())) {
+            return;
+        }
+        if (queueEntry.getIsDeleteMarker()) {
+            return;
+        }
+        const locations = queueEntry.getLocation();
```
DarkIsDude, line 164:
```suggestion
        const locations = queueEntry.getLocations();
```

Hard to refacto ?

extensions/replication/ReplicationQueuePopulator.js
```diff
@@ -124,6 +140,136 @@ class ReplicationQueuePopulator extends QueuePopulatorExtension {
...
+    }
+
+    /**
+     * Local location the object data must be copied to.
+     * @param {ObjectQueueEntry} queueEntry - parsed entry
+     * @param {Object[]} locations - object data locations
+     * @return {String|undefined} target location, undefined if there is none
+     */
+    _getPullReplicationTarget(queueEntry, locations) {
+        const { targetLocation } = locations[0];
```
DarkIsDude: If I remember well we'll keep in location only the one that is usuable before reaching this step (on the source oplog right?)

extensions/replication/ReplicationQueuePopulator.js
DarkIsDude: Why do we want a fallback ? In which case ? I read the description but didn't find the use case (neither in JIRA).

DarkIsDude (changes requested):

extensions/replication/ReplicationQueuePopulator.js
francoisferrand: - this is not about the location the "object belongs to", which we should get from the message. This is really just a _fallback_, in case that the location pointed to in the message has been removed. As indicated in the comment, best to use any _location_ than failing
- this is about having a _default location_, we don't really care which one. Cloudserver use hard-coded `us-east-1` as fallback, I think it is better to just pick the first one. Indicentally `Object.keys(require("locationConfig.json"))` returns the keys in order of definition in the json file, so this ends up being the same.
- should however indeed check the location is not an external backend (e.g. GCP, Azure...) for extra safety (using arsenal's `externalBackends` constant)

francoisferrand:
The 2 uncovered patch lines were `ReplicationQueuePopulator.js` 256 and 262, the `log.error` + `return undefined` inside the `!this.defaultLocalLocation` guard of `_getPullReplicationTarget`. That case was already tested, but by stubbing `_getPullReplicationTarget` itself, so it asserted nothing about the resolution logic; 414c8370 adds a test that drives the real path with no configured location able to hold a local copy, and checks nothing is published to the data-mover topic and no metrics fire.

Worth noting the guard is log-only: without it the fallback returns the same undefined, so the new test has teeth on the skip behaviour (verified: it fails if the guard returns the unusable target instead) but not on which of the two error messages is emitted.

extensions/replication/ReplicationQueuePopulator.js
francoisferrand: it is not a _use case_, it is the _corner case_ where the location is not usable...
- mongo-processor reads the message from Kafka, it writes the ObjMD with the "pointer" location -
  including the location where the data should be written to.
- while the message is in kafka, that location is removed (or renamed)
- queue processor eventually sees the insertion, and triggers replication

--> the queue processor cannot store in that location (it has been removed, or possibly renamed)

What we would like is to use the bucket's default location -i.e. not passing any location in
putObject-, however we need a destination location to the message can be picked'up by a worker; and
we don't have that information in here (nor want to bear the cost of the extra I/O)

The actual target location is not really important (most likely there is only a single "local"
location). We compute it early because we have that information already (in mongo processor), to
avoid a hard coded value; and this also leaves room for implementing custom mapping (if needed at
some point in the future).

But in the end, the most critical part is that the data is replicated, so we use such a fallback in
the corner case above, instead of failing.

francoisferrand: added "us-east-1" as first location we try, as it is the default location (whatever its content) in cloudserver (and AWS SDK).

extensions/replication/ReplicationMetric.js
francoisferrand: `canPublish()` would be misleading: as publish can do nothing more multiple other reasons (no producer, send error...)

In this context, it is also not a good abstraction: we end up just wondering "why wouldn't we", any reader will just go look at the code of the function, and it would actually reduce readability.

I initially tried that, but could not find any fitting name to describe the condition (isLifeycleAction, skipMetrics, isTransitionAction, isCrrReplication, hasOwnMetrics) -but they were all either incorrect or did not provide an abstraction- thus I ended keeping the code simple, inline.

The information the reader actually needs is _why_ those are excluded, but only the comment can say it (the metrics API/routes only implement CRR), the name can't.

extensions/replication/ReplicationQueuePopulator.js
francoisferrand: yes, its' coming from ObjectMD in arsenal... (and looking at the code, it supports older ObjMD model versions which did not support multiple parts/locations)

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
francoisferrand: good catch, but not sure it is as critical as it seems.
We add a miss only if we received an explicit `NoSuchEntity` error from vault; and in that case I don't think the entity can re-appear on its on: could only happen due to some manual operation I guess, and it would require not just the cache but also repairing the objects which failed to transition.
We should have a look though, esp. after we finalized the entities replication (or lack of): to ensure pull replication does not create such situations, and generally to make the code more robust (i.e. 'cap' this list).
→ BB-868

extensions/lifecycle/objectProcessor/LifecycleObjectTransitionProcessor.js
francoisferrand: this is similar to what we do in the other processes, and did not dig so much indeed 🫣
- In other extensions (esp. conductor and bucket processor), concurrency -i.e. number of messages processed in parallel- is good, but transition processed does one lookup per message, with no ordering: so benefit of this value is very limited in theory
- Practically however, pull replication will most likely have a single owner ; so good enough to start
- Fix is acutally more complex, as the cache is currently FIFO: for this use case (and high variability of owners), we should really have an LRU cache

→ BB-869

lib/util/transitionAttempt.js
francoisferrand: No, we need keep `undefined` : this is used to add the user-metadata and show the retry to the user. If not set (or invalid), we don't want to show anything. So the contract is really "give me the transition attempt if any".

(this field is added in the metadata by backbeat, using a value computed by `SorbetCtl` when requeuing from DLQ: sorbetctl builds the new attempt counter, send the "reset/retry" message to backbeat, which writes the counter in MD)

extensions/replication/ReplicationQueuePopulator.js
francoisferrand: in pull replication, metadata is "ingested" first ; then we only need to fetch for empty objects.
these objects should not even reach here: the ingestion should not update their location to trigger pull replication, they should be ingested in the right "state" already.

(added a note in ZKOP-563 / DRCTL-70 so we don't forget)

francoisferrand: also note this guard is not about "empty objects", but inconsisent ones : i.e. objects which should be pull-replicated (location.isCRR=true) but which do not have the expected location entry (pointing to where the data is)
→ really an error, flagged for now should eventually go to the error management system (once we have it)

extensions/replication/ReplicationQueuePopulator.js
francoisferrand: we don't do anything _here_; but the contract of pull replication is that it will be triggered when
the object's storage class is a CRR location (i.e. `locationConfig[objMd.dataStoreName].isCrr`), and
in that case the `objMD.dataLocations` contains a single entry with the "pointer" to the source
object.

So when we reach here, we have checked already that there is indeed one location; which in
particular contains the `targetLocation` where to write the data.

extensions/replication/ReplicationQueuePopulator.js
francoisferrand: should actually be handled by the pipeline (no point triggering a pull replication for empty obejct: no data to replicate, and triggering happens when we already have the metadata).
but updated the code to be resilient against that: cheap insurrance in case it is difficult (or costly) to handle 0-bytes in the ingestion.

extensions/lifecycle/tasks/LifecycleUpdateTransitionTask.js
francoisferrand: the comment is not about the condition (when do we fail - which is redundant with the code), but
about what we do : i.e. keep going, and let the error surface later if account/owner were required.

lib/util/transitionAttempt.js
francoisferrand: I actually was wrong about this field, for the "hot" transition case: the counter is automatically incremented, and metadata is immedately reset once it failed..... which surfaces an existing limitation of transitions:
- when it fails, the metadata is updated immediately (in backbeat), and the transition may thus be restarted immediately.

This is ok-ish for lifeycle (would be retriggered on next scan), but can be an issue for pull replicatoin: we may keep trying forever (fine), but with no throttling, and can quickly degrade if there is a bug/...

--> moved to a followup ticket BB-874

DarkIsDude (approved):
LGTM. Thanks for all improvements 🙏

extensions/lifecycle/LifecycleMetrics.js
```diff
@@ -447,9 +450,21 @@ class LifecycleMetrics {
...
+function getCopyLocationMetricsType(actionEntry) {
+    return actionEntry.getAttribute('metrics.origin') === PULL_REPLICATION ?
+        PULL_REPLICATION : TRANSITION_TYPE;
+}
+
 module.exports = {
     DEFAULT_SCAN_METRIC_RETENTION_S,
     LifecycleMetrics,
     LIFECYCLE_MARKER_METRICS_LOCATION,
+    PULL_REPLICATION,
```
benzekrimaha, line 467: do we really need to export the `PULL_REPLICATION` ?

benzekrimaha: nvm it is used later on , resolving this

benzekrimaha (approved):
LGTM , thanks

PR #2830, opened 2026-08-24
BB-786: trigger cold transition from oplog

francoisferrand:
Support "direct-to-cold" transitions: an object uploaded with a cold storage class must end up archived in that cold location, without waiting for a lifecycle rule.

Cloudserver (out of scope here) stores the data in the hot location on such a PUT, and records the requested cold class in `x-amz-storage-class` together with the transition-in-progress flag, while keeping the ordinary create origin op so that bucket notifications still fire.

The lifecycle queue populator now detects these objects from the MongoDB oplog and publishes the cold archive request, reusing the exact message shape of `ReplicationAPI.sendDataMoverAction`: the whole downstream pipeline (Sorbet -> cold status topic -> `LifecycleColdStatusArchiveTask` -> GC) is unchanged. The populator is strictly publish-only, and never writes object metadata. The discriminator only accepts origin ops that backbeat itself never writes, so a transition is never re-triggered by its own metadata updates; objects already holding `archive.archiveInfo` are excluded, while a pending deferred restore is not.

The transition-in-progress flag must be kept when requeuing such an object, since that flag (together with the cold storage class) is precisely what identifies it as pending: clearing it would hide the object from the populator and break the retry loop after a single attempt. Conversely, a bucket lifecycle rule must not transition an object which is already declared as cold.

Finally, the metadata update completing the transition is stamped with a distinct `s3:LifecycleTransition:Direct` origin op when the object already declared the target cold class, so consumers can tell a direct transition from a lifecycle-driven one.

Issue: BB-786

### DIFF
--- a/extensions/gc/tasks/GarbageCollectorTask.js
+++ b/extensions/gc/tasks/GarbageCollectorTask.js
@@ -286,14 +286,15 @@ class GarbageCollectorTask extends BackbeatTask {
                     version,
                 });
 
+                // Only a direct transition can declare the cold class at this point.
+                const isDirectToCold = objMD.getAmzStorageClass() === newLocation;
+
                 objMD.setLocation()
                     .setDataStoreName(newLocation)
                     .setAmzStorageClass(newLocation)
-                    .setOriginOp('s3:LifecycleTransition')
                     .setTransitionInProgress(false)
-                    .setUserMetadata({
-                        'x-amz-meta-scal-s3-transition-attempt': undefined,
-                    });
+                    .setOriginOp(isDirectToCold ? 's3:LifecycleTransition:Direct' : 's3:LifecycleTransition')
+                    .setUserMetadata({ 'x-amz-meta-scal-s3-transition-attempt': undefined });
                 this._putMetadata(entry, objMD, log, err => {
                     GarbageCollectorMetrics.onS3Request(log, 'putMetadata', 'archive', err);
                     if (!err) {
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -246,9 +246,29 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
     }
 
     /**
-     * A transition is "direct" when the object declares a cold storage class while its data still
-     * sits in a hot location, it has not been archived yet, and cloudserver has flagged it as in
-     * progress.
+     * Whether a put entry designates an object we may act upon: bucket entries carry no object
+     * key, mpu shadow bucket entries are internal, and the master entry of a versioned object
+     * duplicates the version entry, which is processed on its own.
+     *
+     * @param {Object} entry - The record log entry from metadata.
+     * @param {Object} value - The object metadata, already decoded by the caller.
+     * @return {boolean} true if the entry may be dispatched to an object handler.
+     */
+    _isDistinctObjectEntry(entry, value) {
+        if (!entry.key || entry.key.startsWith(mpuBucketPrefix)) {
+            return false;
+        }
+
+        if (this._isVersionedObject(value) && isMasterKey(entry.key)) {
+            this.log.trace('skip processing of object master entry');
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Check if object transition was initiated by direct-to-cold request instead of lifecycle rule.
      *
      * @param {Object} md - The object metadata.
      * @return {boolean} true if a direct transition is pending for this object.
@@ -279,18 +299,11 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
             return;
         }
 
-        if (entry.key.startsWith(mpuBucketPrefix)) {
+        if (!this._isDistinctObjectEntry(entry, value)) {
             return;
         }
 
         if (!this._isDirectToCold(value)) {
-            return;
-        }
-
-        // if entry is a versioned object and is the master entry, skip task as
-        // the non-master entry will be processed
-        if (this._isVersionedObject(value) && isMasterKey(entry.key)) {
-            this.log.trace('skip processing of object master entry');
             return;
         }
 
@@ -363,7 +376,7 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
             return;
         }
 
-        if (entry.key.startsWith(mpuBucketPrefix)) {
+        if (!this._isDistinctObjectEntry(entry, value)) {
             return;
         }
 
@@ -388,13 +401,6 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
 
         if (!value.archive || !value.archive.restoreRequestedAt ||
             !value.archive.restoreRequestedDays) {
-            return;
-        }
-
-        // if entry is a versioned object and is the master entry, skip task as
-        // the non-master entry will be processed
-        if (this._isVersionedObject(value) && isMasterKey(entry.key)) {
-            this.log.trace('skip processing of object master entry');
             return;
         }
 
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -20,9 +20,19 @@ const {
     coldStorageRestoreAdjustTopicPrefix,
     coldStorageRestoreTopicPrefix,
     coldStorageGCTopicPrefix,
+    coldStorageArchiveTopicPrefix,
 } = config.extensions.lifecycle;
 const BackbeatProducer = require('../../lib/BackbeatProducer');
 const locations = require('../../conf/locationConfig.json') || {};
+
+// Object creation is the only operation which may declare a cold storage class, and a retry asks
+// for a new attempt. Any other originOp comes from backbeat itself, and must not re-trigger.
+const transitionOriginOps = [
+    's3:ObjectCreated:Put',
+    's3:ObjectCreated:CompleteMultipartUpload',
+    's3:ObjectCreated:Copy',
+    's3:LifecycleTransition:Retry',
+];
 
 const nSecsPerDay = () => Math.ceil(scaleMsPerDay(config.timeOptions.timeProgressionFactor) / 1000);
 
@@ -100,6 +110,7 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
                     next => this._setupProducer(`${coldStorageRestoreAdjustTopicPrefix}${location}`, next),
                     next => this._setupProducer(`${coldStorageRestoreTopicPrefix}${location}`, next),
                     next => this._setupProducer(`${coldStorageGCTopicPrefix}${location}`, next),
+                    next => this._setupProducer(`${coldStorageArchiveTopicPrefix}${location}`, next),
                 ], done);
             }, cb);
         } else {
@@ -237,6 +248,123 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
      */
     _parseDate(date) {
         return new Date(date.$date || date);
+    }
+
+    _isColdLocation(locationName) {
+        return !!this.locationConfigs[locationName]?.isCold;
+    }
+
+    /**
+     * A transition is "direct" when the object declares a cold storage class while its data still
+     * sits in a hot location, it has not been archived yet, and cloudserver has flagged it as in
+     * progress.
+     *
+     * @param {Object} md - The object metadata.
+     * @return {boolean} true if a direct transition is pending for this object.
+     */
+    _isDirectToCold(md) {
+        return md['x-amz-scal-transition-in-progress']
+            && this._isColdLocation(md['x-amz-storage-class'])
+            && !this._isColdLocation(md.dataStoreName)
+            && !md.archive?.archiveInfo;
+    }
+
+    /**
+     * Handle a "direct-to-cold" transition: cloudserver has written the object data to a hot
+     * location, but the user requested a cold storage class in the PUT request. Cloudserver
+     * flags the object as "transition in progress", and it is up to the queue populator to
+     * trigger the archival, as there is no lifecycle rule (nor lifecycle scan) involved.
+     *
+     * The message published here is strictly the same as the one the lifecycle bucket processor
+     * publishes (c.f. ReplicationAPI.sendDataMoverAction), so that the whole downstream pipeline
+     * (Sorbet, cold status processor, garbage collector) is reused unchanged.
+     *
+     * @param {Object} entry - The record log entry from metadata.
+     * @return {undefined}
+     */
+    _handleTransitionOp(entry) {
+        if (!this.vaultClientWrapper) {
+            return;
+        }
+
+        if (entry.type !== 'put' || entry.key.startsWith(mpuBucketPrefix)) {
+            return;
+        }
+
+        const value = JSON.parse(entry.value);
+        if (!transitionOriginOps.includes(value.originOp) || !this._isDirectToCold(value)) {
+            return;
+        }
+
+        // if entry is a versioned object and is the master entry, skip task as
+        // the non-master entry will be processed
+        if (this._isVersionedObject(value) && isMasterKey(entry.key)) {
+            this.log.trace('skip processing of object master entry');
+            return;
+        }
+
+        const coldLocation = value['x-amz-storage-class'];
+        const attemptHeader = value['x-amz-meta-scal-s3-transition-attempt'];
+        const attempt = attemptHeader ? Number.parseInt(attemptHeader, 10) : undefined;
+        const ownerId = value['owner-id'];
+        this.vaultClientWrapper.getAccountId(ownerId, (err, accountId) => {
+            if (err) {
+                this.log.error('unable to get account', {
+                    method: 'LifecycleQueuePopulator._handleTransitionOp',
+                    ownerId,
+                    err,
+                });
+                return;
+            }
+
+            this.log.trace(
+                'publishing object transition entry',
+                { bucket: entry.bucket, key: entry.key, version: value.versionId, coldLocation },
+            );
+
+            const topic = `${coldStorageArchiveTopicPrefix}${coldLocation}`;
+            const key = `${entry.bucket}/${value.key}`;
+
+            let version;
+            if (value.versionId) {
+                version = encode(value.versionId);
+            }
+
+            const transitionTime = this._parseDate(
+                value['x-amz-scal-transition-time'] || value['last-modified']);
+            const message = JSON.stringify({
+                accountId,
+                bucketName: entry.bucket,
+                objectKey: value.key,
+                objectVersion: version,
+                requestId: uuid(),
+                size: value['content-length'],
+                eTag: `"${value['content-md5']}"`,
+                try: attempt,
+                transitionTime: transitionTime.toISOString(),
+            });
+
+            const producer = this._producers[topic];
+            if (producer) {
+                LifecycleMetrics.onLifecycleTriggered(this.log, 'queuePopulator', 'archive',
+                    coldLocation, Date.now() - transitionTime.getTime());
+
+                const kafkaEntry = { key, message };
+                producer.send([kafkaEntry], err => {
+                    LifecycleMetrics.onKafkaPublish(this.log, 'ColdStorageArchiveTopic', 'queuePopulator', err, 1);
+                    if (err) {
+                        this.log.error('error publishing object transition request entry', {
+                            error: err,
+                            method: 'LifecycleQueuePopulator._handleTransitionOp',
+                        });
+                    }
+                });
+            } else {
+                this.log.error(`producer not available for location ${coldLocation}`, {
+                    method: 'LifecycleQueuePopulator._handleTransitionOp',
+                });
+            }
+        });
     }
 
     _handleRestoreOp(entry) {
@@ -505,6 +633,7 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
         }
 
         this._handleRestoreOp(entry);
+        this._handleTransitionOp(entry);
 
         if (this.extConfig.conductor.bucketSource !== 'zookeeper') {
             this.log.debug('bucket source is not zookeeper, skipping entry', {
--- a/extensions/lifecycle/LifecycleQueuePopulator.js
+++ b/extensions/lifecycle/LifecycleQueuePopulator.js
@@ -25,15 +25,6 @@ const {
 const BackbeatProducer = require('../../lib/BackbeatProducer');
 const locations = require('../../conf/locationConfig.json') || {};
 
-// Object creation is the only operation which may declare a cold storage class, and a retry asks
-// for a new attempt. Any other originOp comes from backbeat itself, and must not re-trigger.
-const transitionOriginOps = [
-    's3:ObjectCreated:Put',
-    's3:ObjectCreated:CompleteMultipartUpload',
-    's3:ObjectCreated:Copy',
-    's3:LifecycleTransition:Retry',
-];
-
 const nSecsPerDay = () => Math.ceil(scaleMsPerDay(config.timeOptions.timeProgressionFactor) / 1000);
 
 class LifecycleQueuePopulator extends QueuePopulatorExtension {
@@ -280,19 +271,19 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
      * (Sorbet, cold status processor, garbage collector) is reused unchanged.
      *
      * @param {Object} entry - The record log entry from metadata.
+     * @param {Object} [value] - The object metadata, already decoded by the caller.
      * @return {undefined}
      */
-    _handleTransitionOp(entry) {
+    _handleTransitionOp(entry, value) {
         if (!this.vaultClientWrapper) {
             return;
         }
 
-        if (entry.type !== 'put' || entry.key.startsWith(mpuBucketPrefix)) {
+        if (entry.key.startsWith(mpuBucketPrefix)) {
             return;
         }
 
-        const value = JSON.parse(entry.value);
-        if (!transitionOriginOps.includes(value.originOp) || !this._isDirectToCold(value)) {
+        if (!this._isDirectToCold(value)) {
             return;
         }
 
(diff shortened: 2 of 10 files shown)


extensions/lifecycle/tasks/LifecycleResetTransitionInProgressTask.js
```diff
@@ -18,13 +22,32 @@ class LifecycleResetTransitionInProgressTask extends LifecycleRequeueTask {
             return false;
         }
         md.setOriginOp('s3:LifecycleTransition:Retry');
-        md.setTransitionInProgress(false);
+        // For a direct transition, the "transition in progress" flag is what the queue populator
+        // keys on to retry the archival: clearing it would both hide this update from the
```
francoisferrand: - not "there is no lifeycle scan to re-pick", but the other way around : we need to set the flag to avoid race condition, i.e. ensure the retry is picked'up ONLY by queue populator and NOT by lifecycle scan --> update the comment for clarity.....
- also comment is quite verbose, stick to the point

(actually, looking at the code below -extra guard on isDirectTransition in LifecycleTask- this is not even true : we guard on direct-2-cold object directly, and the flag is used as an oplog ownership flag, which could be dropped in the future without changing the code here)

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -286,10 +286,14 @@ class GarbageCollectorTask extends BackbeatTask {
                     version,
                 });
 
+                // The object already exposes the location it is transitioned to, so this is a
+                // direct-to-cold transition.
+                const isDirect = objMD.getAmzStorageClass() === newLocation;
```
DarkIsDude: Should we move that to a dedicated function and avoid comment + code duplication ? Like something hasTargetedLocation ?

DarkIsDude: (or better, return directly the target OriginOp ?)

extensions/lifecycle/tasks/LifecycleResetTransitionInProgressTask.js
```diff
@@ -18,13 +19,30 @@ class LifecycleResetTransitionInProgressTask extends LifecycleRequeueTask {
...
 
+    /**
+     * Check whether the object transition was requested directly in the PUT request (as opposed
+     * to being triggered by a lifecycle rule): in that case the requested cold storage class is
+     * declared in the object metadata, while the data still lies in a hot location.
+     *
+     * @param {ObjectMD} md - object metadata
+     * @return {boolean} true if this is a pending direct transition
+     */
+    _isDirectTransition(md) {
```
DarkIsDude:
```suggestion
    _isDirectToColdTransition(md) {
```

extensions/gc/tasks/GarbageCollectorTask.js
DarkIsDude:
```suggestion
                const isDirectToCold = objMD.getAmzStorageClass() === newLocation;
```

extensions/lifecycle/tasks/LifecycleResetTransitionInProgressTask.js
```diff
@@ -18,13 +19,30 @@ class LifecycleResetTransitionInProgressTask extends LifecycleRequeueTask {
...
         md.setUserMetadata({
             'x-amz-meta-scal-s3-transition-attempt': try_,
         });
         return true;
     }
 
+    /**
+     * Check whether the object transition was requested directly in the PUT request (as opposed
+     * to being triggered by a lifecycle rule): in that case the requested cold storage class is
+     * declared in the object metadata, while the data still lies in a hot location.
```
DarkIsDude:
```suggestion
     * Check that the object is a direct to cold transition (without lifecycle trigger)
```

The next of the description describe the inside of the function

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1276,6 +1278,11 @@ class LifecycleTask extends BackbeatTask {
                 if (isObjectCold) {
                     return next(errorTransitionColdObject);
                 }
+                // Skip direct-to-cold objects, whose transition is triggered by the queue
+                // populator and require the transition in progress flag to be set
+                if (locationsConfig[objectMD.getAmzStorageClass()]?.isCold) {
```
DarkIsDude, line 1283: https://github.com/scality/backbeat/pull/2830/changes#diff-37c9ecda10c884ef79419b6744d2491598248b6511813899f07af1909cbf2f43R244 why not reusing that ?

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -239,23 +241,130 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+
+    /**
+     * A transition is "direct" when the object declares a cold storage class while its data still
+     * sits in a hot location, it has not been archived yet, and cloudserver has flagged it as in
+     * progress.
+     *
+     * @param {Object} value - The object metadata.
+     * @return {boolean} true if a direct transition is pending for this object.
+     */
+    _isDirectTransition(value) {
```
DarkIsDude:
```suggestion
    _isDirectToColdTransition(value) {
```

I think it's important because it's how the product call it and we miss context without it.

extensions/gc/tasks/GarbageCollectorTask.js
DarkIsDude: should we also check that it's cold ? Even if it's not useful, should we ? We avoid dispatch the logic about "What is a direct to cold transition" across several piece of code ? https://github.com/scality/backbeat/pull/2830/changes#r3893192734

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -239,23 +241,130 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+     * The message published here is strictly the same as the one the lifecycle bucket processor
+     * publishes (c.f. ReplicationAPI.sendDataMoverAction), so that the whole downstream pipeline
+     * (Sorbet, cold status processor, garbage collector) is reused unchanged.
+     *
+     * @param {Object} entry - The record log entry from metadata.
+     * @param {Object} [value] - The object metadata, already decoded by the caller.
+     * @return {undefined}
+     */
+    _handleTransitionOp(entry, value) {
+        // accountId is mandatory in the message, for Sorbet to assume the bucket owner's role
```
DarkIsDude: Not sure what the goal of this comment ?  why accountId ? It is managed by vaultClientWrapper ? Is it useful to check that here ? As we'll have an error message later

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -504,7 +613,42 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+                bucket: entry.bucket,
+                key: entry.key,
+                error,
+            });
+            return undefined;
+        }
+
+        switch (value.originOp) {
+        // supporting both 's3:ObjectRestore' and 's3:ObjectRestore:Post' to keep compatibility with
+        // older cloudserver versions, the switch to 's3:ObjectRestore:Post' was made to have the
```
DarkIsDude, line 636: Can you define older ? When the switch was done (because if it's an old one, can we consider it as done and just don't manage old message ?)

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -286,10 +286,14 @@ class GarbageCollectorTask extends BackbeatTask {
...
+
                 objMD.setLocation()
                     .setDataStoreName(newLocation)
                     .setAmzStorageClass(newLocation)
-                    .setOriginOp('s3:LifecycleTransition')
+                    .setOriginOp(isDirect ? 's3:LifecycleTransition:Direct' : 's3:LifecycleTransition')
```
DarkIsDude: I miss the arsenal PR maybe to add this event ?

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -239,23 +241,130 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+            const transitionTime = this._parseDate(
+                value['x-amz-scal-transition-time'] || value['last-modified']);
+            const message = JSON.stringify({
+                accountId,
+                bucketName: entry.bucket,
+                objectKey: value.key,
+                objectVersion: version,
+                requestId: uuid(),
+                size: value['content-length'],
+                eTag: value['content-md5'],
```
DarkIsDude: Should we add `"` ? Not sure about this one here

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -239,23 +241,130 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+     *
+     * The message published here is strictly the same as the one the lifecycle bucket processor
+     * publishes (c.f. ReplicationAPI.sendDataMoverAction), so that the whole downstream pipeline
+     * (Sorbet, cold status processor, garbage collector) is reused unchanged.
+     *
+     * @param {Object} entry - The record log entry from metadata.
+     * @param {Object} [value] - The object metadata, already decoded by the caller.
+     * @return {undefined}
+     */
+    _handleTransitionOp(entry, value) {
```
DarkIsDude, line 297: Should we check if the location is paused ? I have the feeling that we already talk about that but was not able to find any trace so asking 🤷. If I remember well we should send then and sorbet will filter them ?

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -239,23 +241,130 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
             return;
         }
 
-        const value = JSON.parse(entry.value);
+        if (!this._isDirectTransition(value)) {
+            return;
+        }
+
+        // if entry is a versioned object and is the master entry, skip task as
+        // the non-master entry will be processed
```
DarkIsDude:
```suggestion
        // Skip it as the key will be processed by the non-master entry
```

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -239,23 +241,130 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+        if (this._isVersionedObject(value) && isMasterKey(entry.key)) {
```
DarkIsDude: This is a copy/paste, should we refactor that to have the logic at one place ?

DarkIsDude (changes requested):

extensions/gc/tasks/GarbageCollectorTask.js
francoisferrand: this is not a bucket notification event, and these are not defined in Arsenal: so no arsenal PR...

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: yes, this will be handled by Sorbet.

DarkIsDude: I remember now, a notification is already send, so sorbet will store all of them in Mongo and we need a rate limiter when the location is enable again. Thanks 🙏

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: not easily. This is not new code, this code (comment) was moved from https://github.com/scality/backbeat/blob/85b3f9f348f4d627368e68bf858eb56b79bbd566/extensions/lifecycle/LifecycleQueuePopulator.js#L384-L386

extensions/gc/tasks/GarbageCollectorTask.js
francoisferrand: my first idea was indeed to extract reusable `isColdTransition()`, however it introduces lots of
complexity (needed extra module to avoid cross-dependency between GC and lifecycle ; GC would need
to access locations ; for 3 lines only...) and it actually would loosen the condition slightly. So
I think it is better stick with the simpler approach, and let each extension manage its own
conditions.

francoisferrand: I was tempted to reuse some generic `isColdTransition` function, but this ends up adding lot of code for very limited benefit - c.f. https://github.com/scality/backbeat/pull/2830/changes#r3901327310

---

Trygin to be less ambitious, and just extracting a function here (→ in utils, since it needs to be shared), it should indeed return the target `originOp`. But I wonder if this is really a benefit: that function would take an object and a location, and return either transition or transition:direct, which does not provide a clean abstraction/semantics: we could deduplicate the code into a `transitionOriginOp()` function, but would still not be so clear, so readers would most likely jump to the utils package - effectively reducing readability.

On top of it, it would introduce a significant risk: having such a function would invite people to update the code like the following, which will not work (as objMD is updated by the first call to `setAmzStorageClass`)
```js
objMD.setLocation()
    .setDataStoreName(newLocation)
    .setAmzStorageClass(newLocation)
    .setOriginOp(transitionOriginOp(objMD, newLocation)) 
```

It would work by using a variable:

```
const originOp = transitionOriginOp(objMD, newLocation);
objMD.setLocation()
    .setDataStoreName(newLocation)
    .setAmzStorageClass(newLocation)
    .setOriginOp(originOp)
```

...but make the ordering constraint very difficult to see - and risk people changing (or trying to change - hopefully we will catch it...) the code later.

All in all in this instance I think it is better to keep the duplication (3 lines x 2) instead of paying the indirection on every read; what do you think?

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: `accountId` is mandatory, and we need vaultClientWrapper to retrieve it from `owner-id`, hence the
early exit.

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: good catch, thanks

extensions/lifecycle/tasks/LifecycleTask.js
francoisferrand: c.f. https://github.com/scality/backbeat/pull/2830/changes#r3904306456

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: In particular, since this comes from oplog: the value could still be needed if a restore is requested by an old Cloudserver, before/during an upgrade - and thus processed by newer backbeat asynchronously.

→ deferring to dedicated cleanup ticket

extensions/lifecycle/tasks/LifecycleResetTransitionInProgressTask.js
```diff
@@ -18,13 +19,27 @@ class LifecycleResetTransitionInProgressTask extends LifecycleRequeueTask {
...
         return true;
     }
 
+    /**
+     * Check if object transition was initiated by direct-to-cold request instead of lifecycle rule.
+     *
+     * @param {ObjectMD} md - object metadata
+     * @return {boolean} true if this is a pending direct transition
+     */
+    _isDirectToCold(md) {
```
benzekrimaha, line 38: not the same check as `LifecycleQueuePopulator._isDirectToCold`, which also requires
`!md.archive?.archiveInfo`. So same name, same doc comment, different body here. From what I get a direct-to-cold object has locations, so `LifecycleColdStatusArchiveTask`
writes `archiveInfo`, keeps the flag set and waits for the GC to flip the location. Which also means that a requeue in that window keeps the flag here, the populator won't republish (archiveInfo),
and `LifecycleTask` now refuses it too making the object stuck for good. on the lifecycle path the flag gets cleared and the next scan retries. do you think that we can share the populator's predicate rather than keep two?

extensions/gc/tasks/GarbageCollectorTask.js
```diff
@@ -286,10 +286,14 @@ class GarbageCollectorTask extends BackbeatTask {
...
 
+                // The object already exposes the location it is transitioned to, so this is a
+                // direct-to-cold transition.
+                const isDirectToCold = objMD.getAmzStorageClass() === newLocation;
+
                 objMD.setLocation()
                     .setDataStoreName(newLocation)
                     .setAmzStorageClass(newLocation)
-                    .setOriginOp('s3:LifecycleTransition')
+                    .setOriginOp(isDirectToCold ? 's3:LifecycleTransition:Direct' : 's3:LifecycleTransition')
```
benzekrimaha: `s3:LifecycleTransition` is in arsenal's `supportedNotificationEvents`,  while `:Direct` isn't and the notification populator uses `originOp` as the event type, dropping anything else silently.
Also meaning that a bucket configured on `s3:LifecycleTransition` is notified for a lifecycle transition
and not for a direct one. is that intended?

extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -1276,6 +1278,11 @@ class LifecycleTask extends BackbeatTask {
...
+                    return next(errorTransitionDeclaredColdObject);
```
benzekrimaha, line 1284: these all land in `log.error('could not apply transition rule')`, c.f. the FIXME just below. The existing ones are per-object exceptions, this one fires for every direct-to-cold object on every scan of a bucket with a transition rule. lot of error
lines for a normal state: can we add the expected-error flag, or at least log this one at debug? This also blocks lifecycle from ever moving an object declaring a cold class, including to a different cold location. Is that deliberate?

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -239,23 +241,142 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+                return;
+            }
+
+            this.log.trace(
+                'publishing object transition entry',
+                { bucket: entry.bucket, key: entry.key, version: value.versionId, coldLocation },
+            );
+
+            const topic = `${coldStorageArchiveTopicPrefix}${coldLocation}`;
+            const key = `${entry.bucket}/${entry.key}`;
```
benzekrimaha: the message matches `sendDataMoverAction`, the key doesn't: there it's `${bucket}/${key}` with the plain key, here it's uri-encoded with the `\0<versionId>` still on it. Meaning we have the same object, different partition depending on which producer sent it. this is consistent with `_handleRestoreOp` below, so maybe fine... but does sorbet key on it? if not I'd recommend dropping the "strictly the same" from the description and from the test name.

extensions/gc/tasks/GarbageCollectorTask.js
francoisferrand: already answered above - this is not a supported notification, but we must still put a value (must change the value to avoid double notification, and we put specific ones to allow troubleshooting)

extensions/lifecycle/tasks/LifecycleTask.js
francoisferrand: no: as you pointed out, this is a FIXME - so unrelated to this PR → BB-863

(and the situation is not worse than before: likely direct-to-cold and lifecycle would not be enabled on the same objects, and direct-to-cold would be used to transition earlier the lifecycle...)

extensions/lifecycle/tasks/LifecycleResetTransitionInProgressTask.js
francoisferrand: Point about reuse already answered above: we cannot share the predicate easily, due to the cross-extension dependency , because it would require both the list of locations and objMD and have a weird name, and finally because it would very easily lead to bad usage.

>  so LifecycleColdStatusArchiveTask writes archiveInfo, keeps the flag set and waits for the GC to flip the location. Which also means that a requeue in that window keeps the flag here, the populator won't republish (archiveInfo)

if archiveInfo was set, the object is actually archived. So it must indeed not be re-archived.
what may be missing is indeed being able to retrigger the missing GC, to finish the transition - unrelated to this PR, but certainly worth confirming → BB-862

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: not an issue today (there should not be 2 identical keys in sorbet topic, so no real ordering need), but best to stay consistent anyway.

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
```diff
@@ -112,10 +112,14 @@ class LifecycleColdStatusArchiveTask extends LifecycleUpdateTransitionTask {
                 objectMD.setOriginOp('s3:LifecycleTransition:SetArchive');
 
                 if (skipLocationDeletion) {
+                    // The object already exposes the location it is transitioned to, so this
```
DarkIsDude: IMO this one is useless

DarkIsDude (approved):
LGTM

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -504,7 +618,42 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+        // supporting both 's3:ObjectRestore' and 's3:ObjectRestore:Post' to keep compatibility with
+        // older cloudserver versions, the switch to 's3:ObjectRestore:Post' was made to have the
+        // correct event type for bucket notifications
+        case 's3:ObjectRestore':
+        case 's3:ObjectRestore:Post':
+        case 's3:ObjectRestore:Retry':
+            this._handleRestoreOp(entry, value);
+            break;
+
+        // Object creation is the only operation which may declare a cold storage class, and a retry
```
benzekrimaha, line 644: this is only true for backbeat writes, a later user put that keeps or sets a create originOp will republish: _isDirectToCold is still true until archiveInfo exists.
As objectPutACL never touches originOp, it stays s3:ObjectCreated:Put and SSE rewrite sets s3:ObjectCreated:Copy
I am actually wondering if a second archive request is safe for Sorbet (same object, new `requestId`), or do we need something stronger than the originOp list? if duplicates are fine, we should mention it in the comment IMO.

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
benzekrimaha: +1 on this and it seems also wrong...
`skipLocationDeletion` is empty locations (0-byte, or already cleared). which is not “the object already exposes the location it is transitioned to”. A regular lifecycle transition of a 0-byte object takes this branch too...
the discriminator is the line below (`getAmzStorageClass() === coldLocation`). We should either drop the comment, or state that.

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -504,7 +618,42 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+            this.log.error('could not parse log entry', {
+                method: 'LifecycleQueuePopulator.filter',
+                bucket: entry.bucket,
+                key: entry.key,
+                error,
+            });
+            return undefined;
+        }
+
+        switch (value.originOp) {
```
benzekrimaha, line 634: `safeJsonParse` only catches SyntaxError. `JSON.parse('null')` succeeds, then `value.originOp` throws and takes down the put path , including the zk bucket update below...

extensions/lifecycle/LifecycleQueuePopulator.js
```diff
@@ -504,7 +618,42 @@ class LifecycleQueuePopulator extends QueuePopulatorExtension {
...
+        case 's3:ObjectRestore':
+        case 's3:ObjectRestore:Post':
+        case 's3:ObjectRestore:Retry':
+            this._handleRestoreOp(entry, value);
+            break;
+
+        // Object creation is the only operation which may declare a cold storage class, and a retry
+        // asks for a new attempt: any other originOp comes from backbeat itself, and must not
+        // re-trigger a transition.
+        case 's3:ObjectCreated:Put':
```
benzekrimaha, line 648: why not `s3:ObjectCreated:Post`? the comment says any object creation, and it is in the notification event list. cloudserver does not emit it today (POST goes through `createAndStoreObject` as Put), so maybe not needed , but then the allowlist is “what cloudserver writes”, not “object creation”...

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: > `objectPutACL` never touches originOp [...] and SSE rewrite sets s3:ObjectCreated:Copy

if that is the case, this is a bug in objectPutACL and SSE rewrite : as it would trigger erroneous bucket notifications.

> I am actually wondering if a second archive request is safe for Sorbet (same object, new requestId), or do we need something stronger than the originOp list?

It is for the most part, at the cost of some extra write (then freed) on tape. It is mentioned in the design, and there is known bug to fix which is in the EPIC.

> if duplicates are fine, we should mention it in the comment IMO.

I don't think so. This is part of the design (and documented there), unrelated to this piece of code.

francoisferrand: objectPutAcl is fine: c.f. https://github.com/scality/cloudserver/blob/ab23592723f8d0a160f5756ea139819cad020c7f/lib/metadata/acl.js

not sure what you called "SSE rewrite", so could not confirm

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: "object creation" is “what cloudserver writes”, since cloudserver is the only one to create objects.
I will create a followup for POST not sending the right bucket notification (in cloudserver)

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: safeJsonParse catches *all* exceptions (but indeed `null` is a valid json):
```js
    try {
        result = JSON.parse(str);
    } catch (err) {
        return { error: err };
    }
```

Nothing new in the PR: before the change, `LifecycleQueuePopulator.js:374` did a raw JSON.parse(entry.value) followed by value.originOp.

extensions/lifecycle/LifecycleQueuePopulator.js
francoisferrand: created https://scality.atlassian.net/browse/CLDSRV-990 to fixing cloudserver

extensions/lifecycle/tasks/LifecycleColdStatusArchiveTask.js
francoisferrand: this comment does not explain the branch (this is always the case of "empty object"), but only the next line : i.e. when/how we identify a direct-to-cold transition...

benzekrimaha (approved):

PR #2833, opened 2026-08-27
Add a D/R mode to the mongo-processor

delthas:
The mongo-processor was written as the out-of-band ingestion consumer, and the D/R metadata sink now reuses it. The two disagree about most of what it does to an object: ingestion rewrites identity and placement to local values, because the source system's accounts and locations do not exist here, while D/R replicates both and so applies what the source-side pipeline sends.

Put those decisions behind a mode, following the shape the notification extension uses for its destinations — an abstract `ProcessorMode`, an implementation per mode, and an index mapping the configured name to the class. `IngestionMode` carries today's behaviour verbatim and `mode` defaults to `ingestion`, so nothing changes for an existing deployment.

### What the mongo-processor does

```mermaid
flowchart LR
    K[("Kafka topic")] --> E["read entry"]
    E --> B["look up the bucket"]
    B --> Z["read the stored object metadata"]
    Z --> A["apply the entry"]
    A --> M[("sink MongoDB")]
```

### Where the two modes diverge

Every decision the two modes disagree about is a hook on `ProcessorMode`, so the surrounding flow stays single-sourced.

```mermaid
flowchart TD
    E["entry"] --> T{"type"}

    T -->|del| G0["shouldProcessDelete"]
    G0 -->|ingestion| I0["only while the stored placement<br/>still matches the bucket"]
    G0 -->|D/R| D0["always"]
    I0 --> DEL[("delete in MongoDB")]
    D0 --> DEL

    T -->|put| G1["needsExistingMetadata"]
    G1 -->|ingestion| I1["only with a scal header or<br/>an enabled replication rule"]
    G1 -->|D/R| D1["always"]

    I1 --> G2["resolveVersionId"]
    D1 --> G2
    G2 -->|ingestion| I2["the scal version id when present"]
    G2 -->|D/R| D2["the entry's own"]

    I2 --> G3["getChangedContent"]
    D2 --> G3
    G3 -->|ingestion| I3["tags"]
    G3 -->|D/R| D3["tags, object lock, and placement<br/>while the data is still remote"]

    I3 --> S{"stored document"}
    D3 --> S

    S -->|absent| G4["applyNewObjectMetadata"]
    S -->|present| G6["replacesExistingMetadata"]
    G6 -->|ingestion| G5["mergeExistingMetadata"]
    G6 -->|"D/R, versioned"| G5
    G6 -->|"D/R, no version id<br/>or a null version"| G4

    G4 -->|ingestion| I4["local owner, location and data part,<br/>ACLs reset"]
    G4 -->|D/R| D4["ACLs reset"]
    G5 -->|ingestion| I5["keep the stored document,<br/>take the tags"]
    G5 -->|D/R| D5["keep the stored document, take tags<br/>and object lock, placement by locality"]

    I4 --> W[("write to MongoDB")]
    D4 --> W
    I5 --> W
    D5 --> W
```

### The hooks

| Hook | Ingestion | D/R | Beyond the ticket |
| --- | --- | --- | --- |
| `needsExistingMetadata` | Reads the stored document only when the entry carries a scal header or the bucket has an enabled replication rule | Always reads it | yes |
| `resolveVersionId` | The `x-amz-meta-scal-version-id` when present, otherwise the entry's | The entry's own, always † | yes |
| `getChangedContent` | Diffs tags only | Diffs tags, object-lock state, and placement while the data is still remote; compares the whole document when the entry replaces it, so a replay is still skipped | yes |
| `applyNewObjectMetadata` | Rewrites owner, `dataStoreName` and the data part to local values, resets ACLs | Resets ACLs, nothing else | no |
| `replacesExistingMetadata` | Never: a stored document is always merged into | True for a write carrying no version id, or a null version — the same key is being rewritten, so the entry replaces the document rather than merging into it | no |
| `mergeExistingMetadata` | Keeps the stored document, takes the entry's tags | Keeps the stored document, takes tags and object-lock state; keeps the stored placement for a localized version, takes the entry's for one still on the remote site | no |
| `shouldProcessDelete` | Skips the delete unless the stored placement still matches the bucket | Applies it always | yes |

The three marked *beyond the ticket* are not in the ticket text but follow from the design:

- **The stored document is always read.** Without it the merge is unreachable — the fetch is skipped when there is no scal header and no enabled replication rule, which is the normal D/R case, so every entry took the create path and an update overwrote the placement. The design requires the merge and asserts replay idempotence but never mentions the read; making it unconditional is our choice, at one `getObject` per entry.
- **Change detection covers the mutable set.** `getContentType` only diffs tags, so a retention or legal-hold change with unchanged tags was dropped as a duplicate before the merge could run. The requirement is the matrix's "Merged on update"; the mechanism is ours.
- **The delete guard is skipped.** Its five-way location test ignores every D/R deletion, since a replicated object's location legitimately differs — cold, on the source, or localized. The design endorses applying deletions directly: "the mongo-processor applies replicated operations -including version deletions- at the metadata level, below the S3 API enforcement".

† **The one decision the design does not cover.** `x-amz-meta-scal-version-id` is ignored in D/R mode. Both call sites used it mode-independently to choose which version to read and to delete. A production object that was itself OOB-ingested or cold-restored carries that header, naming a version of the system *production* ingested from — so honouring it here reads and deletes the wrong version. Version ids are the source's own and identical on both sides, so the entry's own id is authoritative. The design never considers a production site that itself ingests, so this is a guess, and the easiest thing in the PR to reverse.

### What happens to each field

| Field | Ingestion | D/R |
| --- | --- | --- |
| `owner-id`, `owner-display-name` | Replaced with the bucket owner | Kept as sent |
| `dataStoreName` | Replaced with the Zenko location | Kept as sent on a first write; on an update kept from the stored document for a localized version, taken from the entry for one still remote |
| `location` data parts | Rewritten to a single part pointing at the Zenko location | Kept as sent |
| `acl` | Reset to the defaults | Reset to the defaults |
| `tags` | Taken from the entry | Taken from the entry |
| `retentionMode`, `retentionDate`, `legalHold` | Not merged: only tags are | Taken from the entry, cleared values included |
| `versionId` | The scal version id when present | The entry's own |
| `x-amz-meta-scal-version-id` | Honoured | Ignored |
| `replicationInfo` | Reset, shared between the modes | Reset, shared between the modes |
| `content-length`, `content-md5`, and the rest | As sent | As sent |

### Configuration

The mongo-processor now reads two settings from its own extension config instead of from blocks it does not own: the mongo client config, so a D/R sink no longer carries a `queuePopulator` block it never runs, and the service account, so it no longer carries a partial `extensions.ingestion` block. The second was not optional — backbeat validates every extension listed under `extensions` against that extension's own schema, and the ingestion schema also requires a topic, a zookeeper path and a source list, so naming the account there made the whole configuration invalid. Both fall back to the old location for a deployment that enables both extensions, and `auth` is required in D/R mode, which enables no other extension.

Also stops extending the health-check address whitelist when no `server` section is configured. The section is optional, but it was dereferenced unconditionally, so a process serving no API exited before starting.

### Follow-ups and exclusions

| Topic | Ticket | Notes |
| --- | --- | --- |
| GC `deleteData` when a localized version is deleted | BB-813 / BB-815 | Worth being accurate: the design does *not* defer this. It states it as part of deletion handling and it appears in none of its lists of later increments |
| Circuit-breaker rules | BB-822 | The plumbing already exists; the design names the gating signal as local oplog and copy-queue lag |
| `__metastore` and Vault entities | OS-1110 | |
| `transitionInProgress` suppression | BB-819 | Depends on this merge. Keeping the stored document and overlaying a short list is what leaves it open; the non-localized branch restores placement specifically rather than wholesale, so BB-819 adds to a list that already exists |
| The operator's `locationConfig.json` has to stop being `{}` | ZKOP-566 | `isCRRLocation` reads it. Harmless for metadata-only D/R, where every lookup is false and every object is cold, which is the correct answer there |
| Duplicated `lib/util/locations.js` | #2828 | The first commit copies it byte-for-byte rather than stacking on a draft. Whichever of the two lands second has that commit dropped as already applied — verified by rebasing onto #2828 and watching git drop it |
| `replicationInfo` | none | Stays shared and unchanged: the source pipeline resets it and bucket workflows are stripped, so a correct sink bucket has no replication config and this reduces to a reset |
| Metadata RPO metric | none found | The observability section asks for the timestamp of the last metadata write replicated, and alerts on it. `MongoProcessorMetrics` has no such gauge. Every metric here is also `ingestion*`-prefixed, which will read oddly on a D/R dashboard |

Unblocks [ZKOP-562], whose generated configuration sets `mode`, `mongodb` and `auth`, and which cannot start a mongo-processor without this.

Issue: BB-811

### DIFF
--- a/extensions/mongoProcessor/MongoProcessorConfigValidator.js
+++ b/extensions/mongoProcessor/MongoProcessorConfigValidator.js
@@ -1,5 +1,6 @@
 const joi = require('joi');
-const { retryParamsJoi, probeServerJoi, logJoiOptional } = require('../../lib/config/configItems.joi');
+const { retryParamsJoi, probeServerJoi, logJoiOptional, mongoJoi } =
+    require('../../lib/config/configItems.joi');
 const { extensionConfigValidator } = require('../../lib/config/extensionConfigValidator');
 
 const { MAX_QUEUED_DEFAULT }  = require('../../lib/constants').backbeatConsumer;
@@ -7,6 +8,7 @@ const { MAX_QUEUED_DEFAULT }  = require('../../lib/constants').backbeatConsumer;
 const joiSchema = joi.object({
     topic: joi.string().required(),
     groupId: joi.string().required(),
+    mongodb: mongoJoi,
     retry: retryParamsJoi,
     concurrency: joi.number().greater(0).default(1),
     maxQueued: joi.number().greater(0).default(MAX_QUEUED_DEFAULT),
--- a/extensions/mongoProcessor/MongoProcessorConfigValidator.js
+++ b/extensions/mongoProcessor/MongoProcessorConfigValidator.js
@@ -1,6 +1,6 @@
 const joi = require('joi');
 const { modes, defaultMode } = require('./modes');
-const { retryParamsJoi, probeServerJoi, logJoiOptional, mongoJoi } =
+const { authJoi, retryParamsJoi, probeServerJoi, logJoiOptional, mongoJoi } =
     require('../../lib/config/configItems.joi');
 const { extensionConfigValidator } = require('../../lib/config/extensionConfigValidator');
 
@@ -10,7 +10,8 @@ const joiSchema = joi.object({
     topic: joi.string().required(),
     groupId: joi.string().required(),
     mode: joi.string().valid(...Object.keys(modes)).default(defaultMode),
-    mongodb: mongoJoi,
+    mongodb: mongoJoi.when('mode', { is: 'dr', then: joi.required() }),
+    auth: authJoi.when('mode', { is: 'dr', then: joi.required() }),
     retry: retryParamsJoi,
     concurrency: joi.number().greater(0).default(1),
     maxQueued: joi.number().greater(0).default(MAX_QUEUED_DEFAULT),
--- a/extensions/mongoProcessor/MongoProcessorConfigValidator.js
+++ b/extensions/mongoProcessor/MongoProcessorConfigValidator.js
@@ -1,4 +1,5 @@
 const joi = require('joi');
+const { modes, defaultMode } = require('./modes');
 const { retryParamsJoi, probeServerJoi, logJoiOptional, mongoJoi } =
     require('../../lib/config/configItems.joi');
 const { extensionConfigValidator } = require('../../lib/config/extensionConfigValidator');
@@ -8,6 +9,7 @@ const { MAX_QUEUED_DEFAULT }  = require('../../lib/constants').backbeatConsumer;
 const joiSchema = joi.object({
     topic: joi.string().required(),
     groupId: joi.string().required(),
+    mode: joi.string().valid(...Object.keys(modes)).default(defaultMode),
     mongodb: mongoJoi,
     retry: retryParamsJoi,
     concurrency: joi.number().greater(0).default(1),
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -4,11 +4,10 @@ const async = require('async');
 
 const Logger = require('werelogs').Logger;
 const errors = require('arsenal').errors;
-const { replicationBackends, emptyFileMd5 } = require('arsenal').constants;
+const { replicationBackends } = require('arsenal').constants;
 const MongoClient = require('arsenal').storage
     .metadata.mongoclient.MongoClientInterface;
 const { ObjectMD, ReplicationConfiguration } = require('arsenal').models;
-const { VersionID } = require('arsenal').versioning;
 const { extractVersionId } = require('../../lib/util/versioning');
 
 const Config = require('../../lib/Config');
@@ -19,7 +18,7 @@ const ObjectQueueEntry = require('../../lib/models/ObjectQueueEntry');
 const MetricsProducer = require('../../lib/MetricsProducer');
 const { metricsExtension, metricsTypeCompleted, metricsTypePendingOnly } =
     require('../ingestion/constants');
-const getContentType = require('./utils/contentTypeHelper');
+const { modes, defaultMode } = require('./modes');
 const BucketMemState = require('./utils/BucketMemState');
 const MongoProcessorMetrics = require('./MongoProcessorMetrics');
 
@@ -76,6 +75,8 @@ class MongoQueueProcessor {
         this._bootstrapList = null;
         this.logger = new Logger('Backbeat:Ingestion:MongoProcessor');
         this.mongoClientConfig.logger = this.logger;
+        this._mode =
+            new modes[mongoProcessorConfig.mode ?? defaultMode]();
         this._mongoClient = new MongoClient(this.mongoClientConfig);
         this._bucketMemState = new BucketMemState(Config);
 
@@ -252,78 +253,6 @@ class MongoQueueProcessor {
     }
 
     /**
-     * Update ingested entry metadata fields: owner-id, owner-display-name
-     * @param {ObjectQueueEntry} entry - object queue entry object
-     * @param {BucketInfo} bucketInfo - bucket info object
-     * @return {undefined}
-     */
-    _updateOwnerMD(entry, bucketInfo) {
-        // zenko bucket owner information is being set on ingested md
-        entry.setOwnerDisplayName(bucketInfo.getOwnerDisplayName());
-        entry.setOwnerId(bucketInfo.getOwner());
-    }
-
-    /**
-     * Update ingested entry metadata fields: dataStoreName
-     * @param {ObjectQueueEntry} entry - object queue entry object
-     * @param {string} location - owner details
-     * @return {undefined}
-     */
-    _updateObjectDataStoreName(entry, location) {
-        entry.setDataStoreName(location);
-    }
-
-    /**
-     * Update ingested entry metadata location field. Each location change
-     * includes: key, dataStoreName, dataStoreType, dataStoreVersionId
-     * @param {ObjectQueueEntry} entry - object queue entry object
-     * @param {string} zenkoLocation - zenko storage location name
-     * @return {undefined}
-     */
-    _updateLocations(entry, zenkoLocation) {
-        const locations = entry.getLocation();
-        // if version id is undefined, we have a single null object.
-        // To hold reference to this null object, we need to encode "null"
-        // as its dataStoreVersionId
-        const dataStoreVersionId = entry.getVersionId() ?
-            entry.getEncodedVersionId() : 'null';
-        let zenkoDataLocations;
-        if (!locations || locations.length === 0) {
-            zenkoDataLocations = [{
-                key: entry.getObjectKey(),
-                size: 0,
-                start: 0,
-                dataStoreName: zenkoLocation,
-                dataStoreType: 'aws_s3',
-                dataStoreETag: `1:${emptyFileMd5}`,
-                dataStoreVersionId,
-            }];
-        } else {
-            zenkoDataLocations = [{
-                key: entry.getObjectKey(),
-                size: entry.getContentLength(),
-                start: 0,
-                dataStoreName: zenkoLocation,
-                dataStoreType: 'aws_s3',
-                dataStoreETag: `1:${entry.getContentMd5()}`,
-                dataStoreVersionId,
-            }];
-        }
-        entry.setLocation(zenkoDataLocations);
-    }
-
-    /**
-     * Update acl info on ingested object MD
-     * @param {ObjectQueueEntry} entry - object queue entry object
-     * @return {undefined}
-     */
-    _updateAcl(entry) {
-        // reset acl info
-        const objectMDModel = new ObjectMD();
-        entry.setAcl(objectMDModel.getAcl());
-    }
-
-    /**
      * Update replication info on ingested object MD to match Zenko defined
      * replication info.
      * @param {ObjectQueueEntry} entry - object queue entry object
@@ -382,29 +311,17 @@ class MongoQueueProcessor {
         const key = sourceEntry.getObjectKey();
         const entryVersionId = extractVersionId(sourceEntry.getObjectVersionedKey());
 
-        // Use x-amz-meta-scal-version-id if provided, instead of the actual versionId of the object.
-        // This should happen only for restored objects : in all other situations, both the source
-        // and ingested objects should have the same version id (and no x-amz-meta-scal-version-id
-        // metadata).
         const scalVersionId = sourceEntry.getOverheadField('x-amz-meta-scal-version-id');
-        const versionId = scalVersionId ? VersionID.decode(scalVersionId) : entryVersionId;
+        const versionId =
+            this._mode.resolveVersionId(scalVersionId, entryVersionId);
 
         this.logger.debug('processing object delete', { bucket, key, versionId });
 
         async.waterfall([
             cb => this._getZenkoObjectMetadata(log, sourceEntry, versionId, cb),
             (zenkoObjMd, cb) => {
-                // Skip if the object is in a different location, i.e. when the delete was caused
-                // by restored-object expiration or transition. It works because the dataStoreName
-                // is updated before actually sending the object to GC to effectively delete the
-                // data.
-                const encode = versionId => (versionId ? VersionID.encode(versionId) : 'null');
-                if (zenkoObjMd.dataStoreName !== location ||
-                    zenkoObjMd.location?.length !== 1 ||
-                    zenkoObjMd.location[0].dataStoreName !== location ||
-                    zenkoObjMd.location[0].key !== key ||
-                    (zenkoObjMd.location[0].dataStoreVersionId || 'null') !== encode(entryVersionId)
-                ) {
+                if (!this._mode.shouldProcessDelete(zenkoObjMd, location, key,
+                                                   entryVersionId)) {
                     log.end().info('ignore delete entry, transitioned to another location', {
                         entry: sourceEntry.getLogInfo(),
                         location,
@@ -482,19 +399,12 @@ class MongoQueueProcessor {
         this.logger.debug('processing object metadata', { bucket, key, scalVersionId });
 
         const maybeGetZenkoObjectMetadata = cb => {
-            // NOTE: ZenkoObjMD is used for updating replication info, as well as validating the
-            // `x-amz-meta-scal-version-id` header of restored objects. If the Zenko bucket does
-            // not have repInfo set and the header is not set, then we can skip fetching.
-            const bucketRepInfo = bucketInfo.getReplicationConfiguration();
-            if (!scalVersionId && !bucketRepInfo?.rules?.some(r => r.enabled)) {
+            if (!this._mode.needsExistingMetadata(sourceEntry, bucketInfo)) {
                 return cb();
             }
 
-            // Use x-amz-meta-scal-version-id if provided, instead of the actual versionId of the object.
-            // This should happen only for restored objects : in all other situations, both the source
-            // and ingested objects should have the same version id (and not x-amz-meta-scal-version-id
-            // metadata).
-            const versionId = scalVersionId ? VersionID.decode(scalVersionId) : sourceEntry.getVersionId();
+            const versionId = this._mode.resolveVersionId(scalVersionId,
+                sourceEntry.getVersionId());
             return this._getZenkoObjectMetadata(log, sourceEntry, versionId, cb);
         };
 
@@ -509,7 +419,8 @@ class MongoQueueProcessor {
                 return done(err);
             }
 
-            const content = getContentType(sourceEntry, zenkoObjMd);
+            const content =
+                this._mode.getChangedContent(sourceEntry, zenkoObjMd);
             if (content.length === 0) {
                 this._normalizePendingMetric(location);
                 log.end().debug('skipping duplicate entry', {
@@ -522,16 +433,10 @@ class MongoQueueProcessor {
             }
 
             if (zenkoObjMd) {
-                // Keep existing metadata fields, only need to update the tags
-                const tags = sourceEntry.getTags();
-                sourceEntry._data = { ...zenkoObjMd }; // eslint-disable-line no-param-reassign
-                sourceEntry.setTags(tags);
+                this._mode.mergeExistingMetadata(sourceEntry, zenkoObjMd);
             } else {
-                // Update necessary metadata fields before saving to Zenko MongoDB
-                this._updateOwnerMD(sourceEntry, bucketInfo);
-                this._updateObjectDataStoreName(sourceEntry, location);
-                this._updateLocations(sourceEntry, location);
-                this._updateAcl(sourceEntry);
+                this._mode.applyNewObjectMetadata(sourceEntry, location,
+                                                 bucketInfo);
             }
 
             // Try to update replication info, if applicable
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -432,7 +432,8 @@ class MongoQueueProcessor {
                 return done();
             }
 
-            if (zenkoObjMd) {
+            if (zenkoObjMd &&
+                !this._mode.replacesExistingMetadata(sourceEntry, zenkoObjMd)) {
                 this._mode.mergeExistingMetadata(sourceEntry, zenkoObjMd);
             } else {
                 this._mode.applyNewObjectMetadata(sourceEntry, location,
--- a/extensions/mongoProcessor/modes/DRMode.js
+++ b/extensions/mongoProcessor/modes/DRMode.js
@@ -1,0 +1,97 @@
+'use strict';
+
+const { ObjectMD } = require('arsenal').models;
+
+const locations = require('../../../lib/util/locations');
+const ProcessorMode = require('./ProcessorMode');
+const getContentType = require('../utils/contentTypeHelper');
+
+class DRMode extends ProcessorMode {
+    needsExistingMetadata(entry, bucketInfo) { // eslint-disable-line no-unused-vars
+        // always: only the stored document tells a first write from an update,
+        // and a source insert is redelivered on replay and overlaps the
+        // bootstrap dump
+        return true;
+    }
+
+    getChangedContent(entry, zenkoObjMd) {
+        // the shared diff only covers tags, while a replicated object can also
+        // change its object-lock state and, until localized, its placement
+        const content = getContentType(entry, zenkoObjMd);
+        if (!zenkoObjMd || content.length !== 0) {
+            return content;
+        }
+
+        return this._hasMutableChange(entry, zenkoObjMd) ? ['METADATA'] : [];
+    }
+
+    /**
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @param {Object} zenkoObjMd - metadata fetched from mongo
+     * @return {boolean} true if the entry changes mutable metadata
+     */
+    _hasMutableChange(entry, zenkoObjMd) {
+        return entry.getRetentionMode() !== zenkoObjMd.retentionMode ||
+            entry.getRetentionDate() !== zenkoObjMd.retentionDate ||
+            entry.getLegalHold() !== !!zenkoObjMd.legalHold ||
+            (this._isNotLocalized(zenkoObjMd) &&
+                entry.getDataStoreName() !== zenkoObjMd.dataStoreName);
+    }
+
+    /**
+     * A version whose data still lives on the remote site.
+     *
+     * @param {Object} zenkoObjMd - metadata fetched from mongo
+     * @return {boolean} true if the stored version is not localized
+     */
+    _isNotLocalized(zenkoObjMd) {
+        return locations.isCRRLocation(zenkoObjMd.dataStoreName);
+    }
+
+    applyNewObjectMetadata(entry, location, bucketInfo) { // eslint-disable-line no-unused-vars
+        // the source-side pipeline shaped everything else this object keeps;
+        // ACLs are not replicated, so they reset as for an ingested object
+        entry.setAcl(new ObjectMD().getAcl());
+    }
+
+    mergeExistingMetadata(entry, zenkoObjMd) {
+        // an update brings tags and object-lock state, cleared values included
+        const tags = entry.getTags();
+        const retentionMode = entry.getRetentionMode();
+        const retentionDate = entry.getRetentionDate();
+        const legalHold = entry.getLegalHold();
+        const notLocalized = this._isNotLocalized(zenkoObjMd);
+        const dataStoreName = entry.getDataStoreName();
+        const location = entry.getLocation();
+
+        entry._data = { ...zenkoObjMd }; // eslint-disable-line no-param-reassign
+
+        entry.setTags(tags);
+        entry.setRetentionMode(retentionMode);
+        entry.setRetentionDate(retentionDate);
+        entry.setLegalHold(legalHold);
+
+        // a localized version keeps the placement the copy engine gave it; one
+        // still on the remote site takes the entry's
+        if (notLocalized) {
+            entry.setDataStoreName(dataStoreName);
+            entry.setLocation(location);
+        }
+    }
+
+    resolveVersionId(scalVersionId, versionId) {
+        // version ids are identical on both sides, so the entry's is
+        // authoritative; a scal version id names a version of another system
+        // entirely, the one an out-of-band production object was ingested from
+        return versionId;
+    }
+
+    shouldProcessDelete(zenkoObjMd, location, key, versionId) { // eslint-disable-line no-unused-vars
+        // always: a replicated object's location legitimately differs -- cold,
+        // on the source, or localized -- so the ingestion guard would ignore
+        // every deletion
+        return true;
+    }
+}
+
+module.exports = DRMode;
--- a/extensions/mongoProcessor/modes/DRMode.js
+++ b/extensions/mongoProcessor/modes/DRMode.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { isDeepStrictEqual } = require('util');
+
 const { ObjectMD } = require('arsenal').models;
 
 const locations = require('../../../lib/util/locations');
@@ -22,7 +24,38 @@ class DRMode extends ProcessorMode {
             return content;
         }
 
+        // a version is immutable, so the same version id over different
+        // content is an object overwritten in place -- in a bucket that is not
+        // versioned, or whose versioning is suspended
+        if (this.replacesExistingMetadata(entry, zenkoObjMd)) {
+            return this._replacesStoredDocument(entry, zenkoObjMd) ? ['METADATA'] : [];
+        }
+
         return this._hasMutableChange(entry, zenkoObjMd) ? ['METADATA'] : [];
+    }
+
+    /**
+     * Whether writing the entry would leave a different document behind. The
+     * fields compared are the ones the write carries, so a replayed entry is
+     * told from one that changes anything at all, down to a header the older
+     * diff never looked at.
+     *
+     * replicationInfo is left out: the processor resolves it against this
+     * site's own bucket configuration once the entry has been applied, so it is
+     * not a difference the entry answers for.
+     *
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @param {Object} zenkoObjMd - metadata fetched from mongo
+     * @return {boolean} true if the stored document would change
+     */
+    _replacesStoredDocument(entry, zenkoObjMd) {
+        const written = { ...entry.getValue(), acl: new ObjectMD().getAcl() };
+        const stored = { ...zenkoObjMd };
+
+        delete written.replicationInfo;
+        delete stored.replicationInfo;
+
+        return !isDeepStrictEqual(written, stored);
     }
 
     /**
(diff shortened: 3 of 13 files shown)


extensions/mongoProcessor/mongoProcessorTask.js
```diff
@@ -63,10 +64,13 @@ async function handleMetrics(res, log) {
 }
 
 function loadManagementDatabase() {
-    const ingestionServiceAuth = config.extensions.ingestion.auth;
+    // the ingestion extension's auth is the fallback for a deployment that
+    // configures it; a D/R sink enables no extension but this one
+    const serviceAuth =
+        mongoProcessorConfig.auth ?? config.extensions.ingestion.auth;
```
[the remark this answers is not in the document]
delthas: Fixed

extensions/mongoProcessor/MongoProcessorConfigValidator.js
```diff
@@ -1,12 +1,17 @@
...
+    require('../../lib/config/configItems.joi');
 const { extensionConfigValidator } = require('../../lib/config/extensionConfigValidator');
 
 const { MAX_QUEUED_DEFAULT }  = require('../../lib/constants').backbeatConsumer;
 
 const joiSchema = joi.object({
     topic: joi.string().required(),
     groupId: joi.string().required(),
+    mode: joi.string().valid(...Object.keys(modes)).default(defaultMode),
+    mongodb: mongoJoi,
```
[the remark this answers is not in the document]
delthas: Taken, and confirmed: `MongoQueueProcessor` dereferences it at `this.mongoClientConfig.logger = this.logger`, so an undefined config throws exactly there.

`mongodb` is now required in D/R mode, same shape as `auth`. Both constraints sit in the same commit because they can only be expressed once `mode` and `auth` both exist.

extensions/mongoProcessor/modes/IngestionMode.js
```diff
@@ -0,0 +1,184 @@
...
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @return {undefined}
+     */
+    _updateAcl(entry) {
+        // reset acl info
+        const objectMDModel = new ObjectMD();
+        entry.setAcl(objectMDModel.getAcl());
+    }
+
+    /**
```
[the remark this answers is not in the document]
delthas: Fixed, though not by dropping it: that block was `mergeExistingMetadata`'s own doc, which inserting `replacesExistingMetadata` above it displaced. It reads identically at `d9516c48`, the commit before this one, sitting directly above `mergeExistingMetadata` — so `mergeExistingMetadata` was left with no JSDoc at all, and deleting the block would have made that permanent.

Moved back down rather than removed, and amended into the commit that displaced it.

delthas:
Requested @maeldonn in place of Sylvain Senechal, who is currently on PTO.

delthas:
Requested @SylvainSenechal in place of Mael Donnart, who is currently on PTO.

francoisferrand:
nit: PR description is way too long to read, schema way too verobose... so it does not help so much.

> replacesExistingMetadata | [...] | True for a write carrying no version id, or a null version — the same key is being rewritten, so the entry replaces the document rather than merging into it

- if it happens, we could/should use eTag (and/or possibly originOp) to identify if this is a metadata update only (→ keep the locationData) or 
- but for pull-replication, not really an issue: buckets should be be versioned (worth an error/warning/followup ticket, to make sure we never "silently" create an orphan)
- this covers PRA use-case only, so should have no data

→ not sure if/what we should do _in this PR_, but something to consider anyway

extensions/mongoProcessor/MongoProcessorConfigValidator.js
```diff
@@ -1,12 +1,17 @@
...
+    require('../../lib/config/configItems.joi');
 const { extensionConfigValidator } = require('../../lib/config/extensionConfigValidator');
 
 const { MAX_QUEUED_DEFAULT }  = require('../../lib/constants').backbeatConsumer;
 
 const joiSchema = joi.object({
     topic: joi.string().required(),
     groupId: joi.string().required(),
+    mode: joi.string().valid(...Object.keys(modes)).default(defaultMode),
+    mongodb: mongoJoi.when('mode', { is: 'dr', then: joi.required() }),
```
francoisferrand, line 13: mongodb config is already available in the "global" config, should not be duplicated here

extensions/mongoProcessor/modes/DRMode.js
```diff
@@ -0,0 +1,144 @@
...
+
+const { isDeepStrictEqual } = require('util');
+
+const { ObjectMD } = require('arsenal').models;
+
+const locations = require('../../../lib/util/locations');
+const ProcessorMode = require('./ProcessorMode');
+const getContentType = require('../utils/contentTypeHelper');
+
+class DRMode extends ProcessorMode {
```
francoisferrand, line 11: maybe more `PullReplication` than `DR` ?
(e.g. this is actually the _metadata_ part of the pull replication feature of backbeat)

extensions/mongoProcessor/mongoProcessorTask.js
```diff
@@ -18,9 +18,10 @@ const { startProbeServer } = require('../../lib/util/probe');
 const kafkaConfig = config.kafka;
 const mConfig = config.metrics;
 const mongoProcessorConfig = config.extensions.mongoProcessor;
-// TODO: consider whether we would want a separate mongo config
-// for the consumer side
-const mongoClientConfig = config.queuePopulator.mongo;
+// the queue populator's mongo config is the fallback for a deployment that
+// runs one beside this process; a D/R sink runs none, so it configures its own
```
francoisferrand, line 22: I don't understand this comment at all.
In any case, mongo-processor runs next to a mongo instance, and it is passed through the usual config.queuePopulator.mongo...

extensions/mongoProcessor/mongoProcessorTask.js
```diff
@@ -63,10 +64,13 @@ async function handleMetrics(res, log) {
 }
 
 function loadManagementDatabase() {
-    const ingestionServiceAuth = config.extensions.ingestion.auth;
+    // the ingestion extension's auth is the fallback for a deployment that
+    // configures it; a D/R sink enables no extension but this one
+    const serviceAuth =
+        mongoProcessorConfig.auth ?? config.extensions.ingestion?.auth;
```
francoisferrand, line 70: this one may be trickier, not sure if/how we want to hack around...

mongo-processor was (technically) extension, but kind of "part" of ingestion → so using that extension's config.

now you are making the mongoProcessor cover multiple extensions (both replication and ingestion, basically) ; and introducing lots of conditional to address this...
- does this mean it should be separate extensions, instead of "modes" ?
- or should these ternary be handled in the "mode" class ?
- or do we really need to have different service account? (btw, you kept the serviceName: so the difference does not seem so critical?)
- or just handle it at "runtime", e.g. add extra env variable in ZKOP ?

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -482,19 +399,12 @@ class MongoQueueProcessor {
         this.logger.debug('processing object metadata', { bucket, key, scalVersionId });
 
         const maybeGetZenkoObjectMetadata = cb => {
-            // NOTE: ZenkoObjMD is used for updating replication info, as well as validating the
-            // `x-amz-meta-scal-version-id` header of restored objects. If the Zenko bucket does
-            // not have repInfo set and the header is not set, then we can skip fetching.
-            const bucketRepInfo = bucketInfo.getReplicationConfiguration();
-            if (!scalVersionId && !bucketRepInfo?.rules?.some(r => r.enabled)) {
+            if (!this._mode.needsExistingMetadata(sourceEntry, bucketInfo)) {
```
francoisferrand, line 402: nit: I wonder if we need this condition as well... it is an optimization, but not sure how relevant it is...

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -482,19 +399,12 @@ class MongoQueueProcessor {
...
                 return cb();
             }
 
-            // Use x-amz-meta-scal-version-id if provided, instead of the actual versionId of the object.
-            // This should happen only for restored objects : in all other situations, both the source
-            // and ingested objects should have the same version id (and not x-amz-meta-scal-version-id
-            // metadata).
-            const versionId = scalVersionId ? VersionID.decode(scalVersionId) : sourceEntry.getVersionId();
+            const versionId = this._mode.resolveVersionId(scalVersionId,
```
francoisferrand, line 406: semantic of this one is weird : not sure exactly what is _possible_ (in the code), but either scal-verison-id should be purely ingestion specific (→ no scalVersionId in mongoQeueueProcessor.js, everything in "resolveVersionId"), or it needs to stay here and it should just be a condition on the first line (`const scalVersionId = mode.useSclaVersionId() ? sourceEntry.getValue()['x-amz-meta-scal-version-id'] : undefined`);

extensions/mongoProcessor/modes/ProcessorMode.js
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * A processor mode holds the decisions that differ between the streams the
+ * mongo-processor writes: out-of-band ingestion, where identity and placement
+ * are rewritten to local values because the source system's accounts and
+ * locations do not exist here, and D/R, where they are replicated and so are
+ * applied as they arrive.
+ *
+ * Everything else about processing an entry is shared.
+ */
+class ProcessorMode {
```
francoisferrand, line 14: `Mode` is kind a red flag, and the description of the type says the same : just a bunch of behaviors we extracted for the sake of reuse, where we need tweaking...

Can we push this further, so the type (or types, could be multiple "sub-modes") have a real purpose and thus a semantic - and thus can have a real, descriptive name for the class?

In particular, the current approach makes it very difficult to review/read the code of each "mode" (ingesting and DR/pull replication), as there is no context for the function : so need to go all the way to the design (and overly verbose PR description) to try and understand it...

(as it is, kind of look more like just a separate extension, but handled the other way)

lib/util/locations.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Tell whether a location holds data owned by a remote site.
+ *
+ * Data stored on such a location is remote production data: we may read it
+ * (e.g. to copy it locally), but we must never delete it.
+ *
+ * @param {String} dataStoreName - location name
+ * @return {Boolean} true if the location is a CRR (remote) location
+ */
+function isCRRLocation(dataStoreName) {
```
francoisferrand, line 12: does that function really help? It is not much shorter, requires an import as well....but hides the fact that a module loads the locationConfig, and the implicit decision "if the location does not exist, it is not crr"

francoisferrand: (also, not used consistently: since we have this check in many places already...)

lib/Config.js
```diff
@@ -118,9 +118,11 @@ class Config extends EventEmitter {
 
         // whitelist IP, CIDR for health checks
         const defaultHealthChecks = ['127.0.0.1/8', '::1'];
-        const healthChecks = parsedConfig.server.healthChecks;
-        healthChecks.allowFrom =
-            healthChecks.allowFrom.concat(defaultHealthChecks);
+        const healthChecks = parsedConfig.server?.healthChecks;
```
francoisferrand, line 121: not valid?
why change this?

tests/unit/lib/config/Config.spec.js
```diff
@@ -33,6 +33,12 @@ describe('Config', () => {
         assert.doesNotThrow(() => config._parseConfig(testConfig));
     });
 
+    it('should accept a config serving no API, as a D/R sink does', () => {
```
francoisferrand, line 36: no: DR should have a healthcheck anyway (like ingestion's mongo-processor did, AFAIK)

tests/functional/ingestion/MongoQueueProcessor.js
```diff
@@ -1069,3 +1070,400 @@ describe('MongoQueueProcessor', function mqp() {
...
+            const { objVal } = added[0];
+            assert.strictEqual(objVal['owner-id'], 'source-owner-id');
+            assert.strictEqual(objVal['owner-display-name'], 'source-owner');
+            assert.strictEqual(objVal.dataStoreName, 'cold-location');
+            assert.deepStrictEqual(objVal.acl, new ObjectMD().getAcl());
+            done();
+        });
+    });
+
+    it('should keep the stored placement when updating a localized object',
```
francoisferrand, line 1129: line wrap not needed (same for quite a few below, I think)

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -521,17 +432,12 @@ class MongoQueueProcessor {
                 return done();
             }
 
-            if (zenkoObjMd) {
-                // Keep existing metadata fields, only need to update the tags
-                const tags = sourceEntry.getTags();
-                sourceEntry._data = { ...zenkoObjMd }; // eslint-disable-line no-param-reassign
-                sourceEntry.setTags(tags);
+            if (zenkoObjMd &&
+                !this._mode.replacesExistingMetadata(sourceEntry, zenkoObjMd)) {
```
francoisferrand, line 436: why this new "replace metadata" ?
was is actually done in `getContentType` ?

extensions/mongoProcessor/modes/ProcessorMode.js
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Whether the entry's existing metadata has to be read before processing.
+     *
+     * This method must be implemented by subclasses of ProcessorMode
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @param {BucketInfo} bucketInfo - bucket info object
+     * @return {boolean} true if the stored document is needed
+     */
+    needsExistingMetadata(entry, bucketInfo) { // eslint-disable-line no-unused-vars
```
francoisferrand, line 23: should be better the other way : `skipFetchExistingObjMD` ?
the "safe" thing to do is to get metadata and do the merge ; but (typically for optimization purpose) we may want/allow skipping...

(note that it not only _does not fetch_, but it also does not check if object _exists_ at all...)

extensions/mongoProcessor/modes/ProcessorMode.js
```diff
@@ -0,0 +1,122 @@
...
+     * This method must be implemented by subclasses of ProcessorMode
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @param {Object} zenkoObjMd - metadata fetched from mongo
+     * @return {undefined}
+     */
+    mergeExistingMetadata(entry, zenkoObjMd) { // eslint-disable-line no-unused-vars
+        assert(false,
+            'sub-classes of ProcessorMode must implement ' +
+            'the mergeExistingMetadata() method');
+    }
```
francoisferrand, line 88: maybe the logic would be simpler with a single function here `mergeObjMD` : receiving both the entry and zenkoObjMD (which may or may not be null) ?
we may not even need getChangedContent in that case...

extensions/mongoProcessor/modes/ProcessorMode.js
```diff
@@ -0,0 +1,122 @@
...
+     * Whether a delete entry still applies to the object as stored.
+     *
+     * This method must be implemented by subclasses of ProcessorMode
+     * @param {Object} zenkoObjMd - metadata fetched from mongo
+     * @param {string} location - zenko storage location name
+     * @param {string} key - object key
+     * @param {string|undefined} versionId - decoded version id of the entry
+     * @return {boolean} true if the object should be deleted
+     */
+    shouldProcessDelete(zenkoObjMd, location, key, versionId) { // eslint-disable-line no-unused-vars
```
francoisferrand, line 115: same, would it be better to call this function `_skipDelete` ?

extensions/mongoProcessor/modes/DRMode.js
```diff
@@ -0,0 +1,144 @@
...
+     * @return {boolean} true if the stored document would change
+     */
+    _replacesStoredDocument(entry, zenkoObjMd) {
+        const written = { ...entry.getValue(), acl: new ObjectMD().getAcl() };
+        const stored = { ...zenkoObjMd };
+
+        delete written.replicationInfo;
+        delete stored.replicationInfo;
+
+        return !isDeepStrictEqual(written, stored);
```
francoisferrand: that is quite costly, and not really safe anyway (could be a concurrent update in mongo)
- there is no real "replace" : we should always _merge_, i.e. keep the existing replicationInfo, dataLocation, etc...
- in theory could be some corner case with non-versionned: but versioning is required (should be enforced) for pull replication, just like it is for replication

in general, the goal is to have the "same metadata": so I think we shoudl do similar to replication (in cloudserver) -and unlike ingestion-, i.e. keep all the metadata except the few field we must keep: dataLocation, dataStoreName (though a story of its own, e.g. what is the object was transitioned in between?...), replicationInfo, archive status... and not try to optimize mongo writes at all cost

extensions/mongoProcessor/modes/DRMode.js
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @param {Object} zenkoObjMd - metadata fetched from mongo
+     * @return {boolean} true if the entry changes mutable metadata
+     */
+    _hasMutableChange(entry, zenkoObjMd) {
```
francoisferrand, line 66: an efficient way to check for changes could be the microVersionId (added for CRR cascade), bumped on every metadata write which must be replicated?

extensions/mongoProcessor/modes/DRMode.js
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @param {Object} zenkoObjMd - metadata fetched from mongo
+     * @return {boolean} true if the stored version is not localized
+     */
+    _isNotLocalized(zenkoObjMd) {
+        return locations.isCRRLocation(zenkoObjMd.dataStoreName);
+    }
+
+    replacesExistingMetadata(entry, zenkoObjMd) { // eslint-disable-line no-unused-vars
+        // an object with no version of its own is rewritten in place, so the
```
francoisferrand, line 85: - the bucket is versioned. so only MD can change
- if it were not, we could not just replace the object, as it may create an orphan...

overall this point to the point already raised in ProcessorMode: having a single "mergeMetadata", to keep the whole merging logic in each "mode", may be clearer? (and may help defining more precisely what a mode is - or a more precise name/concept)

extensions/mongoProcessor/MongoProcessorConfigValidator.js
delthas: The global one lives under `queuePopulator`, whose `cronRule`, `zookeeperPath`, `logSource` and `probeServer` are all required — a sink that runs no populator would have to invent four fields to carry a mongo endpoint.

extensions/mongoProcessor/modes/DRMode.js
delthas: Agreed — it's `PullReplicationMetadataPolicy` now, under `extensions/mongoProcessor/metadataPolicy/`.

extensions/mongoProcessor/mongoProcessorTask.js
delthas: Reworded to say the blocker outright: a sink runs no queue populator, and that section cannot carry the mongo endpoint alone — it also requires a cron rule, a zookeeper path, a log source and a probe server.

extensions/mongoProcessor/mongoProcessorTask.js
delthas: Same shape: a partial `extensions.ingestion` block is still validated against ingestion's full schema, which also requires a topic, a zookeeper path and a source list. A second extension would duplicate the schema and the task to change four decisions, so I kept the mode and reworded the comment.

extensions/mongoProcessor/MongoQueueProcessor.js
delthas: It predates the PR: the pre-mode code skipped the fetch when there was neither a scal version id nor an enabled replication rule. I only moved it behind the policy, verbatim.

extensions/mongoProcessor/MongoQueueProcessor.js
delthas: Took the second option — the processor reads the header once per path and hands it to the policy. It cannot move inside: the object path reads it from the value, the delete path from the overhead fields.

extensions/mongoProcessor/modes/ProcessorMode.js
delthas: Reworked to `MetadataPolicy`, with `IngestionMetadataPolicy` and `PullReplicationMetadataPolicy`. The processor calls four methods, and applying an entry is one call that decides and writes.

lib/util/locations.js
delthas: Dropped. The policy reads `conf/locationConfig.json` directly, as eleven other consumers do. Only `LifecycleQueuePopulator` also takes it injected, which is why the test location had to go into the shipped file.

lib/Config.js
delthas: `server.healthChecks.allowFrom` is read only by `lib/api/BackbeatServer.js:58`, the API's whitelist. `server` is optional in the schema, so a process serving no API crashed here on `parsedConfig.server.healthChecks`.

tests/unit/lib/config/Config.spec.js
delthas: It does have one — the mongo-processor's probe server, which ZKOP probes. What the sink does not run is the backbeat API, and this whitelist is the API's.

tests/functional/ingestion/MongoQueueProcessor.js
delthas: Done — two were left, the `it()` title and the `getReplicationConfiguration` stub.

extensions/mongoProcessor/MongoQueueProcessor.js
delthas: No — `getContentType` compares tags only once a stored document exists, so an object overwritten in place under the same key diffed to nothing and was dropped. The decision now lives inside the policy's `apply`.

extensions/mongoProcessor/modes/ProcessorMode.js
delthas: Inverted and renamed to `skipsStoredMetadata`, false for pull replication. And yes, it skips the existence check too — that is the pre-mode behaviour, unchanged here.

extensions/mongoProcessor/modes/ProcessorMode.js
delthas: Done: `apply(entry, stored, location, bucketInfo)` returns the content list or null. `changedContent` survives as its first half because the processor still needs that list for `replicationInfo.content`.

extensions/mongoProcessor/modes/ProcessorMode.js
delthas: Renamed to `skipsDelete(stored, location, key, versionId)`, false for pull replication.

extensions/mongoProcessor/modes/DRMode.js
delthas: As discussed offline, keeping it. It fires for an object with no version id, and for the null master a suspended bucket rewrites in place; without it an unversioned object overwritten in place keeps the previous document whole on the sink, md5, length and archive info included. The comparison now runs against the document as mongo holds it, so a replay really is a no-op.

extensions/mongoProcessor/modes/DRMode.js
delthas: The call sites gate on the bucket's replication configuration — `getReplicationInfo` returns undefined without one — so a plain PUT, a transition and a delete marker leave it alone. The exception is the `x-scal-s3-version-id` restore overwrite, which sets it through `metadataStoreObject`, outside `bumpMicroVersionId` entirely.

extensions/mongoProcessor/modes/DRMode.js
delthas: The merging logic is in one place per policy now — `replicationInfo` aside, which the processor still resolves against this site's bucket configuration after the policy has applied.

delthas:
Noting a finding from reviewing this against the ingestion path, for whoever picks up the source pipeline (scality/zenko-operator#631) — nothing actioned here.

`IngestionQueuePopulator._isMasterKeyEntry` drops master-key entries before they reach the processor, except single-null and versioning-suspended ones, and its comment states the invariant the processor's `isNull` handling relies on:

> Versioning suspended entries will have a version id but also a isNull tag. These master keys are considered a version and do not have a duplicate version, we also don't filter this type of entry.

The D/R pipeline has no equivalent, so it emits every matching document. Two consequences:

- **Versioned objects**: the master and the version document both match the filter, and the put envelope keys both by `$value.key`, so they arrive as byte-identical entries. The second is correctly dropped as a duplicate, at the cost of one extra entry and one extra `getObject` per put. `putObjectVerCase4` derives the master from `getLatestVersion()` on this site anyway, so the master entry never carried anything the version entry did not.
- **Suspended objects**: the transition goes through `PUT /_/backbeat/metadata/...?versionId=null`, which resolves to the stored version id and so lands in `putObjectVerCase3` — writing a version document beside the null master. Both then replicate, and since both carry `isNull`, the processor reads and writes both on the master key: the sink keeps the master and never creates the version document. That is the shape a suspended PUT produces canonically, and `?versionId=null` resolves through the master either way, so I do not think anything is lost — but the sink's document set does differ from the source's, and from what the Kafka Connect sink produced.

extensions/mongoProcessor/metadataPolicy/PullReplicationMetadataPolicy.js
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @param {Object} stored - metadata fetched from mongo
+     * @return {boolean} true if the entry changes mutable metadata
+     */
+    _hasMutableChange(entry, stored) {
+        return entry.getRetentionMode() !== stored.retentionMode ||
+            entry.getRetentionDate() !== stored.retentionDate ||
+            entry.getLegalHold() !== !!stored.legalHold ||
```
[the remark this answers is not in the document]
delthas: Arsenal normalises that side already: `getLegalHold()` is `this._data.legalHold || false`, so the entry always yields a boolean. The `!!` is there because `stored` is the raw mongo document rather than an `ObjectMD`, where the field can simply be absent — on a fresh `ObjectMD`, `getLegalHold()` is `false` while `_data.legalHold` is `undefined`, and the comparison already holds.

extensions/mongoProcessor/mongoProcessorTask.js
delthas: Correction to the above: I dropped the comment rather than rewording it. The rationale belongs in the commit message, which carries it.

extensions/mongoProcessor/mongoProcessorTask.js
delthas: Correction to the above: I dropped the comment rather than rewording it. The rationale belongs in the commit message, which carries it.

delthas:
> if it happens, we could/should use eTag (and/or possibly originOp) to identify if this is a metadata update only

The overwrite check uses the archive and `last-modified`; the eTag can't tell, as a PUT of the same bytes keeps it. Versioning isn't enforced: PRA keeps unversioned and suspended buckets, cold objects only, while clean room will require versioning. Orphans on the sink are the GC gap, BB-813 and BB-815, listed in the description.

extensions/mongoProcessor/MongoProcessorConfigValidator.js
delthas: That was my point for editing the config, not just making it look nice, but also this avoids having dummy fields in the config that are unused. Up for discussion :)

extensions/mongoProcessor/modes/DRMode.js
delthas: Wdyt? Reading `archive` and `last-modified` to see if the unversioned master body has changed?  (We can't just look at eg `md5` cause you could recreate an unversioned object with the same body but different fields)

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -382,29 +316,17 @@ class MongoQueueProcessor {
...
-                // data.
-                const encode = versionId => (versionId ? VersionID.encode(versionId) : 'null');
-                if (zenkoObjMd.dataStoreName !== location ||
-                    zenkoObjMd.location?.length !== 1 ||
-                    zenkoObjMd.location[0].dataStoreName !== location ||
-                    zenkoObjMd.location[0].key !== key ||
-                    (zenkoObjMd.location[0].dataStoreVersionId || 'null') !== encode(entryVersionId)
-                ) {
+                if (this._policy.skipsDelete(zenkoObjMd, location, key,
+                                             entryVersionId)) {
```
francoisferrand: not correct, this was actually `versionId`

extensions/mongoProcessor/metadataPolicy/MetadataPolicy.js
```diff
@@ -0,0 +1,115 @@
...
+ * and so are applied as they arrive.
+ *
+ * Everything else about processing an entry is shared.
+ */
+class MetadataPolicy {
+    /**
+     * Whether the metadata already stored for the object can be left unread.
+     *
+     * This method must be implemented by subclasses of MetadataPolicy
+     * @param {string|undefined} scalVersionId - encoded scal version id the
```
francoisferrand: 3 functions here operator only on scalVersionId, the code in MongoQueueProcessor does not do _anything_ with this, and this header/feature is only used by one of the policies.

So it seems `scalVersionId` should be inside "ingestionMetadataPolicy" (maybe we a single "canonicalize entry" function in the policy, if needed? or tweak `skip` functions so they can either skip or return the entry to act one, ...)

(problably a bit more complex, as it requires pulling the _logic/intent_ behind the code, and not just moving the code; but worth the effort to keep the code clean and maintainable)

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -482,24 +404,17 @@ class MongoQueueProcessor {
         this.logger.debug('processing object metadata', { bucket, key, scalVersionId });
 
         const maybeGetZenkoObjectMetadata = cb => {
-            // NOTE: ZenkoObjMD is used for updating replication info, as well as validating the
-            // `x-amz-meta-scal-version-id` header of restored objects. If the Zenko bucket does
-            // not have repInfo set and the header is not set, then we can skip fetching.
-            const bucketRepInfo = bucketInfo.getReplicationConfiguration();
-            if (!scalVersionId && !bucketRepInfo?.rules?.some(r => r.enabled)) {
+            if (this._policy.skipsStoredMetadata(scalVersionId, bucketInfo)) {
```
francoisferrand: this reads very weird (as a contract) that we decide to skip based on `scalVersionId` field, which is most of the time not present...

extensions/mongoProcessor/metadataPolicy/MetadataPolicy.js
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Which version of the object an entry writes, the master following it;
+     * without one, only the master is written.
+     *
+     * This method must be implemented by subclasses of MetadataPolicy
+     * @param {ObjectQueueEntry} entry - object queue entry object
+     * @return {string|undefined} version id to write, or undefined for the
+     *   master only
+     */
+    putVersionId(entry) { // eslint-disable-line no-unused-vars
```
francoisferrand: name read very weird: looks like this would function would actually put a new versoin

extensions/mongoProcessor/metadataPolicy/IngestionMetadataPolicy.js
```diff
@@ -0,0 +1,169 @@
...
+        // Versioning suspended entries will have a version id but also a isNull tag.
+        // These master keys are considered a version and do not have a duplicate version,
+        // we don't specify the version id and repairMaster in this case
+        if (entry.getVersionId() && !entry.getIsNull()) {
+            return entry.getVersionId();
+        }
+        return undefined;
+    }
+
+    apply(entry, stored, location, bucketInfo) {
```
francoisferrand: hard to read, not clear what "stored" is: either use temporal distinction (oldEntry / newEntry), or the "kind" of object (entry -> new stuff coming through kafka ; objMD -> what is already stored)

extensions/mongoProcessor/metadataPolicy/IngestionMetadataPolicy.js
```diff
@@ -0,0 +1,169 @@
...
+        // master keys with a 'null' version id comming from
+        // a versioning suspended bucket are considered a version
+        // we should not specify the version id in this case
+        if (entry.getIsNull()) {
+            return undefined;
+        }
+        return this.resolveVersionId(scalVersionId, entry.getVersionId());
+    }
+
+    putVersionId(entry) {
```
francoisferrand: putVersionId, storedVersionId, resolveVersionId, and the scalVersionId handled in MongoQueueProcessor all point to the same thing : we probably need just one function to do the mapping, and call it consistently... (even if it was not the case before)

extensions/mongoProcessor/MongoProcessorConfigValidator.js
francoisferrand: probeServer should be needed anyway; and cronRule will likely come here too.
so only zookeeperPath and probeServer remain.

What i am concerned if that we create 2 paths to configure the same module, and end up with duplication (like in ZKOP or config, if it ends up being shared) or inconsistencies...

extensions/mongoProcessor/metadataPolicy/MetadataPolicy.js
SylvainSenechal: I don't fully understand the comment, you want to remove some parameters because they are only use in one child class but not the other ?

If so I don't know I think this class here is fine declaring what functions are required and which parameters they receive regardless of whether we use them or not

extensions/mongoProcessor/metadataPolicy/IngestionMetadataPolicy.js
SylvainSenechal: I agree it can be a bit hard to follow whats a kafka message from an object metadata, and whats existing/coming. This is also reflected in the commit messages

extensions/mongoProcessor/metadataPolicy/IngestionMetadataPolicy.js
SylvainSenechal: I agree there might be something weird.
For example in pullReplicationMetadataPolicy,
storedVersionId return putVersionID directly.

But still not sure all of this is refactorable because there are some differences, and with the current implementation at least mongoProcessor can call the method without having to care about whch class its working on

PR #2835, opened 2026-09-01
Answer every rebalance callback raised during shutdown

delthas:
librdkafka requires every rebalance callback to be answered, and the shutdown path answered none of them. Each case fails differently, and each leaves the client parked in the rebalance, so the disconnect wedges on it and the process can never exit.

```mermaid
flowchart TD
    R["rebalance callback raised<br/>after the shutdown began"] --> Q{"which"}
    Q -->|revoke| U["un-assign — answers it"]
    Q -->|grant| A["decline with an empty assign"]
    A --> F{"disconnect already started?"}
    F -->|"yes — assign is refused"| U2["un-assign — still permitted"]
    F -->|no| D["declined, nothing to revoke later"]
    U --> OK["client leaves the rebalance,<br/>disconnect completes"]
    U2 --> OK
    D --> OK
```

## Changes

**A revoke** was left for `close()` to answer by un-assigning later. That holds for the revoke `unsubscribe()` itself raises, but one arriving *after* `close()` has already un-assigned has nothing left to answer it. Answering here does not cut the drain short: `close()` still waits for the in-flight work, the partitions are just handed back sooner, and the offsets that drain exists to commit (BB-758) are unaffected.

**A grant** was accepted, which left the disconnect an assignment to revoke all over again. It is declined instead — but `assign()` is refused once `disconnect()` has started, since the binding gates it on `isConnected()` while permitting `unassign()` for the whole close, so the decline falls back to `unassign()`. librdkafka coerces an assign into a full unassign during termination anyway.

## Verification

Three unit tests: a grant during shutdown is declined, a grant is still answered when `assign()` is refused mid-close, and a revoke during shutdown is answered. Each was checked against the previous implementation to confirm it fails there.

This is the defect that decides whether the process exits at all, and it was found by measurement rather than review. Across 8 full runs of the lib suite per arm on CI:

| | before this PR | with it |
| --- | --- | --- |
| suite runs where the process exits | 1 / 8 | **8 / 8** (matches 9.5) |
| disconnect bound fires → process fails to exit | 40 / 40 | bound no longer fires |

The bounded disconnect was returning from `close()` looking successful while leaving a client whose destructor blocks. Answering the callbacks is what stops it firing, so the bound is a backstop rather than load-bearing.

The pod-level census found the grant case independently: the callback still accepted a partition grant *after* `close()` had handed the partitions back, so the disconnect had to revoke them again, wedged, hit its 5 s bound, and returned without a `LeaveGroup` — the surviving members then waited out `session.timeout.ms`. That was 6 of 69 iterations; declining took the SIGKILL rate to zero and halved the residual. No message was lost and nothing was committed past unprocessed work across 240 iterations.

Issue: BB-833

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -819,9 +819,28 @@ class BackbeatConsumer extends EventEmitter {
         if (err.code === kafka.CODES.ERRORS.ERR__ASSIGN_PARTITIONS) {
             this._log.info('rdkafka.assign', { assignment });
 
-            if (!this._shuttingDown) {
-                this._setDrain(null);
+            // close() has already handed the partitions back. Accepting a
+            // fresh grant now leaves the client holding an assignment the
+            // disconnect has to revoke all over again, which wedges it and
+            // costs us the LeaveGroup the shutdown exists to send.
+            if (this._shuttingDown) {
+                // assign() is refused once disconnect() has started -- the
+                // binding gates it on isConnected(), which is already false --
+                // while unassign() is explicitly permitted while closing, and
+                // librdkafka coerces an assign into a full unassign during
+                // termination anyway. Leaving the callback unanswered is what
+                // parks the client in the rebalance.
+                this._bestEffort('assign.declined', () => {
+                    try {
+                        this._consumer.assign([]);
+                    } catch (e) { // eslint-disable-line no-unused-vars
+                        this._consumer.unassign();
+                    }
+                });
+                return;
             }
+
+            this._setDrain(null);
 
             try {
                 this._consumer.assign(assignment);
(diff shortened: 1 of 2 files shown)


lib/BackbeatConsumer.js
```diff
@@ -819,10 +819,29 @@ class BackbeatConsumer extends EventEmitter {
...
+            // fresh grant now leaves the client holding an assignment the
+            // disconnect has to revoke all over again, which wedges it and
+            // costs us the LeaveGroup the shutdown exists to send.
+            if (this._shuttingDown) {
+                // assign() is refused once disconnect() has started -- the
+                // binding gates it on isConnected(), which is already false --
+                // while unassign() is explicitly permitted while closing, and
+                // librdkafka coerces an assign into a full unassign during
+                // termination anyway. Leaving the callback unanswered is what
+                // parks the client in the rebalance.
```
francoisferrand, line 832: based on this, shoud we not just _always_ call `unassign()` in that case, instead of trying to `assign([])` first?

lib/BackbeatConsumer.js
```diff
@@ -819,10 +819,29 @@ class BackbeatConsumer extends EventEmitter {
...
+                    try {
+                        this._consumer.assign([]);
+                    } catch (e) { // eslint-disable-line no-unused-vars
+                        this._consumer.unassign();
+                    }
+                });
+                return;
             }
 
+            this._setDrain(null);
```
francoisferrand, line 843: should not be needed, c.f. first PR in the stack: a new assign cannot happen during an earlier unassign, so setDrain should stay managed in a single place (during partition revoke).

francoisferrand (changes requested):
could this have been created by the first PR, which changes the unsubscribe sequence?
If we are not subscribed, we should not get assigned: so doing it in the shutdown avoids the issue? Or is it just the race condition where we unsubscribe right before processing the assign partitions event?

PR #2836, opened 2026-09-01
Answer every caller when close() is called more than once

delthas:
The services install their SIGTERM handlers with `process.on` rather than `once`, so a repeated signal calls `close()` again. The second call started its own drain wait, overwriting the single drain slot the first was waiting on.

```mermaid
sequenceDiagram
    participant A as first caller
    participant C as BackbeatConsumer
    participant B as second caller
    A->>C: close()
    Note right of C: drain starts, first caller<br/>waits on the single drain slot
    B->>C: close() — SIGTERM repeated
    Note right of C: second wait overwrites the slot
    C-->>B: answered when the drain completes
    Note over A: never answered — released only by<br/>its own timeout, minutes after the<br/>consumer had already left the group
```

## Changes

Coalesce instead: the first call runs the shutdown, later ones attach to it, and every caller is answered once it completes.

## Verification

One unit test: closing twice answers both callers exactly once. It was checked against the previous implementation to confirm it fails there.

Issue: BB-833

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -181,6 +181,7 @@ class BackbeatConsumer extends EventEmitter {
         // a deferred un-assign gives up when this no longer matches
         this._rebalanceId = 0;
         this._shuttingDown = false;
+        this._closeCallbacks = null;
 
         this._messagesConsumed = 0;
         // this variable represents how many kafka messages have been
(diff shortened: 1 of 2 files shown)


delthas:
Requested @francoisferrand in place of Sylvain Senechal, who is currently on PTO.

francoisferrand (commented):
Is the issue really the _callbacks_, or really just that the second call to `close()` should immediately abort (i.e. not trigger further drain/...) ?

i.e. find with that approach, but I wonder if just switch to once() (or a guard in the event handler) would not be simpler and more appropriate?

PR #2837, opened 2026-09-02
Guard the deferred un-assign against a superseding rebalance

delthas:
A consumer that defers answering a partition revoke can lose an assignment it was granted in the meantime, and stall forever.

## The bug

When a revoke arrives while tasks are in flight, `BackbeatConsumer` does not call `unassign()` right away: it waits for the processing queue and the offset ledger to drain. While we owe librdkafka that answer, the rebalance protocol is effectively on hold — but a **subscribed topic that looks changed in metadata** is handled outside of it: librdkafka rejoins the group immediately, without waiting for our answer. The new generation can grant the partitions right back, and we apply them… then the drain completes, and the deferred `unassign()` throws that fresh assignment away.

```mermaid
sequenceDiagram
    participant BB as BackbeatConsumer
    participant RK as librdkafka
    participant B as broker

    RK->>BB: revoke [p0..p4]
    Note over BB: tasks in flight →<br/>defer unassign() until drained
    Note over RK: waiting for unassign()…
    B-->>RK: metadata: subscribed topic looks changed
    Note over RK: …but a topic change makes librdkafka<br/>rejoin now, revoke still unanswered
    RK->>B: JoinGroup + SyncGroup
    B-->>RK: new generation, same partitions
    RK->>BB: assign [p0..p4]
    BB->>RK: assign() — consuming again
    Note over BB: drain completes
    BB->>RK: unassign() — discards the fresh assignment
    Note over BB,B: broker: member still owns p0..p4 · local: nothing assigned<br/>membership unchanged → no rebalance will ever fix it
```

**What makes a topic "look changed"**: librdkafka compares consecutive snapshots of its local metadata cache, restricted to the subscription, on topic name, partition count and (when `client.rack` is set) replica racks. Anything that makes two snapshots differ counts:

- a partition count change (what the test below uses);
- the subscribed topic being created or deleted;
- replica rack changes, on consumers configured with `client.rack` (follower fetching);
- a transiently wrong answer from a metadata-stale broker during a rolling restart: a topic-level `UNKNOWN_TOPIC_OR_PART` for an existing topic plants a short-lived negative cache entry, so the topic "vanishes" for one observation and "reappears" at the next — two spurious rebalances with no actual change to the topic;
- `TOPIC_AUTHORIZATION_FAILED` flaps (ACL churn), same mechanism with a longer-lived negative entry;
- on librdkafka 2.3.0, nothing at all was needed — see Versions below.

Leader changes, ISR changes, and refreshes that return the same data do **not** trigger it.

The end state is silent and permanent: the consumer is connected, a group member, owns every partition at the broker and passes every liveness check — and consumes nothing. Observed in a CTST run as a GC pod idling 2h35m with 188 of 199 messages stranded.

## The fix

Only the newest rebalance may act:

- on every rebalance event, disarm the previous revoke's pending drain triggers and watchdog, so a superseded un-assign can no longer fire. As a side effect this also stops a second revoke from leaving an orphaned watchdog behind, which would disconnect a healthy consumer ~5 minutes later;
- the un-assign continuation can already be in flight when the newer event lands (through the asynchronous offset publish), so it also checks a rebalance counter before releasing the assignment.

## Reproduction

The new functional test manufactures the interleaving deterministically against a real broker: one held task as the in-flight work, a partition-count bump as the topic change. It is red on `development/9.5` and green with this fix — in two variants: the change observed by an application `getMetadata()` call, and by librdkafka's own periodic `topic.metadata.refresh.interval.ms` refresh, i.e. with no application metadata call involved at all.

## Versions

On librdkafka 2.3.0 the trigger fired constantly: every full metadata refresh wiped the metadata cache, so even a completely static topic read as "vanished" then "reappeared" on each refresh (fixed upstream in 2.4.0 by confluentinc/librdkafka#4677). The rejoin-while-a-revoke-is-unanswered behaviour itself is unchanged through 2.12.0, where the test reproduces the stall with a genuine topic change.

Replaces #2818.

Issue: BB-835

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -170,6 +170,9 @@ class BackbeatConsumer extends EventEmitter {
         // installed below delegates through this flag, so we don't
         // need to add/remove ledger listeners on every rebalance.
         this._drainCallback = null;
+        // Bumped on every rebalance event; a deferred un-assign
+        // captured under an older value must not run.
+        this._rebalanceId = 0;
         this._offsetLedger.on('empty', topic => {
             if (topic === this._topic && this._drainCallback) {
                 this._drainCallback();
@@ -758,6 +761,15 @@ class BackbeatConsumer extends EventEmitter {
      * @returns {void}
      */
     _onRebalance(err, assignment) {
+        // A new rebalance supersedes whatever a previous revoke left
+        // pending: disarm its drain triggers and watchdog, and stamp
+        // this event so a continuation already in flight (the async
+        // offset publish below) can tell it has been superseded.
+        const rebalanceId = ++this._rebalanceId;
+        this._setDrain(null);
+        clearTimeout(this._drainProcessQueueTimeout);
+        this._drainProcessQueueTimeout = null;
+
         if (err.code === kafka.CODES.ERRORS.ERR__ASSIGN_PARTITIONS) {
             this._log.info('rdkafka.assign', { assignment });
 
@@ -804,6 +816,17 @@ class BackbeatConsumer extends EventEmitter {
                 }
 
                 const doUnassign = () => {
+                    if (rebalanceId !== this._rebalanceId) {
+                        // a newer rebalance landed while offsets were
+                        // being published: the assignment is no longer
+                        // ours to release
+                        this._log.info('skipping superseded un-assign', {
+                            topic: this._topic,
+                            rebalanceId,
+                            currentRebalanceId: this._rebalanceId,
+                        });
+                        return;
+                    }
                     this._resumePausedPartitions();
 
                     try {
--- a/tests/functional/lib/BackbeatConsumerRebalanceSupersede.js
+++ b/tests/functional/lib/BackbeatConsumerRebalanceSupersede.js
@@ -1,0 +1,263 @@
+const assert = require('assert');
+const { promisify } = require('util');
+const kafka = require('node-rdkafka');
+const sinon = require('sinon');
+
+const BackbeatProducer = require('../../../lib/BackbeatProducer');
+const BackbeatConsumer = require('../../../lib/BackbeatConsumer');
+const { withTopicPrefix } = require('../../../lib/util/topic');
+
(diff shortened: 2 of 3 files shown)


lib/BackbeatConsumer.js
```diff
@@ -758,6 +761,15 @@ class BackbeatConsumer extends EventEmitter {
      * @returns {void}
      */
     _onRebalance(err, assignment) {
+        // A new rebalance supersedes whatever a previous revoke left
+        // pending: disarm its drain triggers and watchdog, and stamp
+        // this event so a continuation already in flight (the async
+        // offset publish below) can tell it has been superseded.
+        const rebalanceId = ++this._rebalanceId;
```
francoisferrand, line 768: what events can we get, is this condition not too strong?

- if we receive UNASSIGN while already unassigned, we can just ignore it (and continue). It's weird, but best to finish what was already going on
- if we receive ASSIGN, then we should abort indeed and print a message ; however if wonder if that aborting should not immediately call "doUnassign()" : so we resume partitions, unassign the customer (but not the newly assigned partitions...) and emit signal...

lib/BackbeatConsumer.js
```diff
@@ -804,6 +816,17 @@ class BackbeatConsumer extends EventEmitter {
...
+                        // being published: the assignment is no longer
+                        // ours to release
+                        this._log.info('skipping superseded un-assign', {
+                            topic: this._topic,
+                            rebalanceId,
+                            currentRebalanceId: this._rebalanceId,
+                        });
+                        return;
+                    }
                     this._resumePausedPartitions();
```
francoisferrand, line 830: should we not resume paused partitions as well?
this may happen when breakbeat engages - if we don't do it everything is stuck...

francoisferrand (commented):
I think we should point clearly that this actually _breaks_ our rebalance handling: we would still be processing previous tasks, yet not be able to commit them...

Would be safer to fit these "unexpected" (and buggy) events in the same automaton, e.g. sending the same events eventually/etc ; and possibly fail like when our own timeout is reached...

PR #2838, opened 2026-09-02
Leave the consumer group promptly on shutdown

delthas:
## Problem

`close()` unsubscribed then waited for the rebalance callback to un-assign before disconnecting:

```js
this._consumer.unsubscribe();
this.once('unassign', () => next());   // unbounded
```

librdkafka delivers no such callback when the consumer holds no assignment, and postpones the unsubscribe outright while a rebalance is in progress. So the wait never completed: `close()` never returned, the SIGTERM handler never reached `process.exit`, and the pod was SIGKILLed with the member still registered at the broker. The group then held zero partitions until `session.timeout.ms` (45s) evicted it — the takeover stall this ticket is about.

## Mechanism

`rd_kafka_consumer_close()` (reached via `disconnect()`) already performs the whole protocol exit from any join state. Its `terminate0` calls the *internal* unsubscribe with `leave=true`, which **bypasses the postponement** that blocks the application-level `unsubscribe()`, and it sends the `LeaveGroup` on its own. Its one requirement is that any rebalance callback it delivers on the close queue is answered — with `unassign()` (`assign()` is refused once the close has started).

```mermaid
sequenceDiagram
    participant App as close()
    participant RK as node-rdkafka
    participant CG as librdkafka cgrp
    App->>RK: unsubscribe()
    Note over CG: postponed while a rebalance is in progress
    App->>RK: disconnect()
    RK->>CG: consumer_close() then terminate0
    Note over CG: internal unsubscribe(leave=true)<br/>bypasses the postponement
    CG-->>App: REVOKE (if partitions still held)
    App->>RK: unassign() — answered at once
    CG->>CG: LeaveGroup sent
    RK-->>App: 'disconnected'
```

This corrects the `F_LEAVE_ON_UNASSIGN_DONE` framing in #2819: the leave comes from `terminate0`, and the explicit pre-disconnect unsubscribe/unassign are only there to drain in-flight work first, not to arm the leave.

## Approach

Rather than gate `disconnect()` on an un-assign that may never arrive:

- **Wait for the drain only when a revoke is actually coming** — one already parked and draining, or about to be triggered by the unsubscribe because we still hold partitions — and bound that wait. A revoke genuinely in progress drains under the full budget (its own watchdog emits `unassign` at `maxPollIntervalMs - 1000`); a revoke that never *starts* (an unsubscribe postponed by an in-flight rebalance) is given only a short grace, then we let `disconnect()` perform the leave. Waiting the full budget there would run past the pod's grace period.
- **Answer every rebalance callback raised once shutting down** — decline a grant, un-assign a revoke — so `disconnect()` cannot wedge on an unanswered one.
- **Bound the disconnect wait itself.**
- **Stop fetching once the shutdown has started**, so the pipeline the drain is waiting on cannot refill.
- **Bound the wait for an in-flight offset publish**, rather than rescheduling `close()` indefinitely if the ZooKeeper publish never calls back.

The first commit migrates `close()` to async/await (a behavior-preserving refactor, `close(cb)` kept as a thin shim), per the repo's [async/await migration guidance](https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration); the second is the fix above.

The un-assign wait uses a guarded `_getAssignments()` (mirroring `_getSubscription()`), since `assignments()` throws `ERR__STATE` on a closing client (BB-845).

### `close()` flow

```mermaid
flowchart TD
    A["close(cb)"] --> B["_closing = true<br/>stop fetch loop, cron, circuit breaker"]
    B --> C{"consumer connected?"}
    C -->|no| Z["disconnect step"]
    C -->|yes| D["best-effort unsubscribe()"]
    D --> E{"waiting on a drain<br/>or partitions held?"}
    E -->|no| Z
    E -->|yes| W["wait for 'unassign' (bounded)"]
    W -->|"'unassign' emitted<br/>(drain done or watchdog)"| Z
    W -->|"grace elapsed,<br/>revoke in progress: keep waiting"| W
    W -->|"grace elapsed,<br/>no revoke started"| G["give up: unassign()"]
    G --> Z
    Z --> H["_disconnecting = true<br/>disconnect(), bounded 5s"]
    H --> I["cb()"]
```

### Rebalance callbacks during shutdown

Only the disconnect phase short-circuits the drain; the same single drain path handles a revoke while running and while shutting down.

```mermaid
stateDiagram-v2
    direction LR
    [*] --> Running
    Running --> ShuttingDown : close() called
    ShuttingDown --> Disconnecting : drain done or bound hit
    Disconnecting --> [*] : disconnected

    note right of Running
        ASSIGN -> assign(partitions)
        REVOKE -> drain, then unassign()
    end note
    note right of ShuttingDown
        fetch loop stopped
        ASSIGN -> decline: unassign()
        REVOKE -> drain, then unassign()
    end note
    note right of Disconnecting
        ASSIGN -> decline: unassign()
        REVOKE -> unassign() now, no drain
    end note
```

## Relationship to the earlier stack

Supersedes the substance of #2819 / #2834 / #2835, rebuilt as one change on `development/9.5` (not stacked on #2818). It keeps a single drain path — the existing revoke handler — rather than adding a second one for shutdown. The double-`close()` coalescing (#2836) is intentionally left out.

## Known residual

When the leaving member has its own `JoinGroup` in flight (it was revoked and is rejoining), its explicit `LeaveGroup` is transport-blocked behind that request. `close()` still returns within its bound and the remaining members take over via a normal rebalance; only that member's fast explicit leave is delayed. This is a librdkafka transport-ordering limitation, not specific to this change; lowering `session.timeout.ms` (BB-844) bounds its cost.

## Tests

- Unit tests covering every shutdown path (the un-assign wait and its grace/re-arm, the bounded disconnect, declining a grant and answering a revoke during shutdown, and stopping consumption).
- Two real-broker departure tests: prompt takeover when the group is settled, and `close()` returning within its bound when called during a rebalance.

Issue: BB-833

### DIFF
--- a/lib/BackbeatConsumer.js
+++ b/lib/BackbeatConsumer.js
@@ -19,6 +19,8 @@ const {
     backbeatConsumer: {
         CONCURRENCY_DEFAULT,
         MAX_QUEUED_DEFAULT,
+        DISCONNECT_TIMEOUT_MS,
+        SHUTDOWN_DRAIN_GRACE_MS,
     }
 } = require('./constants');
 
@@ -31,10 +33,6 @@ const { withTopicPrefix } = require('./util/topic');
 
 // Resolve after `ms` (call-time setTimeout so test fake timers apply).
 const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
-// Resolve when `emitter` next emits `event` (errors are ignored, matching
-// the callback-based waits this replaced).
-const waitForEvent = (emitter, event) =>
-    new Promise(resolve => emitter.once(event, resolve));
 
 /**
  * Stats on how we are consuming Kafka
@@ -201,6 +199,15 @@ class BackbeatConsumer extends EventEmitter {
         // since the last consume, and is used to trigger a new consume
         // immediately after the current one
         this._tasksCompletedSinceLastConsume = false;
+        // Set at close() entry: stops fetching, declines new grants
+        this._closing = false;
+        // Set at the disconnect step: rebalance callbacks are answered
+        // at once, since the client's close blocks on them
+        this._disconnecting = false;
+        // True between receiving a revoke and answering it (draining)
+        this._waitingDrain = false;
+        // Bounds close()'s wait for the drain
+        this._closeDrainTimeout = null;
 
         /** @type {ConsumerStats} */
         this.consumerStats = { lag: {} };
@@ -436,6 +443,13 @@ class BackbeatConsumer extends EventEmitter {
         if (this._tryConsumedTimeout) {
             clearTimeout(this._tryConsumedTimeout);
             this._tryConsumedTimeout = null;
+        }
+
+        // stop fetching once closing: new entries would refill the
+        // pipeline close() is draining, and this ends the
+        // self-rescheduling consume loop that would poll a closed client
+        if (this._closing) {
+            return undefined;
         }
 
         // use non-flowing mode of consumption to add some flow
@@ -765,6 +779,19 @@ class BackbeatConsumer extends EventEmitter {
      */
     _onRebalance(err, assignment) {
         if (err.code === kafka.CODES.ERRORS.ERR__ASSIGN_PARTITIONS) {
+            if (this._closing) {
+                // Decline the grant, but still answer the callback or the
+                // client stays parked. assign() is refused mid-disconnect,
+                // so answer with unassign().
+                this._log.info('rdkafka.assign declined, shutting down',
+                    { assignment });
+                this._bestEffort('unassign',
+                    () => this._consumer.unassign());
+                if (!this._waitingDrain) {
+                    this.emit('unassign', unassignStatus.SHUTDOWN);
+                }
+                return;
+            }
             this._log.info('rdkafka.assign', { assignment });
 
             try {
@@ -785,6 +812,19 @@ class BackbeatConsumer extends EventEmitter {
                 ledger: this._offsetLedger.getProcessingCount(this._topic),
             });
 
+            if (this._disconnecting) {
+                // The client's close delivers this revoke and blocks until
+                // answered; entries can no longer commit, so answer at
+                // once instead of draining.
+                KafkaBacklogMetrics.onRebalance(this._topic, this._groupId,
+                    unassignStatus.SHUTDOWN);
+                this._bestEffort('unassign',
+                    () => this._consumer.unassign());
+                this.emit('unassign', unassignStatus.SHUTDOWN);
+                return;
+            }
+
+            this._waitingDrain = true;
             const unassign = jsutil.once(status => {
                 this._log.info(`processing queue ${status}, un-assigning`, {
                     queueLen: this._processingQueue.length(),
@@ -820,6 +860,7 @@ class BackbeatConsumer extends EventEmitter {
                         logger.bind(this._log)('rdkafka.unassign failed', { e: e.toString(), assignment });
                     }
 
+                    this._waitingDrain = false;
                     this.emit('unassign', status);
                 };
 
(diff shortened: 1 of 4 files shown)


lib/BackbeatConsumer.js
```diff
@@ -1231,30 +1318,80 @@ class BackbeatConsumer extends EventEmitter {
         this._circuitBreaker.stop();
         return async.waterfall([
```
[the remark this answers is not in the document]
delthas: Rewritten

lib/BackbeatConsumer.js
```diff
@@ -759,6 +779,19 @@ class BackbeatConsumer extends EventEmitter {
      */
     _onRebalance(err, assignment) {
         if (err.code === kafka.CODES.ERRORS.ERR__ASSIGN_PARTITIONS) {
+            if (this._closing) {
```
francoisferrand, line 782: not needed, since we `unsubscribe()` first ?

lib/BackbeatConsumer.js
```diff
@@ -779,6 +812,19 @@ class BackbeatConsumer extends EventEmitter {
                 ledger: this._offsetLedger.getProcessingCount(this._topic),
             });
 
+            if (this._disconnecting) {
+                // The client's close delivers this revoke and blocks until
+                // answered; entries can no longer commit, so answer at
+                // once instead of draining.
```
francoisferrand, line 818: correct in term of kafka (will be replayed anyway); but it is better to abort the current task or finish them (even if we don't commit)?
e.g. most of our tasks are idempotent, and replaying them is fine: for exemple aborting replication wastes some bandwidth, while replaying it after it completes is mostly no-op...

lib/BackbeatConsumer.js
```diff
@@ -1221,42 +1309,111 @@ class BackbeatConsumer extends EventEmitter {
...
+    async _close() {
+        this._closing = true;
         if (this._publishOffsetsCronTimer) {
             clearInterval(this._publishOffsetsCronTimer);
             this._publishOffsetsCronTimer = null;
         }
-        if (this._publishOffsetsCronActive) {
-            return setTimeout(() => this.close(cb), 1000);
+        // Wait briefly for an in-flight offset publish to finish, but never
+        // block shutdown on it: after the grace, proceed and let it error out.
```
francoisferrand, line 1324: why would be it delayed more than 1s?
(what is the cron timer period)

lib/BackbeatConsumer.js
```diff
@@ -1221,42 +1309,111 @@ class BackbeatConsumer extends EventEmitter {
...
+        const publishDeadline = Date.now() + SHUTDOWN_DRAIN_GRACE_MS;
+        while (this._publishOffsetsCronActive && Date.now() < publishDeadline) {
+            await delay(1000);
```
francoisferrand, line 1327: if we go that road, should we not use some sort of even (or callback), instead of a polling loop?
(and this is probably a different PR than handling close during rebalance)

lib/BackbeatConsumer.js
```diff
@@ -1221,42 +1309,111 @@ class BackbeatConsumer extends EventEmitter {
      * @return {undefined}
      */
     close(cb) {
+        // Keep the callback signature for existing callers; errors are
+        // swallowed as before (cb takes no arguments).
+        this._close().then(() => cb(), () => cb());
```
francoisferrand, line 1314:
```suggestion
        this._close().then(cb, () => cb());
```

lib/BackbeatConsumer.js
```diff
@@ -1221,42 +1309,111 @@ class BackbeatConsumer extends EventEmitter {
...
-                    this._consumer.once('disconnected', () => next());
-                } else {
-                    process.nextTick(next);
+
+        if (this._consumer?.isConnected()) {
+            // Best-effort unsubscribe: a no-op when not subscribed, and the
+            // trigger for the drain revoke when we hold partitions.
+            // Unconditional so a transient ERR__STATE on subscription() can't
+            // skip it.
+            this._bestEffort('unsubscribe',
```
francoisferrand, line 1336: nit: this `bestEffort` function really does not help readability....

lib/BackbeatConsumer.js
```diff
@@ -195,6 +199,15 @@ class BackbeatConsumer extends EventEmitter {
...
         this._tasksCompletedSinceLastConsume = false;
+        // Set at close() entry: stops fetching, declines new grants
+        this._closing = false;
+        // Set at the disconnect step: rebalance callbacks are answered
+        // at once, since the client's close blocks on them
+        this._disconnecting = false;
+        // True between receiving a revoke and answering it (draining)
+        this._waitingDrain = false;
+        // Bounds close()'s wait for the drain
+        this._closeDrainTimeout = null;
```
francoisferrand, line 210: does not seem like 4 states : but a single state/automaton

francoisferrand (commented):
> librdkafka delivers no such callback when the consumer holds no assignment

this is why we check if there are assignments before:
```
if (subscription !== null) {
    this._consumer.unsubscribe();
    this.once('unassign', () => next());   // unbounded
```

> postpones the unsubscribe outright while a rebalance is in progress

that is the real condition we are trying to handle, right?
...but AFAIK (still not fully clear, the comment is very verbose and PR quite long; so not fully sure I understand) if rebalance is in progress, we would have received the revoke callback and `unassign` event is "guaranteed" to happen already (i.e. once we have finished draining)

So not sure what the exact condition (scenario) you are trying to fix...

PR #2841, opened 2026-09-05
Index ingestion readers by bucket and location

francoisferrand:
`IngestionPopulator` held its readers in arrays, so every lookup was a
linear scan. Removing a source walked both reader lists; pausing or
resuming a location walked them again to match on location constraint.
With one bucket per location and deployments reaching thousands of
locations, a configuration change or a pause event became quadratic work
on the event loop.

Readers are now keyed by zenko bucket, with a separate index listing the
buckets of each location for the operations that act on a whole location.
`_pausedLocations` becomes a map as well, so membership tests stop
enumerating its keys. Every lookup, insertion and removal is now constant
time, and the logic of each code path is unchanged.

Two latent bugs disappear as a side effect: `_removePausedReaders()`
spliced `logReaders` while iterating it with `forEach`, skipping the
element after each removal, and `_resumeService()` could push a reader
that was already active. Both were masked by the one-bucket-per-location
deployment, but neither invariant was enforced anywhere; the map
representation makes them unrepresentable.

Issue: BB-865

### DIFF
--- a/lib/queuePopulator/IngestionPopulator.js
+++ b/lib/queuePopulator/IngestionPopulator.js
@@ -496,28 +496,18 @@ class IngestionPopulator {
         // ]
 
         const buckets = config.getIngestionBuckets();
-        const configuredSources = Object.keys(this._ingestionSources);
+        const staleSources = new Set(Object.keys(this._ingestionSources));
         buckets.forEach(bucket => {
             const { zenkoBucket } = bucket;
-            // Get ingestion source information
             const sourceInfo = this._getSourceInformation(bucket);
-            // if the zenko bucket has already been setup and is already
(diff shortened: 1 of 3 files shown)


DarkIsDude (commented):
The JIRA mention key by location and some other place key by bucket, which one should we trust ? Otherwise look good to me. Let me know if you need a re-review if some stuff change 🙏

DarkIsDude (approved):

francoisferrand:
> The JIRA mention key by location and some other place key by bucket, which one should we trust ? Otherwise look good to me. Let me know if you need a re-review if some stuff change 🙏

The first (and only) comment in the ticket explains the change/refinement.
Nothing to change AFAIK

maeldonn (approved):

PR #2842, opened 2026-09-07
Publish no metrics when none are configured

delthas:
We now deploy mongo-processor for PRA, which is supposed to consume from the production Kafka instance it might only have **read access** to.

But currently it would try to setup automatically some metrics producer at startup, which are not used at all by it. (The metrics producer is normally used to write some metrics in a topic, then get them back later, put them into Redis, serve them for an API, but that's not used by the mongo processor).

So here we make the metrics producer entirely optional if the metrics configuration block is not passed. To avoid crashing the mongo processor.

## Why the producer is still constructed

Optional here means it connects to nothing and publishes nothing, rather than not existing. The producer is handed down to the replication tasks and the log readers, and `CopyLocationTask` pulls the raw `BackbeatProducer` out of it with `getProducer()` from its own constructor. Leaving it null instead would need a guard at ten call sites, one of them a constructor, and a missed one is a crash.

Three callers reach past `publishMetrics()` and are handled where they are: the mongo-processor schedules no reporting interval, the two populators start no `MetricsConsumer` (it needs a topic and a redis of its own), and `ReplicationMetric` sends nothing.

Issue: BB-853

### DIFF
--- a/extensions/mongoProcessor/MongoQueueProcessor.js
+++ b/extensions/mongoProcessor/MongoQueueProcessor.js
@@ -63,7 +63,8 @@ class MongoQueueProcessor {
      *  backoff factor
      * @param {number} [mongoProcessorConfig.concurrency] - consumer concurrency
      * @param {Object} mongoClientConfig - config for connecting to mongo
-     * @param {Object} mConfig - metrics config
+     * @param {Object} [mConfig] - metrics config, no metrics are published
+     *   when omitted
      */
     constructor(kafkaConfig, mongoProcessorConfig, mongoClientConfig, mConfig) {
         this.kafkaConfig = kafkaConfig;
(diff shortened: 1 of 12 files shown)


benzekrimaha (approved):
LGTM

extensions/mongoProcessor/MongoQueueProcessor.js
```diff
@@ -82,9 +83,11 @@ class MongoQueueProcessor {
...
-            this._sendMetrics();
-        }, METRIC_REPORT_INTERVAL_MS);
+        if (mConfig) {
+            setInterval(() => {
+                this._sendMetrics();
+            }, METRIC_REPORT_INTERVAL_MS);
+        }
     }
 
     _setupMetricsClients(cb) {
```
francoisferrand, line 87: should not create producer if mConfig is not set.
(and setInterval may be set in this function instead, after the producer has been created?)

lib/MetricsProducer.js
francoisferrand, line 1: makes no sense to create a producer with no config?
do we really need this?

extensions/mongoProcessor/MongoQueueProcessor.js
francoisferrand: ok, found it (after opening the repo and search in the code) : the useful info in the PR message would have been to understand why we keep the producer, and it is half-said in the overly verbose section...

francoisferrand (approved):

francoisferrand:
> ## Human TL;DR
> [...]
> 
> ## Claude
> [...]

if it's too long and you need a human message, then drop the Claude verbose explanation. The PR message is addressed to the humans whom you ask for reviews 🙏

lib/MetricsProducer.js
delthas: It is a producer that connects to nothing and publishes nothing, rather than one built against a config it does not have — the alternative to it existing is `null`, and `null` travels.

The instance is handed down into the replication tasks and the log readers, and `CopyLocationTask` takes the raw `BackbeatProducer` out of it via `getProducer()` in its own constructor. So not constructing it means a guard at ten sites — `UpdateReplicationStatus`, `ReplicateObject` (x2), `MultipleBackendTask` (x2), `LogReader`, `IngestionReader`, the mongo-processor's two publish paths, and that constructor — where one missed guard is a crash rather than a missing counter.

Keeping the object confines the decision to the one class that owns the topic. I have written the reasoning into the PR description, which is where you should have found it.

extensions/mongoProcessor/MongoQueueProcessor.js
delthas: `setInterval` fixed.

delthas: I agree the PR body was bad. I should have rewritten it completely instead of just adding a TL;DR to avoid reading it.

delthas:
!done 3h

delthas:
[`BB-853`](https://scality.atlassian.net/browse/BB-853): logged `3h`; transitioned to Done (`Fixed`)

PR #2843, opened 2026-09-09
copy location task reads the source data from a remote site

francoisferrand:
A DR mirrors buckets from a production site it has no data path to: the
objects' metadata is replicated, but their data still lives on
production, in a CRR location Cloudserver has no data client for. So the
copy cannot go through the usual multiple-backend route, and reads the
data over S3 on the production site itself.

The location describes how to get there -- the servers to reach it, and
the STS to assume roles on -- and the object's location part says where
the data landed there and which role to assume to read it back, one role
per production account owning objects. Clients come from the replication
ClientManager, which already does this for replication: it refreshes the
assumed-role credentials, forgets the accounts that went quiet, and
supports an STS key read from a file.

A production site is Scality too, so both reads are the same GetObject
and only differ in where they point and whether a location constraint
means anything; they carry the same request uids, so a copy can be
followed across the two sites.

Failing to read the data is reported as an error rather than skipped, so
the lifecycle side gets to decide what it means for the transition and
the object is not left flagged in transition for good. A location that
is not reachable, or that holds more than the single part it is meant
to, fails the same way instead of quietly reading the wrong bytes.

Issue: BB-812
Co-authored-by: sylvain senechal <senechalsylvain.dev@gmail.com>

### DIFF
--- a/extensions/replication/queueProcessor/QueueProcessor.js
+++ b/extensions/replication/queueProcessor/QueueProcessor.js
@@ -229,6 +229,10 @@ class QueueProcessor extends EventEmitter {
         this.logger = new Logger(
             `Backbeat:Replication:QueueProcessor:${this.site}`);
 
+        // clients to read data straight from the sites we replicate to,
+        // keyed by endpoint and role, shared by all copy location tasks
+        this.sourceClientManagers = {};
+
         // global variables
         if (sourceConfig.transport === 'https') {
             this.sourceHTTPAgent = new HttpsAgent.Agent({
@@ -698,6 +702,7 @@ class QueueProcessor extends EventEmitter {
             destHTTPAgent: this.destHTTPAgent,
             vaultclientCache: this.vaultclientCache,
             accountCredsCache: this.accountCredsCache,
+            sourceClientManagers: this.sourceClientManagers,
             replicationStatusProducer: this.replicationStatusProducer,
             mProducer: this._mProducer,
             logger: this.logger,
--- a/extensions/replication/tasks/CopyLocationTask.js
+++ b/extensions/replication/tasks/CopyLocationTask.js
@@ -6,7 +6,7 @@ const { ObjectMD } = models;
 
 const BackbeatMetadataProxy = require('../../../lib/BackbeatMetadataProxy');
 const BackbeatTask = require('../../../lib/tasks/BackbeatTask');
-const { 
+const {
     BackbeatRoutesClient,
     GetObjectCommand,
     MultipleBackendPutObjectCommand,
@@ -24,8 +24,11 @@ const { getAccountCredentials } =
           require('../../../lib/credentials/AccountCredentials');
 const RoleCredentials =
           require('../../../lib/credentials/RoleCredentials');
+const ClientManager = require('../../../lib/clients/ClientManager');
+const { authTypeAssumeRole } = require('../../../lib/constants');
 const { metricsExtension, metricsTypeQueued, metricsTypeCompleted } =
     require('../constants');
+const locations = require('../../../conf/locationConfig.json') || {};
 
 const MPU_GCP_MAX_PARTS = 1024;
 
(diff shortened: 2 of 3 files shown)


extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +104,154 @@ class CopyLocationTask extends BackbeatTask {
...
+     */
+    async _sendGetObject(actionEntry, objMD, range, log, abortController) {
+        // read from the site holding the data, where it is native and there
+        // is no location to redirect to, or locally through Cloudserver
+        const { client, target: { bucket, key, version }, dataStoreName } =
+            this._getSourceLocationClient(objMD, log) ?? {
+                client: this.backbeatClient,
+                target: actionEntry.getAttribute('target'),
+                dataStoreName: objMD.getDataStoreName(),
+            };
```
maeldonn, line 238: nit: Hard to read the first time

```suggestion
        const source = this._getSourceLocationClient(objMD, log) ?? {
			client: this.backbeatClient,
            target: actionEntry.getAttribute('target'),
            dataStoreName: objMD.getDataStoreName(),
		};

		const { client, target, dataStoreName } = source;
		const { bucket, key, version } = target;
```

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +104,154 @@ class CopyLocationTask extends BackbeatTask {
...
+     * @param {String} roleArn - the role ARN carried by the location part
+     * @param {Werelogs} log - the logger instance
+     * @return {BackbeatRoutesClient} the client
+     * @throws {ArsenalError} AccessDenied (retryable) if credentials
+     * could not be obtained for the role
+     */
+    _getAssumedRoleS3Client(siteConfig, roleArn, log) {
+        const { transport, endpoint, sts } = siteConfig;
+        // a location may carry its servers without an explicit port
+        const [host, port = transport === 'https' ? 443 : 80] = endpoint.split(':');
```
maeldonn, line 120:
```suggestion
        const defaultPort = transport === 'https' ? 443 : 80;
		const [host, port = defaultPort] = endpoint.split(':');
```

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +104,154 @@ class CopyLocationTask extends BackbeatTask {
...
+     * reach it, no single part to read, no role to assume, or no credentials
+     * for that role
+     */
+    _getSourceLocationClient(objMD, log) {
+        const site = objMD.getDataStoreName();
+        if (!locations[site]?.isCRR) {
+            return null;
+        }
+        // The location itself carries how to reach the site holding the data:
+        // the servers to read from, and the STS to assume roles on.
```
maeldonn:
```suggestion
```

extensions/replication/tasks/CopyLocationTask.js
```diff
@@ -101,6 +104,154 @@ class CopyLocationTask extends BackbeatTask {
...
+            // retryable: a config fix should be picked up without losing the action
+            const err = errors.InternalError.customizeDescription(
+                `no endpoint to reach source location ${site}`);
+            err.retryable = true;
+            throw err;
+        }
+        // A CRR location holds the whole object in a single part, written by
+        // the source site's own replication: it carries where the data landed
+        // there, and the role to assume to read it back. Ranged reads below
+        // assume that, and would read the wrong bytes if it did not hold.
```
maeldonn:
```suggestion
        // Ranged reads below assume the whole object is in a single part.
```

maeldonn (approved):

extensions/replication/tasks/CopyLocationTask.js
francoisferrand: Not the same, this always computes the default...

extensions/replication/queueProcessor/QueueProcessor.js
```diff
@@ -229,6 +229,10 @@ class QueueProcessor extends EventEmitter {
         this.logger = new Logger(
             `Backbeat:Replication:QueueProcessor:${this.site}`);
 
+        // clients to read data straight from the sites we replicate to,
+        // keyed by endpoint and role, shared by all copy location tasks
+        this.sourceClientManagers = {};
```
delthas, line 234: nit: Should we ever prune old cache entries? Otherwise that's unbounded over time. Though I guess the volume will be very low.

delthas (approved):

extensions/replication/queueProcessor/QueueProcessor.js
francoisferrand: Issue already identified, shall be looked at in https://scality.atlassian.net/browse/BB-868

PR #2844, opened 2026-09-09
Honor transitionOneDayEarlier in v1 lifecycle transitions

maeldonn:
The v1 eligibility pre-filter and the noncurrent version transition apply compared raw rule days, ignoring transitionOneDayEarlier, so eligible objects were silently skipped while the current version apply stage honors the flag. Compute eligibility with LifecycleDateTime, as the apply stage does.

Issue: BB-867

### DIFF
--- a/extensions/lifecycle/tasks/LifecycleTask.js
+++ b/extensions/lifecycle/tasks/LifecycleTask.js
@@ -816,9 +816,10 @@ class LifecycleTask extends BackbeatTask {
      * @param {number} daysSinceInitiated - Days passed since entity (object or version) last modified
      * NOTE: entity is not an in-progress MPU or a delete marker.
      * @param {number} currentDate - current date
+     * @param {string} lastModified - entity last modified date
      * @return {boolean} true if rule applies - false otherwise.
      */
-    _isRuleApplying(rule, daysSinceInitiated, currentDate) {
+    _isRuleApplying(rule, daysSinceInitiated, currentDate, lastModified) {
         if (rule.Expiration && this._supportedRules.includes('Expiration')) {
             if (rule.Expiration.Days !== undefined && daysSinceInitiated >= rule.Expiration.Days) {
                 return true;
@@ -835,14 +836,11 @@ class LifecycleTask extends BackbeatTask {
 
         if (rule.Transitions && rule.Transitions.length > 0
             && this._supportedRules.includes('Transition')) {
+            // Same computation as the apply stage, so that
+            // transitionOneDayEarlier is honored.
             return rule.Transitions.some(t => {
-                if (t.Days !== undefined && daysSinceInitiated >= t.Days) {
-                    return true;
-                }
-                if (t.Date && t.Date < currentDate) {
-                    return true;
-                }
-                return false;
+                const transitionTime = this._lifecycleDateTime.getTransitionTimestamp(t, lastModified);
+                return transitionTime !== null && transitionTime <= currentDate;
             });
         }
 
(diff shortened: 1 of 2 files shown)


extensions/lifecycle/tasks/LifecycleTask.js
```diff
@@ -816,9 +816,10 @@ class LifecycleTask extends BackbeatTask {
      * @param {number} daysSinceInitiated - Days passed since entity (object or version) last modified
      * NOTE: entity is not an in-progress MPU or a delete marker.
      * @param {number} currentDate - current date
+     * @param {string} lastModified - entity last modified date
      * @return {boolean} true if rule applies - false otherwise.
      */
-    _isRuleApplying(rule, daysSinceInitiated, currentDate) {
+    _isRuleApplying(rule, daysSinceInitiated, currentDate, lastModified) {
```
francoisferrand, line 822: it seems weird to add `lastModified` here: since `daysSinceInitiated` is already computed from `lastModified`...

looking at the code in `LifecycleDateTime.findDaysSince()`, it seems the issue is really that `findDaysSince()` uses `expireOneDayEarlier` systematically: so that flag affects both transition & expiration, instead of only expiration and the other flag affecting transitions...

francoisferrand (commented):
I wonder if we should really fix this:
- `transitionOneDayEarlier` is not really needed, since we can configure the system to transition after 0 day (`expireOneDayEarlier` is more useful, since the minimum delay for expiration is 1 day)
- both flags are kind of deprecated, `timeProgressionFactor` should be used instead
- while sound in principle (e.g. use `getTransitionTimestamp` consistently to filter), I fear it may be complicated to insert cleanly in the code, without breaking abstractions levels...

extensions/lifecycle/tasks/LifecycleTask.js
SylvainSenechal: Looking at the code structure of isRuleApplying : Before the change, the code was similar for both expiration and transition. Now transition uses Arsenal helper while expiration manually checks the date.
It's a bit annoying to create another ticket in arsenal, but maybe we should have a getExpirationTimestamp helper in arsenal ? and this way, this fuction isRuleApplying will probably only need 2 parameters rule + lastModified I guess

SylvainSenechal:
> I wonder if we should really fix this:
> 
> * `transitionOneDayEarlier` is not really needed, since we can configure the system to transition after 0 day (`expireOneDayEarlier` is more useful, since the minimum delay for expiration is 1 day)
> * both flags are kind of deprecated, `timeProgressionFactor` should be used instead
> * while sound in principle (e.g. use `getTransitionTimestamp` consistently to filter), I fear it may be complicated to insert cleanly in the code, without breaking abstractions levels...

I just saw this comment : I wonder, are these only used for our own testing, or do we have clients using them ? 
I know timeProgressionFactor is used in the zenko cucumber tests.

If we can confirm that these 2 expire/transitionOneDayEarlier are only used for easier testing and not by clients, then yeah imo we should leave or deprecate them more actively

SylvainSenechal:
> * transitionOneDayEarlier

in zenko operator, they are called "xxxforTesting" :

	transitionsOneDayEarlier := cr.Spec.Backbeat.LifecycleBucketProcessor.TriggerTransitionsOneDayEarlierForTesting
	expirationsOneDayEarlier := cr.Spec.Backbeat.TriggerExpirationsOneDayEarlierForTesting
	timeProgressionFactor := cr.Annotations[timeProgressionFactorAnnotation]

Probably worth to take a look and see if we can just remove that code and just keep time progression factor 🤔

SylvainSenechal (changes requested):
discussion, see comments

PR #2845, opened 2026-09-11
Keep only the targeted backend in the replica metadata

maeldonn:
toReplicaEntry cloned the source entry and only updated the targeted backend, so the destination metadata inherited the statuses of the source's other destinations, frozen at queue time and never refreshed. Keep the targeted backend alone, copying it since clone() shares replicationInfo with the source.

Issue: BB-875

### DIFF
--- a/lib/models/ObjectQueueEntry.js
+++ b/lib/models/ObjectQueueEntry.js
@@ -232,9 +232,15 @@ class ObjectQueueEntry extends ObjectMD {
 
     toReplicaEntry(backend) {
         const newEntry = this.clone();
+
+        // clone() shares replicationInfo with the source: detach before mutating.
+        newEntry.setReplicationInfo(this.getReplicationInfo());
+        const matched = newEntry._findBackend(backend);
+
         newEntry
             .setAccountId(this.getAccountId())
             .setBucket(this.getReplicationTargetBucket(backend))
+            .setReplicationBackends(matched ? [{ ...matched }] : [])
             .setReplicationSiteStatus(backend, 'REPLICA')
             .setReplicationStatus('REPLICA');
         return newEntry;
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backbeat",
-  "version": "9.5.2",
+  "version": "9.5.3",
   "description": "Asynchronous queue and job manager",
   "main": "index.js",
   "scripts": {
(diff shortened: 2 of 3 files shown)


lib/models/ObjectQueueEntry.js
```diff
@@ -232,9 +232,15 @@ class ObjectQueueEntry extends ObjectMD {
 
     toReplicaEntry(backend) {
         const newEntry = this.clone();
+
+        // clone() shares replicationInfo with the source: detach before mutating.
```
francoisferrand, line 236: should this be fixed in `clone()` instead? Or are there places where we rely on that shallow cloning behavior?

lib/models/ObjectQueueEntry.js
```diff
@@ -232,9 +232,15 @@ class ObjectQueueEntry extends ObjectMD {
...
+        newEntry.setReplicationInfo(this.getReplicationInfo());
+        const matched = newEntry._findBackend(backend);
```
francoisferrand, line 238: why should we send this: our "local" replication status makes no sense on the remote site, should not even be sent ?

SylvainSenechal: I'm guessing it's important since the pr and jira ticket were probably created after an issue was found 🤔  I think the current implementation is messing up with collectResponseHeader in cloudserver, right @maeldonn ?

lib/models/ObjectQueueEntry.js
```diff
@@ -232,9 +232,15 @@ class ObjectQueueEntry extends ObjectMD {
 
     toReplicaEntry(backend) {
```
SylvainSenechal, line 233: I think in the end the required change is not that complicated, but it took me a fair bit of discussion with ai to understand whats happening and why we are doing this change. 

The comment "// clone() shares replicationInfo with the source: detach before mutating." didn't really help me understand what was happening


I think the comment can be more straightforward, if I understood correctly, the issue is : 
A source replication site can now be used to replicate to multiple destination. When creating the destination entry, we only care about writing into that destinationEntry the data about where it was replicated, and not the other site, thus the findBackend and matched new thing.

So if I understood correctly, I think it should be more made more clear by the comment   .

As always, I think it's not helping that the function has variables named replicationBackend**S**(its a location?)/site, there is also another setReplicationBackends without an s

Anyways, I think this probably just needs one more comment around findBackend to explain why and thats it

lib/models/ObjectQueueEntry.js
SylvainSenechal: Needs more investigation but AI is hinting at directly changing Arsenal

PR #2846, opened 2026-09-12
Initiate deferred restores when a direct transition completes

francoisferrand:
A `RestoreObject` may be accepted on a direct-to-cold object before its archive
completes: cloudserver records the request in the metadata, but nothing can be
retrieved yet, and the object is still in its hot location. The restore has to
be initiated later, once the object is really archived.

The archive completion write carries the pending request and is itself an oplog
entry, so routing `s3:LifecycleTransition:Direct` to the restore handler is
enough to initiate it — no extra metadata write, no new publisher. The restore
still goes through Sorbet, which creates the restore-expiration job.

Two related issues showed up on the way:

- the completion write replaced the whole archive block with the archiveInfo
  returned by the cold backend, dropping the pending request, so the restore
  would have stayed pending forever;
- garbage collection keyed on the presence of the archive block, which now
  exists without archiveInfo while the object awaits its archive, and would
  have asked Sorbet to remove an archive that never existed.

Issue: BB-804


SylvainSenechal (approved):

DarkIsDude (approved):
LGTM

PR #2849, opened 2026-09-14
Bootstrap TypeScript compilation

francoisferrand:
New modules should be writable in TypeScript, but nothing in the toolchain understands it today. This wires up the compiler, the linter, the image build and the test suites, then converts one module to prove the whole path works.

The guiding constraint was that existing code must not notice. 379 JavaScript files and 155 test files are untouched: no test renamed, no spec migrated, no script repointed.

## What each commit does

**Type checking** — `tsconfig.json` plus a `typecheck` script, run in the existing lint job rather than a job of its own. `allowJs` with `checkJs` off keeps the current sources in the program so module resolution works, while leaving them unchecked, so the strict settings only ever apply to new TypeScript. typescript-eslint is scoped to `**/*.ts`.

**Compiling into the image** — the image now ships the build output. The contents of `dist/` are flattened into the working directory, so `lib/`, `bin/` and `extensions/` keep the paths they have today and every package script and entrypoint stays as it is. Compilation happens in its own stage so the devDependencies it needs never reach the runtime image.

`conf/` is copied separately because the compiler only emits JSON it sees `require`d, and `conf/config.json` is read with `readFileSync`. The IAM policies are not copied at all: they ship as their own `backbeat-policies` artifact, built from the checkout.

**Loading TypeScript in tests** — registering ts-node in `.mocharc.json` covers every suite at once, so the thirteen mocha invocations stay as they are and a spec keeps requiring its subject without an extension, whichever language it is written in.

**First module** — `lib/util/buffer.js` becomes `lib/util/buffer.ts`. It uses `export =`, not `export default`, so requiring it is unchanged; `export default` would force every caller to reach through `.default`.

## Worth knowing for the next migration

`export =` is the rule, not a preference. It emits a plain `module.exports = {...}`, which is what every existing caller expects.

`strict` implies `alwaysStrict`, so the compiler would otherwise add `"use strict"` to all 201 emitted files, 148 of which run sloppy today. Since mocha runs the sources and never the build output, no test would have caught the difference. It is explicitly off.

When migrating, the deletion of the `.js` has to land in the same commit as the `.ts`. With both present the compiler emits the `.ts` into `dist/` while node resolves the `.js` — silently, so the image would run different code from the tests.

`declaration` is off: backbeat ships as a service image, not a consumed library, so nothing reads its declarations. Turning it on forces type annotations into JavaScript sources to satisfy a `.d.ts` nobody loads.

## Checks

The build is a no-op for existing code: all 201 emitted files parse to the same syntax tree as their source, differing only in whitespace and the sourcemap comment.

The unit suite gives 1950 passing / 1 pending / 13 failing, identical with and without the loader registered, and identical again after the migration. The 13 need Kafka, Mongo and Redis.

`tests/unit/lib/util/buffer.spec.js` is unchanged and loads the TypeScript module, as do both plain-JavaScript callers. nyc attributes coverage to `buffer.ts` by name with line numbers from the source map.

The image was built and compared against one from the previous Dockerfile: same behaviour, and services still read `/usr/src/app/conf/`.

Test renames, the jest migration and re-enabling skipped tests are deliberately out of scope; they are tracked separately.

Issue: BB-295


.github/workflows/tests.yaml
```diff
@@ -290,10 +292,10 @@ jobs:
     - name: run ballooning tests for lifecycle conductor
       run: yarn mocha tests/performance/lifecycle/conductor-check-memory-balloon.js --exit
       env:
-        # Constrain heap long-lived heap size to 150MB, so that pushing 200K messages
+        # Constrain heap long-lived heap size to 175MB, so that pushing 200K messages
         # will crash if they end up in memory all at the same time (circuit breaking
         # ineffective) while waiting to be committed to the kafka topic.
-        NODE_OPTIONS: '--max-old-space-size=150'
+        NODE_OPTIONS: '--max-old-space-size=175'
```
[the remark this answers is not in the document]
francoisferrand: no, the point is to bootstrap _now_. So we load the typescript module (which takes ±20MB), and we increase the heap size. This way it is ready when we start really using typescript, and don't need to remember and adjust then.

Dockerfile
```diff
@@ -53,8 +64,13 @@ RUN apt-get update && \
...
 
-# Keep the .git directory in order to properly report version
-COPY . /usr/src/app
+# conf/config.json is read with readFileSync rather than required, so it is
+# never emitted and has to come from the sources.
+COPY conf/ ./conf/
+
+# Flattened on purpose: lib/, bin/ and extensions/ keep the paths they have in
+# the sources, so every entrypoint and package script works unchanged.
+COPY --from=compiler /usr/src/app/dist/ ./
```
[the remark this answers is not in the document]
francoisferrand: the script was renamed to `ensureServiceUser.js`, and a link is manually created in docker image for compatibility

.nycrc
```diff
@@ -2,8 +2,11 @@
...
         "bin/**/*.js",
+        "bin/**/*.ts",
         "lib/**/*.js",
-        "extensions/**/*.js"
+        "lib/**/*.ts",
+        "extensions/**/*.js",
+        "extensions/**/*.ts"
     ],
     "exclude": [
         "**/*.spec.js"
```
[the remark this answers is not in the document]
francoisferrand: Added `**/*.spec.ts` to `exclude`, for symmetry with the `.js` entry.

Worth noting it is belt-and-braces either way: all 86 specs live under `tests/`, which `include` never covers, so neither exclude pattern is load-bearing today. It costs nothing and keeps the two languages treated the same.

.nycrc
```diff
@@ -3,6 +3,7 @@
     "include": [
         "bin/**/*.js",
         "lib/**/*.js",
+        "lib/**/*.ts",
```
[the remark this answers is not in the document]
francoisferrand: Added both patterns — `.nycrc` `include` now carries `bin/**/*.ts` and `extensions/**/*.ts` alongside the `.js` ones.

They match nothing today, since the only TypeScript module is `lib/util/buffer.ts`, so this is purely so the next migration in either directory is measured rather than silently dropping out of coverage.

eslint.config.mjs
```diff
@@ -64,4 +65,21 @@ export default [...compat.extends("scality"), {
...
     },
+}, ...tseslint.configs.recommended.map(config => ({
+    ...config,
+    files: ["**/*.ts"],
+})), {
+    files: ["**/*.ts"],
+
+    languageOptions: {
+        parserOptions: {
+            project: "./tsconfig.json",
```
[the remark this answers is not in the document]
francoisferrand: Good catch, and reproduced before fixing. Dropped `parserOptions.project` entirely rather than adding a second tsconfig.

Verified both halves of the report:

Adding a `.spec.ts` under `tests/` did break `yarn lint`, exactly as described:

```
0:0  error  Parsing error: "parserOptions.project" has been provided for @typescript-eslint/parser.
The file was not found in any of the provided project(s): tests/unit/lib/util/probe.spec.ts
```

And `project` buys nothing today — `recommended` carries 46 rules, while the 27 type-aware ones live only in `recommendedTypeChecked`, which we do not extend.

Chose removal over a `tsconfig.eslint.json` because the second option adds a file to support a capability we are not using. If type-checked rules are enabled later, `project` comes back together with whatever config they need, as one deliberate change.

Confirmed the rules still fire afterwards, so this isn't quietly turning linting off:

```
1:7  error  'unusedVar' is assigned a value but never used  @typescript-eslint/no-unused-vars
```

`lib/util/buffer.ts` and a `tests/**/*.spec.ts` both lint clean now, and `yarn typecheck` is unaffected — type checking was always `tsc`'s job here, never eslint's.

eslint.config.mjs
```diff
@@ -64,4 +65,15 @@ export default [...compat.extends("scality"), {
...
         }
     },
+}, ...tseslint.configs.recommended.map(config => ({
+    ...config,
+    files: ["**/*.ts"],
+})), {
+    files: ["**/*.ts"],
+
+    rules: {
+        // Redundant with the compiler, which already rejects undefined names
```
DarkIsDude, line 75: Good point but why not let it here ? Compiler and linter can overlapse ?

tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+        "moduleResolution": "node",
+        "rootDir": "./",
+        "outDir": "./dist",
+        "allowJs": true,
+        "checkJs": false,
+        "resolveJsonModule": true,
+        // Backbeat ships as a service image, not as a consumed library: nothing
+        // reads its declarations, and emitting them would force type
+        // annotations into JavaScript sources to keep the compiler happy.
+        "declaration": false,
```
DarkIsDude, line 14: This is the default behaviour (most project don't need that), so I guess we can drop it ?

tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+        "sourceMap": true,
+        "strict": true,
+        // Existing modules run in sloppy mode and the test suite exercises the
+        // sources rather than the build output, so emitting the directive would
+        // change their semantics with nothing to catch it.
+        "alwaysStrict": false,
```
DarkIsDude, line 20: Did we check if we have issue with tests right now ? (add strict mode, run test, check the output) ? If it's not the case, it can be a good follow-up ?

Dockerfile
```diff
@@ -36,6 +36,17 @@ RUN yarn install --ignore-engines --frozen-lockfile --production --network-concu
     && rm -rf ~/.node-gyp \
     && rm -rf /tmp/yarn-*
 
+################################################################################
+FROM builder AS compiler
+
+# Compiling needs the devDependencies, which must not go in the runtime image.
+# Their install scripts are skipped: the compiler reads sources, it never loads
+# any of them.
```
DarkIsDude, line 44:
```suggestion
```

It's really a standard behaviour for a NodeJS docker image

package.json
```diff
@@ -102,6 +103,7 @@
     "nyc": "^17.1.0",
     "sinon": "^10.0.0",
     "typescript": "^5.7.3",
```
SylvainSenechal, line 105: Can keep this but there is also typescript 7 recently available

Dockerfile
```diff
@@ -53,11 +64,17 @@ RUN apt-get update && \
...
 COPY --from=builder /usr/src/app/node_modules ./node_modules/
 COPY --from=builder /usr/local/bin/dockerize /usr/local/bin/
 
+# Keep the extensionless entrypoint for backwards compatibility: deployments
+# spawn it by that path. The compiler emits the script like any other module,
+# without the executable bit it is spawned with.
+RUN chmod +x bin/ensureServiceUser.js \
+    && ln -s ensureServiceUser.js bin/ensureServiceUser
+
 ENV AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE=1
```
DarkIsDude, line 78: should we keep it as we migrated to v3 ?

Dockerfile
```diff
@@ -53,11 +64,17 @@ RUN apt-get update && \
...
-# Keep the .git directory in order to properly report version
-COPY . /usr/src/app
+COPY conf/ ./conf/
+COPY --from=compiler /usr/src/app/dist/ ./
 COPY --from=builder /usr/src/app/node_modules ./node_modules/
 COPY --from=builder /usr/local/bin/dockerize /usr/local/bin/
 
+# Keep the extensionless entrypoint for backwards compatibility: deployments
+# spawn it by that path. The compiler emits the script like any other module,
+# without the executable bit it is spawned with.
```
DarkIsDude, line 74:
```suggestion
# Expose the script without extention for backwards compatibility
```

I guess we can plan to remove it in a few versions ?

DarkIsDude (approved):

SylvainSenechal (approved):
I don't have much to say, if it works it works and its cool 🤷 

Typescript 7 is available since July, but when I tried using it in Zenko last month there was an issue with eslint still not ready to accept it, so I guess it can wait.

Maybe there are some more config that we will want to add or change but no need to block on this pr.

We may want to do a release now and use it in Zenko to make sure nothing is broken when building the image ?

Dockerfile
francoisferrand: I wish, but I want to avoid having to add (yet another) dependency on the version in operator (+ federation)
Instead of a link, should we maybe just _rename_ the script in the Docker image? So we never need to migrate?

(longer term: maybe we will handle this differently with the new Vault implementation... which would solve the problem in a different way)

Dockerfile
francoisferrand: I thought the same, though it's unrelated to this PR. I will check if it runs and/or still shows the error.

francoisferrand: the migration is not finished 😭

```
├─┬ @zenko/cloudserver@9.1.4 invalid: "git+https://github.com/scality/cloudserver#9.1.4" from the root project
│ ├─┬ arsenal@8.2.32 invalid: "git+https://github.com/scality/Arsenal#8.2.32" from node_modules/@zenko/cloudserver
│ │ └── aws-sdk@2.1692.0 deduped
│ ├── aws-sdk@2.1692.0
│ ├─┬ bucketclient@8.2.5 invalid: "scality/bucketclient#8.2.5" from node_modules/@zenko/cloudserver
│ │ └─┬ arsenal@8.2.4 invalid: "git+https://github.com/scality/Arsenal#8.2.4" from node_modules/@zenko/cloudserver/node_modules/bucketclient
│ │   └── aws-sdk@2.1692.0 deduped
│ └─┬ utapi@8.2.4 invalid: "github:scality/utapi#8.2.3" from node_modules/@zenko/cloudserver
│   ├─┬ arsenal@8.2.28 invalid: "git+https://github.com/scality/Arsenal#8.2.28" from node_modules/utapi
│   │ └── aws-sdk@2.1692.0 deduped
│   ├── aws-sdk@2.1692.0 deduped
│   └─┬ bucketclient@8.2.5 invalid: "scality/bucketclient#8.2.5" from node_modules/utapi
│     └─┬ arsenal@8.2.4 invalid: "git+https://github.com/scality/Arsenal#8.2.4" from node_modules/utapi/node_modules/bucketclient
│       └── aws-sdk@2.1692.0 deduped
└─┬ bucketclient@8.2.4 invalid: "scality/bucketclient#8.2.4" from the root project
  └─┬ arsenal@8.2.4 invalid: "git+https://github.com/scality/Arsenal#8.2.4" from node_modules/bucketclient
    └── aws-sdk@2.1692.0 deduped
```
--> creating followups BKTCLT-50 and BB-883

eslint.config.mjs
francoisferrand: the problem is not overlap, but false positive (as it is evaluate with JS engine, some things are not available) ; but it is actually alreayd disabled by `tseslint.configs.recommended`

tsconfig.json
francoisferrand: no issue in the tests, which run on the source code (so breakage cannot be seen easily). Filed followup ticket BB-884

package.json
francoisferrand: Typescript 7 is the golang rewrite of typescript, not supported by typescript-eslint (<6.1.0) and ts-node.
Typescript 6 could be possible, but raises warnings due to deprecated features.
→ followup BB-885

PR #2852, opened 2026-09-15
Skip change stream events that carry no documentKey

delthas:
Some events have no `documentKey` , such as `drop`, `rename`, `dropDatabase` and `invalidate` events, and we don't filter them out. Then some reads on the field were not null-guarded.

This adds a guard. We preserve incrementing the processed metric.

Issue: BB-879


SylvainSenechal (approved):
Ok, discussing with ai it seemed that there may be some more problems around the resumeToken handling and the way certain events are handled, but still better than a full crash.
Anyways this whole part should disappear or be heavily modified with the new oplog design right ?

extensions/oplogPopulator/OplogPopulator.js
```diff
@@ -174,42 +174,51 @@ class OplogPopulator {
...
-                break;
+        // events such as "drop" or "invalidate" carry no documentKey
+        const bucketName = change.documentKey?._id;
+        if (bucketName === undefined) {
+            this._logger.info('Skipping change stream event without documentKey', {
+                method: 'OplogPopulator._handleChangeStreamChange',
+                type: change.operationType,
+            });
+        } else {
+            const isListeningToBucket = this._allocator.has(bucketName);
```
francoisferrand: just a simple ternary would be enough?

```suggestion
            const isListeningToBucket = change.documentKey?._id ? this._allocator.has(bucketName) : false;
```

or better, do not compute these 2 bool every time, but only when needed? 

```suggestion
        const isListeningToBucket = change => this._allocator.has(change.documentKey._id);
        // no fullDocument field in delete events
        const isBackbeatEnabled = change => this._isBucketBackbeatEnabled(change.fullDocument.value);

     switch (change.operationType) {
            case 'delete':
                if (isListeningToBucket()) {
                    await this._allocator.stopListeningToBucket(change.documentKey._id);
                }
                [...]
```
(guessing -but I may be wrong- that `change.fullDocument` is always present ?)

delthas: Done.

`change.fullDocument` is empty for deletes, and can be empty for `updateLookup`s right before a delete (since the lookup races with the delete). I moved `isBackbeatEnabled` to the only case that uses it anyway.

francoisferrand (approved):

delthas:
!done 3h

delthas:
[`BB-879`](https://scality.atlassian.net/browse/BB-879): logged `3h`; transitioned to Done (`Fixed`)

PR #2854, opened 2026-09-16
Move fcntl to 0.3.1

francoisferrand:
0.2.2 is stuck on NAN and pins node-gyp 8 internally, which no longer builds on recent Node. 0.3.0 is the N-API rewrite arsenal already depends on, so this aligns backbeat with what the tree resolves anyway and takes one blocker off the Node 24 upgrade.

Nothing else pinned 0.2.2 on this line, so it drops out of the lockfile along with the node-gyp 8 subtree it pulled in — hence the large but very one-sided `yarn.lock` diff (334 deletions). No package changes version and none are added; every removed entry is attributable to 0.2.2's tree. `node-gyp` 8 is gone entirely, only the `^11.1.0` that 0.3.0 uses remains.

The binding surface is unchanged. `index.js` is byte-identical between the two tags and still exports the same `posixFadvise(fd, offset, len, advice)`. Diffing the NAN and N-API `fcntl.cpp` gives the same signature, same return value and the same short-circuit to 0 on macOS/Windows; the only difference is the error object built on the `res < 0` branch, which `posix_fadvise` doesn't take since it returns the errno instead of -1. backbeat never requires fcntl directly — the only consumer is arsenal's `releasePageCacheSync`, which just checks `ret !== 0`.

Deliberately separate from the Node 24 runtime bump (BB-888): this is a NAN to N-API swap of a native module, so it should be visible on its own rather than tangled with a runtime change. Nothing here bumps Node, the Dockerfile, CI or `engines` — it lands and is green on Node 22.

Verified on Node 22: `yarn install --frozen-lockfile` clean, native module builds and loads, real `posix_fadvise` exercised on linux (macOS short-circuits to 0) returning 0 for DONTNEED/WILLNEED and EBADF for a bad fd, arsenal's `releasePageCacheSync` works, 1984 unit tests passing, lint clean.

Issue: BB-887


package.json
```diff
@@ -67,7 +67,7 @@
     "breakbeat": "scality/breakbeat#v1.0.3",
     "bucketclient": "scality/bucketclient#8.2.4",
     "commander": "^13.0.0",
-    "fcntl": "github:scality/node-fcntl#0.2.2",
+    "fcntl": "github:scality/node-fcntl#0.3.0",
```
delthas: We could do `0.3.1` now that https://github.com/scality/node-fcntl/pull/22 was merged?

delthas (approved):

DarkIsDude (approved):

package.json
francoisferrand: Good call. 0.3.1 exists, keeps engines at >=22, and only adds the smoke-test/CI refresh while leaving index.js and src/fcntl.cpp unchanged, so I bumped the PR to 0.3.1.

PR #2855, opened 2026-09-17
Bump package.json to 9.6.0-preview.1

delthas:
## Summary

Bump backbeat to `9.6.0-preview.1` so `development/9.6` — which carries BB-808, the removal of `docker-entrypoint.sh` — can be consumed by Zenko for integration testing.

No released tag carries BB-808 today, and without it Zenko cannot use `backbeat.configurationOverrides` at all: on a 9.5.x image, setting *any* jq-mapped variable makes the entrypoint rewrite `conf/config.json` in place, which fails because the operator mounts that directory read-only.

```
/usr/src/app/docker-entrypoint.sh: line 321: conf/config.json.tmp: Read-only file system
```

The container then crash-loops and the Zenko CR never becomes `Available`. That is what blocks scality/Zenko#2535, and it is why `9.6.0-preview.1` specifically is the version to cut — zenko-operator already gates config-override support on `>= v9.6.0-preview.1` (`pkg/controller/zenko/backbeat.go:180`).

Issue: BB-808


DarkIsDude (approved):

benzekrimaha (approved):

delthas:
!done 1h

delthas:
!release

delthas:
Release summary:
* `development/9.6`: releasing [`9.6.0-preview.1`](https://github.com/scality/backbeat/actions/runs/35583129149)

delthas:
[`BB-808`](https://scality.atlassian.net/browse/BB-808): logged `1h`; already Done — left as is

PR #2857, opened 2026-09-21
Evict the client managers that stopped being used

francoisferrand:
The queue processor caches a `ClientManager` per destination host and role, and per source endpoint and role, but only ever added to those caches. A manager does drain its own credentials and clients once they expire, so what was left behind was the shell: the object, its two keepAlive agents and its sweep timer, one per endpoint and role for the life of the process. A tenant that replicates once leaves one there for good.

Cardinality is bounded by (endpoints × roles), so it grows with tenant count rather than with objects — not a runaway leak, but it never comes back down.

A manager now tells whoever holds it when a sweep left it with nothing, and a small cache owns them, dropping and closing it when that happens. An in-flight task may still be holding one: closing only releases the sweep and the idle sockets, the clients it handed out keep working.

The get-or-create wiring that `ReplicateObject` and `CopyLocationTask` each had is now in that cache, and `close()` also destroys the keepAlive agents so eviction actually reclaims the sockets.

Left out of scope: credentials whose STS refresh never succeeded keep a null `expireTime`, so they are never swept and would pin their manager. That is negative caching, closer to BB-868.

Issue: BB-886


DarkIsDude (approved):
LGTM 🙏

delthas (approved):

